PR #22, opened 2015-07-22
Split interfaces into smaller, reusable and compose-able units.

iocanel:
Existing interfaces that are used to compose the DSL are structured  / named based on their position in method chain rather than their behaviour. This makes it a bit difficult to understand and is restrictive too.

I split the interfaces into the smallest possible pieces (i.e. Listable, Deleteable, Editable etc) and re-composed the DSL using those building blocks.


jimmidyson:
Missing license headers - you can run `mvn license:format` IIRC

jimmidyson:
@iocanel I knew you had something in mind ;)

jimmidyson:
[merge]

PR #27, opened 2015-07-24
Rootpaths

iocanel:


iocanel:
[merge]

jimmidyson:
Could we move `RootPaths` in kubernetes-model into model package? No need for separate package for root.

iocanel:
No problem with me.

jimmidyson:
Fixed in kubernetes-model. I'm not working today (as you can see ;)) so would you be able to release the model 1.0.2 & update client accordingly?

PR #42, opened 2015-07-30
update method for HasMetadata types now respects resourceVersion. upd…

iocanel:
…ate method for Service also applies clusterIP.


iocanel:
[merge]

PR #48, opened 2015-08-03
Fixes #39: Remove DNS domain from kube client default URL

### LINKED ISSUE #39: looks like we should be able to remove the DNS domain from the hard coded DNS name
see https://github.com/openshift/origin/issues/3670#issuecomment-125600469

so hopefully this should be enough https://kubernetes.default.svc

jimmidyson:


jimmidyson:
[merge]

PR #54, opened 2015-08-04
Initial rolling update - image update only

### LINKED ISSUE #46: Rolling updates for RC
This should scale down existing RC while scaling up new RC until old RC pods == 0 & new RC pods == required replicas.

This should also support just updating the image rather than the whole RC as this is a common thing to do (& supported by kubectl).

jimmidyson:
Fixes #46


jimmidyson:
[merge]

PR #81, opened 2015-09-04
Move examples to their own module.

iocanel:
I have mixed feeling about kubernetes-client having logging impliementation artifacts in its pom even as provided dependencies.

I think that the client itself should have just slf4j-api. The implementations only make sense for the examples and the examples should move to their own module if they are to bring additional deps.


iocanel:
[merge]

iocanel:
[merge]

PR #82, opened 2015-09-06
Add load from file, process parameters for templates

jimmidyson:


jimmidyson:
@iocanel Fancy reviewing this before I merge? Moved `Processable` stuff to `openshift-client`. Added `load(InputStream)` to all types, supporting both JSON & YAML input streams.

Main reason to ask you to look at this is `templates.process` now returns `KubernetesList` but no way to create `KubernetesList` items. Wonder if it makes sense to add a `client.list().{load, create}` to loop through `items` & create each in turn? Sadly no atomic way to do so.

Perhaps something like:

```
t = client.templates().inNamespace('ns').withName('tmpl').process();
client.list().inNamespace('ns').withItems(t.items()).create();

t = client.templates().inNamespace('ns').load(is).process();
client.list().inNamespace('ns').withItems(t.items()).create();
```

My other problem with this is that `load` is under `inNamespace` as it then allows `get`, `create`, etc (`client.services().inNamespace('ns').load(is).create()`). Without namespace, could just allow `get` like:

```
s = client.services().load(is).get();
client.services().inNamespace('ns').create(s);
```

What do you think?

iocanel:
non-namespaced load
---
Clear enough. I like it! +1!

namespaced load
---
I assume that its there to support "create" from load. Or maybe even update from load etc.
As this could get quite nasty and quite complex, I wonder if it makes sense to just modify the existing Createable interface and add a method like:

    T create(InputStream is);

client list operations
---

If we are to treat KubernetesList as the rest of the resources, I would expect (for consistency) something like:

   KubernetesList t = client.templates().inNamespace('ns').withName('tmpl').process();
   client.list().inNamespace('ns').create(t);

instead of:

    KubernetesList t = client.templates().inNamespace('ns').withName('tmpl').process();
    client.list().inNamespace('ns').withItems(t.items()).create();

same applies for the load version of it.

On the other hand KubernetesList is not a resource we can query, delete etc as the rest of the resources (e.g. Pod, ReplicationController, Service etc), so I am wondering if this is an overkill. Not sure though.

Maybe something like: 

    client.templates().inNamespace('ns').withName('tmpl').processAndCreate();
    client.templates().inNamespace('ns').load(is).processAndCreate();

jimmidyson:
## Non-namespace load

Implemented - only allows get right now.

## Namespaced load

Namespaced load, etc is working fine. IMO it's not just for create, but is for all ops on a resource, name taken from input stream contents. So e.g. the following is valid:

```
client.services().inNamespace('ns').load(is).edit()....done();
```

Would fire a replace off rather than a create of course.

## Kubernetes list

Lists aren't only useful for creation from templates (also list of generic resources) so think it makes more sense to split out the `list()` into separate type, but only `Createable`, `Replaceable` perhaps?

Would still mean you'd have to do:

```
l = client.templates()....process();
client.list().inNamespace(...).create(l);
```

iocanel:
There are a lot of LoadXXX interfaces with the only thing being different being the return type. We could reduce the number of interfaces by using a generic return type.

jimmidyson:
@iocanel Very probably, but I can't figure it out right now :(

jimmidyson:
@iocanel TBH the interface structure is getting quite difficult to work with now. Hope there's not much more to add.

PR #87, opened 2015-09-08
Use cascading() method in the dsl, rather than using a cascading argu…

iocanel:
…ment. Fix cascading in mocks.


jimmidyson:
Cascading should be default behaviour.

iocanel:
[merge]

Issue #92, opened 2015-09-09
show a nicer error message if the client is not authenticated


jstrachan:
the following is a little cryptic:

```
Caused by: com.fasterxml.jackson.core.JsonParseException: Unrecognized token 'Unauthorized': was expecting ('true', 'false' or 'null')
 at [Source: io.fabric8.kubernetes.client.internal.org.jboss.netty.buffer.ChannelBufferInputStream@73a19967; line: 1, column: 14]
	at com.fasterxml.jackson.core.JsonParser._constructError(JsonParser.java:1581)
	at com.fasterxml.jackson.core.base.ParserMinimalBase._reportError(ParserMinimalBase.java:533)
	at com.fasterxml.jackson.core.json.UTF8StreamJsonParser._reportInvalidToken(UTF8StreamJsonParser.java:3448)
	at com.fasterxml.jackson.core.json.UTF8StreamJsonParser._handleUnexpectedValue(UTF8StreamJsonParser.java:2607)
	at com.fasterxml.jackson.core.json.UTF8StreamJsonParser._nextTokenNotInObject(UTF8StreamJsonParser.java:841)
	at com.fasterxml.jackson.core.json.UTF8StreamJsonParser.nextToken(UTF8StreamJsonParser.java:737)
	at com.fasterxml.jackson.databind.ObjectReader._initForReading(ObjectReader.java:378)
	at com.fasterxml.jackson.databind.ObjectReader._bindAndClose(ObjectReader.java:1494)
	at com.fasterxml.jackson.databind.ObjectReader.readValue(ObjectReader.java:1102)
	at io.fabric8.kubernetes.client.dsl.internal.BaseOperation.handleResponse(BaseOperation.java:433)
	at io.fabric8.kubernetes.client.dsl.internal.BaseOperation.handleGet(BaseOperation.java:463)
	at io.fabric8.kubernetes.client.dsl.internal.BaseOperation.get(BaseOperation.java:100)
```

wonder if we can give the user a hint they need to login via `oc login` or `kubectl login`?

PR #93, opened 2015-09-09
#92: Add assertResponseCode to have a single point for checking statu…

iocanel:
…s and throwing exceptions. Add custom error messsage as part of the Config object. Use different custom messages auth exception message when using external and internal clients. Allow using different message per http status.


iocanel:
[merge]

PR #114, opened 2015-09-15
fixes #113 so that we can easily get a pods log

### LINKED ISSUE #113: provide an API to get a pods log

jstrachan:
fixes #113 so that we can easily get a pods log

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/DefaultKubernetesClient.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/DefaultKubernetesClient.java
@@ -21,50 +21,12 @@ import com.ning.http.client.AsyncHttpClient;
 import com.ning.http.client.AsyncHttpClientConfig;
 import com.ning.http.client.ProxyServer;
 import com.ning.http.client.Realm;
+import com.ning.http.client.Response;
 import com.ning.http.client.filter.FilterContext;
 import com.ning.http.client.filter.FilterException;
 import com.ning.http.client.filter.RequestFilter;
 import com.ning.http.client.providers.netty.NettyAsyncHttpProviderConfig;
-import io.fabric8.kubernetes.api.model.DoneableEndpoints;
-import io.fabric8.kubernetes.api.model.DoneableEvent;
-import io.fabric8.kubernetes.api.model.DoneableNamespace;
-import io.fabric8.kubernetes.api.model.DoneableNode;
-import io.fabric8.kubernetes.api.model.DoneablePersistentVolume;
-import io.fabric8.kubernetes.api.model.DoneablePersistentVolumeClaim;
-import io.fabric8.kubernetes.api.model.DoneablePod;
-import io.fabric8.kubernetes.api.model.DoneableReplicationController;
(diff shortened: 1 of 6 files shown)


jstrachan:
merge

jstrachan:
[merge]

jimmidyson:
Read the error message first ;)

`License check failed: run mvn -N license:format to update all licenses, commit, squash & force push please.`

jstrachan:
@jimmidyson the 2nd push fixed that error - read the 2nd commit and passing error message? :)

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/DefaultKubernetesClient.java
```diff
@@ -317,6 +281,23 @@ public RootPaths rootPaths() {
   }
 
   @Override
+  public String getPodLog(String namespace, String podName, String containerName, boolean pretty) {
```
jimmidyson: This isn't at all inline with our beautiful DSL :(

`pods().inNamespace(...).withName(...).logs()` would be better, have to think about how to handle container & prettiness in there.

@iocanel Thoughts?

jimmidyson: This should also only be in `PodOperations...`.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/DefaultKubernetesClient.java
```diff
@@ -317,6 +281,23 @@ public RootPaths rootPaths() {
...
+    String url = getMasterUrl() + "namespaces/" + namespace + "/pods/" + podName + "/log?pretty=" + pretty;
```
jimmidyson: URLs can be constructed using the helpers.

jimmidyson:
Aha! Whoops!

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/DefaultKubernetesClient.java
jstrachan: only saw those helpers in BaseOperations though - was there somewhere else?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/DefaultKubernetesClient.java
jstrachan: tried adding it there first but got loads of compile errors as pods().withNamespace(x).withName(y) doesn't return PodOperations so there's no way to add extra pod specific operations?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/DefaultKubernetesClient.java
jstrachan: TBH this is a single function with a single result with 4 parameters; not sure how valuable it is to be in the pods DSL but if you wanna refactor stuff so it can be there its not a huge biggie either way

PR #120, opened 2015-09-18
Fix broken getLogs()

iocanel:


iocanel:
[merge]

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java
```diff
@@ -55,13 +55,13 @@ public String getLog(String containerId) {
...
   public String getLog(String containerId, Boolean isPretty) {
     StringBuilder sb = new StringBuilder();
-    sb.append("log?pretty=").append(isPretty);
+    sb.append("/log?pretty=").append(isPretty);
     if (containerId != null && !containerId.isEmpty()) {
       sb.append("&container=").append(containerId);
     }
 
     try {
-      URL url = new URL(getResourceUrl(), sb.toString());
```
jimmidyson: Why didn't this work? I assume because `getResourceUrl` didn't end in a slash? You probably should check that here, sometimes `//` will cause issues.

iocanel: You right! Maybe we need to fix resourceUrl() so that it always has a trailing slash

jstrachan: or use a helper method to join paths to check there's exactly one / between 2 path strings? e.g.  like https://github.com/jstrachan/fabric8/blob/cdelivery/components/fabric8-utils/src/main/java/io/fabric8/utils/URLUtils.java#L178-178

jimmidyson: Why is this just not part of the language... Java, eh...

PR #138, opened 2015-09-23
Fix broken openshift operations. Add basic tests for namepsace and no…

iocanel:
…n-namespaced openshift operations.


iocanel:
[merge]

PR #144, opened 2015-09-25
Split mock servers for kubernetes and openshfit from test bases and m…

iocanel:
…ake them reusable from downstream projects.


iocanel:
[merge]

PR #146, opened 2015-09-25
Tests that need to use kube config now ignore service account (cause …

iocanel:
…of failures inside kubernetes).


jimmidyson:
Hmm wonder if we should handle this differently in `Config`?

iocanel:
[merge]

iocanel:
[merge]

PR #155, opened 2015-10-06
Fix wrong url in Template process operation. Align OpenshiftConfig wi…

iocanel:
…th Config on handling of the root url. Add unit test.


iocanel:
[merge]

PR #156, opened 2015-10-06
Add support for deleting lists of items.

iocanel:


iocanel:
it's not ready yet.
needs more work.

iocanel:
@jimmidyson: I'd like your feedback here.

The point where I am not 100% sure is, if the multi delete should also be available to namespaced operations. 

Currently, multi delete is available in both:

    client.xxx().delete(xxx1,xxx2,xxx3...,xxxn);
    client.xxx().inNamespace("ns").delete(xxx1,xxx2,xxx3...,xxxn);

In the former case, the namespace on the item will be used. If it doesn't exists the default namespace will be used.

In the later case, for each item in the list, the namespace on the item will be used. If it doesn't exists the namespace on the operation will be used.

jimmidyson:
Do we use default namespace for other ops if not specified?

iocanel:
yes

iocanel:
[merge]

PR #179, opened 2015-10-20
Use consistent handling of error messages across operation. Provide m…

iocanel:
…essages that distinguish between error and failed requests.


iocanel:
[merge]

PR #185, opened 2015-10-23
Http Mock Server tests are now using SSL. Introduce SSL utils for reu…

iocanel:
…sing ssl related utilities in tests. Add unit test to test the behavior of ssl related problems.


iocanel:
[merge]

PR #203, opened 2015-11-02
Misc

iocanel:


jimmidyson:
Why is the SSL guessing required now? Do we allow master URLs without scheme?

iocanel:
@jimmidyson: it would be nice to align with "oc" in that matter.

jimmidyson:
Didn't know oc did that

iocanel:
[merge]

PR #212, opened 2015-11-10
How about client.adapt(OkHttpClient.class)

iocanel:
Since we no longer expose the client, but there are ppl still want to use it, I am wondering if it makes sense to do it via adapt().

This pull request makes all the required changes to adapt to support adapting things other than just clients. It also implements support of OkHttpClient adaptor (as part of the main adapt impl, not via service loader). Of course we could change that if we'd like.

This is more for review guys.@jimmidyson @alesj. Does it solve the issue we discussed yesterday?

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/BaseClient.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/BaseClient.java
@@ -94,20 +94,24 @@ public class BaseClient implements Client {
   }
 
   @Override
-  public <C extends Client> Boolean isAdaptable(Class<C> type) {
+  public <C> Boolean isAdaptable(Class<C> type) {
     ExtensionAdapter<C> adapter = Adapters.get(type);
     if (adapter != null) {
       return adapter.isAdaptable(this);
-    } else {
+    } else if (type.isAssignableFrom(OkHttpClient.class)) {
+      return true;
+    }else {
       return false;
     }
   }
 
   @Override
-  public <C extends Client> C adapt(Class<C> type) {
+  public <C> C adapt(Class<C> type) {
     ExtensionAdapter<C> adapter = Adapters.get(type);
     if (adapter != null) {
       return adapter.adapt(this);
+    } else if (type.isAssignableFrom(OkHttpClient.class)) {
+      return (C) httpClient;
     }
     throw new IllegalStateException("No adapter available for type:" + type);
   }
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Adapters.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Adapters.java
@@ -35,15 +35,15 @@ public final class Adapters {
     }
   }
 
-  public static <C extends Client> void register(ExtensionAdapter<C> adapter) {
+  public static <C> void register(ExtensionAdapter<C> adapter) {
     EXTENSION_ADAPTER_MAP.put(adapter.getExtensionType(), adapter);
   }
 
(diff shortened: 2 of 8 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/BaseClient.java
```diff
@@ -94,20 +94,24 @@ public Config getConfiguration() {
   }
 
   @Override
-  public <C extends Client> Boolean isAdaptable(Class<C> type) {
+  public <C> Boolean isAdaptable(Class<C> type) {
     ExtensionAdapter<C> adapter = Adapters.get(type);
     if (adapter != null) {
       return adapter.isAdaptable(this);
-    } else {
+    } else if (type.isAssignableFrom(OkHttpClient.class)) {
```
alesj: Why don't you simply internally register ExtensionAdapter for OkHttpClient,
and handle everything the same way -- via ExtensionAdapter.

alesj:
@iocanel, @jimmidyson this would work for me.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/BaseClient.java
iocanel: That's an option too, though it would either have to use reflection or have access to the client

alesj: Add inner class (OkHttpClientExtensionAdapter) to BaseClient (or where httpClient field is),
and register it against Adapters in BaseClient ctor -- no reflection here.

alesj: Or simply register anonymous ExtensionAdapter in BaseClient ctor.

jimmidyson:
Personally I don't really like it. It's not adapting a client to a related type, it's exposing an implementation detail.

The only reason I see for exposing the HTTP client directly is to use it for stuff that we're missing in the kube client. I'd rather just implement that stuff properly...

iocanel:
@jimmidyson: Its not that I am crazy about it either, but it is an option. The other option which requires more effort is to watcher closer how @alesj is using the client and try to support in our client without exposing our internals.

jimmidyson:
This is a client to kubernetes. There is a known set of operations, some of which we have not implemented yet (e.g. proxy, exec) due to previous HTTP client constraints that aren't there now.

I feel like exposing the client is like allowing people to use com.sun packages... but with a realistic chance of us changing it & breaking users' stuff.

iocanel:
@jimmidyson: You purist ...

iocanel:
:-)

alesj:
Like I said, the fully configured client is *very* useful.
And as much as I agree with @jimmidyson on the impl details,
I cannot see how this is gonna work out -- supporting every need that I might have.
This is always gonna be a catch-up game, plus probably adding some stuff that really don't belong into K8s/OS client code.

jimmidyson:
@alesj We don't need to support every need _you_ might have, we need to provide a client that covers all kubernetes/openshift operations. Turnaround on missing operations has been very quick so far when requested so shouldn't be a problem I wouldn't say. And in this particularly library we won't be `adding some stuff that really don't belong into K8s/OS client code`. If you want that then you should create your own higher level client that wraps this REST client. We might look at doing that in a separate module in future but right now it seems the REST API is enough for most needs.

alesj:
Imo, I'm not asking for much :-), a simple handle to existing underlying http client.
It's not like people will abuse it and then complain.

I'm totally fine if / when this impl detail changes.

PR #216, opened 2015-11-10
Fix watch single resource & close functionality

### LINKED ISSUE #208: Can't watch specific resource entity
When I try to set the watch for a specific resource entity, like pod:

        Watch watch = client.pods().inNamespace("default")
                .withName(podName).watch(new Watcher<Pod>() {
.....
        });

I see an error in the log:
watch.onClose io.fabric8.kubernetes.client.KubernetesClientException: Connection unexpectedly closed

Most probably it happens, because the watch as query parameter could be set only for resource entity list, but not the specific resource, to set a watch for a specific resource different URL has to be used, like:
http://host:port/api/v1/watch/namespaces/default/pods/pod_name

jimmidyson:
Fixes #208


jimmidyson:
[merge]

Issue #218, opened 2015-11-11
Reuse http client when adapting client


jimmidyson:
Currently when we adapt a client to e.g. `OpenShiftClient` we create a new instance of `OkHttpClient`. This creates a separate `ThreadPoolExecutor` which may be wasteful considering the client is using the same config to create the `OkHttpClient` instance.

One intricacy to deal with if we do this is making sure we don't close the underlying `ThreadPoolExecutor` until every client using it is closed. Could get tricky.

Thoughts @iocanel?

iocanel:
@jimmidyson: You are going to love it !!!

The main reason why the adapt method doesn't reuse the same client, is that it doesn't have visibility to it.

jimmidyson:
Sort it out! ;)

iocanel:
return new DefaultOpenShiftClient(client.adapt(OkHttpClient.class), OpenShiftConfig.wrap(client.getConfiguration()));

jimmidyson:
That'll work...

How to deal with not closing the thread pool until we're really done with it?

jimmidyson:
Complete!

PR #220, opened 2015-11-11
Fixes for #218

iocanel:


jimmidyson:
So how do we ensure the underlying `ThreadPoolExecutor` doesn't get closed when apapted client gets closed?

iocanel:
We don't! The question is should we?

jimmidyson:
Yes! We close the thread pool on `client.close()` so any adapted client close would close thread pool for all clients. This is not a good thing.

iocanel:
The only problem I see in this is that we change the existing behaviour it might break code people upgrading. Besides that I think that this is the idea way of working (more like a cast not less like a factory).

jimmidyson:
As long as we protect the thread pool then I don't see how it would break existing behaviour?

iocanel:
and how do you propose we do that?

iocanel:
I think I have an idea.

iocanel:
which was not a very good one.

honestly I can't think of anything that doesn't classify as "Micky Mouse" hacks.

jimmidyson:
I knew you would if I waited long enough :-b

jimmidyson:
Damn. All I could think of is some kind of reference counter. Or perhaps we forget about doing this & not worry about multiple thread pools being created...

jimmidyson:
If it's helpful we could create our own `ExecutorService` & pass it to the first original `OkHttpClient` created.

jimmidyson:
So a custom `ExecutorService` that holds a reference count that we could increment/decrement  on clone/close & shutdown when zero?

iocanel:
I'd rather knowingly break forward compatibility, than do it by mistake while counting refs :-)

jimmidyson:
What's your idea that would break forwards compatibility?

jimmidyson:
Discussed on IRC - happy with this approach now as long as we also document that only the original client should be `close`d.

iocanel:
[merge]

jimmidyson:
:+1:

PR #224, opened 2015-11-12
Validate name passed to withName

### LINKED ISSUE #207: using .withName(null) has interesting side effects
So using client.pods().inNamespace("myns").withName(null).get():

Returns a Pod object, which is actually not a pod object but a PodList deserialised as Pod. 

This applies pretty much to all types.

jimmidyson:
Fixes #207


jimmidyson:
[merge]

PR #228, opened 2015-11-13
Misc

iocanel:


jimmidyson:
Instead of disabling it can we instead use the `/apis` endpoint to discover enabled API groups?

iocanel:
[merge]

PR #230, opened 2015-11-14
Adapt should clone client so per-connection settings can be applied

jimmidyson:


iocanel:
so we don't want to create a new client (as we used to) but we also don't want to use the same?

not sure I get it.

jimmidyson:
So new client = new thread & connection pool = waste of resources. Cloning client means settings (e.g. interceptors, read timeouts, etc) for the client are copied over to the new client but can be changed on the new client without affecting the old client, while reusing the same thread & connection pool. As an example, we do this in `WatchConnectionManager`: we change the read timeout on the cloned client as otherwise the watch is cancelled (effectively times out), but we retain the read timeout on the original client to protect against slow API server.

iocanel:
ok makes sense

jimmidyson:
[merge]

Issue #213, opened 2015-11-10
provide an equilvalent to .get() which fails if the resource is missing with a meaningful error message describing the namespace/name/labels being filtered


jstrachan:
I like the default of say

```
kubernetes.replicationControllers().withName("cheese").get()
```

returning null if not found.

It would be nice to add a version of this which returns a non-null value or throws a meaningful exception.

e.g. if you do something like...

```
ReplicationController rc = kubernetes.replicationControllers().withName("cheese").require()
```

if `cheese` doesn't exist it would be nice to get some exception like `new ResourceNotFoundException("ReplicationController cheese is not found in default namespace")`.

It then avoids doing these kinds of assertions at a higher level. These are kinda handy when writing test cases etc

iocanel:
Ooops! I accidentially closed this one!

PR #234, opened 2015-11-17
#213: Add getRequired() method to the Gettable interface.

iocanel:


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Gettable.java
```diff
@@ -23,4 +25,11 @@
    */
   T get();
 
+
+  /**
+   * @return the item or throws Exception if the item doesn't exist.
+   * @throws io.fabric8.kubernetes.client.KubernetesClientException if an error occurs
+   */
+  T getRequired() throws ResourceNotFoundException;
```
jimmidyson: How about `mustGet`?

iocanel: TBH, I don't like mustGet, I also don't like getRequired(), getMandatory() or even require().

I was tempted to make it part of the DSL as an additional method like: .failingIfNotExists().get(); But that also seems like an overkill. So not sure ....

iocanel: made up my mind..... it should be getOrBarf() :-)

jstrachan: getNotNull()? Guava has [checkNotNull()](http://guava-libraries.googlecode.com/svn-history/r14/trunk/javadoc/com/google/common/base/Preconditions.html#checkNotNull(T)) though thats more of a precondition check. how about getAndNotNull()?

iocanel: Let's summarise our options:

i) require()
ii) xxx.getRequired()
iii) xxx.getMandatory()
iv) xxx.mustGet()
v) xxx.getAndNotNull()
vi) xxx.failingIfMissing().get();

jimmidyson:
What's happening with this?

iocanel:
Once, I find some spare time, I I'll go with @davsclaus suggestion, which is much cleaner.

iocanel:
almost a year old, lets drop it

PR #235, opened 2015-11-17
225

iocanel:


jimmidyson:
Looks like FullExample has been mangled.

iocanel:
Damn! Let me fix that!

jimmidyson:
Is exec actually working? I can't see how you're hooking up the websocket to the streams.

jimmidyson:
Oh hang on not looking properly - sorry!

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Ttyable.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.fabric8.kubernetes.client.dsl;
+
+public interface Ttyable<T> {
+
+    T allocatingTerminal();
```
jimmidyson: `withTty`?

kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/ExecExample.java
```diff
@@ -0,0 +1,61 @@
...
+        try (final KubernetesClient client = new DefaultKubernetesClient(config)) {
+            watch = client.pods().withName(podName)
+                    .usingInput(System.in)
+                    .usingOut(System.out)
+                    .usingError(System.err)
+                    .withTTY()
+                    .exec();
+            Thread.sleep(10 * 1000);
+        } finally {
+            if (watch != null) {
```
jimmidyson: Looks a bit nicer (& shorter) with try-with-resources.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
```diff
@@ -0,0 +1,157 @@
...
+
+    @Override
+    public void onOpen(WebSocket webSocket, Response response) {
+        webSocketcRef.set(webSocket);
+        executorService.submit(pumper);
+    }
+
+    @Override
+    public void onFailure(IOException ioe, Response response) {
+        LOGGER.error(response != null ? response.message() : "Exec Failure.", ioe);
```
jimmidyson: Do we need to return this to the caller?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
```diff
@@ -0,0 +1,157 @@
...
+
+    private void send(byte[] bytes) throws IOException {
+        if (bytes.length > 0) {
+            WebSocket ws = webSocketcRef.get();
+            if (ws != null) {
+
+                try (BufferedSink sink = ws.newMessageSink(WebSocket.PayloadType.BINARY)) {
+                    sink.write(new byte[]{0});
+                    sink.write(bytes);
+                    sink.flush();
```
jimmidyson: Is this redundant? Doesn't `close` flush in this case?

iocanel: ohhh yes its try-with-resources.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
iocanel: The only way I see this being possible is by allowing the user to pass his own web socket listener and have ExecWebSocketListener delegate to it, if needed.

jimmidyson: We had to do this with `watch` otherwise there's little info for the client to work out why a call was rejected.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java
```diff
@@ -18,60 +18,168 @@
...
+        return new PodOperationsImpl(client, getConfig(), apiVersion, namespace, name, isCascading(), getItem(), containerId, in, out, err, withTTY).getLog(isPretty);
+    }
+
+    @Override
+    public ContainerResource<String, InputStream, OutputStream, OutputStream, String, Watch> inContainer(String containerId) {
+        return new PodOperationsImpl(client, getConfig(), apiVersion, namespace, name, isCascading(), getItem(), containerId, in, out, err, withTTY);
+    }
+
+    @Override
+    public TtyExecable<String, Watch> usingError(OutputStream err) {
```
jimmidyson: I understand the reasoning behind `OutputStream` so we can write to it, but shouldn't we try to follow the same approach as [`java.lang.Process`](http://docs.oracle.com/javase/7/docs/api/java/lang/Process.html) & swap this round so stdout is actually an `InputStream` for client's to read from, & stdin is an `OutputStream` for clients to write to? I know conceptually it seems weird, but that's the only way I can see this being usable programatically rather than just piping to `System.out`, etc

jimmidyson:
The shell example works great but I'm just worried about this being usable programatically - see https://github.com/fabric8io/kubernetes-client/pull/235/files#r45176125

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java
iocanel: From my POV its not the same thing as a Process, since we advise exec where to place the output, rather than asking exec from where to read the output.

Still, if it makes life easier to provide "additional" methods for the piping, then why not?

jimmidyson: From my PoV this is exactly the same as Process, just remote, so I'd like to support it if it's easy,

jimmidyson:
`ExecPipesExample` is pretty horrible isn't it? I was kind of hoping to return `InputStream`s for stdout/stderr & `OutputStream` for stdin to the caller without them setting up their own resources. Any chance of that do you think?

iocanel:
On stdin one can pass an input stream, a piped output stream or nothing.
On stdout/stdeer one can pass an output stream, a piped input stream or nothing.

In case of nothing the client creates and manages the required piped streams, which are passed back to the caller via ExecWatch item.

I am not sure what's exactly that you don't like in the ExecPipesExample, but my guess is that you don't like the pumping action. The user wouldn't need to use any kind of pump, he would just need to read the input stream.

jimmidyson:
I don't like the fact the caller has to create so much stuff to pass in. Very verbose. I'm not sure why we can't return the streams to the caller?

iocanel:
We CAN just return the streams. And we DO return the input stream: https://github.com/iocanel/kubernetes-client/blob/225/kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/ExecPipesExample.java#L64

We could do the same for the output stream, too. I choose not to so that we can demonstrate all possible options. Maybe that was not a very bright idea :-)

jimmidyson:
[merge]

PR #237, opened 2015-11-19
Improve the messages shown in exceptions. Only display the Status as …

iocanel:
…part of the message if its send by the server (still is included in the exception). Fix some inaccurate custom messages. Add test for the above.


iocanel:
[merge]

PR #241, opened 2015-11-23
OSFUSE-143: Use snakeyaml 1.8 on the redhat branches

davsclaus:


davsclaus:
Okay snake 1.8 is not an osgi bundle, so we have to use 1.16 etc.

iocanel:
Wrap protocol isn't installed OOTB in the minimal distribution of karaf that is used for testing.
We need to configure the test distro to include the wrap protocol in the startup, so that we can get this working. 

Something like this: https://github.com/iocanel/kubernetes-client/commit/9f3261bbaf1e5b0d3aa497d962e6e903c800affe

jimmidyson:
I think it best we upgrade to align with EAP7's dependency version: 1.15

davsclaus:
closing as 1.8 is not osgi compliant

PR #256, opened 2015-12-02
add method to allow to set namespaceRequired in BaseOperation.java

yybear:
some k8s create api doesn't need set namespace, for example: "create a PersistentVolume", the url is "POST /api/v1/persistentvolumes". so allow BaseOperation to set namespaceRequired

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/ClientMixedOperation.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/ClientMixedOperation.java
@@ -29,4 +29,6 @@ public interface ClientMixedOperation<T, L, D, R extends ClientResource<T, D>>
   extends ClientResource<T, D>,
   ClientOperation<T, L, D, R>,
   ClientNonNamespaceOperation<T, L, D, R> {
+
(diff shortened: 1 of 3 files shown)


iocanel:
Indeed! Some operations do not require a namespace. 

Currently it is up to the operation implementation to return the correct value in isNamespaceRequired().

So, PersistentVolumeOperationsImpl should override isNamespaceRequired() and return false.

But I don't agree that this should be exposed as method e.g. setNamespaceRequired(). An operation either requires a namespace or not, its not something that the user needs to tackle, neither its something that changes on runtime.

Could you please, remove the  setNamespaceRequired() and just override the isNamespaceRequired() where it makes sense?

yybear:
yeah, overriding isNamespaceRequired is better.

iocanel:
great catch btw

yybear:
this is my first pull request, so i need to close this comment and create a new pull request?

iocanel:
You can edit this pull request, by pushing things into the branch you used to create the pull reuquest (in your case yybear:master).

In situations like this, I usually, squash my commits into a single and force push into the branch.

something like this:

git add ....
git commit -m "New commit"
git rebase -i HEAD~2 
git push -f <remote repo> <remote branch>

note that squashing into a single commit (the rebase thing is optional).

Issue #260, opened 2015-12-04
Add additional options to get logs of a pod which already provides by Kubernetes rest api


kasundsilva:
I have been trying get logs of a running pod with fabric8 client. 

As I understand we can get logs as following,
As I understand we can get logs as following,

String log = kube.pods().inNamespace(<namespace>).withName(<pod_name>).getLog(true);

Here through the getLog() methos it provides only snapshot logs. But from Kubernetes rest api we have more flexibility to do stuff.

String url = "https://10.245.1.2/api/v1/namespaces/<namespace>/pods/<pod_name>/log?container=tomcat-server&follow=false&previous=false&timestamps=false";

As with the **URL parameters** we have more flexibility on the output to 
- Get previous terminated pod logs
- Stream the logs of pods
- Display only the most recent #of lines of pods
- Show all logs from pods written in the certain time period

davsclaus:
+1 sounds like great functionality

iocanel:
Merged

kasundsilva:
When will this be released ? This is not available in v2.2.72, the latest version.

iocanel:
most probably later today

kasundsilva:
Great!

PR #261, opened 2015-12-05
#260: Add additional options to get logs of a pod.

iocanel:


iocanel:
Streaming the logs works in exactly the same way as exec does. You can either pass an output stream or nothing. In the later case the returned handle contains a ref to an input stream which you can use to read logs.

iocanel:
[merge]

khogeland:
@iocanel What's with all the "pretty" logic in this? The `pretty` parameter doesn't do anything for the v1 `/log` operation because it's not JSON formatted.

PR #269, opened 2015-12-17
Fix exec with upgraded okhttp

### LINKED ISSUE #268: exec and writing to stdin doesn't work with okhttp 2.6.0+
We either need to investigate and fix the root cause, or rollback to 2.5.0, as this is a blocker.

jimmidyson:
Fixes #268


jimmidyson:
[merge]

PR #282, opened 2016-01-11
Dependency upgrades

jimmidyson:
@iocanel Any issues with these upgrades?


jimmidyson:
ping @iocanel

jimmidyson:
In the absence of what would, I'm sure, be an awesome review by @iocanel, I'm merging - can always revert later if we need to.

PR #285, opened 2016-01-21
Add withResourceVersion() implementation.

bjornswift:
The `withResourceVersion()` signature was added in 049e1077, but was missing an implementation. This is an attempt to add one, since the old `watch(resourceVersion, watcher)` method was marked deprecated in 6e89a5a8.


iocanel:
ok to test

iocanel:
[merge]

iocanel:
Hi Bjorn and thanks for the contribution!

I think that there are a couple of more places that need to be updated. 

- ReplicationControllerOperationsImpl 
- BuildConfigOperationsImpl

those two have slightly different constructors so the reflection trick won't work there.

I will merge pull request as is, and raise an issue so that we don't forget. You are more than welcome to pick it up. Thanks again!

bjornswift:
Thanks for the quick turnaround!

PR #311, opened 2016-02-20
manually updating to next development version.  I dont know why this …

rawlingsj:
…wasnt automatically done.  It only seem to happen with kubernetes-client, all six other repos have their poms updated correctly.  The CD process will change very soon which will help each project be responsible for its own pipelibe, hopefully that will help


rawlingsj:
[merge]

PR #321, opened 2016-03-07
Adding grace period parameter to delete operations

sarvigalava:


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/GracePeriodConfigurable.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.fabric8.kubernetes.client;
+
+public interface GracePeriodConfigurable<T>
+{
+    T withGracePeriodSeconds(Long gracePeriodSeconds);
```
jimmidyson: Can we just call it `withGracePeriod`?

Such a shame we can't use Java 8's `java.time.Duration`. Why use Long? Is int not big enough? Do you want to be able to pass nulls?

sarvigalava:
The parameter in DeleteOptions model has the type Long, so I reused that. (as well as the name of the parameter 'gracePeriodSeconds')

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationSupport.java
```diff
@@ -173,12 +170,19 @@ public URL getResourceUrl() throws MalformedURLException {
...
-    handleDelete(getResourceUrl(checkNamespace(resource), checkName(resource)));
+  protected <T> void handleDelete(T resource, Long gracePeriodSeconds) throws ExecutionException, InterruptedException, KubernetesClientException, IOException {
+    handleDelete(getResourceUrl(checkNamespace(resource), checkName(resource)), gracePeriodSeconds);
   }
 
-  protected void handleDelete(URL requestUrl) throws ExecutionException, InterruptedException, KubernetesClientException, IOException {
-    Request.Builder requestBuilder = new Request.Builder().delete(null).url(requestUrl);
+  protected void handleDelete(URL requestUrl, Long gracePeriodSeconds) throws ExecutionException, InterruptedException, KubernetesClientException, IOException {
+    RequestBody requestBody = null;
+    if (gracePeriodSeconds >= 0) {
```
jimmidyson: Is 0 a valid grace period value? Guess that's the default meaning delete immediately?

sarvigalava:
A parameter of a negative value means to use the default value, 0 stands for immediately

sarvigalava:
If the requestBody is not provided, kubernetes handles request with gracePeriod = -1

jimmidyson:
Thanks very much for this PR @sarvigalava. There's a problem with the mocks (see circleci build) & would prefer the method to be renamed as suggested, but LGTM as soon as build passes.

@iocanel - if you have time, please can you cast your eye over this in case I missed anything?

sarvigalava:
ok, I will fix the mocks and rename the method

iocanel:
I am looking at it.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/GracePeriodConfigurable.java
jimmidyson: From @sarvigalava:

> The parameter in DeleteOptions model has the type Long, so I reused that. (as well as the name of the parameter 'gracePeriodSeconds')

OK let's leave it as a Long, although I don't really like that tbh - would prefer it to be an int, but generated type is a Long due to source type being go's int64.

Definitely rename though.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationSupport.java
jimmidyson: Possible NPE?

iocanel:
so does grace period make sense for anything else other than pods and replication controllers?

sarvigalava:
I think that it makes sense at lest for pods, rcs, namespaces and daemonSets but all endpoints accept this parameter even it is not used (maybe for future considerations)

sarvigalava:
I have renamed the method and changed the type of the parameter to long.

jimmidyson:
Looking good now.

Circleci tells me that the OpenShift operations also need updating - would you be able to do that too please? They reuse the same Delete semantics.

sarvigalava:
I will take a look at it tomorrow

jimmidyson:
Thanks very much.

sarvigalava:
Done, all tests passed now :)

jimmidyson:
Wow is it tomorrow already? :-b

jimmidyson:
OK to test

jimmidyson:
Awesome! Can you squash commits please? Then good to merge - thanks!

iocanel:
since its globally accepted then I am happy to merge.

sarvigalava:
I have squashed the commits, thanks guys

jimmidyson:
[merge]

jimmidyson:
Thanks @sarvigalava!

PR #322, opened 2016-03-08
fixing the url by removing the namespace

KurtStam:
The users url on kubernetes is 

http://<kubernetes_master//oapi/v1/users, 

but if the namespace is not set to null is results in 

http://<kubernetes_master//oapi/v1/namespace/default/users 

which then fails. Please verify since there are no unit tests to check I'm not breaking something else.


jimmidyson:
Hmm uses should be non-namespaced so this isn't quite the right fix I don't think. Will have a look tomorrow.

jimmidyson:
This looks like it will need to be done for any non-namespaced resource...

KurtStam:
https://github.com/fabric8io/kubernetes-client/issues/326

iocanel:
So, the BaseOperation contains a method called isNamespaceRequired(). The initial intention was to override this method in operation implementations that involve namespaced resources and use it inside getResourceUrl() and getNamespacedUrl().

iocanel:
Thanks for spotting that Kurt. I am closing this and will create a new one with a global fix.

PR #341, opened 2016-03-21
bug fix for @AfterSuite not running after Arquillian integration tests.

yongk802:
Fix @Aftersuite exception after running Arquillian integration Tests. Someone should take a look and see if this overloaded constructor is the best way to fix this.


jstrachan:
ok to test

jstrachan:
[merge]

Issue #344, opened 2016-03-24
Setting User-Agent Header


lwander:
Hey,

Is it possible to set the User-Agent header for requests sent to ones Kubernetes cluster in this API? I'm asking for the purpose of recording usage metrics.

Thanks,
Lars

jimmidyson:
Yeah was thinking of adding that but haven't got around to it. Would be happy for a contribution if you're able - should just be a case of adding option to Config & setting in client create.

lwander:
I'll take a look!

PR #345, opened 2016-03-24
Allow user to specify user-agent header for k8s requests

lwander:
@jimmidyson for #344 

It's not clear to me how to build this project, so this could be totally broken. Is there a getting started guide for developers?

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
@@ -22,6 +22,7 @@ import io.fabric8.kubernetes.api.model.Context;
 import io.fabric8.kubernetes.client.internal.KubeConfigUtils;
 import io.fabric8.kubernetes.client.internal.SSLUtils;
 import io.fabric8.kubernetes.client.utils.Utils;
+import io.fabric8.kubernetes.client.Version;
 import io.sundr.builder.annotations.Buildable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -68,6 +69,8 @@ public class Config {
   public static final String KUBERNETES_ALL_PROXY = "all.proxy";
   public static final String KUBERNETES_NO_PROXY = "no.proxy";
 
+  public static final String KUBERNETES_USER_AGENT = "fabric8-kubernetes-client/" + Version.clientVersion() ;
+
   public static final Long DEFAULT_ROLLING_TIMEOUT = 15 * 60 * 1000L;
   public static final int DEFAULT_LOGGING_INTERVAL = 20 * 1000;
 
@@ -98,6 +101,7 @@ public class Config {
   private String httpProxy;
   private String httpsProxy;
   private String[] noProxy;
+  private String userAgent;
 
   private Map<Integer, String> errorMessages = new HashMap<>();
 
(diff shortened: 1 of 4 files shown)


jimmidyson:
We need to add a contributing/building guide but you should just be able to run `mvn clean install` & that should be it.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
```diff
@@ -68,6 +68,8 @@
   public static final String KUBERNETES_ALL_PROXY = "all.proxy";
   public static final String KUBERNETES_NO_PROXY = "no.proxy";
 
+  public static final String KUBERNETES_USER_AGENT = "fabric8-kubernetes-client";
```
jimmidyson: Wonder if we can compile in the version number somehow? That would help with debugging & issue tracking. Something like fabric8-kubernetes-client/version would be awesome. Wonder if @iocanel has any idea how we can do that?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/HttpClientUtils.java
```diff
@@ -139,6 +139,16 @@ public Response intercept(Chain chain) throws IOException {
                 }
             }
 
+            if (config.getUserAgent() != null && !config.getUserAgent().isEmpty()) {
+                httpClient.interceptors().add(new Interceptor() {
```
jimmidyson: Should this be added as a network interceptor target than a normal interceptor?

jimmidyson:
Could you add this new option to the readme as well?

Thank you for the contribution. Looking forward to merging this when the few nits are sorted.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
iocanel: @jimmidyson: We could have a filtered resource under src/main/resources that would contains ${project.version} this would be substituted during build and could be loaded by Config in a static block.

T

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/HttpClientUtils.java
lwander: I'm not familiar with the difference, this is the only documentation on okHttp interceptors I can find: http://square.github.io/okhttp/3.x/okhttp/

lwander:
Do you guys normally squash commits before merging?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
jimmidyson: That sounds perfect.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/HttpClientUtils.java
jimmidyson: See more at https://github.com/square/okhttp/wiki/Interceptors.

jimmidyson:
I like to review with separate commits (easier to track changes from feedback) & ask for commits to be squashed where appropriate before merge.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
jimmidyson: Don't mind adding the version in subsequent PR though.

jimmidyson: Added version available as `Version.clientVersion()` so can we add the default user agent as `"fabric8-kubernetes-client/" + Version.clientVersion()`.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/HttpClientUtils.java
jimmidyson: Yeah let's add it as a network interceptor please - seen it done in a few places for user agent & seems to be the accepted way. Shame it's not just a simple case of configuring http client but it works...

lwander:
Hey @jimmidyson, I changed the interceptor to a NetworkInterceptor (the link you sent was very helpful in explaining the difference), and I added the version to the UA.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/HttpClientUtils.java
```diff
@@ -139,6 +139,16 @@ public Response intercept(Chain chain) throws IOException {
                 }
             }
 
+            if (config.getUserAgent() != null && !config.getUserAgent().isEmpty()) {
+                httpClient.networkInterceptors().add(new Interceptor() {
+                    @Override
+                    public Response intercept(Chain chain) throws IOException {
+                        Request agent = chain.request().newBuilder().addHeader("User-Agent", config.getUserAgent()).build();
```
jimmidyson: This should just use `header` target than `addHeader` to ensure only single value for the header. Sorry didn't notice that before.

jimmidyson:
OK to test

jimmidyson:
Happy to merge when changed to `header` & commits squashed please. Thank you!

lwander:
All set!

jimmidyson:
LGTM [merge]

Thanks!

lwander:
Thanks for the help!

PR #346, opened 2016-03-25
Compile in client project version

jimmidyson:
@iocanel, @lwander This enables the client version to be compiiled in (no properties files - yay!) available as ``Version.clientVersion()``
that you can use in the default user agent string.


jimmidyson:
[merge]

PR #355, opened 2016-03-31
Upgrade model version to use strings for API versions

### LINKED ISSUE #354: Move apiversion getter to common interface
Currently all resources have a 'getApiVersion' but they method signature is not common to all of them.

jimmidyson:
Also move ConfigMaps to v1 API.

Fixes #354


jimmidyson:
[merge]

PR #366, opened 2016-04-07
Add ReplicaSet support

jimmidyson:


jimmidyson:
[merge]

rhuss:
@jimmidyson cool, thanks. If i understand ReplicaSet right, it the same as ReplicationController except that you are more flexible wrt the selector. Is it already available in OpenShift origin 1.1.x ? 

wonder whether I should use it directly for the plugin or whether still to wait a bit.

jimmidyson:
It's still beta API so I'd wait by default, but in time...

jimmidyson:
& yes it is available in OpenShift, but doubt it's properly integrated with auth etc so probably best to avoid for now.

PR #370, opened 2016-04-08
Fixes #369, no namespace for persistent volumes

### LINKED ISSUE #369: PersistentVolume wrongly requires namespace
@jimmidyson @iocanel 

- Persistent volumes in Kubernetes are not namespace specific. They are not created in a particular namespace.
- However, the client errors out when I don't provide a namespace while creating a persistent volume.
- This was not the case earlier with version 2.2.85
- I'm seeing this issue after upgrading to 2.2.101

Code:
```java
client.persistentVolumes().load(is).create();
```
"is" above is an input stream.

Error:

```
Caused by: com.xxx.xxx.xxx.exception.ResourceException: ResourceException (0): Namespace not specified. But operation requires namespace.
	at com.xxx.xxx.xxx.postPersistentVolume(KubeClient.java:240) ~[na:na]
	at com.xxx.xxx.xxx.postPersistentVolume(KubeClient.java:232) ~[na:na]
```

parikhyash:
@jimmidyson @iocanel PTAL

- Fixed issue where persistent volume creation requires a namespace

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/DefaultKubernetesClient.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/DefaultKubernetesClient.java
@@ -142,8 +142,8 @@ public class DefaultKubernetesClient extends BaseClient implements KubernetesCli
   }
 
   @Override
-  public ClientMixedOperation<PersistentVolume, PersistentVolumeList, DoneablePersistentVolume, ClientResource<PersistentVolume, DoneablePersistentVolume>> persistentVolumes() {
-    return new PersistentVolumeOperationsImpl(httpClient, getConfiguration(), getNamespace());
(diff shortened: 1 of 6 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/DefaultKubernetesClient.java
```diff
@@ -143,7 +143,7 @@ public DefaultKubernetesClient(String masterUrl) throws KubernetesClientExceptio
 
   @Override
   public ClientMixedOperation<PersistentVolume, PersistentVolumeList, DoneablePersistentVolume, ClientResource<PersistentVolume, DoneablePersistentVolume>> persistentVolumes() {
```
iocanel: I think that this should change to ClientNonNamespaceOperation in the DSL.

parikhyash:
@iocanel PTAL

- Changed to ClientNonNamespaceOperation in the DSL and other places as well

jimmidyson:
ok to test

jimmidyson:
LGTM [merge]

Thanks!

PR #372, opened 2016-04-11
Fix field selector query param

### LINKED ISSUE #371: Error when using mulitple field selectors, comma missing from the concatenated query string
If withField() or withFields() is called on an operation with multiple selectors, the query string becomes invalid, e.g. contains
`?fieldSelector=name1=value1name2=value2`
instead of
`?fieldSelector=name1=value1,name2=value2`

The solution seems simple: the if block appending the comma to the StringBuilder in BaseOperation#getFieldQueryParam() should be moved inside the for loop.

jimmidyson:
Fixes #371


jimmidyson:
[merge]

PR #391, opened 2016-04-29
Do not modify the item before the operation succeeds

jwendell:
handleReplace() can fail and generate an exception, then we should
make sure that, in this case, we do not touch the original item.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/HasMetadataOperation.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/HasMetadataOperation.java
@@ -78,8 +78,10 @@ public class HasMetadataOperation<T extends HasMetadata, L extends KubernetesRes
         if (got == null) {
           return null;
         }
+
+        item = handleReplace(item);
         item.getMetadata().setResourceVersion(got.getMetadata().getResourceVersion());
-        return handleReplace(item);
+        return item;
       } catch (KubernetesClientException e) {
         caught = e;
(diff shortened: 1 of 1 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/HasMetadataOperation.java
```diff
@@ -78,8 +78,10 @@ public T replace(T item) {
         if (got == null) {
           return null;
         }
+
+        item = handleReplace(item);
```
jimmidyson: It's important the resource version is updated before sending the request to increase the chances of successful replace. Agree the original item shouldn't be touched. Can we possibly switch to the doneable approach used in edit above, but updating resource version & calling done in this method, visitor actually calling handleReplace? Hope that makes sense!

jwendell: -ETOOMUCHJAVA for me :)

I could study all this stuff and try to make something, but if you have the ability, time and knowledge to make that (100x faster than me BTW), please, go ahead!

jimmidyson: Haha sorry it's badly worded (it's late here!). This is definitely an issue & I'll take a look asap at fixing.

jimmidyson:
Closed in favour of #398. Thank you for pointing out the issue @jwendell.

PR #393, opened 2016-05-03
Support for 'exists' type label selector

gyoetam:
Allow filtering for the presence/lack of a label without specifying a value to check.
Providing a null value parameter to the `withLabel(key, value)` and `withoutLabel(key,value)` methods results in queries like: `?labelSelector=key1,key2,!key3`. Convenience methods `withLabel(key)` and `withoutLabel(key)` are also added to the interface.

Note that to be consistent with the other methods of the class, special values are not checked in the key, making it possible to abuse these methods with calls like `withLabel("key=value,key2")`...

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Filterable.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Filterable.java
@@ -29,7 +29,11 @@ public interface Filterable<T> {
 
   T withLabel(String key, String value);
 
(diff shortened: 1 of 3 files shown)


jimmidyson:
LGTM. Some of the mocks need updating too if you don't mind. Thanks!

gyoetam:
I modified the related BaseMockOperation class, did I miss something else?

jimmidyson:
ok to test

jimmidyson:
LGTM - thanks! [merge]

Issue #394, opened 2016-05-03
DefaultKubernetesClient namespace resolution


luismoramedina:
Hi, we are working on getting services of the current namespace from a pod, we are using the *DefaultKubernetesClient*. This client does not load automatically the namespace because is using a KUBERNETES_NAMESPACE environment variable that is not present on the openshift projects.

We have read about another ways to resolve the namespace:

1. Use /var/run/secrets/kubernetes.io/serviceaccount/namespace file (http://kubernetes.io/docs/user-guide/accessing-the-cluster/#accessing-the-api-from-a-pod). For kubernetes 1.2 and later.

2. The POD_NAMESPACE environment variable, for earlier versions. See the kubernetes client code: https://github.com/kubernetes/kubernetes/blob/master/pkg/client/unversioned/clientcmd/client_config.go#L341-L342

What kubernetes version supports the fabric8 kubernetes-client?

Are you planning to support the way kubernetes 1.2 resolves current namespace (as described in point 1.)? Are you interested in a pull request with this behaviour?

Thank you.

iocanel:
@luismoramedina: the idea is to support the latest version possible and pull request are always welcome :-)

jimmidyson:
Fixed by #395

PR #395, opened 2016-05-04
namespace from path if no environment variable or system property is defined

javiboo:
Recovers the k8s namespace from file /var/run/secrets/kubernetes.io/serviceaccount/namespace, according to the product [documentation](http://kubernetes.io/docs/user-guide/accessing-the-cluster/#accessing-the-api-from-a-pod).

For backwards compatibility, only checks this file if no environment variable or system property is provided.

#394


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
```diff
@@ -59,6 +60,8 @@
   public static final String KUBERNETES_ROLLING_TIMEOUT_SYSTEM_PROPERTY = "kubernetes.rolling.timeout";
   public static final String KUBERNETES_LOGGING_INTERVAL_SYSTEM_PROPERTY = "kubernetes.logging.interval";
 
+  // avoid inlining by calling new String("") in order to change value with reflection for test purposes
+  public static final String KUBERNETES_NAMESPACE_PATH = new String("/var/run/secrets/kubernetes.io/serviceaccount/namespace");
```
jimmidyson: Remove `new String` - it is a String :)

jimmidyson: Ah sorry - now I read the comment above :) Hmm have to think about that.

jimmidyson:
LGTM ok to test

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
```diff
@@ -23,6 +23,7 @@
 import io.fabric8.kubernetes.client.internal.SSLUtils;
 import io.fabric8.kubernetes.client.utils.Utils;
 import io.sundr.builder.annotations.Buildable;
+import org.apache.commons.lang.StringUtils;
```
iocanel: We would like to avoid runtime dependency on commons-lang.

If I am not mistaken, its currently a transitive dependency of sundrio (which is compile time dependency). But no other code should directly use it.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
iocanel: We had a similar issue with the location of ``.kube/config`` the way we address is by allowing those to be overridden via system properties. We should use the same approach here.

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/ConfigTest.java
```diff
@@ -195,6 +209,36 @@ public void testWithKubeConfigAndSytemPropertiesAndBuilder() {
     assertEquals("testns2", config.getNamespace());
   }
 
+  @Test
+  public void testWithNamespacePath() {
```
iocanel: In reality all cases are "with namespace path" but the ones that explicitly specify a namespace via system/env.

This is an other reason why I prefer the default namespace path to be overridden via system/env as its easier to isolate.

iocanel:
Overall it looks great, with some minor points for discussion.

Personally, I don't mind merging it as is, and working on the commented points later on.

iocanel:
@jimmidyson: wdyt?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
jimmidyson: Yeah there's no reason this couldn't be overridden, even if the default is probably going to be used 99.9% of the time.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
jimmidyson: Ooo good point. Can't have this dep - it's caused issues in the past for adoption/integration. It's only one simple method that we can dupe in if there's no other way around it.

jimmidyson:
@iocanel I'd prefer the comments to be fixed up as part of the merge, either by @javiboo (if you have time) or by one of us in separate PR, pulling in @javiboo's commits.

javiboo:
Hi @iocanel & @jimmidyson,

I can fix the issues commented. In short, they consist in:

1. Remove new String in constant `KUBERNETES_NAMESPACE_PATH`
2. Remove runtime dependency to commons-lang 
3. `KUBERNETES_NAMESPACE_PATH` value can be overridden via system properties like the location of `.kube/config`
4. Adapt tests

Anything else?

jimmidyson:
@javiboo SGTM - thanks!

iocanel:
@javiboo: sounds good to me too! :-)

javiboo:
Hi @jimmidyson, can you review the changes in the PR? thanks!

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/ConfigTest.java
```diff
@@ -195,6 +195,66 @@ public void testWithKubeConfigAndSytemPropertiesAndBuilder() {
...
+
+  @Test
+  public void testWithEmptyNamespacePath() {
+    System.setProperty(Config.KUBERNETES_NAMESPACE_FILE, TEST_EMPTYNAMESPACE_FILE);
+    System.setProperty(Config.KUBERNETES_MASTER_SYSTEM_PROPERTY, "http://somehost:80");
+
+    Config config = new Config();
+    assertNotNull(config);
+    assertEquals("http://somehost:80/", config.getMasterUrl());
+    assertEquals("", config.getNamespace());
```
jimmidyson: Is an empty namespace valid?

jimmidyson:
One nit about empty namespace but otherwise LGTM.

ok to test

javiboo:
Hi @jimmidyson, I remove test for empty namespace. Can you check it, please? Thanks!

jimmidyson:
LGTM thanks! [merge]

jimmidyson:
ok to test

javiboo:
@jimmidyson it's necessary anything else to merge the PR? thanks!

jimmidyson:
@javiboo It was supposed to be automerged already - sorry!

jimmidyson:
[merge]

jimmidyson:
Phew worked this time :) Thanks @javiboo!

Issue #396, opened 2016-05-04
Unable to edit Pods when Kubernetes cluster is running on AWS


lwander:
Hey,

When I install a Kubernetes cluster using [these steps](http://kubernetes.io/docs/getting-started-guides/aws/) and the default profile, I'm unable to make edits to pods.

Here is code that breaks (in groovy). It's a degenerate example for the sake of being short.

```
def client = new DefaultKubernetesClient()
client.pods().inNamespace("default").withName("exiting-pod").edit().done() 
```

Originally, this threw the error message:

```Forbidden! User null doesn't have permission.```

which looked promising since the username wasn't populated, but that's simply because the default kubeconfig generated when creating an AWS kubernetes cluster doesn't populate a `users.user.username` field and only relies on the client certs. I switched the default context in the kubeconfig file to use the install-script populated basic auth pair and the error message changed to 

```Forbidden! User admin doesn't have permission.```

What's odd is that this exact code works on when the cluster is running on GKE, all operations on replication controllers work against this cluster on AWS, and I can run 

```kubectl edit pod existing-pod --namespace default```

without running into any permission errors, although it's using the same credentials that fail for that same command with this client library.

I'm using library version `kubernetes-api:2.2.102`.

lwander:
The patch I submitted addresses the `null` in the first 403 message. (Unfortunately doesn't fix the issue).

iocanel:
@lwander: It sounded extremely weird that it was working using ``kubectl`` but not by this client, so I followed your steps and managed to reproduce that myself.

After bumping my head against the wall for a while, I realised that kubectl edit is using the "PATCH" method, while the kubernetes-client is using the "PUT" method. 

Now, if we want to also use patch in the client I could see 2 possible approaches here:

1. Do a get, edit, compare and patch.
2. Don't get and just patch.

(1): is what we currently do, but without the compare part, which I am not sure if its something we can easily do or not.

(2): is having the edit command use a blank object so that we can create a patch out of it. This is also tricky.

@jimmidyson: Thoughts.

jimmidyson:
@iocanel So PUT doesn't work but PATCH does? That seems a bit weird to me. Documentation on PATCH semantics are https://github.com/kubernetes/kubernetes/blob/master/docs/devel/api-conventions.md#patch-operations. From those docs PUT is only supported when sending through the whole resource. Can you see if the whole resource is actually being sent through?

iocanel:
[...]

PR #397, opened 2016-05-04
Correctly show incorrect API access by username

lwander:
The `username` field in `Config` isn't necessarily set, since it's only used by basic auth. Instead, `user` in `Context` is the name of the user in kubeconfig with the full authentication details specified. This makes the 403 more understandable.

This fixes one issue encountered in: https://github.com/fabric8io/kubernetes-client/issues/396

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
@@ -270,7 +270,7 @@ public class Config {
               config.setPassword(currentAuthInfo.getPassword());
 
               config.getErrorMessages().put(401, "Unauthorized! Token may have expired! Please log-in again.");
-              config.getErrorMessages().put(403, "Forbidden! User "+config.getUsername()+ " doesn't have permission.");
+              config.getErrorMessages().put(403, "Forbidden! User "+currentContext.getUser()+ " doesn't have permission.");
             }
             return true;
           }


iocanel:
ok to test

iocanel:
[merge]

iocanel:
Thanks Lars!

I'll follow up in the #396.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
```diff
@@ -270,7 +270,7 @@ private boolean tryKubeConfig(Config config) {
               config.setPassword(currentAuthInfo.getPassword());
 
               config.getErrorMessages().put(401, "Unauthorized! Token may have expired! Please log-in again.");
-              config.getErrorMessages().put(403, "Forbidden! User "+config.getUsername()+ " doesn't have permission.");
+              config.getErrorMessages().put(403, "Forbidden! User "+currentContext.getUser()+ " doesn't have permission.");
```
jimmidyson: From https://github.com/kubernetes/kubernetes/blob/f4beccf000c706fc66644422062e91391576da8a/pkg/client/unversioned/clientcmd/api/types.go#L109-L110 `currentContext.getUser()` should return the `name of the authInfo for this context`, not the username:

> // AuthInfo is the name of the authInfo for this context
	AuthInfo string `json:"user"`

The username is embedded in the `AuthInfo` object that this references - if there is no matching `AuthInfo` object then there is something wrong with the kube config file. I would prefer to log that fact & keep the username logic the same as it currently is.

jimmidyson:
Re https://github.com/fabric8io/kubernetes-client/pull/397#discussion_r62301607 I think this should be reverted as I don't think it doesn't reflects the correct username that is being used for auth. I think it is actually hiding a problem in the kube config file.

iocanel:
@jimmidyson: I don't follow....

I am not sure what the types that the client is using are, but I do see that our classes and how they do map to the ~/.kube/config and I don't see a problem. Care to provide some more info?

jimmidyson:
`currentContext.getAuthInfo()` returns a String which references an `AuthInfo` object. To provide an example:

A context:

```
- context:                                                                      
    cluster: localhost:8443                                                     
    namespace: demo                                                             
    user: admin/localhost:8443                                                  
  name: demo/localhost:8443/admin
```

`context.getUser()` returns `admin/localhost:8443`. That is not a username, but the name of an `AuthInfo` object. In my kube config file this referenced `AuthInfo` object looks like:

```
- name: admin/localhost:8443                                                                  
  user:                                                                         
    token: SSSSHHH-THIS-IS-SECRET
```

Notice that has no username - this is using an oAuth token. However, the username _can_ be in the `AuthInfo` object - here's the definition upstream:

```
// AuthInfo contains information that describes identity information.  This is use to tell the kubernetes cluster who you are.
type AuthInfo struct {
	// LocationOfOrigin indicates where this object came from.  It is used for round tripping config post-merge, but never serialized.
	LocationOfOrigin string
	// ClientCertificate is the path to a client cert file for TLS.
	ClientCertificate string `json:"client-certificate,omitempty"`
	// ClientCertificateData contains PEM-encoded data from a client cert file for TLS. Overrides ClientCertificate
	ClientCertificateData []byte `json:"client-certificate-data,omitempty"`
	// ClientKey is the path to a client key file for TLS.
	ClientKey string `json:"client-key,omitempty"`
	// ClientKeyData contains PEM-encoded data from a client key file for TLS. Overrides ClientKey
	ClientKeyData []byte `json:"client-key-data,omitempty"`
	// Token is the bearer token for authentication to the kubernetes cluster.
	Token string `json:"token,omitempty"`
	// Impersonate is the username to act-as.
	Impersonate string `json:"act-as,omitempty"`
	// Username is the username for basic authentication to the kubernetes cluster.
	Username string `json:"username,omitempty"`
	// Password is the password for basic authentication to the kubernetes cluster.
	Password string `json:"password,omitempty"`
	// AuthProvider specifies a custom authentication plugin for the kubernetes cluster.
	AuthProvider *AuthProviderConfig `json:"auth-provider,omitempty"`
	// Extensions holds additional information. This is useful for extenders so that reads and writes don't clobber unknown fields
	Extensions map[string]runtime.Object `json:"extensions,omitempty"`
}
```

So I guess we need to actually handle the auth error message better: username is only available for basic auth connections. But `context.getUser()` never returns an actual username in any case.

lwander:
I figured this was at the very least more helpful as an error message, since `Forbidden! User null doesn't have permission.` seems to hint at a deeper problem than the error message is surfacing.

jimmidyson:
Agreed - wonder if we should put more context in the message to say what kind of auth was attempted, masking sensitive info I guess, but enough info to be able to better diagnose the issue?

lwander:
That sounds best to me

PR #398, opened 2016-05-06
Ensure replace doesn't modify the original resource

jimmidyson:
@iocanel Can you think of a simpler way to do this? We don't know the builder type at this point sadly.

@jwendell This supercedes #391.


jimmidyson:
@iocanel Any comments?

iocanel:
@jimmidyson: I like it. The only other feasible solution is to add some generics-porn :-) 

So let's merge it.

jimmidyson:
[merge]

PR #402, opened 2016-05-18
Make jobs, deployments & replica sets scaleable, and replica sets also support rolling updates

jimmidyson:


jimmidyson:
[merge]

PR #409, opened 2016-05-23
#408: The OkHttpClient adapter should a clone of the client of the sp…

### LINKED ISSUE #408: Unable to use `.extensions()` when multiple clients are configured with SSL
When multiple clients are registered, each adaptor is registered [here](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Adapters.java#L43). Unfortunately, it's registering the adaptor by [class type](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/BaseClient.java#L60), which means if multiple adaptors with the same class (i.e. multiple clients exist) are added, the old ones are overwritten. 

My inclination is to register the adaptor with a string combining details about the cluster with the adaptor class to facilitate this. @iocanel I believe you wrote that adaptor code, WDYT?

iocanel:
…ecified instance.

Fixes #408

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/BaseClient.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/BaseClient.java
@@ -25,7 +25,7 @@ import io.fabric8.kubernetes.client.utils.Utils;
 import java.net.URL;
 
 
-public abstract class BaseClient implements Client {
+public abstract class BaseClient implements Client, HttpClientAware {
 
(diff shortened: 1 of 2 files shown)


lwander:
Thanks for the fix!

iocanel:
[merge]

iocanel:
@lwander: I should be thanking you for the fix :-)

PR #411, opened 2016-05-24
Fix list deletes for cascadeable resources

### LINKED ISSUE #410: Deleting deployment does  not delete associated pods
I am trying to delete deployment and its associated pods using following code 

`kubernetesClient
                .extensions()
                .deployments()
                .withLabels(labels).delete();`

It doesn't delete pods associated with it . Is there a alternative to delete pods too ?

where as when I delete using kubectl it deletes pods as well.

jimmidyson:
Fixes #410


jimmidyson:
[merge]

PR #414, opened 2016-05-25
Disable read timeout for log watching

### LINKED ISSUE #368: Log watch streaming does not work as expected
When a client reads log using LogWatch then the reading is blocked and never resumes if the client reads the data faster than the log file generates them.

The issue is probably hidden in InputStreamPumper line

while (keepReading && !Thread.currentThread().isInterrupted() && (length = in.read(buffer)) != -1)

when the client is at the end of the log then  in.read(buffer) probably returns -1 which causes the cycle ended and method terminated. Thus the reading from pod log is stopped.

When I have monitored the behaviour with debugger it is clearly visible that the executor thread process run() method of the pumper and suddenly it waits for new task that never arrives.

jimmidyson:
Fixes #368


jimmidyson:
[merge]

Issue #418, opened 2016-05-25
scaling down a deployment resource to 0 never returns


ravilr:
@jimmidyson 

looks like there is some issue with scaling down a deployment to 0.  the underlying Replicaset resource itself is scaled down to 0 and the pods are getting deleted. but, it waits on waitUntilDeploymentIsScaled forever..

at java.util.concurrent.CountDownLatch.await(CountDownLatch.java:231)
at io.fabric8.kubernetes.client.dsl.internal.DeploymentOperationsImpl.waitUntilDeploymentIsScaled(DeploymentOperationsImpl.java:92)
at io.fabric8.kubernetes.client.dsl.internal.DeploymentOperationsImpl.scale(DeploymentOperationsImpl.java:62)
at io.fabric8.kubernetes.client.dsl.internal.DeploymentOperationsImpl.scale(DeploymentOperationsImpl.java:39)

ravilr:
It seems the deployment controller manager doesn't write all the fields in deploymentStatus object, after scaling down to 0. i see only observedGeneration, but no replicas/updatedReplicas/availableReplicas fields.

stripped down output of `kubectl get deployment -o yaml` after scale down to 0.
```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  .......
  generation: 8
......
spec:
  minReadySeconds: 30
  replicas: 0
.........
status:
  observedGeneration: 8
```

jimmidyson:
I thought I'd handled that properly in https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/DeploymentOperationsImpl.java#L80.

ravilr:
@jimmidyson Thanks again :)  the above change went in yesterday.  i'm still running release 1.3.92. will upgrade once there is a new release available on maven central.

jimmidyson:
Ah thanks for confirming - couldn't figure out what I'd broken :)

ravilr:
can we add a timeout to the countdownlatch await. same as rollingUpdater?

jimmidyson:
PR would be welcome - not sure we'll have time to get to that soon.

ravilr:
@jimmidyson  Will try to send a pr adding a scaleTimeout to all ClientScalableResources. 

Also, was looking at kubectl implementation of deleting a deployment resource:
[...]

PR #423, opened 2016-05-26
Fix deployment reaping & upgrade model

jimmidyson:
See https://github.com/fabric8io/kubernetes-client/issues/418#issuecomment-221744856

@ravilr This should fix the deployment reaping to be consistent with kubectl.


jimmidyson:
[merge]

Issue #432, opened 2016-05-27
fabric8 client API version 1.2.2  order of create rep controller / create watch didn't matter. in 1.3.91  it does matter.


buildlackey:
Hi.

i filed an issue on stack overflow here -> 
http://stackoverflow.com/questions/37475114/does-fabric8-and-kubernetes-1-2-4-not-support-watches-on-replication-controllers

which illustrates the steps to reproduce this issue, using a small scala program.   My colleague Vijay Samuel (mad respect to him)  helped me find a solution to the problem, which is expressed via a working java program in that same stack overflow post. 

Initially we thought the key tweak to making things work was to upgrade from 1.3.x of client API to 1.4-SNAPSHOT, but it turns out that this is immaterial.  The main thing one must do is to make sure that watches are created before the object that is being watched is created.

For example, in Vijay's if you switched the order of these statements

```java
client.replicationControllers().inNamespace(namespace).withLabel("l", "v").watch(watcher);
createRc(client, namespace, podName, image);
```

to this:

```java
createRc(client, namespace, podName, image);
client.replicationControllers().inNamespace(namespace).withLabel("l", "v").watch(watcher);
```

the program would cease to work.      Switching the order would have been fine in 1.2.2 as far as i can tell from the testing i have done.

**Vijay's working example**

```java
import com.fasterxml.jackson.databind.ObjectMapper;
import com.typesafe.scalalogging.StrictLogging;
import io.fabric8.kubernetes.api.model.Quantity;
import io.fabric8.kubernetes.api.model.ReplicationController;
import io.fabric8.kubernetes.client.Watcher.Action;
import io.fabric8.kubernetes.client.*;

import java.util.HashMap;
import java.util.Map;


public class Vijay {


  public static DefaultKubernetesClient getConnection ()  {
    ConfigBuilder
         configBuilder =   Config.builder() ;
    Config config =
      configBuilder.
        withMasterUrl("http://localhost:8080").
        build();
    return new DefaultKubernetesClient(config);
  }


  public static void main(String[] args) throws Exception {
    DefaultKubernetesClient client = getConnection();

    String namespace = "junk6";
    String podName = "prom";
    String image = "nginx";

    Watcher<ReplicationController> watcher = new Watcher<ReplicationController>() {

      @Override
      public void onClose(KubernetesClientException cause) {
        // TODO Auto-generated method stub

      }

      @Override
      public void eventReceived(Action action, ReplicationController resource) {
[...]

PR #433, opened 2016-05-27
Dont send fieldSelector unless theres actually a field to select on...

jimmidyson:
Found while helping out with #432


jimmidyson:
[merge]

PR #439, opened 2016-06-06
Add patch support

jimmidyson:
This change adds `Patchable.patch()` to all resources & also switches `edit()`
to call `patch` instead of `replace`.

@iocanel This could cause problems for users so if I'm not about & it does cause
issues we should probably revert the `edit()` change.


jimmidyson:
[merge]

PR #440, opened 2016-06-06
Add errorReceived callback for watches, fix reconnection, handle outdated resource versions requests

jimmidyson:


jimmidyson:
[merge]

PR #445, opened 2016-06-11
Delete job always taking "scaleTimeout" time to delete a job

salilwagh:
countDownLatch.countDown() is never getting called.
job.getStatus().getActive() is returning null if job is scalled down to 0
The reason here is that the api server does not send "active" field inside the response JSON.
Hence jackson deserializes it to null. Only if the job is active is the active field set to some int.
Hence delete job always takes getConfig().getScaleTimeout() amount of time to terminate.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/JobOperationsImpl.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/JobOperationsImpl.java
@@ -94,7 +94,11 @@ public class JobOperationsImpl extends HasMetadataOperation<Job, JobList, Doneab
       public void run() {
         Job job = getMandatory();
         atomicJob.set(job);
-        if (Objects.equals(job.getSpec().getParallelism(), job.getStatus().getActive())) {
+        Integer activeJobs  = job.getStatus().getActive();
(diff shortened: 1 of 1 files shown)


salilwagh:
A request to admins to verify the fix. Thanks!

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/JobOperationsImpl.java
```diff
@@ -94,6 +94,8 @@ private void waitUntilJobIsScaled() {
       public void run() {
         Job job = getMandatory();
         atomicJob.set(job);
+        if (job.getStatus().getActive() == null)
```
jimmidyson: I'd rather not mutate the job object here - can you extract `getActive` to an Integer variable & zero it if null? Then use that in the comparison below?

And please use curlies around blocks please.

jimmidyson:
Sorry for the delay! Looks good, with just some small stylistic comments to update.

jimmidyson:
Thanks!

ok to test

jimmidyson:
[merge] Thanks for the PR @salilwagh!

salilwagh:
@jimmidyson no problem at all. Thanks!

PR #452, opened 2016-06-29
Componentstatus is not namespaced. Removing namespace from query

vjsamuel:
APIResourceList shows that componentstatuses endpoint is not namespaced and hence queries to the componentstatuses is failing.

```
{
  "kind": "APIResourceList",
  "groupVersion": "v1",
  "resources": [
    {
      "name": "bindings",
      "namespaced": true,
      "kind": "Binding"
    },
    {
      "name": "componentstatuses",
      "namespaced": false,
      "kind": "ComponentStatus"
    },
    
....
```

Hence removing namespace support for componentstatuses query.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/DefaultKubernetesClient.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/DefaultKubernetesClient.java
@@ -113,7 +113,7 @@ public class DefaultKubernetesClient extends BaseClient implements NamespacedKub
 
   @Override
   public ClientMixedOperation<ComponentStatus, ComponentStatusList, DoneableComponentStatus, ClientResource<ComponentStatus, DoneableComponentStatus>> componentstatuses() {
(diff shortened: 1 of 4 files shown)


davsclaus:
Thanks for the PR. 

I think @iocanel and @jimmidyson should review this as they are working on this project the most.

jimmidyson:
Thanks for the PR.

You'll need to also add in the same as https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/NodeOperationsImpl.java#L39-L42 to indicate this isn't a namespaced operation.

There are some test failures too - can you please update those & make sure that `mvn clean install` works locally?

vjsamuel:
@jimmidyson apologies for not noticing the failed test cases. I have fixed them, it caught a bug in my code :)

jimmidyson:
LGTM ok to test

jimmidyson:
[merge]

jimmidyson:
Thanks for the fix!

PR #454, opened 2016-06-30
Fix build config reaping

jimmidyson:
@bparees Hope that looks a bit better! Shame there's no way to filter on Build's `status.config.name` & `namespace` - would be safer than relying on labels for this I guess.

### DIFF
--- a/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/BuildConfigOperationsImpl.java
+++ b/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/BuildConfigOperationsImpl.java
@@ -31,6 +31,7 @@ import io.fabric8.kubernetes.client.utils.URLUtils;
 import io.fabric8.openshift.api.model.Build;
 import io.fabric8.openshift.api.model.BuildConfig;
 import io.fabric8.openshift.api.model.BuildConfigList;
+import io.fabric8.openshift.api.model.BuildList;
 import io.fabric8.openshift.api.model.BuildRequest;
 import io.fabric8.openshift.api.model.DoneableBuildConfig;
 import io.fabric8.openshift.api.model.WebHookTrigger;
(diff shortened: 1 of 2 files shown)


openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/BuildConfigOperationsImpl.java
```diff
@@ -135,7 +135,7 @@ public Void trigger(WebHookTrigger trigger) {
   }
 
   private void deleteBuilds() {
-    new BuildOperationsImpl(client, (OpenShiftConfig) config, namespace).withLabel("buildconfig", "edam").delete();
+    new BuildOperationsImpl(client, (OpenShiftConfig) config, namespace).withLabel("buildconfig", getName()).delete();
```
bparees: so buildconfig is actually our deprecated label at this point.  Unfortunately we have a bit of a mess:
https://github.com/openshift/origin/blob/master/pkg/build/api/types.go#L15
https://github.com/openshift/origin/blob/master/pkg/build/api/types.go#L39-L45

The issue is that labels are limited to 63 characters.  So we actually have to truncate the buildconfig name to put it in the label, if it's 64+ chars.  As a result, we now do the following:
1) always add the annotation (no limit on length) 
2) truncate the name if needed and add it as a label (using both the current, and deprecated, label keys).

So to *find* the builds associated w/ a buildconfig you really need to:
1) check if the buildconfig name is > 63 chars.  if not, you can get all the builds using both labels (since some builds may only have the old deprecated label) and merge the set.
2) if it is longer than 63 chars, you really need to get builds using the truncated label value (again getting using both the old and new label), then filter the results using the annotation to make sure what you got back actually matches your buildconfig name and not a different buildconfig that shares the same first 63 chars.

@jwforres just went through this w/ the web console, so she can double check my logic.

(sorry)
(i feel your pain)

to be fair you can probably ignore the deprecated label since there's no way someone will have a pipeline build that only has the deprecated label applied.

jimmidyson: Thank you @bparees - I'll fix this up tomorrow.

jimmidyson:
@bparees Updated as per your logic - would appreciate a review if you have time.

bparees:
lgtm, thanks @jimmidyson!

jimmidyson:
Thanks @bparees!

[merge]

PR #483, opened 2016-08-02
Force TLS 1.2 even on Java 7

jimmidyson:
@danushkaf This is the PR with the fixes. The problem is that in Java 7 the default SSL context is TLSv1.0 only, in Java 8 it is TLSv1.2. This PR forces TLSv1.2 SSL context for both versions. See https://github.com/fabric8io/kubernetes-client/compare/v1.3.x...jimmidyson:tls-version?expand=1#diff-be9cbbacc6126b5bebe96713829dc752R101 & https://github.com/fabric8io/kubernetes-client/compare/v1.3.x...jimmidyson:tls-version?expand=1#diff-b049f296ce77937d1b28c472e0707370R77.

This will be in the 1.3.x release stream, but fixes will also be included in 1.4.x with the upgraded okhttp3. If you could try it out & confirm that it fixes the issue for you, then I will get it merged & a 1.3 release out ASAP.


jimmidyson:
retest this please

danushkaf:
Will try first thing tomorrow morning.

jimmidyson:
OK. I'm happy with the change so I'm going to actually get a release of this out - we can always investigate more if it doesn't work for you, but of course I'm sure it will :)

jimmidyson:
[merge]

PR #493, opened 2016-08-12
Enable editing of SCC

### LINKED ISSUE #491: Add support for editing Security Context Constrains
Current impl doesn't override `edit()` method so default one from `BaseOperation` is used - `KubernetesClientException: Cannot edit read-only resources` is thrown.

I'll happily submit a PR, but don't know where to start.
I believe that the key part of implementing the edit() method is to create `DoneableSecurityContextConstraints` object, but so far failing to do so.

Thx a lot.

https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/SecurityContextConstraintsOperationsImpl.java

jimmidyson:
Fixes #491


jimmidyson:
[merge]

PR #501, opened 2016-08-25
Configure writeTimeout for buildconfigs

jamesnetherton:


openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/BuildConfigOperationsImpl.java
```diff
@@ -235,7 +235,10 @@ public void writeTo(BufferedSink sink) throws IOException {
         }
       };
 
-      OkHttpClient newClient = client.newBuilder().readTimeout(timeout, timeoutUnit).build();
+      OkHttpClient newClient = client.newBuilder()
+        .readTimeout(timeout, timeoutUnit)
+        .writeTimeout(timeout, timeoutUnit)
```
jimmidyson: Can we make this independently configurable?

jamesnetherton: Yes, can do. I'll take a look.

jamesnetherton: Struggling to find a clean way to implement this. How do we want the the BuildConfigs DSL to look?

At the moment you'd do something like:

```
buildConfigs()
  .withName("foo")
  .instantiateBinary()
  .withTimeout(60, TimeUnit.SECONDS)
```

Any ideas on the best way to introduce the writeTimeout value?

Simplest way would probably be to extend withTimeout (for this builder) to include a writeTimeout arg, so you're always forced to specify both a read & write timeout value.

jimmidyson: It may be that I messed up the `withTimeout` DSL - IIRC `instantiateBinary` should be terminal.

@iocanel Do you have some time to look at this? You are Mr DSL after all...

iocanel:
@jimmidyson: Sorry, I didn't see that, before @jamesnetherton pointed me at it!

Will have a look asap!

iocanel:
The client DSL, uses some cross cutting methods (that apply to all kinds of operations) and some resource specific ones. Usually the resource speClientBucific ones are implemented at the resource level (e.g. ClientPodResource, ClientScallableResource and so on). 

So, I think that we just need to update the BuildConfigResource and make it possible to add with timeout before that instantiate thing. With a first glance this could be as simple as adding Timeoutable<Instantiateable<BuildRequest, N>> to the list of interfaces the resource extends and then align the implementations accordingly.

iocanel:
Forget my previous comment. I was intially confused by @jimmidyson: comment (so I blame Jimmy!).

It seems to me that instantiateBinary() is not a terminal method and the way its currently implemented is ok.

      Build build = client.buildConfigs().inNamespace("ns1").withName("bc2").instantiateBinary()
        .withCommitterName("committer name")
        .withCommitterEmail("committer@someorg.com")
        .withAuthorName("author name")
        .withAuthorEmail("author@someorg.com")  
        .withTimeout(5, TimeUnit.SECONDS)

        .fromInputStream(dummy);

jimmidyson:
I am to blame... Sorry! I got mixed up thinking instantiateBinary was terminal (in my defence, it sounds terminal to me...).

jimmidyson:
So let's circle  back to the original request I made:

> Can we make [read & write timeouts] independently configurable?

Now I'm questioning whether we actually need to... Thoughts @iocanel @jamesnetherton?

jamesnetherton:
My initial thoughts are probably not. Would folks need control over both the HTTP client read / write timeout values? I'm guessing probably not.

Setting the them both to the same value seems a fair compromise (if somewhat lazy).

jimmidyson:
I'm OK with that.

ok to test

jimmidyson:
add to whitelist

iocanel:
I am not 100% sure if this is accurate, but I think that not even all http clients themselves expose read timeout and write timeout differently, so yeah it does seem overkill to me.

jimmidyson:
BTW @jamesnetherton what problem did this cause for you? Do you know what the default writeTimeout was set to?

jamesnetherton:
@jimmidyson I was testing the f-m-p with S2I binary builds and started seeing timeouts on the first build run. That led me to https://issues.jboss.org/browse/OSFUSE-258 and a realisation that the client was timing out on write, while it waited for OpenShift to pull the S2I image.

Think we were just using the default OkHttpClient default read / write timeouts. Which IIRC is 10 seconds.

jimmidyson:
@jamesnetherton AHA! I'd seen that too but didn't think of it - nice catch! Thanks!

LGTM [merge]

PR #509, opened 2016-09-15
Expose cascading to external resources.

iocanel:


jimmidyson:
Bit more description of what you're trying to do would be good :)

PR #514, opened 2016-09-19
Misc

iocanel:
This pull request adds:

 Cascading support for external resource. (e.g. client.resource(myresource).cascading(true).delete().

Fallback to items api version if one is not specified. Currently we always fallback to v1 which is problematic when we use extensions. So if an existing object is specified, we can fallback to its version.


jimmidyson:
[merge]

Issue #515, opened 2016-09-20
Split apiGroup and apiVersion


iocanel:
The client internally uses the notion of apiGroup and apiVersion. When fed with externally created objects both group and version will be part of the apiGroup field, so the client needs split them and read the correct version.

PR #516, opened 2016-09-20
#515: Fix issues with mixed apiGroup and apiVersion.

iocanel:


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
```diff
@@ -124,13 +124,29 @@ protected BaseOperation(OkHttpClient client, Config config, String apiGroup, Str
...
 
+  /**
+   * Separates apiVersion for apiGroup/apiVersion combination.
+   * @param apiVersion  The apiVersion or apiGroup/apiVersion combo.
+   * @return            Just the apiVersion part without the apiGroup.
+     */
+  private static String trimVersion(String apiVersion) {
+    if (apiVersion == null) {
+      return null;
+    } else if (apiVersion.contains("/")) {
```
jimmidyson: You can always split regardless of whether it contains a `/`.

jimmidyson:
Why close this?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
iocanel: If you don't check for contains("/") you'd need to check the length of the array afterwards.
So complexity wise its the same thing. Now performance wise, in the worst case (no / is found) both approach should behave the same. In all other cases its possibly cheaper to check the length of an array compare to check contains. And thank you so much for spotting it :-)

iocanel:
Because, I the same commit is included in an other PR

iocanel:
More specifically: #519

PR #517, opened 2016-09-21
Use periodic websocket pings to check that server connection is still valid

jimmidyson:
This handles reconnections is the server crashes & TCP connections are not properly closed.

@iocanel Would you mind reviewing please?

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
@@ -16,14 +16,6 @@
 package io.fabric8.kubernetes.client.dsl.internal;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import okhttp3.HttpUrl;
-import okhttp3.OkHttpClient;
-import okhttp3.Request;
-import okhttp3.Response;
-import okhttp3.ResponseBody;
-import okhttp3.ws.WebSocket;
-import okhttp3.ws.WebSocketCall;
-import okhttp3.ws.WebSocketListener;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.KubernetesResourceList;
 import io.fabric8.kubernetes.api.model.Status;
@@ -33,6 +25,14 @@ import io.fabric8.kubernetes.client.Watch;
 import io.fabric8.kubernetes.client.Watcher;
 import io.fabric8.kubernetes.client.dsl.base.BaseOperation;
 import io.fabric8.kubernetes.client.dsl.base.OperationSupport;
+import okhttp3.HttpUrl;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+import okhttp3.ResponseBody;
+import okhttp3.ws.WebSocket;
+import okhttp3.ws.WebSocketCall;
+import okhttp3.ws.WebSocketListener;
 import okio.Buffer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -43,8 +43,9 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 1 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
```diff
@@ -124,12 +128,26 @@ private final void runWatch() throws MalformedURLException, ExecutionException,
...
         queue.add(true);
+        pingFuture = pingExecutor.scheduleAtFixedRate(
+          new Runnable() {
+            @Override
+            public void run() {
+              try {
+                webSocket.sendPing(new Buffer().writeUtf8("Alive?"));
+              } catch (IOException e) {
+                logger.error("Failed to send ping", e);
+                onFailure(e, null);
```
iocanel: I always thought that the onFailure() was referring to the upgrade operation and was not expected after a successful upgrade. Still the javadoc doesn't verify my assumption. 

Need to be careful, if there is code written using this assumption.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
```diff
@@ -124,12 +128,26 @@ private final void runWatch() throws MalformedURLException, ExecutionException,
       private final Logger logger = LoggerFactory.getLogger(this.getClass());
 
       @Override
-      public void onOpen(WebSocket webSocket, Response response) {
+      public void onOpen(final WebSocket webSocket, Response response) {
         logger.debug("WebSocket successfully opened");
         webSocketRef.set(webSocket);
         currentReconnectAttempt.set(0);
         started.set(true);
+        queue.clear();
```
iocanel: why do we clear the queue? Do we expect onOpen() to be called multiple times?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
```diff
@@ -139,11 +157,13 @@ public void onFailure(IOException e, Response response) {
           Status status = OperationSupport.createStatus(response);
           logger.error("Exec Failure: HTTP:" + status.getCode() + ". Message:" + status.getMessage(), e);
           if (!started.get()) {
+            queue.clear();
```
iocanel: This is an example of code (written by me) that assumes that onFailure() can only occur during open.

Still I am not sure if it makes sense to clear the queue. After all the queue is used like a CountDownLatch(1) that you can pass an exception to the guy that waits.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
```diff
@@ -64,11 +66,13 @@
   private final int reconnectInterval;
   private final AtomicInteger currentReconnectAttempt = new AtomicInteger(0);
   private final AtomicReference<WebSocket> webSocketRef = new AtomicReference<>();
-  private final ExecutorService executor = Executors.newSingleThreadExecutor();
+  private final ExecutorService reconnectExecutor = Executors.newSingleThreadExecutor();
```
iocanel: Why are we using two executors? A single ScheduledExecutorService thread pool with 2 threads would do. And most probably it could work with a single thread too, given that reconnect and ping are not supposed to happen in parallel (but that may be too much).

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
jimmidyson: It was being... it was a bit weird & this worked :)

jimmidyson: Though I lazily didn't dig deeper.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
jimmidyson: Hmm yeah. Let me revisit it & maybe think this time.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
jimmidyson: Good idea. Changing to single executor with 2 threads.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
jimmidyson: Good point. It wasn't doing anything anyway so switched to onClose.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
jimmidyson: It doesn't make sense.

jimmidyson:
Updated - care to look over once more & merge if you're happy. Want to get a release of this out to fix Jenkins sync plugin issues.

PR #520, opened 2016-09-23
Upgrade to kube model 1.0.54

jimmidyson:


jimmidyson:
Self-merging as only upgrading model version [merge]

PR #522, opened 2016-09-26
fix for re-assign trustCerts variable

chy168:
fix for re-assign trustCerts variable in Config()


jimmidyson:
Ah I see it's assigned twice - thanks for the update! As the current functionality doesn't cause any problems, I won't cut a release with this fix - it will just be included when we cut the next release.

ok to test

jimmidyson:
[merge]

jimmidyson:
Thanks @chy168!

Issue #523, opened 2016-09-26
have an .apply() method that wraps up create() or replace() based on if the resource already exists


jstrachan:
we've got apply when using the 

```java
kubernetesClient.resource(foo).apply()
```

but we don't have an apply when using the DSL to find a resource:

```java
kubernetes.extensions().deployments().inNamespace("foo").withName("bar").apply(resource);
```
it'd be nice to be able to apply stuff generically

iocanel:
Now both styles support the following methods:

createOrReplace() and createOrReplaceNew() for inlining.

apply() from resource has been deprecated.

PR #538, opened 2016-10-06
#523: Deprecate apply() in favor of createOrReplace(). Add createOrRe…

iocanel:
…place() and createOrReplaceWithNew() in all operations.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
@@ -302,6 +302,36 @@ public class BaseOperation<T, L extends KubernetesResourceList, D extends Doneab
     }
   }
 
+
+  @Override
+  public D createOrReplaceWithNew() throws KubernetesClientException {
+    final Function<T, T> visitor = new Function<T, T>() {
+      @Override
+      public T apply(T resource) {
+        try {
+          return createOrReplace(resource);
+        } catch (Exception e) {
+          throw KubernetesClientException.launderThrowable(e);
+        }
+      }
+    };
+
+    try {
+      return getDoneableType().getDeclaredConstructor(Function.class).newInstance(visitor);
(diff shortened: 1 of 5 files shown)


iocanel:
[merge]

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
```diff
@@ -302,6 +302,36 @@ public T apply(T resource) {
...
+    try {
+      return getDoneableType().getDeclaredConstructor(Function.class).newInstance(visitor);
+    } catch (InvocationTargetException | NoSuchMethodException | IllegalAccessException | InstantiationException e) {
+      throw KubernetesClientException.launderThrowable(e);
+    }
+  }
+
+  @Override
+  public T createOrReplace(T item) {
+    if (get() == null) {
```
sayedasadali: @iocanel question regarding `resource().createOrReplace(resourceObj);`

the `get()` will return a `null` value if and only if none of the namespaces already contain the `Resource`?
which means that the `create` workflow will only be invoked during the first call to create a `Resource` in any `namespace`. All subsequent calls to create a similar `Resource` in a different `namespace` will invoke the `replace` workflow and subsequently fail because the `Resource` doesn't exist?

example scenario: ( will succeed only if there are no ResourceQuotas in any of the namespaces. )
`client.ResourceQuotas().createOrReplace(resourceQuota); `

this attempts a replace as well: (the get() returns the loaded object from context and fails with PUT on non-existent resource)
`client.resourceQuotas().inNamespace(namespace).load(resourceQuotaInputStream).createOrReplace(resourceQuota);`

is there a way to support the following operation?
`client.ResourceQuotas().inNamespace(namespace).createOrReplace(resourceQuota);`

iocanel: @sayedasadali: It was intended to support `client.ResourceQuotas().inNamespace(namespace).createOrReplace(resourceQuota)`, If not, then its probably a bug.

PR #541, opened 2016-10-07
539

iocanel:


iocanel:
@jimmidyson: I want your view on this. From what I understand the problem is okhttp doesn't prevent any way of removing an interceptor. So the original interceptor always hides the openshift oauth one.

openshift-client/src/main/java/io/fabric8/openshift/client/DefaultOpenShiftClient.java
```diff
@@ -147,17 +146,22 @@ public DefaultOpenShiftClient(String masterUrl) throws KubernetesClientException
...
+      this.delegate = new DefaultKubernetesClient(clientWithOpenShiftOAuhtInterceptor(httpClient), config);
       this.openShiftUrl = new URL(config.getOpenShiftUrl());
     } catch (MalformedURLException e) {
       throw new KubernetesClientException("Could not create client", e);
     }
   }
 
-  private void addOpenShiftAuthInterceptor() {
-    httpClient = httpClient.newBuilder().authenticator(Authenticator.NONE).build();
-    httpClient = httpClient.newBuilder().addInterceptor(new OpenShiftOAuthInterceptor(httpClient,  OpenShiftConfig.wrap(getConfiguration()))).build();
```
jimmidyson: Does this work for you:

```java
private void addOpenShiftAuthInterceptor() {
    OkHttpClient.Builder builder = httpClient.newBuilder();
    builder.authenticator(Authenticator.NONE);
    builder.interceptors().clear();
    builder.interceptors().add(new OpenShiftOAuthInterceptor(httpClient,  OpenShiftConfig.wrap(getConfiguration())));
    httpClient = builder.build();
  }
```

iocanel: The list is unmodifiable :-)

jimmidyson: Can you do `builder.interceptors().add(0, new OpenShiftOAuthInterceptor(httpClient,  OpenShiftConfig.wrap(getConfiguration())));` without the clear?

jimmidyson: Problem is creating a new client will create a new thread pool - old client will need to be cleaned up.

iocanel: I like the idea of a single interceptor with switchable behavior.

jimmidyson: That happens automatically if we use a builder from the existing client so we should stick to that pattern but figure out how we either can swap interceptors or switch the mode of the interceptor. How about letting the auth interceptor check if server is OpenShift on creation & enabling the oauth flow if it is?

iocanel: If we use the builder from the existing client, we are getting stuck with an un-modifiable list of interceptors. So, instead we can just create a new builder and control what pieces we want to keep from the old one. 

Regarding `swapable behavior`, I like the idea, but it is really overkill for the particular case.

Regarding `using a single interceptor`, why spoil something that is so cleanly separated?

iocanel: @jimmidyson: Added a second commit of what I think its best.

jimmidyson: I'm confused. I've just run a test of swapping interceptors & don't get any problem with unmodifiable list:

```java
OkHttpClient client = HttpClientUtils.createHttpClient(new ConfigBuilder().build());
    System.out.println(client.interceptors());
    OkHttpClient.Builder builder = client.newBuilder();
    builder.interceptors().clear();
    System.out.println(builder.build().interceptors());
```

iocanel: Let me try that again! Maybe I did something stupid like trying to clear interceptors on the client rather than the builder.

iocanel:
[merge]

jimmidyson:
After all our discussion & you merge without me reviewing final??? You've forgotten to clear authenticator I think...

iocanel:
You're right!

jimmidyson:
I know! :tada:

iocanel:
I said you're right, not Han Solo!

PR #542, opened 2016-10-10
Do clear the authenticator when building DefaultOpenShiftClient.

iocanel:


openshift-client/src/main/java/io/fabric8/openshift/client/DefaultOpenShiftClient.java
```diff
@@ -155,7 +156,8 @@ protected DefaultOpenShiftClient(OkHttpClient httpClient, OpenShiftConfig config
   private OkHttpClient clientWithOpenShiftOAuhtInterceptor(OkHttpClient httpClient) {
     OkHttpClient.Builder builder = httpClient.newBuilder();
     builder.interceptors().clear();
-    builder.interceptors().add(new OpenShiftOAuthInterceptor(httpClient, OpenShiftConfig.wrap(getConfiguration())));
+    builder.authenticator(Authenticator.NONE)
+           .interceptors().add(new OpenShiftOAuthInterceptor(httpClient, OpenShiftConfig.wrap(getConfiguration())));
```
jimmidyson: Think you need to build a new http client without any authenticator first... like we used to do :-b

iocanel:
[merge]

PR #550, opened 2016-10-20
Remove unused felix.scr.annotations version property

jamesnetherton:
I didn't find any references to this property. `scr.annotations.version` seems to be the one that's used.


jimmidyson:
LGTM thanks [merge]

PR #551, opened 2016-10-20
Fix ResourceCompare.equals()

davidxia:
We aren't using the `right` param.
This causes equals to always return true.


davidxia:
@iocanel

iocanel:
Great catch!
LGTM!

davidxia:
@iocanel Can you merge?

jstrachan:
ok to test

jstrachan:
[merge]

PR #553, opened 2016-10-26
Upgraded kubernetes-model version to 1.0.60

ucosty:
Update to use the version 1.0.60 of kubernetes-model. I also had to fix up some method calls, as editSelector must have become editExtensionsSelector at some stage.


jstrachan:
gonna close this - just upgraded to the new model 1.0.61. We've had all kinds of sundrio issues so sticking on the same version 0.3.3 for now

PR #556, opened 2016-10-27
Add configurable websocket read timeout and ping intervals

jimmidyson:
@iocanel @jbguerraz As discussed on IRC. Would be great if you could try it out
and see if it works for you.


jbguerraz:
I still get an exception (https://gist.github.com/jbguerraz/2eb57cad29287af0f577628a3c44cfb1) even with that 5s timeout (I tried with 10 seconds and still get the exception); no exception only when timeout is set to 0.

jimmidyson:
So the problem with Kubernetes is present in 1.3.x. This is a known issue with Kubernetes 1.3 not closing websocket connections properly & doesn't affect 1.2 or 1.4 (or any version of OpenShift as it's patched correctly there). I'm going to [merge] this PR & think about how to handle force closing ws connections - this is a bit more involved as it requires changes to okhttp so we'll only do it if required, but really recommend upgrading.

jbguerraz:
Thx @jimmidyson ! Will upgrade to 1.4.3 ;) and at least we got configurable websocket timeout & ping interval now :)

jimmidyson:
Yeah! Releasing as we speak...

Issue #548, opened 2016-10-19
[openshift-client] Deleting a DeploymentConfig does not delete the deployer pod


donovanmuller:
Similar to #518 in that if you (`cascading = true`) delete a dc, everything (dc, rc, pods) gets deleted _except_ the deployer pod. 

The deployer pod also does not have the labels that the dc, rc or pods do, so finding the corresponding deployer pod for a dc is rather nasty. I currently have this which works but feels dirty:

``` java
    // if appId relates to an app with name 'some-app'
    client.deploymentConfigs().withName(appId).cascading(true).delete();
    ...
    // clean up the deployment pods. E.g. some-app-1-deploy
    for (Pod pod : client.pods().withLabel("openshift.io/deployer-pod-for.name").list().getItems()) {
        if (pod.getMetadata().getName().startsWith(appId))  {
            client.pods().withName(pod.getMetadata().getName()).cascading(true).delete();
    }
}
```

A cascading delete for a `DeploymentConfig` should also delete the deployer pods in OpenShift.

iocanel:
I'll have a look asap.

PR #567, opened 2016-11-18
#548: DeploymentConfig reaper now also reaps the deployer pod.

iocanel:


jimmidyson:
Could you check if that's consistent with openshift experience?

iocanel:
@jimmidyson: My understanding is that if the deployer is present at the time we delete the `dc`, it won't get deleted, by the command line client.

Still, I don't really see a reason to keep it around.

jimmidyson:
Have you had a look at the OpenShift client code to see what it does on delete?

jimmidyson:
See https://github.com/openshift/origin/blob/master/pkg/client/deploymentconfigs.go#L85 - just deletes the DC itself, no reaping at all. I'm kind of confused why the deployer pod stays around tbh. From what I see, the deployer pod is GCed after a short period, just like any terminated pod.

jimmidyson:
Plus take a look at http://kubernetes.io/docs/user-guide/garbage-collection/. Not sure 100%if it applies to openshift yet, but a better approach in general.

iocanel:
The deployer may be left behind, if there is an error, or the DeploymentConfig gets wiped while the deployer is running. I don't know if there are other cases too.

Maybe there are other cases too, but I don't see any hard in deleting it.

PR #576, opened 2016-11-25
Reaper now reports if the target resource has been deleted. 

iocanel:
Long story short: When deleting a `dc` the rc is not removed. It seems that Openshift will not remove the `rc` if `dc` is still there *(This also applies to oc)*.

So we change the order of things and we delete the DC from within the reaper.

### DIFF
--- a/kubernetes-examples/src/main/java/io/fabric8/openshift/examples/DeploymentConfigExamples.java
+++ b/kubernetes-examples/src/main/java/io/fabric8/openshift/examples/DeploymentConfigExamples.java
@@ -27,6 +27,9 @@ import io.fabric8.kubernetes.client.DefaultKubernetesClient;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.openshift.api.model.Build;
 import io.fabric8.openshift.api.model.BuildRequestBuilder;
+import io.fabric8.openshift.api.model.DeploymentTriggerPolicy;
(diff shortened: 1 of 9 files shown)


iocanel:
@jimmidyson: Can you have a look please?

kubernetes-examples/src/main/java/io/fabric8/openshift/examples/DeploymentConfigExamples.java
```diff
@@ -76,11 +88,13 @@ public static void main(String[] args) throws InterruptedException {
...
       log("Created pods:", client.pods().inNamespace("thisisatest").list().getItems());
-      client.deploymentConfigs().inNamespace("thisisatest").withName("nginx").scale(0);
+      client.deploymentConfigs().inNamespace("thisisatest").withName("nginx").delete();
+      log("Pods:", client.pods().inNamespace("thisisatest").list().getItems());
+      log("Replication Controllers:", client.replicationControllers().inNamespace("thisisatest").list().getItems());
 
       log("Done.");
     }finally {
-      client.namespaces().withName("thisisatest").delete();
+     // client.projects().withName("thisisatest").delete();
```
jimmidyson: Why not delete this?

jimmidyson:
Let's merge this for now & look at async GC later.

jimmidyson:
[merge]

PR #577, opened 2016-11-30
Provide more options to for loading and using resources.

iocanel:


jimmidyson:
Can you summarise what these extra options are?

iocanel:
For load() we now have additional methods for loading File, URL and String pointing to the path of the file.
For resource() we now have the ability to pass the string representation of the object.

No biggies, in practice we just hide serialization and exception handling which in some cases its really useful. (e.g. inside a pipeline).

iocanel:
@jimmidyson: can u have a look and merge please?

jimmidyson:
LGTM [merge]

jimmidyson:
[merge]

PR #591, opened 2016-12-20
enable okhttp proxyAuthenticator

kaimago:


kaimago:
What are next steps to pass "test" and "license-check" checks ?

We really like to use the kubernetes-client, but for that we need the proxy-Authenticator !

iocanel:
ok to test

iocanel:
@kaimago we just had to whitelist the pr

jimmidyson:
Can you explain your use case please? Would it be enough to have configurable proxy username & password instead of a complete proxy authenticator? Would be easier to configure via env vars/system properties if so.

kaimago:
In our case username & password would be enough and of course header needs to be "Proxy-Authorization". 
We don't need env vars/system properties.

Should I change the PR ?

iocanel:
@kaimago: yes, please

kaimago:
Ok, that way ? 
I skipped the junittests, because I have compiling errors locally with the Fluent-classes :(

iocanel:
@kaimago: It looks good to me! 

What kind of issues do you have with the Fluent classes?

jimmidyson:
We also need to add env vars/system properties for these new config options, but we can do that in a future PR before we release this, unless @kaimago feels like adding them in this PR?

jimmidyson:
[merge]

kaimago:
@iocanel 
Regarding the fluent compile errors:
Genereated Sources are looking like this:

import **package** java.lang.Integer;
import **package** okhttp3.TlsVersion;
import io.fabric8.kubernetes.api.builder.Fluent;
import **package** java.lang.Long;
import package java.lang.String;
import org.slf4j.Logger;
import package java.util.Map;
import package io.sundr.builder.annotations.Buildable;

public interface ConfigFluent<A extends ConfigFluent<A>> extends Fluent<A>{


    public boolean isTrustCerts();
    public A withTrustCerts(boolean trustCerts);
    public package java.lang.String getMasterUrl();



Any Idea ?
I'm using Intellij Idea 2016.3.3

PR #594, opened 2016-12-22
remove WebSocket close in ExecWebSocketListener close function

fanhaozzu:
the webSocket have bean closed when dkc.pods().withName("xxxxx").xxxxx.exec(""), so it need't close the socket again.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
@@ -16,6 +16,7 @@
 
 package io.fabric8.kubernetes.client.dsl.internal;
 
+import okhttp3.OkHttpClient;
 import okhttp3.RequestBody;
 import okhttp3.Response;
 import okhttp3.ResponseBody;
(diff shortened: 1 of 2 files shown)


jimmidyson:
ok to test

iocanel:
@fanhaozzu: what happens in the case where the user decides to call close() on the handle but the exec isn't over yet? Should we leave the websocket open?

I think that we should keep the webscoket.close() but we need to check if its closed already. 

@jimmidyson: wdyt?

jimmidyson:
> I think that we should keep the webscoket.close() but we need to check if its closed already.

Sounds good to me, don't want to leave any potential connection leaks.

iocanel:
@fanhaozzu: do you fancy updating you pull request as discussed?

fanhaozzu:
@iocanel  @jimmidyson  it's my pleasure, i will have a try.

fanhaozzu:
@iocanel @jimmidyson i have improve the pull request, in the new pull request 

https://github.com/fabric8io/kubernetes-client/pull/622

iocanel:
@fanhaozzu my understanding is that since you created #622 we can close this one, right?

Please, feel free to reopen, if I misunderstood.

PR #605, opened 2017-01-05
Wait until scaled methods, do not fail if the resource is gone.

iocanel:


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ReplicaSetOperationsImpl.java
```diff
@@ -126,15 +126,25 @@ public ReplicaSet scale(int count, boolean wait) {
...
         try {
-          ReplicaSet rc = getMandatory();
+          ReplicaSet rc = get();
+          //If the rs is gone, we shouldn't wait.
+          if (rc == null) {
+            if (count == 0) {
+              countDownLatch.countDown();
+              return;
+            } else {
+              return;
```
jstrachan: so we still block forever if we try to scale up and the resource has gone now right? Why not pass an exception back to the caller

iocanel: I agree in the long run we should just pass the exception back to the caller.

jstrachan: this is separate to passing the exception - by returning and not counting down the latch we'll block the caller thread forever right (exception or no exception)?

jstrachan:
[merge]

PR #607, opened 2017-01-09
Misc

iocanel:


openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/BuildConfigOperationsImpl.java
```diff
@@ -229,6 +229,9 @@ public void writeTo(BufferedSink sink) throws IOException {
           try {
             source = Okio.source(inputStream);
             sink.writeAll(source);
+          } catch (IOException e) {
+            throw KubernetesClientException.launderThrowable("Can't instantiate binary build, due to error reading stream. "
+              + "Can be caused if stream points to a missing resource or closed externally.", e);
           } finally {
```
rhuss: Could the IoError happens also because of **writing** to the sink ? I guess that the usual error source, because the input stream is typically from a local file and less likely to break.

iocanel: You are right! Let me fix that!

iocanel:
[merge]

PR #608, opened 2017-01-09
Fix createOrReplace for Loadables

### LINKED ISSUE #602: Ambiguity with BaseOperation createOrReplace()
[relevant code](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java#L352-L362)

```
  @Override
  public T createOrReplace(T item) {
    if (Utils.isNullOrEmpty(name) && item instanceof HasMetadata) {
      return withName(((HasMetadata)item).getMetadata().getName()).createOrReplace(item);
    }
    if (get() == null) {
      return create(item);
    } else {
      return replace(item);
    }
  }

```
For the case of first-time creation of a resource. The `get()` here constructs a url using `getNamespacedUrl` which results in something like: `http://<kube_master>/api/v1/<resource>`. This url usually returns a `<resource>List` and the condition `get() == null` is never true!

meaning that the replace workflow is invoked. which predictably fails at a PUT on the non-existent resource.

example usage (this will fail): `client.resourceQuotas().inNamespace(namespace).load(resourceQuotaInputStream).createOrReplace(resourceQuota);`

relevant PR: #538 
kubernetes-client version tested with: v1.4.16 (however, this code hasn't been updated in the subsequent releases)

am I understanding this wrong? or will the `createOrReplace()` workflow never behave as expected?

jimmidyson:
Fixes #602

This adds more tests as well as sneaking in a change to allow no args `createOrReplace` following `load` to allow `client.load(is).createOrReplace()` which is currently not possible.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
@@ -350,11 +350,22 @@ public class BaseOperation<T, L extends KubernetesResourceList, D extends Doneab
   }
 
   @Override
-  public T createOrReplace(T item) {
+  public T createOrReplace(T... items) {
+    T item = getItem();
+    if (items.length > 1) {
+      throw new IllegalArgumentException("Too many items to create.");
+    } else if (items.length == 1) {
+      item = items[0];
+    }
+
+    if (item == null) {
+      throw new IllegalArgumentException("Nothing to create.");
+    }
+
     if (Utils.isNullOrEmpty(name) && item instanceof HasMetadata) {
       return withName(((HasMetadata)item).getMetadata().getName()).createOrReplace(item);
     }
-    if (get() == null) {
+    if (fromServer().get() == null) {
       return create(item);
     } else {
       return replace(item);
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationSupport.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationSupport.java
@@ -282,7 +282,11 @@ public class OperationSupport {
       } else if (response.message() != null) {
         statusMessage = response.message();
       }
-      return JSON_MAPPER.readValue(statusMessage, Status.class);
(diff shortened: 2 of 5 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
```diff
@@ -350,11 +350,22 @@ public T apply(T resource) {
...
+
+    if (item == null) {
+      throw new IllegalArgumentException("Nothing to create.");
+    }
+
     if (Utils.isNullOrEmpty(name) && item instanceof HasMetadata) {
       return withName(((HasMetadata)item).getMetadata().getName()).createOrReplace(item);
     }
-    if (get() == null) {
+    if (fromServer().get() == null) {
```
jimmidyson: This was the sneaky missing `fromServer` btw.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
```diff
@@ -350,11 +350,22 @@ public T apply(T resource) {
   }
 
   @Override
-  public T createOrReplace(T item) {
+  public T createOrReplace(T... items) {
+    T item = getItem();
```
iocanel: why do you store getItems() to item, if you are going to do nothing with it and replace it a few lines below? This lines seems uneeded.

iocanel: Unless you want to support items.length == 0. Eitherway, its a bit confusing

iocanel:
[merge]

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
jimmidyson: Yeah that is exactly why. This enables `client.pods().load(...).createOrReplace()` whereas before you'd have to pass something in to `createOrReplace`.

PR #620, opened 2017-01-14
WIP: Support for `Expect: 100-continue` header

rhuss:
This support is required when uploading data for binary OpenShift builds since this is used by OpenShift for validating a request before accepting the request upload stream.

For this to work, the latest, yet non-released version of OkHttp 3.6.0-SNAPSHOT is required. Support for this feature came in with https://github.com/square/okhttp/commit/a589b8170333476233d48476587a1d4363c90bd0

Since since OkHttp 3.5.0 the WebSocket handling has changed completely, this library needs to be updated for using this new WS scheme. Currently the tests are not running:

* fabric8/mockwebserver needs to be updated to use at least 3.5.0 as well since the tests are using a mixing fabric8/mockwebserver (depending on 3.4.x) and square/okhttp's MockServer (from 3.5.x) which doesn not mix at all.

Also the kubernetes clients needs to removed the depedency on okhttp-ws since the WS support is now directly available within OkHttp directly.


kubernetes-client/pom.xml
```diff
@@ -170,7 +170,7 @@
     <dependency>
       <groupId>com.squareup.okhttp3</groupId>
       <artifactId>okhttp-ws</artifactId>
-      <version>${okhttp.version}</version>
+      <version>3.4.1</version>
```
iocanel: why?

pom.xml
```diff
@@ -75,7 +75,8 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <mockwebserver.version>0.0.4</mockwebserver.version>
-    <okhttp.version>3.4.1</okhttp.version>
+    <okhttp.version>3.6.0-SNAPSHOT</okhttp.version>
```
iocanel: Ohhh that's why :-)

rhuss:
@iocanel okhttp 3.6.0 is out and it would be awesome if we could upgrade kubernetes-client. I tried my best, but we need probably some refactoring for the mock server as well as the API havs changed. 

People are eagerly waiting for better error reporting in f-m-p ;-) --> https://issues.jboss.org/browse/OSFUSE-557

kubernetes-client/pom.xml
rhuss: I also think that okhttp-ws might not be needed anymore as it has been added to okhttp directly.

iocanel:
@rhuss: sorry for the dealy. I completely missed that! Let me try to merge now.

iocanel:
Merged your changes to the latest master as part of #702. So I am closing this one.

PR #622, opened 2017-01-16
shutdown okhttp thread(include websocket connection) when close execWatch

fanhaozzu:
shutdown okhttp thread(include websocket connection) when close execWatch, okhttp thread will remain a minutes if don't this.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
@@ -65,8 +65,10 @@ public class ExecWebSocketListener implements ExecWatch, WebSocketListener, Auto
     private final AtomicBoolean started = new AtomicBoolean(false);
     private final ArrayBlockingQueue<Object> queue = new ArrayBlockingQueue<>(1);
     private final ExecListener listener;
-
-    public ExecWebSocketListener(InputStream in, OutputStream out, OutputStream err, PipedOutputStream inputPipe, PipedInputStream outputPipe, PipedInputStream errorPipe, ExecListener listener) {
+    private final OkHttpClient client;
+    
+    public ExecWebSocketListener(InputStream in, OutputStream out, OutputStream err, PipedOutputStream inputPipe, PipedInputStream outputPipe, PipedInputStream errorPipe, ExecListener listener, OkHttpClient client) {
+        this.client = client;
         this.listener = listener;
         this.in = inputStreamOrPipe(in, inputPipe);
         this.out = outputStreamOrPipe(out, outputPipe);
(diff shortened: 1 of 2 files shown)


fanhaozzu:
This will close okhttp threads, of  course it will also close the webSocket connection. If we just only close webSocket, the okhttp threads remain running(of course, the okhttp threads and connections that are held will be released automatically if they remain idle.)

fanhaozzu:
@iocanel

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
```diff
@@ -95,14 +97,13 @@ public void close() {
...
             }
-        } catch (Throwable t) {
-            throw KubernetesClientException.launderThrowable(t);
-        }
-
-        WebSocket ws = webSocketRef.get();
-        try {
-            if (ws != null) {
-                ws.close(1000, "Closing...");
+            if (client != null) {
```
fanhaozzu: i just want send a command to  apiserver to get result, i don't want keep watching, i need to shutdown all the running thread.

ws.close() is not equivalent to calling client.close(), client.close will shutdown okhttp thread.
If just ws.close(), okhttps threads are still alive.Of course the okhttp thread will close automaticlly after a minutes if there is no request.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
```diff
@@ -95,14 +98,13 @@ public void close() {
...
```
jimmidyson: The lines below effectively (evict idle & shutdown dispatcher) close the client.

fanhaozzu: i we use evict idle & shutdown dispatcher, i think ws.close is not nessary

jimmidyson: But that shuts down the dispatcher for the whole client so no more calls can be made using the client following the websocket close. Client must continue to be usable after websocket close.

fanhaozzu: yes, so i think we shoud remove the ws.close, let it released automaticly after a minutes

jimmidyson: Why? The caller had requested the websocket be closed so we should close the websocket but leave the client usable.

iocanel:
I feel that the changes in this pull request, reflect specific use cases and can't be applied in the client.

@fanhaozzu: I am going to close the pull request for now, as the suggested changes are going to cause quite a few issues to existing users. If you want to rework the pull request, feel free to reopen.

PR #628, opened 2017-01-17
OSFUSE-481: allow using short OPENSHIFT_URL env variable

nicolaferraro:
PR for `master` branch. Some environments (like CDK) put the root URL in the `OPENSHIFT_URL` environment variable, instead of the `oapi` base URL. With this fix, both versions will be supported.

### DIFF
--- a/openshift-client/src/main/java/io/fabric8/openshift/client/OpenShiftConfig.java
+++ b/openshift-client/src/main/java/io/fabric8/openshift/client/OpenShiftConfig.java
@@ -23,7 +23,11 @@ import io.fabric8.kubernetes.client.utils.Utils;
 import io.sundr.builder.annotations.Buildable;
 import io.sundr.builder.annotations.BuildableReference;
 
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.util.Map;
+
+import org.apache.commons.lang.StringUtils;
 
 public class OpenShiftConfig extends Config {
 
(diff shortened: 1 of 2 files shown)


openshift-client/src/main/java/io/fabric8/openshift/client/OpenShiftConfig.java
```diff
@@ -96,7 +96,16 @@ private static String getDefaultOapiVersion(Config config) {
   }
 
   private static String getDefaultOpenShiftUrl(Config config) {
-    return Utils.getSystemPropertyOrEnvVar(OPENSHIFT_URL_SYTEM_PROPERTY, URLUtils.join(config.getMasterUrl(), "oapi", getDefaultOapiVersion(config)));
+    String openshiftUrl = Utils.getSystemPropertyOrEnvVar(OPENSHIFT_URL_SYTEM_PROPERTY);
+    if (openshiftUrl != null) {
+      // The OPENSHIFT_URL environment variable may be provided without the 'oapi' path in some configurations
+      if (!openshiftUrl.contains("/oapi/")) {
```
jimmidyson: Should this check should really be is the URL path empty or `/` only? Either way I think it will be mostly be fine I guess.

nicolaferraro:
Something like this @jimmidyson ?

jimmidyson:
Looks good - wonder if we can get a test in for that?

nicolaferraro:
Sure

jimmidyson:
[merge]

Issue #613, opened 2017-01-12
Add wait functionality for the resource that it makes sense


iocanel:
There are lots of places where we need to wait for resource to become ready, to name a few:

- arquillian extension
- jenkins plugins
- etc

and in each place we use different code.

Resources that we could add wait functionality:

- Deployment
- ReplicaSet
- Pod
- Service
- ReplicationController
- DeploymentConfig

The result coud look like:

    Pod p = client.pods().inNamespace("myns").withName("mypod").await(5, TimeUnit.MINUTES);

or using a resource

    HasMetadata item = client.resouce(myresource).await(5, TimeUnit.MINUTES);

or using load

    List<HasMetadata> items = client.load(is).waitingUntilReady(5, TimeUnit.MINUTES).createOrReplace();

In case of failure to become ready we could throw an unchecked version of a TimeoutException.

@jimmidyson: Thoughts?B

jimmidyson:
Yeah great idea.

Prefer to be consistent in naming, await perhaps isn't expressive enough? ` client.pods().inNamespace("myns").withName("mypod").await(5, TimeUnit.MINUTES)` - what am I waiting for? I assume `Ready` status but would be clearer to be named `waitUntilReady` perhaps? But then what does ready mean for a deployment for example?

With `client.resouce(myresource).await(5, TimeUnit.MINUTES)` if the resource type isn't "waitable" what should happen? Throw exception, or return immediately?

It's a shame that `createOrReplace` has to be the terminal method in `client.load(is).waitingUntilReady(5, TimeUnit.MINUTES).createOrReplace()` - doesn't read so well.

iocanel:
`waitUntilReady` seems good to me.

In all cases waitUntilReady should mean that the required number of pods is ready.  
So for deployments specifically we should check the matching replica set and ensure that the the ready replicas match the spec.

If a resource is not waitable we should return asap for two reasons:
- an non waitable resource is ready anyway.
- I think that in most cases no one will just wait on a single resource, but in a list of resources (waitable and non waitables). So its a one way really.

I agree that these are not great: 

    client.load(is).waitingUntilReady(5, TimeUnit.MINUTES).createOrReplace()
    client.load(is).createOrReplace(5, TimeUnit.MINUTES)

so how about:

    client.load(us).createOrReplaceAnd().waitUntilReady(5, TimeUnit.MINUTES);

jimmidyson:
All sounds good to me!

[...]

PR #632, opened 2017-01-21
Introduce Waitable.

iocanel:
We have discussed those changes as part of #613

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/readiness/Readiness.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/readiness/Readiness.java
@@ -1,0 +1,140 @@
+/**
+ * Copyright (C) 2015 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.fabric8.kubernetes.client.internal.readiness;
+
+import io.fabric8.kubernetes.api.model.EndpointSubset;
+import io.fabric8.kubernetes.api.model.Endpoints;
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.fabric8.kubernetes.api.model.Pod;
+import io.fabric8.kubernetes.api.model.PodCondition;
+import io.fabric8.kubernetes.api.model.ReplicationController;
+import io.fabric8.kubernetes.api.model.ReplicationControllerSpec;
+import io.fabric8.kubernetes.api.model.ReplicationControllerStatus;
+import io.fabric8.kubernetes.api.model.extensions.Deployment;
+import io.fabric8.kubernetes.api.model.extensions.DeploymentSpec;
+import io.fabric8.kubernetes.api.model.extensions.DeploymentStatus;
+import io.fabric8.kubernetes.api.model.extensions.ReplicaSet;
+import io.fabric8.kubernetes.api.model.extensions.ReplicaSetSpec;
+import io.fabric8.kubernetes.api.model.extensions.ReplicaSetStatus;
+import io.fabric8.kubernetes.client.utils.Utils;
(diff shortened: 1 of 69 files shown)


iocanel:
@jimmidyson: since you have some spare, I thought it might be a good idea to keep you off boredom :-)

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/readiness/Readiness.java
```diff
@@ -0,0 +1,140 @@
...
+    return spec.getReplicas().intValue() == status.getReadyReplicas();
+  }
+
+
+  public static boolean isDeploymentConfigReady(DeploymentConfig d) {
+    Utils.checkNotNull(d, "Deployment can't be null.");
+    DeploymentConfigSpec spec = d.getSpec();
+    DeploymentConfigStatus status = d.getStatus();
+
+    return spec.getReplicas().intValue() == status.getReplicas() &&
```
jimmidyson: `status.getReadyReplicas()`?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/readiness/Readiness.java
```diff
@@ -0,0 +1,140 @@
...
+      throw new IllegalArgumentException("Item needs to be one of [Deployment, ReplicaSet, Pod, DeploymentConfig, ReplicationController], but was: [" + item.getKind() + "]");
+    }
+  }
+
+  public static boolean isDeploymentReady(Deployment d) {
+    Utils.checkNotNull(d, "Deployment can't be null.");
+    DeploymentSpec spec = d.getSpec();
+    DeploymentStatus status = d.getStatus();
+
+    return spec.getReplicas().intValue() == status.getReplicas() &&
```
jimmidyson: `status.getReadyReplicas()`?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/readiness/Readiness.java
```diff
@@ -0,0 +1,140 @@
...
+      spec.getReplicas().intValue() <= status.getAvailableReplicas();
+  }
+
+  public static boolean isReplicationControllerReady(ReplicationController r) {
+    Utils.checkNotNull(r, "ReplicationController can't be null.");
+    ReplicationControllerSpec spec = r.getSpec();
+    ReplicationControllerStatus status = r.getStatus();
+
+    return spec.getReplicas().intValue() == status.getReplicas();
```
jimmidyson: `status.getReadyReplicas()`?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/readiness/Readiness.java
iocanel: not available.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/readiness/Readiness.java
iocanel: Not available.

jimmidyson: Sorry just noticed that's only available in OpenShift 1.5 :(

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/readiness/Readiness.java
jimmidyson: Sorry just noticed that's only available in Kubernetes master :(

jimmidyson:
[merge]

palmerabollo:
Do you know why it is not implemented for Service Accounts? Is it a bug or a feature? Thanks!

rohanKanojia:
@palmerabollo : Could you please elaborate your use case? Does it make sense for ServiceAccounts to be Waitable?

palmerabollo:
Hi @rohanKanojia, THANK YOU for your prompt reply. As far as I know, when you create a Service Account in Kubernetes, there is another process (the Token Controller, see https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#token-controller) that is in charge of observing Service Account creations + creating the Secret to allow API access.

I've faced some cases where the Service Account is created but the token/secret is still not associated to it, so the Service Account is not yet usable. This is why I think it should be `Waitable`, to be able to wait for the token/secret to be ready, too. Does it make sense?

Currently I have to implment a custom polling mechanism to get the Service Account status and wait until the secret is associated to the Service Account before continuing. 

I might be missing something or perhaps there is a cleaner workaround I'm not aware of.

PR #634, opened 2017-01-26
Various fixes for WatchConnectionManager

gyoetam:
- Removed never thrown exceptions from the ctor signature
- Only validate the watch URL once on startup
- **Changed reconnect timing to an exponential backoff capped at 32*baseInterval**
- **Now using a single threaded executor for serial event handling (this eliminates an issue where in some cases the connection gets duplicated)**
- Don't try to send ping while the websocket is inactive during a reconnect backoff
- Delay the first ping with the ping interval, don't send right away on open
- Added/modified level of some log statements
- **Ensure that websocket is:**
  - **always properly closed (e.g. ping failure) to avoid leaked connections**
  - **never double closed (this removes some unnecessary exceptions from the log)**
- Properly queue startup exceptions and don't reconnect in that case
- Fixed some type warning issues
- Introduced thread-renaming tasks for the executor for better logging
- Added test case for Watch

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
@@ -39,14 +39,15 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
-import java.net.ConnectException;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.Objects;
 import java.util.concurrent.ArrayBlockingQueue;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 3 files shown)


gyoetam:
Hi @jimmidyson, would you mind checking out this PR? Seems to work well on my side.

foxish:
Is there a reason we want websockets to do watches? They have some support lacking (https://github.com/kubernetes/kubernetes/issues/25126). @jimmidyson, did the project consider using a plain hanging HTTP GET instead?

iocanel:
@foxish: It seems that this question is quite popular lately.

Personally, I wouldn't mind using hanging gets as an option _(keeping the web-sockets as the default out of the box experience)_. 

Is this something that you would consider contributing?

foxish:
@iocanel I can try my hand at a patch over the next couple of days. Has there been interest in supporting the `gcp` auth provider as well?

/cc @iyanuobidele

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/WatchTest.java
```diff
@@ -0,0 +1,125 @@
...
+      .build();
+  static final WatchEvent outdatedEvent = new WatchEventBuilder().withStatusObject(outdatedStatus).build();
+
+  @Test
+  public void testDeletedAndOutdated() throws InterruptedException {
+    logger.info("testDeletedAndOutdated");
+    KubernetesClient client = server.getClient().inNamespace("test");
+
+    // DELETED event, then history outdated
+    server.expect()
```
iocanel: Nice testing Fu!

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
```diff
@@ -66,21 +67,35 @@
...
   private final int reconnectInterval;
+  // TODO maybe make this configurable, either as is (an exponent) or a max time
+  private final static int maxIntervalExponent = 5; // max 32x slowdown from base interval
   private final long websocketTimeout;
   private final long websocketPingInterval;
   private final AtomicInteger currentReconnectAttempt = new AtomicInteger(0);
   private final AtomicReference<WebSocket> webSocketRef = new AtomicReference<>();
-  private final ScheduledExecutorService executor = Executors.newScheduledThreadPool(2);
+  // single threaded serial executor
+  private final ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor(new ThreadFactory() {
```
iocanel: I like the ThreadFactory part. Really nice addition.

iocanel:
ok to test

iocanel:
[merge]

gyoetam:
@iocanel Awesome, thanks! Sorry for not splitting up the changes into smaller PRs, next time I'll keep that in mind. :)

Issue #639, opened 2017-02-02
WatchConnectionManager : if the connection is dropped, the watcher keep creating new connections


lburgazzoli:
If the connection breaks, the WatchConnectionManager appears to create additional connections, thus the client is notified twice (or more)

- 1st disconnection (2 connections are created):
```
2017-02-02 13:04:47,298 | ERROR | 8.42.34:8443/... | WatchConnectionManager           | 67 - io.fabric8.kubernetes-client - 1.4.33 | Exec Failure
java.net.SocketTimeoutException: timeout

...

2017-02-02 13:04:55,455 | DEBUG | 8.42.34:8443/... | WatchConnectionManager           | 67 - io.fabric8.kubernetes-client - 1.4.33 | WebSocket successfully opened
2017-02-02 13:04:55,465 | DEBUG | 8.42.34:8443/... | WatchConnectionManager           | 67 - io.fabric8.kubernetes-client - 1.4.33 | WebSocket successfully opened
```

- 2nd disconnection  (3 connections are created)::
```
2017-02-02 13:08:29,457 | ERROR | 8.42.34:8443/... | WatchConnectionManager           | 67 - io.fabric8.kubernetes-client - 1.4.33 | Exec Failure
java.net.SocketTimeoutException: timeout

...

2017-02-02 13:08:33,577 | DEBUG | 8.42.34:8443/... | WatchConnectionManager           | 67 - io.fabric8.kubernetes-client - 1.4.33 | WebSocket successfully opened
2017-02-02 13:08:33,583 | DEBUG | 8.42.34:8443/... | WatchConnectionManager           | 67 - io.fabric8.kubernetes-client - 1.4.33 | WebSocket successfully opened
2017-02-02 13:08:33,583 | DEBUG | 8.42.34:8443/... | WatchConnectionManager           | 67 - io.fabric8.kubernetes-client - 1.4.33 | WebSocket successfully opened

```

lburgazzoli:
Fixed by #634

PR #640, opened 2017-02-02
WatchConnectionManager : if the connection is dropped, the watcher keep creating new connections #639

lburgazzoli:
This issue appear to be caused by a call to onClose from the ping task which triggers a new connections a attempt but at the same time, the error causing the ping failure is also intercepted by onFailure which triggers a new connection too.

Not sure if removing the call to onClose from ping task try/catch would result in other cases not more handled so a review would be very welcome.


iocanel:
Just merged #634 which among other addresses this issue too, with a slightly different approach. Can you check if this indeed solves the issue?

lburgazzoli:
Yes it does, much better fix :-)

Issue #647, opened 2017-02-07
Client should accept valid response status codes range


nikhiljindal:
From https://github.com/kubernetes/kubernetes/issues/33196

Looks like the client only accepts status ok (http code 200) as success for deletion: https://github.com/fabric8io/kubernetes-client/blob/112fdc2974f8ebd04839bd44595c9c836f808c4d/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationSupport.java#L202
We want to update the apiserver to send 202 in case of asynchronous deletion.

Please fix the client so that the apiserver change does not break it.

Please feel free to comment on the original issue in case of any questions

jimmidyson:
This should be applied across all requests.

iocanel:
We will delegate to okhttp clients isSuccesfull() which literaly does:

     return code >= 200 && code < 300;

jimmidyson:
Perfect

jimmidyson:
Done.

PR #649, opened 2017-02-07
#647: Delegate to okhttp's isSuccesfull() for checking of request was…

iocanel:
… successfully executed (don't check status codes).


iocanel:
@jimmidyson: can you have a quick look here too, so that we can get a release out?

jimmidyson:
[merge]

Issue #636, opened 2017-01-30
Authentication with GKE


foxish:
The client does not know how to use oauth to authenticate with GKE clusters.
Error thrown is invariably:
```
 Message: Unauthorized! Token may have expired! Please log-in again..
```

iocanel:
It should be fine now! You will still need to use gcloud to obtain the token, but the client now will be able to read it and use it.

PR #650, opened 2017-02-07
#636: Check if authprovider is present in .kube/config file and try t…

iocanel:
…o use the access-token if available.


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
```diff
@@ -17,6 +17,7 @@
 package io.fabric8.kubernetes.client;
 
 import io.fabric8.kubernetes.api.model.ConfigBuilder;
+import io.fabric8.zjsonpatch.internal.guava.Strings;
```
jimmidyson: Wrong import I assume?

iocanel:
Fixed import so let's [merge]

PR #651, opened 2017-02-07
A couple more improvements on waitables.

iocanel:
1. client.resource(<instance of hasmetadata>) should make use of the type parameter (rather than using HasMetadata). 

2. We should be able to use waitUntilReady() on  exisiting resource.
3. waitUntilReady() should return the instance on which the readiness check was performed rather than performing internally a new get().
4. Reduce the boilerplate code that is used internally in waitUntilReady().

5. Added a test that for client.resource(pod).waitUntilReady() and client.resource(pod).createAnd().waitUntilReady().


jimmidyson:
[merge]

PR #652, opened 2017-02-07
Allow watching resources over HTTP if watching using websocket fails

foxish:
Falling back to using persistent HTTP Get. 
The code could be DRYer. Getting this out there to get some initial opinions.

@iocanel @jimmidyson @iyanuobidele

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
@@ -40,6 +40,7 @@ import io.fabric8.kubernetes.client.dsl.Gettable;
 import io.fabric8.kubernetes.client.dsl.Reaper;
 import io.fabric8.kubernetes.client.dsl.Watchable;
 import io.fabric8.kubernetes.client.dsl.internal.WatchConnectionManager;
+import io.fabric8.kubernetes.client.dsl.internal.WatchHTTPManager;
 import io.fabric8.kubernetes.client.utils.URLUtils;
 
 import java.io.File;
@@ -649,6 +650,22 @@ public class BaseOperation<T, L extends KubernetesResourceList, D extends Doneab
       return watch;
     } catch (MalformedURLException e) {
       throw KubernetesClientException.launderThrowable(e);
+    } catch (KubernetesClientException ke) {
+      WatchHTTPManager watch = null;
+      try {
+        watch = new WatchHTTPManager(
+          client,
+          this,
+          resourceVersion,
+          watcher,
+          config.getWatchReconnectInterval(),
+          config.getWatchReconnectLimit(),
+          config.getConnectionTimeout()
+        );
+      } catch (MalformedURLException e) {
+        throw KubernetesClientException.launderThrowable(e);
+      }
+      return watch;
     }
   }
 
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
@@ -18,6 +18,7 @@ package io.fabric8.kubernetes.client.dsl.base;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.client.internal.readiness.Readiness;
 import io.fabric8.kubernetes.client.utils.Utils;
+import java.net.ProtocolException;
 import okhttp3.HttpUrl;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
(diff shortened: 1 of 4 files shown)


foxish:
Will fix tests shortly.

iocanel:
ok to test

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
```diff
@@ -649,6 +650,22 @@ public Watch watch(String resourceVersion, final Watcher<T> watcher) throws Kube
       return watch;
     } catch (MalformedURLException e) {
       throw KubernetesClientException.launderThrowable(e);
+    } catch (KubernetesClientException ke) {
```
jimmidyson: Should we check the type of the exception? For example, there would be no point in falling back to HTTP watch in e.g. auth error. Perhaps we should only fall back if we receive a 200 instead of a 101 for example?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
```diff
@@ -649,6 +650,22 @@ public Watch watch(String resourceVersion, final Watcher<T> watcher) throws Kube
...
+      WatchHTTPManager watch = null;
+      try {
+        watch = new WatchHTTPManager(
```
jimmidyson: Simply `return new WatchHTTP....`

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchHTTPManager.java
```diff
@@ -0,0 +1,278 @@
...
+      .url(httpUrlBuilder.build())
+      .addHeader("Origin", requestUrl.getProtocol() + "://" + requestUrl.getHost() + ":" + requestUrl.getPort())
+      .build();
+
+    Response response = null;
+    try {
+      response = clonedClient.newCall(request).execute();
+      BufferedSource source = response.body().source();
+      while (!source.exhausted()) {
+        String message = source.readUtf8LineStrict();
```
jimmidyson: Is each event new line separated then? Not chunked response? I haven't looked to be sure but will trust you :)

jimmidyson:
Thanks for this PR. It's looking good, I like the approach, just some minor questions and we need a test too if possible.

jimmidyson:
Oh and this breaks some tests too - see build logs please.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchHTTPManager.java
foxish: Yes, the chunk size is sent at first, but does not correspond to the event boundaries. The newlines are reliable separators.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
foxish: Done, added a check for `ke.getCode()` to be 200.

foxish:
@jimmidyson @iocanel PTAL. Addressed comments, added tests.

jimmidyson:
When  I try this locally, I get a 5 second delay between 200 response from websocket watcher request and the HTTP watcher kicking in:

```
09:15:48.719 [main] DEBUG io.fabric8.kubernetes.client.dsl.internal.WatchConnectionManager - Connecting websocket ... io.fabric8.kubernetes.client.dsl.internal.WatchConnectionManager@e5041ef
09:15:53.830 [OkHttp http://localhost:8001/...] WARN  io.fabric8.kubernetes.client.dsl.internal.WatchConnectionManager - Exec Failure: HTTP 200, Status: 200 -
```

Any ideas why?

iocanel:
[merge]

jimmidyson:
@iocanel I was waiting for feedback on https://github.com/fabric8io/kubernetes-client/pull/652#issuecomment-278273383...

iocanel:
@jimmidyson: you've approved the request though.

iocanel:
I aborted the job on jenkins until you get feedback.

jimmidyson:
@iocanel That is a good point. I had approved the code review, but was testing locally before merging. I will ensure from now on that review approval means a complete yes, not just code review approval. Sorry for the misunderstanding!

foxish:
Thanks for the quick review. A 5 second delay seems quite substantial, and I don't remember running into this. I'll look into it shortly.

jimmidyson:
@foxish Yeah it was unexpected, it could be environmental but I'm not sure how. Let me know if you have any ideas.

foxish:
@jimmidyson It seems like the websocket call takes ~5 seconds to fail. The delay comes from the `DEFAULT_WEBSOCKET_TIMEOUT` which is set to 5 seconds. 
Specifying it in the form: `Config config = new ConfigBuilder().withMasterUrl("http://127.0.0.1:8001").withWebsocketTimeout(1000).build();` for example, brings the delay down to 1s.

jimmidyson:
OK so sounds like we need to address the websocket timeout separately. Getting back a 200 should prevent the websocket from trying to connect and immediately return.

@foxish Would you have time to look at that fix? I'm happy to merge this in & get a fix in for that in a separate PR.

foxish:
That sounds good. I'll send a follow-up PR for the fix. Would cutting a release be possible sometime early next week? We'd like to use these changes in our Spark integration efforts.

jimmidyson:
We can do the release as soon as we have the fix for the delay.

jimmidyson:
[merge] thanks @foxish!

PR #655, opened 2017-02-08
Fix high latency when websocket connection fails

foxish:
Follow up to https://github.com/fabric8io/kubernetes-client/pull/652 based on https://github.com/fabric8io/kubernetes-client/pull/652#issuecomment-278395155

/cc @jimmidyson @iocanel


iocanel:
ok to test

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
```diff
@@ -193,6 +194,13 @@ public void onFailure(IOException e, Response response) {
           return;
         }
 
+        if (response != null && response.code() == HTTP_OK) {
```
iocanel: I would group this block with the rest of the `if then else` that take place when `response != null`. 

Definitely not a blocker though!

jimmidyson:
Thanks again! [merge]

We'll kick off a release as soon as this is merged.

PR #656, opened 2017-02-09
Make watching via http async

foxish:
The watching via http get was previously happening on the main thread. 
This one ensures that it is executed in a background thread, so that we don't block. This behavior is similar to the websockets client.

P.S. Missed this earlier :) Sorry!
/cc @iocanel @jimmidyson


foxish:
ping

jimmidyson:
LGTM ok to test

jimmidyson:
Thanks! [merge]

PR #658, opened 2017-02-13
Fix timeout exception constructor params order

### LINKED ISSUE #657: Format String error in KubernetesClientTimeoutException
When you call KubernetesClientTimeoutException.<init>(HasMetadata, long, TimeUnit) the String.format method call will throw an exception, because the milliseconds are passed as the last argument but it should be the first argument.

jimmidyson:
Fixes #657


iocanel:
[merge]

Issue #631, opened 2017-01-21
Allow per-request configuration changes


jimmidyson:
From @iocanel in https://github.com/redhat-ipaas/ipaas-rest/pull/65#discussion_r97195525:

If we break the Config object, in two parts:

properties that affect the connection (e.g. url, keys, certificates)
properties that don't affect the connection (token, credentials, timeout values)
we could possibly allow partial client reconfiguration:

DSL style

```java
 client.withRequstConfiguration(config).pods().list();
```

Functional Style

```java
PodList list =  client.withRequstConfiguration(config, new Function<KubernetesClient, PodList> {
        public PodList apply(KubernetesClient client) {
              return client.pods().list();
        }
 });
```

Lamda syle

```java
PodList list =  client.withRequstConfiguration(config, c -> c.pods.list());
```

While I love the Lamda style personally, as we're still on Java 7 for now, we're probably going to have to go with DSL style.

iocanel:
@jimmidyson: The question is how we perform the split without breaking existing clients.

PR #660, opened 2017-02-14
Add the ability to call a function with custom request configuration

iocanel:
This is a fix for #631.

It actually allows the call of a function with custom request configuration. In detail:

1. It moves the request specific configuration to a new object called RequestConfig _(forward compatible change)_.
2. The config object internally delegates to RequestConfig (to limit the code changes to the minimum).
3. The RequestConfig delegate is overridable by a thread local variable.

The code actually looks like:  

       Pod pod1 = client.withRequestConfig(new RequestConfigBuilder().withOauthToken("TOKEN").build()).call(new Function<NamespacedKubernetesClient, Pod>() {
       @Override
       public Pod apply(NamespacedKubernetesClient c) {
         return c.pods().inNamespace("test").withName("pod1").get();
       }
    });

Worths mentioning that despite the functional feel, it works with JDK7 too as it uses the Function interface that is generated  as part of the kubernetes-model (generated by sundrio).

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Adapters.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Adapters.java
@@ -28,6 +28,8 @@ import java.util.Map;
 import java.util.ServiceLoader;
 import java.util.Set;
 
+import okhttp3.OkHttpClient;
+
 public final class Adapters {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(Adapters.class);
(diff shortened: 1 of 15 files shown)


jimmidyson:
Why do we need ThreadLocal rather than using the same strategy to wrap the config as we did with `OpenShiftConfig.wrap` and creating new client with that and existing http client?

iocanel:
We want to reuse the same instance of the client, but apply the new `request configuration`.

The `request configuration` also affects things, like interceptors and authenticators, which are already configured for the `OkHttpClient` instance we use and will not pick up the new configuration (unless of course we recreate the `OkHttpClient`).

So we would either have to somehow change the state of these objects, or use something like a ThreadLocal.

Makes sense?

jimmidyson:
[merge]

PR #663, opened 2017-02-16
Drop watch logging from info to debug

ash211:
You would only look at WebSocket open/close logging when debugging an issue

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
@@ -155,7 +155,7 @@ public class WatchConnectionManager<T extends HasMetadata, L extends KubernetesR
         if (response != null && response.body() != null) {
           response.body().close();
         }
-        logger.info("WebSocket successfully opened");
+        logger.debug("WebSocket successfully opened");
         webSocketRef.set(webSocket);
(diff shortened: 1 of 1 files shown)


ash211:
cc @foxish

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
```diff
@@ -293,7 +293,7 @@ public void onPong(Buffer buffer) {
 
       @Override
       public void onClose(final int code, final String reason) {
-        logger.info("WebSocket close received. code: {}, reason: {}", code, reason);
```
foxish: Thanks @ash211.
we want the same change in WatchHTTPManager.java as well I think.

foxish:
cc @jimmidyson @iocanel PTAL

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
ash211: WatchHTTPManager already has the normal path for open and close logging at debug level instead of info.  I'm looking at lines 111 and 272 in WatchHTTPManager.java

jimmidyson:
LGTM thanks! ok to test

jimmidyson:
Thanks again! [merge]ing...

PR #664, opened 2017-02-16
Allow from load(), resource() and resourceList() to access the specific resource operation.

iocanel:
The goal here is to be able to do something like:

    client.load(is).as(TemplateResource.class).process(parameters)    (implemented)

or stuff like:

    client.load(is).as(PodResource).log() (not implemented yet).

So this pull request adds the base for all that and adds an implementation specific to Template.

This will eventually allow us to clean the `kubernetes-client` module from openshift specific code: https://github.com/fabric8io/kubernetes-client/blob/c64d7e15ac0cb088d19a905e4a80e30a58f99820/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.java#L380

It will also allow tooling e.g. arquillian to be able to easily load and handle resources as templates (when running against openshift) in a clean way.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/ResourceFactories.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/ResourceFactories.java
@@ -1,0 +1,70 @@
+/**
+ * Copyright (C) 2015 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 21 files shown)


jimmidyson:
While I like the idea, this DSL is really starting to grow and I just want to make sure it makes sense to add these things. If you know it's a Template, then why wouldn't you use:

```java
client.templates().load(is).process(...)
```

iocanel:
`client.templates().load()` is used when you know you load a template.
`client.load().as(TemplateOperation)` is when you don't know that its a template but want to handle it as one.

So its pretty much a shortcut for https://github.com/fabric8io/kubernetes-client/pull/664/commits/cf8f148cd5e6984ae4da440bb014566753d355ef#diff-eb210085613096eea51a2e8484d13403R41

iocanel:
@jimmidyson: on second though, it might be simpler if I merge both features under:

     client.templates().load()

So I am closing it for now....

PR #662, opened 2017-02-15
Clenaup

iocanel:
A while back we in order to support the xxx-mock modules, we added some extra generic parameters across resources. 

Now that the xxx-mock are removed we can simplify our internal interfaces. So, this pull request remove all redundant interfaces and makes code way more simple.

PR #676, opened 2017-02-24
Set version to 2.1-SNAPSHOT.

iocanel:


iocanel:
Apparently, [this](https://github.com/fabric8io/kubernetes-client/pull/662) brought up changes that cause compatibility issues. 

Even though the interfaces modified were intermediate, in some cases where used as return types in the DSL, which seems to be causing issues.

The first release with those changes is 2.0.6 and it seems that libraries built against 2.0.6 don't play nicely with stuff built with 2.0.5. 

We should at least bump to 2.1.0 so that the version itself can hint that.

jimmidyson:
[merge]

PR #681, opened 2017-02-27
complete top-level thirdPartyResource impl

iyanuobidele:
**What's In this PR ?**
Completes the implementation of the top-level ThirdPartyResources as a nonNamespaceOperation based on [k8s-doc](https://kubernetes.io/docs/user-guide/thirdpartyresources/#creating-a-thirdpartyresource)

This is a pre-req to a downstream [project ](https://github.com/apache-spark-on-k8s/spark) tracked at [issue](https://github.com/apache-spark-on-k8s/spark/issues/153). Also a pre-req to supporting CustomObjects (future PR).

**How it was tested**
Unit tests

The final goal is to add support for CustomObjects in k8s through the client. I feel I might need to add the Builder/Model classes to the kubernetes-model project to get this done. I will open an issue at the kubernetes-model to open discussions on this.

/cc @foxish @iocanel @jimmidyson @mccheah @ash211

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/ExtensionsAPIGroupClient.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/ExtensionsAPIGroupClient.java
@@ -36,11 +36,7 @@ import io.fabric8.kubernetes.api.model.extensions.ReplicaSet;
 import io.fabric8.kubernetes.api.model.extensions.ReplicaSetList;
 import io.fabric8.kubernetes.api.model.extensions.ThirdPartyResource;
 import io.fabric8.kubernetes.api.model.extensions.ThirdPartyResourceList;
-import io.fabric8.kubernetes.client.dsl.MixedOperation;
-import io.fabric8.kubernetes.client.dsl.Resource;
-import io.fabric8.kubernetes.client.dsl.RollableScallableResource;
-import io.fabric8.kubernetes.client.dsl.ScalableResource;
-import io.fabric8.kubernetes.client.dsl.ExtensionsAPIGroupDSL;
+import io.fabric8.kubernetes.client.dsl.*;
 import io.fabric8.kubernetes.client.dsl.internal.DaemonSetOperationsImpl;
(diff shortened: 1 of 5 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/KubernetesClient.java
```diff
@@ -109,6 +112,8 @@
 
   NonNamespaceOperation<PersistentVolume, PersistentVolumeList, DoneablePersistentVolume, Resource<PersistentVolume, DoneablePersistentVolume>> persistentVolumes();
 
+  NonNamespaceOperation<ThirdPartyResource, ThirdPartyResourceList, DoneableThirdPartyResource, Resource<ThirdPartyResource, DoneableThirdPartyResource>> tprs();
```
iocanel: I'd prefer the method called thirdpartyresources() instead of trps().

jimmidyson: Agreed. Perhaps we could keep tprs as an alias though?

iocanel:
Currently the method for thirdpartyresources() is present under the extensions.

The pull request moves it under the main client (not sure if we should), but don't see it being removed from the extensions.

If we need to make thrid party resources as non namespaced resource, why don't we just do it under the extensions api?

jimmidyson:
> If we need to make thrid party resources as non namespaced resource, why don't we just do it under the extensions api?

Agreed this is what should happen, just updating the existing existing resource in extensions API, not moving to core client.

iyanuobidele:
@iocanel @jimmidyson Thanks for the quick response and comments. I'll address your comments very shortly.

iyanuobidele:
@iocanel @jimmidyson PTAL. I addressed all your comments.

jimmidyson:
ok to test

jimmidyson:
Can you fix up license headers please with `mvn -N license:format`?

iyanuobidele:
@jimmidyson could you please take a look at [issue](https://github.com/fabric8io/kubernetes-model/issues/197#issuecomment-283173408)

jimmidyson:
Thanks! [merge]

PR #684, opened 2017-02-28
OpenShiftOAuthInterceptor will now try with oauth token, if original …

iocanel:
…request is unauthorized and request using credentials is also unauthorized.

### DIFF
--- a/openshift-client/src/main/java/io/fabric8/openshift/client/internal/OpenShiftOAuthInterceptor.java
+++ b/openshift-client/src/main/java/io/fabric8/openshift/client/internal/OpenShiftOAuthInterceptor.java
@@ -64,41 +64,28 @@ public class OpenShiftOAuthInterceptor implements Interceptor {
         Response response = chain.proceed(request);
 
         //If response is Forbidden or Unauthorized, try to obtain a token via authorize() or via config.
-        if (response.code() != 401 && response.code() != 403) {
+        if (!isUnauthorized(response)) {
           return response;
-        } else if (Utils.isNotNullOrEmpty(config.getUsername()) && Utils.isNotNullOrEmpty(config.getPassword())) {
+        }
+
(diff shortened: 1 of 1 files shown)


jimmidyson:
Can you explain a bit more about what the use case is for this? I assume that if a user has decided to use the username/password then that should be the only auth used, not fall back to a token.

iocanel:
@jimmidyson: That would be the case if username / password were only set explicitly, but this is not the case. A username/password may have been picked up from system / environment / .kube/config and will hide oauth regardless of how it was set (even if it has been set explicitly).

jimmidyson:
I would say that is fine behaviour though. If the env has stuff that config is set from, then that should be respected, not fall back to oauth token. Personally I would find that surprising behaviour.

iocanel:
You can tell that to the person that writes the code bellow and gets an Exception.

    Config config = new ConfigBuilder().withMasterUrl(master).withOauthToken(token).build();
    OpenshiftClient client = new DefaultOpenshiftClient(config);
    client.doStuff();

jimmidyson:
Now that is different. We should probably fix the fact that if any auth stuff is explicitly set through builder then that takes precedence over discovered from environment.

openshift-client/src/main/java/io/fabric8/openshift/client/internal/OpenShiftOAuthInterceptor.java
```diff
@@ -122,5 +109,37 @@ private  String authorize() {
...
+  }
+
+  private static void setAuthHeader(Request.Builder builder, String token) {
+    if (token != null) {
+      builder.header(AUTHORIZATION, String.format("Bearer %s", token));
+    }
+  }
+
+  private static boolean isUnauthorized(Response response) {
+    return response.code() == 401 || response.code() == 403;
```
jimmidyson: Please use `HttpURLConnection.HTTP_OK` and `HttpURLConnection.HTTP_FORBIDDEN` just to be a bit clearer what these mean.

jimmidyson:
Minor nit but as discussed on IRC earlier, let's merge this before next release.

Issue #633, opened 2017-01-25
Cannot get limit amount of data returned from pod logs


pmcq:
Pod operations also does not support limitBytes query parameter:
https://kubernetes.io/docs/api-reference/v1/operations/

From what I can tell this would mean that for a long-running pod trying to get logs would result in creating a massive string.

https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java#L110

iocanel:
absolutely right! We need to add support for `limitBytes`

serhii-samoilenko:
@iocanel, can you please take a look at my PR?

serhii-samoilenko:
The fix for this issue was included in version 2.2.0.
I think, it can be closed.

jimmidyson:
Thanks!

PR #686, opened 2017-03-01
#633: Added support for limitBytes parameter on pod logs request

serhii-samoilenko:
Made a fix for issue #633.
Now it is possible to limit log output for pod, like:
`kubernetesClient.pods().withName("pod").limitBytes(1024).getLog();`

Also, please take a look at `PodOperationsImpl`, lines 279-300.
Looks like there was a copy-paste error, so that `withTTY` option where always set to true in irrelevant config methods: `withPrettyOutput()`, `tailingLines()`, `sinceTime()`, `sinceSeconds()`, `terminated()`. I took the liberty of fixing this too.


iocanel:
ok to test

iocanel:
LGTM [merge]

Issue #674, opened 2017-02-23
Watch's onClose method not being called when the Watch closes


ash211:
Between 1.4.34 and 2.0.3 of this library we're seeing a regression in the behavior of the Watch class.  After that change, the `onClose` method of a Watch object is no longer called when the Watch closes.

I'm seeing this in my logs as well mentioning that the onClose is being ignored:

```
2017-02-15 16:06:22 INFO  WatchConnectionManager:158 - WebSocket successfully opened
2017-02-15 16:06:22 INFO  WatchConnectionManager:158 - WebSocket successfully opened
2017-02-15 16:06:22 INFO  WatchConnectionManager:158 - WebSocket successfully opened
2017-02-15 16:06:22 INFO  WatchConnectionManager:158 - WebSocket successfully opened
2017-02-15 16:06:29 INFO  WatchConnectionManager:296 - WebSocket close received. code: 1000, reason:
2017-02-15 16:06:29 WARN  WatchConnectionManager:298 - Ignoring onClose for already closed/closing websocket
2017-02-15 16:06:29 INFO  WatchConnectionManager:296 - WebSocket close received. code: 1000, reason:
2017-02-15 16:06:29 WARN  WatchConnectionManager:298 - Ignoring onClose for already closed/closing websocket
2017-02-15 16:06:29 INFO  WatchConnectionManager:296 - WebSocket close received. code: 1000, reason:
2017-02-15 16:06:29 WARN  WatchConnectionManager:298 - Ignoring onClose for already closed/closing websocket
2017-02-15 16:07:16 INFO  WatchConnectionManager:296 - WebSocket close received. code: 1000, reason:
2017-02-15 16:07:16 WARN  WatchConnectionManager:298 - Ignoring onClose for already closed/closing websocket
```

Can we please debug the cause of this regression and restore the `onClose` call?  My project relied on it being called during shutdown.

Downstream project links:
- First discovered in the downstream project at https://github.com/apache-spark-on-k8s/spark/issues/120#issuecomment-280185028
- Partially worked around by manually replicating the `onClose` behavior in https://github.com/apache-spark-on-k8s/spark/pull/121
- Continuing to cause problems at https://github.com/apache-spark-on-k8s/spark/issues/143

cc @kimoonkim @foxish

PR #688, opened 2017-03-01
Fix watch close

foxish:
Fixes https://github.com/fabric8io/kubernetes-client/issues/674.
It was a regression introduced by https://github.com/fabric8io/kubernetes-client/pull/634 which made the `onClose()` event not get called when `close()` is called. 

cc @iocanel @jimmidyson @ash211 @gyoetam

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
@@ -149,7 +149,7 @@ public class WatchConnectionManager<T extends HasMetadata, L extends KubernetesR
       @Override
       public void onFailure(WebSocket webSocket, Throwable t, Response response) {
         if (forceClosed.get()) {
-          logger.warn("Ignoring onFailure for already closed/closing websocket", t);
+          logger.debug("Ignoring onFailure for already closed/closing websocket", t);
           // avoid resource leak though
           if (response != null && response.body() != null) {
             response.body().close();
@@ -187,7 +187,7 @@ public class WatchConnectionManager<T extends HasMetadata, L extends KubernetesR
         }
 
         if (currentReconnectAttempt.get() >= reconnectLimit && reconnectLimit >= 0) {
-          watcher.onClose(new KubernetesClientException("Connection failure", t));
+          closeEvent(new KubernetesClientException("Connection failure", t));
           return;
         }
 
@@ -223,8 +223,8 @@ public class WatchConnectionManager<T extends HasMetadata, L extends KubernetesR
               webSocketRef.set(null); // lose the ref: closing in close() would only generate a Broken pipe
                                       // exception
               // shut down executor, etc.
+              closeEvent(new KubernetesClientException(status));
               close();
-              watcher.onClose(new KubernetesClientException(status));
               return;
             }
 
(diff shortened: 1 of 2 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
```diff
@@ -289,8 +289,8 @@ public void execute() {
                 // An unexpected error occurred and we didn't even get an onFailure callback.
                 logger.error("Exception in reconnect", e);
                 webSocketRef.set(null);
+                closeEvent(new KubernetesClientException("Unhandled exception in reconnect attempt", e));
```
ash211: is the specific fix this reordering before the `close()`?  the other changes look like a refactoring / log level adjustments

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
```diff
@@ -308,7 +308,7 @@ public void waitUntilReady() {
   @Override
   public void close() {
     logger.debug("Force closing the watch {}", this);
-    forceClosed.set(true);
+    closeEvent(null);
```
ash211: this `null` looks like it could make it through to a parameter of the `watcher.onClose` callback.  Is the contract that the cause can be nullable?  I'm guessing not all custom watches check for null in their `onClose` methods

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
```diff
@@ -323,6 +323,15 @@ public void close() {
     }
   }
 
+  private void closeEvent(KubernetesClientException cause) {
+    if (forceClosed.get()) {
```
mccheah: One small tweak is to use `getAndSet(true)` to remove the `set(true)` call later.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
foxish: https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Watcher.java#L27

It is part of the contract.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
foxish: The reordering was because we allow `onClose()` to be called exactly once. The fix is `closeEvent(null);` within the `close()` method, which ensures that in case of a user initiated closure, we do try and call `onClose()` if we haven't previously.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
foxish: Thanks! Done.

jimmidyson:
OK to test

jimmidyson:
Will take a better look later but initial look LGTM. Thanks for this PR.

foxish:
I'll likely have a similar fix for the HTTP watcher sometime in the near future, but it isn't high priority.
Would you also be able to cut a release after this PR is merged? Thanks!

jimmidyson:
Great work and thanks for the test! [merge]

jimmidyson:
@foxish Released 2.2.1 including this fix.

PR #698, opened 2017-03-10
fix inconsistency between Job api version in model and api path in client

### LINKED ISSUE #697: Inconsistency between Job api version in model and api path in client
Using the kubernetes-client version 2.2.1 and so the kubernetes-model version 1.0.67 (which is a dependency of kubernetes-client version 2.2.1) it's not possible to manage kubernetes job anymore : the kubernetes api server complains `the API version in the data (batch/v2alpha1) does not match the expected API version (extensions/v1beta1)`

nb: relates to #689 and #611

jbguerraz:
Fix #697


jstrachan:
ok to test

iocanel:
LGTM [merge]

PR #702, opened 2017-03-12
WIP: Support for `Expect: 100-continue` header

iocanel:
This support is required when uploading data for binary OpenShift builds since this is used by OpenShift for validating a request before accepting the request upload stream.

For this to work, the latest, yet non-released version of OkHttp 3.6.0-SNAPSHOT is required. Support for this feature came in with https://github.com/square/okhttp/commit/a589b8170333476233d48476587a1d4363c90bd0

Since since OkHttp 3.5.0 the WebSocket handling has changed completely, this library needs to be updated for using this new WS scheme. Currently the tests are not running:

* fabric8/mockwebserver needs to be updated to use at least 3.5.0 as well since the tests are using a mixing fabric8/mockwebserver (depending on 3.4.x) and square/okhttp's MockServer (from 3.5.x) which doesn not mix at all.

Also the kubernetes clients needs to removed the depedency on okhttp-ws since the WS support is now directly available within OkHttp directly.


iocanel:
[merge]

Issue #705, opened 2017-03-15
Feature request: Pod port-forwarding


rhuss:
No idea whether it's a realistic feature but it would be awesome if kubernetes-client would support a functionality like `kubectl port-forward` in order to access a pod directly. 

Concrete use cases are from fabric8-maven-plugin:

* Open a remote debug port to a pod (currently: Starting an external kubectl)
* Open a port to a spring-boot app to allow hot-deployment with the devtools (currently: Done via a service and requiring exposecontroller)

Issue #534, opened 2016-09-30
Implement portforward


iocanel:
We already use it in a handful of places, using the binaries. It would be awesome if the client supported it.

iocanel:
@rcernich, @jwendell, @alesj, @spolti  have already done a great job implementing something like this, here: https://github.com/arquillian/arquillian-cube/blob/master/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/PortForwarder.java

I am pretty sure, that it wouldn't be that hard to port that code to use okhttp3 and the rest of the facilities we use in the `kubernetes-client`.

Guys would you be interested / have the cycles to port the port forwarding logic directly into the client?

rcernich:
@iocanel I couldn't figure out how to make this work with okhttp.  I needed access to both the request and response streams so I could transfer between two connections, i.e. I couldn't figure out how to handle long running requests.  Also, the k8s spdy implementation is not complete, so we needed to add some workarounds to get it to work with undertow.  Specifically, it doesn't send window updates, so the client would eventually stop sending data once it exhausted the send window.

iocanel:
@rcernich: I have no clue on how portforward works internally, so I can't really understand the challenges :-)

Let me have a look...

rcernich:
@iocanel, it looked like okhttp was setup to send a request, then receive a response.  For port forwarding, you need to keep the request and response streams open simultaneously, since the application protocol may be something other than simple http requests, e.g. eap management, jdbc, jms, etc.

kolleroot:
Hi there, are there any updates regarding this issue?
It would be really great to have this feature in the library, because a hack with `kubectl port-forward` would be very messy.

I think, that #417 is also relevant to this issue, because in the kubectl command they also use the sub resource construct to access the port forward resource (https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/portforward.go#L194-L198). But I think it just appends `/portforward` to the pod resource url. So #417 is not a requirement but would generalize the accessing of sub resources.

PR #707, opened 2017-03-20
Implementing port-forward (websockets)

nicolaferraro:
This is related to #705 and #534.
I've played a bit and managed to implement port-forward, but only for kubernetes 1.6+.

Kubernetes uses SPDY as handshaking protocol for establishing data channels, and recently has introduced support for websockets. Websockets are just used to establish the connection, then binary data is exchanged.

I implemented 2 versions:
1) Bind a specific or random local port to a remote port in a pod (equivalent of `kubectl port-forward`)
2) Exchange data directly with the pod using java input/output channels

I tested it with HTTP resources but also binary channels (remote mysql database connected with local mysql workbench). I used Minikube with Kubernetes 1.6.beta3.

It's not difficult to implement the SPDY version (to support current versions of kubernetes/openshift), but OkHttp has dropped support for SPDY, so we should add another SPDY capable client library if we want it... Is there a way to put SPDY back?


iocanel:
[merge]

PR #721, opened 2017-04-02
Move server-mock objects into their own modules, so that we separate …

iocanel:
…test code, from code that is meant to be used by users.


iocanel:
[merge]

Issue #720, opened 2017-04-01
Provide a shaded version of the client that downstream projects can use internally.


iocanel:
Every now and then we get into a discussion about how to handle version mismatches between downstream projects.

In some of those case it makes sense to `shade` the client in order to avoid such conflicts (e.g. the [arquillian-cube](https://github.com/arquillian/arquillian-cube/) community has decided that this is the way to go).

The problem is that in recent versions of maven shading a project in one maven module and using the shaded module in an other module of a single maven project doesn't really work (see below). 

Related issues: 

https://issues.apache.org/jira/browse/MSHADE-206
https://issues.apache.org/jira/browse/MNG-5899

So, it would make sense if we provided the shaded version for them.

PR #723, opened 2017-04-02
Add an uberjar and an uberjar with versioned packages (majro / minor).

iocanel:
This addresses #720.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -94,6 +94,7 @@
 
     <scr.annotations.version>1.9.12</scr.annotations.version>
     <sundrio.version>0.3.12</sundrio.version>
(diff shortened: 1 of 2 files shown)


iocanel:
@jimmidyson: If you have the cycles I'd like you to have a very quick look.

jimmidyson:
How does this handle transitives?

iocanel:
@jimmidyson: Transitives are maintained (as transitives) without being included in the uberjar itself. We could possibly include `okhttp` inside the uberjar, though since okhttp already includes the major version inside the package, we are relatively safe.

jimmidyson:
OK let's try this for first cut and see if we need to shade more in future.

[merge]

PR #724, opened 2017-04-03
Add a dummy interface inside the uberjar in order to force generation…

iocanel:
… of sources and javadoc jars.


iocanel:
[merge]

PR #725, opened 2017-04-04
fixes #511 to add a withSelector(LabelSelector) method on the DSL

### LINKED ISSUE #511: should we have a pods().withSelector(LabelSelector)
we have various withLabel() methods like `withLabels(Map<String,String>)`, `withLabel(String, String)` but should we have a withSelector(LabelSelector) which then wraps up iterating through the LabelSelector turning it into matchLabels etc?

jstrachan:


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Filterable.java
```diff
@@ -38,5 +42,7 @@
   T withFields(Map<String, String> labels);
 
   T withField(String key, String value);
+
+  T withSelector(LabelSelector selector);
```
jimmidyson: Can we call it `withLabelSelector` please, just to be consistent with other method names?

jstrachan: sure

jimmidyson:
Thanks! Will merge on green :)

jimmidyson:
[merge]

PR #727, opened 2017-04-06
Allow non-required fields to be unset when processing templates locally

lulf:
When switching from remote to local template processing using processLocal, I noticed a difference in how template parameters are handled when they are marked as required or not.

With remote processing, non-required parameters with unset values are allowed, while the local processing throws an exception. 

This fix checks the required field before throwing the exception, and sets the value to an empty string if not (as done when doing remote processing).


jimmidyson:
LGTM - ok to test

lulf:
Is there anything else needed to have this merged?

iocanel:
[merge]

Issue #711, opened 2017-03-27
Certificates from system keystore are ignored by kubernetes client


ibuziuk:
There are two possible ways for providing certificates while creating OpenShift / K8S client:

- `KUBERNETES_CERTS_CA_FILE` - using the **ENV VAR** and creating client using default constructor
- `withCaCertFile()` - providing path to the certificate file if client is created via` ConfigBuilder`

However, if certificate is already installed in the system, client will ignore it and would expect certificate to be explicitly set during  creation [1]

It looks like the kubernetes client creates an empty keystore and install the provided certificate into it.
Basically, this approach ignores system keystore and valid certificates from it are simply not taken into account.

The issue should be reproducible for everyone who uses k8-client to talk to a different cluster, as it seems to assume only this clusters key is valid and ignore all other root chains.

[1] https://github.com/redhat-developer/che-starter/blob/master/src/main/java/io/fabric8/che/starter/openshift/OpenShiftClientWrapper.java#L64

ethanfrogers:
I'm dealing with a similar issue. From within a pod, I'm not able to communicate with the cluster using the cert from the ServiceAccount. The cert is self-signed.

bartoszmajsak:
I cannot assign, but @matousjobanek is working on it.

ibuziuk:
@bartoszmajsak awesome!!

bartoszmajsak:
@iocanel can you assign @matousjobanek to this issue?

MatousJobanek:
Hi,
I think that this is caused by the fact that there is no file that would be loaded when the KeyStore is created:
https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/CertUtils.java#L68
If I try (locally) to change the logic to:
~~~java
    char[] passphrase = "secrets".toCharArray();
    trustStore.load(new FileInputStream("/home/mjobanek/.keystore"),passphrase);
~~~
then it works.

Similar for the keystore: https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/CertUtils.java#L100

@iocanel could you shed more light on this, please? Is there any other way to load the truststore/keystore? If not, would be possible to load a default truststore/keystore file or some other file that would be specified using a parameter?

iocanel:
[...]

PR #738, opened 2017-04-19
issue #711 KeyStore tries to load both default keystore files

MatousJobanek:
When KeyStore is initialized it tries to load keystore file "~/.keystore" with the default passphrase "changeit"
if the file is not present or if there is any problem when the file is loaded, it tries to load the file "java_jre_path/lib/security/cacerts"
if none of the files can be loaded, then an empty KeyStore is created.


jimmidyson:
This is looking really good - sorry for the slow review!

I just read the following details from http://docs.oracle.com/javase/8/docs/technotes/guides/security/jsse/JSSERefGuide.html#InstallationAndCustomization and more further down on that page (sorry not linkable...):

```
If a null KeyStore parameter is passed to the SunJSSE PKIX or SunX509 TrustManagerFactory, then the factory uses the following process to try to find trust material:

If the javax.net.ssl.trustStore property is defined, then the TrustManagerFactory attempts to find a file using the file name specified by that system property, and uses that file for the KeyStore parameter. If the javax.net.ssl.trustStorePassword system property is also defined, then its value is used to check the integrity of the data in the truststore before opening it.

If the javax.net.ssl.trustStore property is defined but the specified file does not exist, then a default TrustManager using an empty keystore is created.

If the javax.net.ssl.trustStore system property was not specified, then:
if the file java-home/lib/security/jssecacerts exists, that file is used;
if the file java-home/lib/security/cacerts exists, that file is used;
if neither of these files exists, then the SSL cipher suite is anonymous, does not perform any authentication, and thus does not need a truststore.
For information about what java-home refers to, see The Installation Directory.

The factory looks for a file specified via the javax.net.ssl.trustStore security property or for the jssecacerts file before checking for a cacerts file. Therefore, you can provide a JSSE-specific set of trusted root certificates separate from ones that might be present in cacerts for code-signing purposes.
```

So this should really support system properties `javax.net.ssl.trustStore` and `javax.net.ssl.keyStore` before looking for defaults paths. Does that sound reasonable?

MatousJobanek:
Good point @jimmidyson 
Please, take a look at my latest commit - I've added a check of the system properties there.
The only thing I'm not sure about is if we should check the `~/.keystore` file automatically. It is a default location for the `keytool`, but there is no mention about the file in the document. WDYT?

iocanel:
ok to test

jimmidyson:
retest this please

jimmidyson:
@MatousJobanek Sorry for the delay on this. I would prefer to stick to just the documented keystore/truststore locations so can you please remove the $HOME/.keystore default? Thanks!

jimmidyson:
Thanks @MatousJobanek! [merge]

PR #740, opened 2017-04-21
A set of fixes and improvements around the ExecWebSocketListener

iocanel:


iocanel:
@amisevsk: This pull request should address two of the issues you recently raised so your feedback here is more than welcome.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
```diff
@@ -90,33 +107,39 @@ public void close() {
...
-      if (!executorClosed.compareAndSet(false, true)) {
+  /**
+   * Performs the cleanup tasks:
+   * 1. closes the InputStream pumper
+   * 2. closes all internally managed closeables (piped streams).
+   *
+   * The order of these tasks can't change or its likely that the pumper will through errors,
+   * if the stream it uses closes before the pumper it self.
+   */
+  private void cleanUp() {
```
amisevsk: As far as I can tell, cleanUp() is called only if `ExecWatch.close()` is called. I feel like at least some clean up should be done if the WebSocket we're listening on is closed (e.g. close the write end of the output pipes).

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
```diff
@@ -218,7 +246,9 @@ public void onClosed(WebSocket webSocket, int code, String reason) {
         webSocketClosed.set(true);
         LOGGER.debug("Exec Web Socket: On Close with code:[{}], due to: [{}]", code, reason);
         try {
-            closeExecutor();
+            if (cleanupOnClose.get()) {
```
amisevsk: This will be false if the websocket is closed before the ExecWatch, and so `cleanUp()` won't be called. I think that `out` and `err` should be closed if applicable on WebSocket close, regardless of whether `EWSL.close()` has been called.

This would signal to `InputStreamPumper` that no more output is coming, and so that thread can complete.

kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/ExecLoopExample.java
```diff
@@ -45,22 +48,34 @@ public static void main(String[] args) throws InterruptedException, IOException
...
+          InputStreamPumper pump = null;
+          try {
+               watch = client.pods().withName(podName).redirectingOutput().exec("date");
+               pump = new InputStreamPumper(watch.getOutput(), new SystemOutCallback());
+               executorService.submit(pump);
+               Future<String> outPumpFuture = executorService.submit(pump, "Done");
+               executorService.scheduleAtFixedRate(new FutureChecker("Pump " + (i + 1), outPumpFuture), 0, 2, TimeUnit.SECONDS);
+
+               //We need to wait or the pumper (ws -> System.out) will not be able to print the message.
+               Thread.sleep(1000);
```
amisevsk: If instead we do 

``` java
while(!outPumpFuture.isDone()) {
    Thread.sleep(1000);
}
```

here, execution gets stuck in that while loop until the pipe dies. This makes it hard to deal with exec commands where we don't necessarily know how long to wait before closing.

kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/ExecLoopExample.java
```diff
@@ -45,22 +48,34 @@ public static void main(String[] args) throws InterruptedException, IOException
...
-              executorService.submit(pump);
-              Thread.sleep(2000);
+          ExecWatch watch = null;
+          InputStreamPumper pump = null;
+          try {
+               watch = client.pods().withName(podName).redirectingOutput().exec("date");
+               pump = new InputStreamPumper(watch.getOutput(), new SystemOutCallback());
+               executorService.submit(pump);
+               Future<String> outPumpFuture = executorService.submit(pump, "Done");
+               executorService.scheduleAtFixedRate(new FutureChecker("Pump " + (i + 1), outPumpFuture), 0, 2, TimeUnit.SECONDS);
```
amisevsk: This seems to log "Future [...] is not done" until we reach the `finally` block and close things manually.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
amisevsk: To add: if I remove the conditional check and just naively call `cleanUp()`, I get the behaviour I expect -- the future returned by submitting the `InputStreamPumper` is done as soon as the exec is finished in the pod.

amisevsk:
For context, [here's](https://github.com/eclipse/che/blob/4b14c8aadf4329c289ac77d27c0e9e95f0fa3d2c/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java#L877) how I'm trying to use `ExecWatch`. The command being executed in the pod can be fairly complex, so the `Thread.sleep()` of 2500ms is kind of arbitrary. I'd like to be able to call e.g. `outFuture.isDone()` in a loop until the exec is finished, and then proceed. If this isn't the intended use case, I'm happy to take feedback.

kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/ExecLoopExample.java
iocanel: This Future (the pumper) will never finish unless explicitly requested (via close) or until an error occurs. That doesn't mean that we have to use an error just to signal the user. I think that this must be handled gracefully and if the user needs to get signal about the exec status, he should use an `ExecListener`.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
iocanel: If we close the streams before the pumper, the pumper will consider this an error and will propagate Exception messages like `Pipe closed`, so we have to always close the pumper first and then the streams.

Now, if we implicitly close the streams, the user code may be still reading and receive a `Pipe closed` message.

I think that since we are handing the user a stream and a handle, we should not close the stream until the user decides to.

iocanel:
@amisevsk: We can't make assumptions on how/when the user is actually using the watch stream, so I think that we can't/shouldn't close the stream implicitly. 

So, I think that its not the InputStreamPumpers role to signal, when the execution is over. I mentioned in a previous comment that I would use the ExecListener:


          ExecWatch watch = null;
          InputStreamPumper pump = null;
          final CountDownLatch latch = new CountDownLatch(1);
          watch = client.pods().withName(podName).redirectingOutput().usingListener(new ExecListener() {
            @Override
            public void onOpen(Response response) {
            }

            @Override
            public void onFailure(Throwable t, Response response) {
              latch.countDown();
            }

            @Override
            public void onClose(int code, String reason) {
              latch.countDown();
            }
          }).exec("date");
          pump = new InputStreamPumper(watch.getOutput(), new SystemOutCallback());
          executorService.submit(pump);
          Future<String> outPumpFuture = executorService.submit(pump, "Done");
          executorService.scheduleAtFixedRate(new FutureChecker("Pump " + (i + 1), outPumpFuture), 0, 2, TimeUnit.SECONDS);

          latch.await(5, TimeUnit.SECONDS);
          watch.close();
          pump.close();


I am open to discussion though. So if you feel that I am missing something or anything, let's schedule a call or something to directly talk about it.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
amisevsk: I'm not sure I see the issue with closing the write end of the stream returned by watch.getOutput(). In my mind, once the websocket is closed, we're not going to get any more data to write to the stream, and so it should be safe to close the write end at that point (the rest of the ExecWatch can remain open). InputStreamPumper should not throw an exception in that case, since we can pull remaining data out of the buffer and should hit `-1` on read once all data has been obtained. As it stands, the last read will block, since the write end stays open after the websocket is closed. 

This was kind of the sticking point in my use case, as it's not clear that the InputStreamPumper thread will never finish on its own. By closing the write end of the output and error streams, we could signal to InputStreamPumper that we're done writing, and once all input in the buffer has been read, it's safe to close.

kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/ExecLoopExample.java
amisevsk: Shouldn't the pumper finish when it has pumped all of the input? Is there a way to restart writing output to the `out` pipe after the websocket is closed? I see your point, and realize I'm probably missing a use case somewhere here.

amisevsk:
I think I see the issue. The ExecListener example may be the answer to my problem, it seems to work well. I still think I'm missing a piece of this use case, though (probably due to lack of familiarity). If you would like to discuss further, we can arrange a good time for a call -- I'll send you an email.

Either way, I'm happy with this result, and using an ExecListener should work well enough.

iocanel:
[merge]

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
pearj: I can confirm that this is indeed true. I had to call ExecWatch.close inside an ExecListener which is less than ideal (because close calls the listener again). Otherwise I've found it deadlocks if you read the stdout on the PipeInputStream and no data has flowed and the websocket behind closes.  For example if the exec causes stderr only and no stdout, but something is already reading on the stdout PipeInputStream. 
I think the PipeOutputStream that is connected to PipeInputStream should always be closed when the onClose method is called. The problem is that ExecWatch.close() closes both the Pipe Input and Output Stream. So it's not safe to call in all conditions.

PR #748, opened 2017-04-28
edit() / getMandatory() now return a more meaningful error on 404

iocanel:
Fixes https://github.com/redhat-ipaas/ipaas-rest/issues/328. Well not exactly fixes, but it does provide a better error message.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
@@ -18,11 +18,10 @@ package io.fabric8.kubernetes.client.dsl.base;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.LabelSelector;
 import io.fabric8.kubernetes.api.model.LabelSelectorRequirement;
-import io.fabric8.kubernetes.api.model.ReplicationController;
-import io.fabric8.kubernetes.client.KubernetesClientTimeoutException;
+import io.fabric8.kubernetes.client.OperationInfo;
 import io.fabric8.kubernetes.client.dsl.Resource;
+import io.fabric8.kubernetes.client.dsl.internal.DefaultOperationInfo;
 import io.fabric8.kubernetes.client.internal.readiness.Readiness;
-import io.fabric8.kubernetes.client.internal.readiness.ReadinessWatcher;
 import io.fabric8.kubernetes.client.utils.Utils;
 import okhttp3.HttpUrl;
 import okhttp3.OkHttpClient;
@@ -31,7 +30,6 @@ import io.fabric8.kubernetes.api.builder.Function;
 import io.fabric8.kubernetes.api.model.Doneable;
 import io.fabric8.kubernetes.api.model.KubernetesResourceList;
 import io.fabric8.kubernetes.api.model.RootPaths;
-import io.fabric8.kubernetes.api.model.StatusBuilder;
 import io.fabric8.kubernetes.client.Config;
 import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.Watch;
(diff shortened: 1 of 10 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
```diff
@@ -24,6 +24,7 @@
 import io.fabric8.kubernetes.client.internal.readiness.Readiness;
 import io.fabric8.kubernetes.client.internal.readiness.ReadinessWatcher;
 import io.fabric8.kubernetes.client.utils.Utils;
+import io.fabric8.zjsonpatch.internal.guava.Strings;
```
jimmidyson: Is this the right `Strings`? Not in OSGi exported package.

jimmidyson: Actually it's unused so can be removed.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
```diff
@@ -222,7 +223,9 @@ public T getMandatory() {
     if (item != null) {
       return item;
     }
-    throw new KubernetesClientException("Resource not found!", 404, new StatusBuilder().withCode(404).withMessage("Not found.").build());
```
jimmidyson: Do we not get a nice Status back for a 404? Pretty sure we should do now - maybe this was just a legacy thing?

An example response:

```
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "pods \"jimmi\" not found",
  "reason": "NotFound",
  "details": {
    "name": "jimmi",
    "kind": "pods"
  },
  "code": 404
}
```

jimmidyson: Ah get returns null. Perhaps we should refactor that so `get` calls `getMandatory` instead and `get` returns null if it's a 404? That way we can use the proper error from the `GET` rather than this synthetic error.

iocanel: Actually getMandatory() calls get() so we can't

jimmidyson: Like I said, flip it round so we can :)

iocanel: Not sure what you mean by better status.

If you are talking about the status message that is part of KubernetesClientException, then yes it does have a nicer message now.

jimmidyson: What I mean is if we change this round so that `get` calls `getMandatory` then we can return the proper error from Kubernetes as we'll have the status from response body.

iocanel: yes, which doesn't contain the namespace

iocanel: and its pretty common to get a 404, due to pointing to the wrong namespace, right?

iocanel: @jimmidyson: ping ^^^

jimmidyson: So perhaps we should include details of the source operation in the exception? Wonder if that would be easy enough to do? So we can use the raw message back from kubernetes, but with details of what the caller was trying to do?

iocanel: @jimmidyson: We can modify the `KubernetesClientException` in order to accept an additional object, that contains (name, namespace, kind, and type of opertation). The 3 first are available to all operations as the are part of `BaseOperation` the type of operation is method specific and needs to be passed when we call the `launderException`.

This can work and in fact I have implemented something like this locally and plays nicely. However, this requires a minor version bump. So how about we merge it as is, for 2.2.x and I'll implement it as described for 2.3.

iocanel:
[merge]

PR #751, opened 2017-05-04
fixes #746 to support loading templates with number parameters

### LINKED ISSUE #746: Provide option to parse Templates without really parsing the underying objects using the typesafe DTO model
Issue:  When users specify replicas as a parameter in the templates 

replicas: '${REPLICAS}'
 
 and below line is executed
 
`List<Template> tlist = sourceClient().templates().inNamespace(namespace).withLabels(labelMap).list().getItems();`
 
it fails with an error 

```
caused by: com.fasterxml.jackson.databind.exc.InvalidFormatException: Can not deserialize value of type java.lang.Integer from String "${REPLICAS}": not a valid Integer value
at [Source: N/A; line: -1, column: -1] (through reference chain: io.fabric8.openshift.api.model.TemplateList["items"]->java.util.ArrayList[0]->io.fabric8.openshift.api.model.Template["objects"]->java.util.ArrayList[0]->io.fabric8.openshift.api.model.DeploymentConfig["spec"]->io.fabric8.openshift.api.model.DeploymentConfigSpec["replicas"])

```

We are using 2.2.7 version

jstrachan:
now the processLocal(map) allows the numeric parameters to be replaced so that the template can be properly loaded providing the numeric properties are in the supplied map

### DIFF
--- a/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/TemplateOperationsImpl.java
+++ b/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/TemplateOperationsImpl.java
@@ -22,6 +22,8 @@ import io.fabric8.kubernetes.api.model.KubernetesList;
 import io.fabric8.kubernetes.api.model.KubernetesListBuilder;
 import io.fabric8.kubernetes.api.model.KubernetesResourceList;
 import io.fabric8.kubernetes.client.KubernetesClientException;
+import io.fabric8.kubernetes.client.dsl.MixedOperation;
+import io.fabric8.kubernetes.client.utils.Serialization;
 import io.fabric8.kubernetes.client.utils.URLUtils;
 import io.fabric8.kubernetes.client.utils.Utils;
 import io.fabric8.openshift.api.model.DoneableTemplate;
@@ -50,6 +52,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
+import java.util.concurrent.ExecutionException;
 
 public class TemplateOperationsImpl
   extends OpenShiftOperation<Template, TemplateList, DoneableTemplate, TemplateResource<Template, KubernetesList, DoneableTemplate>>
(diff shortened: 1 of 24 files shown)


openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/TemplateOperationsImpl.java
```diff
@@ -214,4 +220,22 @@ private URL getProcessUrl() throws MalformedURLException {
     return new TemplateOperationsImpl(client, OpenShiftConfig.wrap(config), null, namespace, null, false, template, null, false, 0L,
       null, null, null, null, null);
   }
+
+  @Override
+  protected <T> T unmarshal(Class<T> type, InputStream is) throws IOException {
+    if (replaceValueStream != null) {
```
iocanel: Apparently, I missing something here. `load()` is a method which I expect it to be called before `processLocally()` (which is the method that sets value to `replaceValueStream`) . So I don't see how its possible this not to be null.

openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/TemplateOperationsImpl.java
```diff
@@ -139,6 +140,8 @@ public KubernetesList processLocally(ParameterValue... values) {
   }
 
   public KubernetesList processLocally(Map<String, String> valuesMap)  {
+    this.replaceValueStream = new ReplaceValueStream(valuesMap);
```
iocanel: To answer my own question: `It's becase load() may call get() which may call unmarshal`.

openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/TemplateOperationsImpl.java
iocanel: Never mind, I figured it out :-)

jstrachan:
@iocanel BTW my first attempt was to try custom deserialisers; turns out thats kinda complex as there are many numeric, date and boolean types in Java that all need changing - plus some Strings validation won't work with template expressions either. So a filter seems the simplest thing that could possibly work.

I hear you on the mutability thing though. Maybe we need a separate new method on TemplateOperations to `expandProperties(Map)` or something?

jstrachan:
OK have added a `replaceParameters(Map)` method now that returns a new immutable `TemplateOperation` which is then internally reused by the `processLocally(Map)` method. It also means folks can load the template with parameter replacement too:
```
Template template = client.templates().withName("foo").replaceParameters(map).get();
```

iocanel:
@jstrachan: I've implemented something like this on top of your pull request.

    client.templates().withParameters(myparms).list()
    client.templates().withParameters(myparams).withName('cheese').get()
    client.templates().withParameters(myparams).inNamespace('mynamespace').withName('cheese').get()

The `withParameters` is exclusive to template operations.

wdyt?

jstrachan:
@iocanel looks awesome thanks!

jstrachan:
@iocanel am happy to merge if you are?

iocanel:
[merge]

PR #755, opened 2017-05-09
Deprecate empty constructor of Config, so that we can remove its functionality in future releases

iocanel:
The problem is simple. Our generated builder make use of the default constructor internally. By autoconfioguring Config as part of the default constructor we make it impossible to distinguish what's user provided and what's autoconfigured.

We **MUST** remove everything from the default constructor but first we need to warn people about it. So let's depracate for now and we remove as of 2.4.


iocanel:
[merge]

PR #756, opened 2017-05-09
Use decodeUrl in CertUtilsTest to ensure that the test runs ok even w…

iocanel:
…hen url escaped characters are present in the path.


jimmidyson:
Weird one but ok... [merge]

PR #768, opened 2017-05-30
Fixes #767: adding lock replace

### LINKED ISSUE #767: Unable to leverage optimistic locking during patch/replace
Currently:
- `.patch()` ignores the `resourceVersion` field while computing the diff
- `.replace()` updates the resource version with latest version found on server (trying 10 times to get the most recent one).

There seems to be no way to do **optimistic locking** of resources (e.g. the first one who tries to update a resource wins).

Libraries doing leader election like [camel-master](https://github.com/fabric8io/fabric8-ipaas/blob/master/camel-master/src/main/java/io/fabric8/camel/master/KubernetesLock.java#L183) simply cannot work: once the first leader loses the leadership, anybody becomes the leader... this is elsewhere called *anarchy*. 

Deleting the resource and recreating it is not an option in the case of camel-master, since it's shared.

@iocanel, @jstrachan thoughts?

nicolaferraro:
Adding a `.lockResourceVersion("...")` to replace to perform optimistic locking.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Lockable.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Lockable.java
@@ -13,11 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package io.fabric8.kubernetes.client.dsl;
 
-public interface CascadingEditReplacePatchDeletable<I,D,T,B> extends
-  EditReplacePatchDeletable<I,D,T,B>,
-  Cascading<EditReplacePatchDeletable<I,D,T,B>>,
-  Lockable<Replaceable<I, I>> {
+public interface Lockable<T> {
+
+  T lockResourceVersion(String resourceVersion);
 }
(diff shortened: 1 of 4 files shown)


nicolaferraro:
Ping @iocanel

iocanel:
@nicolaferraro: Sorry for the delay! 

Now, this will have to go in 2.5.x since it introduces and uses a new interface in the DSL.

nicolaferraro:
It's related to https://github.com/fabric8io/kubernetes-client/issues/767

iocanel:
This should have been part of 2.5 but, that didn't happen so I am bumping to 2.6.

nicolaferraro:
Is there something I can do to get this merged (it has been waiting a lot :wink: now)?

I don't think it's such a breaking change in the DSL, since it just adds a new method. Anyway, I can also bump global version to 2.6-SNAPSHOT in this PR if it's needed.

oscerd:
Maybe you can switch to 2.6-SNAPSHOT and then we can merge.

iocanel:
I bumped to 2.6-SNAPSHOT, so I think we are good to go

iocanel:
@nicolaferraro: sorry for the delay. This should have made it to the 2.5.x but it slipped.

Regarding, the change not being breaking. It doesn't really matter how big or small a change is. Changes to the public api needs at least a minor version bump.

nicolaferraro:
Thanks @iocanel, I understand. It stayed here for 2 months mainly because I forgot about it :)

Thanks for meeting it.

Issue #763, opened 2017-05-22
apiVersion in group batch of Job should support both v1 and v2alpha1


lvjiangzhao:
related [611](https://github.com/fabric8io/kubernetes-client/issues/611), [697](https://github.com/fabric8io/kubernetes-client/issues/697), [698](https://github.com/fabric8io/kubernetes-client/pull/698)
I wonder why we support v2alpha1 with v1 missing. At the moment, we could not use batch/v1 Job with this client. Is there a plan to involve this ?

piyush1594:
Right now in the latest client, you can use batch/v1 and extensions/v1beta1 Hence closing this.

PR #779, opened 2017-06-13
use v1 instead of v2alpha1 for Job api

popyy0101:
Related #763


joan38:
We need to merge https://github.com/fabric8io/kubernetes-model/pull/215 before or it will break

joan38:
Can anyone review this? It's been open since 13 Jun.

I understand the maintainers are busy and do it as volunteer, that's why we fix stuff ourself here.
If the number of maintainers is not enough then we should probably look in for new volunteer maintainer?

Kubernetes is a project moving fast, I don't think it's reasonable to have a client laying multiple versions beind.

Cheers

joan38:
This can be closed now as it's been superseded by https://github.com/fabric8io/kubernetes-client/pull/827

oscerd:
Closed.

oscerd:
Thanks @popyy0101

Issue #762, opened 2017-05-16
Delete Statefulset never returned because the k8s server never return the observedGeneration value


kennyzli:
`k8client.apps().statefulSets().inNamespace(namespace.serviceName).withName(name).delete()` which will keep waiting for ever. look into the code and find out the code find out the generation = 2 and observedGeneration never returned. which will leave the scale waiting forever

`RollableScalableResourceOperation -> waitUntilScaled
          long generation = t.getMetadata().getGeneration() != null ? t.getMetadata().getGeneration() : -1;
          long observedGeneration = getObservedGeneration(t);
          if (observedGeneration >= generation && Objects.equals(desiredReplicas, currentReplicas)) {
            queue.put(true);
          }
`

 kctl get statefulset -o yaml

`- apiVersion: apps/v1beta1
  kind: StatefulSet
  metadata:
    creationTimestamp: 2017-05-16T18:11:43Z
    generation: 2
    labels:
    ---
    namespace: default
    resourceVersion: "984668"
  spec:
    replicas: 0
    selector:
      matchLabels:
    template:
      metadata:
        creationTimestamp: null
        labels:
          app: node-mgr-pod
      spec:
        ------
  status:
    replicas: 0
`

khogeland:
Fix for this: https://github.com/fabric8io/kubernetes-client/pull/787

These values are optional (for all of RC, RS, SS), so there's no guarantee the current code won't break again in a future version of Kubernetes.

PR #787, opened 2017-06-15
Remove generation check from RollableScalableResourceOperation

khogeland:
[Generation](https://github.com/kubernetes/kubernetes/blob/master/pkg/api/types.go#L107-L110) and [ObservedGeneration](https://github.com/kubernetes/kubernetes/blob/master/pkg/api/types.go#L2454-L2456) are both optional, and as seen in https://github.com/fabric8io/kubernetes-client/issues/762, some versions of k8s may leave those fields unpopulated, causing a scale operation to hang forever. The scale-only operation shouldn't care about the generation anyway– if the current status has the correct number of replicas, the scale operation is complete.


khogeland:
Closing because this appears to be a bug in k8s 1.5. The API server is returning a garbage status with no generation and 0 replicas.

PR #790, opened 2017-06-19
OpenShiftOAuthInterceptor.authorize() should only throw IOException

gyoetam:
The authorize method should never wrap an underlying checked IOException
into an unchecked KubernetesClientException. The IOException is declared
in the calling method (intercept()) and is handled by okhttp's RealCall$
AsyncCall as part of the Interceptor contract. But a RuntimeException is
not handled and okhttp's Dispatcher thread simply exits in this case,
instead of notifying the listener through the onFailure() callback.

Specific scenario that fails without this fix:
A watch is connected via websocket and the authorization token times
out, so the server closes the connection. When the watch first tries to
reconnect, a socket read timeout occurs during the authorization phase.
- Without the fix. this timeout kills the Dispatcher thread and the
  WatchConnectionManager is never notified via the onFailure() callback.
- With the fix, RealCall$AsyncCall properly catches the IOException and
  calls onFailure(), allowing WatchConnectionManager to retry.

For other exceptions that may occur: KCE.launderThrowable() only wraps
checked exception; RuntimeException-s and Error-s simply passed through
the catch clause anyway. All checked exceptions thrown in the method are
IOExceptions now, which we are supposed to throw anyway.

### DIFF
--- a/openshift-client/src/main/java/io/fabric8/openshift/client/internal/OpenShiftOAuthInterceptor.java
+++ b/openshift-client/src/main/java/io/fabric8/openshift/client/internal/OpenShiftOAuthInterceptor.java
@@ -21,7 +21,6 @@ import okhttp3.Interceptor;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
-import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.utils.URLUtils;
 import io.fabric8.kubernetes.client.utils.Utils;
 import io.fabric8.openshift.client.OpenShiftConfig;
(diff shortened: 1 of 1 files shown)


gyoetam:
@iocanel: Could you take a look at the pr-merge build for this on jenkins, maybe restart it manually? Clicking the details just gives "Not found" for me.

jimmidyson:
ok to test

jimmidyson:
retest this please

gyoetam:
I see the same "Something is wrong with the build of this commit" message for PR-800, while all other PRs have a "This commit cannot be built " message. Is this even related to the actual PR, or is it some general issue with the jenkins jobs?

iocanel:
retest this please

piyush1594:
Hey, @gyoetam Would you like to please rebase and push again. Thanks

piyush1594:
@gyoetam Would you like to please take a look at this?

rohanKanojia:
I've opened a new PR #1293 since I wasn't able to update this, hope you don't mind.

PR #795, opened 2017-06-21
fix(apigroup): adds API Group support

### LINKED ISSUE #716: support the new OpenShift API Groups for using OpenShift APIs
the OpenShift API is now using API groups:

* authorization.openshift.io
* build.openshift.io
* apps.openshift.io
* template.openshift.io
* image.openshift.io
* project.openshift.io
* user.openshift.io
* oauth.openshift.io
* network.openshift.io
* route.openshift.io
* quota.openshift.io
* security.openshift.io

Origin `3.6.0-alpha.0 ` has the new API Groups for the OpenShift APIs: https://github.com/openshift/origin/releases/tag/v3.6.0-alpha.0 

So we should update the openshift client to detect if these API groups are present and if so use those instead of the `oapi/` path.

So we map `oapi/` -> `/apis/apps.openshift.io/` etc depending on the resource kind

jstrachan:
lets add support for discovery and use of the new API Groups for OpenShift resources if using OpenShift 3.6.x
otherwise stick to the legacy /oapi REST APIs
fixes #716

### DIFF
--- a/kubernetes-tests/src/test/java/io/fabric8/openshift/client/server/mock/BuildConfigs.java
+++ b/kubernetes-tests/src/test/java/io/fabric8/openshift/client/server/mock/BuildConfigs.java
@@ -1,0 +1,46 @@
+/**
+ * Copyright (C) 2015 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.fabric8.openshift.client.server.mock;
+
+import io.fabric8.kubernetes.client.KubernetesClientException;
+import io.fabric8.openshift.api.model.BuildConfig;
+import io.fabric8.openshift.api.model.BuildConfigList;
+import io.fabric8.openshift.client.DefaultOpenShiftClient;
+import io.fabric8.openshift.client.OpenShiftClient;
+
+import java.util.List;
+
+public class BuildConfigs {
+  public static void main(String[] args) {
+    try {
+      OpenShiftClient client = new DefaultOpenShiftClient();
+      BuildConfigList list = client.buildConfigs().list();
+      if (list == null) {
+        System.out.println("ERROR no list returned!");
+        return;
+      }
+      List<BuildConfig> items = list.getItems();
+      for (BuildConfig item : items) {
+        System.out.println("BuildConfig " + item.getMetadata().getName() + " has version: " + item.getApiVersion());
+      }
+    } catch (KubernetesClientException e) {
+      System.out.println("Failed: " + e);
+      e.printStackTrace();
+    }
+  }
+
+}
--- a/openshift-client/src/main/java/io/fabric8/openshift/client/DefaultOpenShiftClient.java
+++ b/openshift-client/src/main/java/io/fabric8/openshift/client/DefaultOpenShiftClient.java
@@ -149,6 +149,10 @@ public class DefaultOpenShiftClient extends BaseClient implements NamespacedOpen
     this(new OpenShiftConfigBuilder().build());
   }
 
+  public DefaultOpenShiftClient(boolean noServer) throws KubernetesClientException {
+    this(new OpenShiftConfigBuilder().withNoServer(noServer).build());
+  }
+
   public DefaultOpenShiftClient(String masterUrl) throws KubernetesClientException {
     this(new OpenShiftConfigBuilder().withMasterUrl(masterUrl).build());
   }
--- a/openshift-client/src/main/java/io/fabric8/openshift/client/DefaultOpenShiftClient.java
+++ b/openshift-client/src/main/java/io/fabric8/openshift/client/DefaultOpenShiftClient.java
@@ -15,10 +15,10 @@
  */
 package io.fabric8.openshift.client;
 
-import io.fabric8.kubernetes.api.model.DoneableLimitRange;
-import io.fabric8.kubernetes.api.model.KubernetesResourceList;
-import io.fabric8.kubernetes.api.model.LimitRange;
-import io.fabric8.kubernetes.api.model.LimitRangeList;
+import io.fabric8.kubernetes.api.model.*;
+import io.fabric8.kubernetes.api.model.apiextensions.CustomResourceDefinition;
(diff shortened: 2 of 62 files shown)


jstrachan:
ok to test

jstrachan:
the OpenShiftClient now uses the `/apis` to detect if its OpenShift and whether OpenShift API Groups are available. Adding this extra check by default also leaded to some test failures that either
* were not expecting a REST query to `/apis` in the mock server (I wonder if we can disable that easily in the openshift mock server to avoid future issues in tests?)
* did not assume any server at all (e.g. loading resources from disk or locally processing local templates).

For the former I tried to lazy evaluate the API Groups stuff when getting the openshift client from the mock server. For the latter I added a new constructor `new DefaultOpenShiftClient(true)` to indicate `noServer` which disables the API Groups check

openshift-client/src/main/java/io/fabric8/openshift/client/DefaultOpenShiftClient.java
```diff
@@ -311,102 +325,102 @@ public KubernetesListMixedOperation lists() {
 
   @Override
   public MixedOperation<Build, BuildList, DoneableBuild, BuildResource<Build, DoneableBuild, String, LogWatch>> builds() {
-    return new BuildOperationsImpl(httpClient, OpenShiftConfig.wrap(getConfiguration()), getNamespace());
+    return new BuildOperationsImpl(httpClient, OpenShiftConfig.withApiGroup(this, BUILD, OpenShiftConfig.wrap(getConfiguration())), getNamespace());
```
iocanel: 1. All operations can either get instantiated by the client, or by a ResourceHandler. So, adding handling of the api groups here only covers half of the cases. 

So I think that this needs to move inside the Operation impl itself.

2. All operations have a getConfiguration() method and in some edge cases we do use it. So, this change can potential break things (because it passes to the operation a config with custom urls).

If we move the login inside the operation itself and keep the config object clean, I think we should be fine.

kubernetes-tests/src/test/java/io/fabric8/openshift/client/server/mock/BuildConfigs.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.List;
+
+public class BuildConfigs {
+  public static void main(String[] args) {
+    try {
+      OpenShiftClient client = new DefaultOpenShiftClient();
+      BuildConfigList list = client.buildConfigs().list();
+      if (list == null) {
+        System.out.println("ERROR no list returned!");
```
jstrachan: the API should never return null

openshift-server-mock/src/main/java/io/fabric8/openshift/client/server/mock/OpenShiftServer.java
```diff
@@ -40,6 +41,16 @@ public KubernetesClient getKubernetesClient() {
   }
 
   public NamespacedOpenShiftClient getOpenshiftClient() {
+    if (!new OpenshiftAdapterSupport().isAdaptable(client)) {
+      throw new IllegalArgumentException("Client does not support OpenShift!");
```
jstrachan: just pushed a code comment to be more clear why - the issue is the lazy call to the `/apis` REST API (to detect which API groups are supported on the kubernetes/openshift cluster) was causing tests to fail due to the extra REST call appearing the mock server results; hence the eager query

openshift-client/src/main/java/io/fabric8/openshift/client/OpenShiftConfig.java
```diff
@@ -155,4 +205,48 @@ public long getBuildTimeout() {
...
+
+  public void setNoServer(boolean noServer) {
+    this.noServer = noServer;
+  }
+
+  /**
+   * Updates the list items if they have missing or default apiVersion values and the resource is currently
+   * using API Groups with custom version strings
+   */
+  public void updateApiVersion(KubernetesResourceList list) {
```
jstrachan: there are 2 methods; one for lists (KubernetesResourceList) and one for resources (HasMetadata) - that covers Template and TemplateList right?

jstrachan: have moved these methods into a separate `ApiVersionHelpers` class

openshift-client/src/main/java/io/fabric8/openshift/client/DefaultOpenShiftClient.java
jstrachan: just starting that - turns out to be a fairly massive change as the code needs the OpenShiftClient so that means all constructors for all operations + all the handlers need to change....

jstrachan: ah phew - I can just create another OpenShiftClient from the httpClient + config ;) - panic over!

openshift-client/src/main/java/io/fabric8/openshift/client/DefaultOpenShiftClient.java
```diff
@@ -140,6 +140,10 @@ public DefaultOpenShiftClient() throws KubernetesClientException {
     this(new OpenShiftConfigBuilder().build());
   }
 
+  public DefaultOpenShiftClient(boolean disableApiGroupCheck) throws KubernetesClientException {
```
iocanel: Personally, I wouldn't add an new constructor just for this. That's what the Config objects is here for. For things that are TOO common and TOO obvious (e.g. the master url) we can make an exception, but I don't think that this is the case here.

openshift-client/src/main/java/io/fabric8/openshift/client/OpenShiftConfig.java
```diff
@@ -55,6 +58,14 @@ public OpenShiftConfig(Config kubernetesConfig) {
     );
   }
 
+  public OpenShiftConfig(Config kubernetesConfig, String apiGroupUrl, String apiGroupVersion) {
+    this(kubernetesConfig,
+      getDefaultOpenShiftUrl(kubernetesConfig), getDefaultOapiVersion(kubernetesConfig), DEFAULT_BUILD_TIMEOUT
+    );
+    this.openShiftUrl = apiGroupUrl;
```
iocanel: ApiGroupUrl and apiGroupVersion are not exaclty user configuration and should not be handled here.
Also using the openshiftUrl as carrier for this, is dangerous and will cause issues.

The only thing that does make sense to be here is `disableApiGroupCheck`. Everything else should move into the implementation classes.

Each implementation class has a methods like `getResourceUrl()` and `getNamesacedUrl()` I would override them so that they handle the logic there. `OpenShiftOperation` might be good place for this.

jstrachan: OK I've moved the apiGroup / apiGroupVersion state out of the OpenShiftConfig class and into the OpenShiftOperation now with a little helper class `ConfigAndApiGroupsInfo` used to minimise disruption to all the constructors of operations & handlers

jstrachan:
ok to test

Issue #798, opened 2017-06-28
Support for Retry to k8s.default.svc


kameshsampath:
When using the client with Istio service mesh, the application is not able to connect to kubernetes.default.svc because of Envoy proxy (sidecar) is taking few seconds to get initialised, hence to avoid connection failures which will fade away after proxy comes up we can add some kind of retry mechanism

wang-xiaowu:
i have the same scenario to use this func,
sometimes i can't create or delete an istio sidecar with istioclient,
i think with this fun i can retry the operation when this problem occurs

manusa:
Retry is already built into the client. This issue is from 2017.
What version of the client are you using?
Are you sure it's the same issue, the client will automatically retry in several periods until the limit given it receives a `429` (Too Many Requests) or `5xx` (Server Error) HTTP status code:

https://github.com/fabric8io/kubernetes-client/blob/9e9e650f456aabaf5b8bf419b5f9744654e986e1/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/StandardHttpClient.java#L169

PR #800, opened 2017-06-28
Issue#789 Resolution

### LINKED ISSUE #789: Does this client supports drain command?
Does this client supports drain command?

https://kubernetes.io/docs/user-guide/kubectl/v1.6/#drain

If so, someone please share a sample with me.

kameshsampath:
Fixes #789  Added retry logic

 Required parameters in Config

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/HttpClientUtils.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/HttpClientUtils.java
@@ -1,12 +1,12 @@
 /**
  * Copyright (C) 2015 Red Hat, Inc.
- *
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
- *         http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/HttpClientUtils.java
```diff
@@ -18,167 +18,184 @@
...
                 Request authReq = chain.request().newBuilder().addHeader("Authorization", "Bearer " + config.getOauthToken()).build();
-                return chain.proceed(authReq);
+                response = chain.proceed(authReq);
+              }
+            } catch (ConnectException e) {
+              LOGGER.warn("Request failed retry {} ", retries);
+              try {
+                SECONDS.sleep(retryDelay);
+              } catch (InterruptedException e1) {
+                //dont worry
```
iocanel: We shouldn't ignore the fact that the thread gets interrupted.
Since the method signature does throw IOEXception, I would throw java.io.InterruptedIOException to signal that the interceptor was interrupted.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/HttpClientUtils.java
```diff
@@ -18,167 +18,184 @@
...
               }
-              return chain.proceed(request);
+              retries++;
+            } catch (IOException e) {
```
iocanel: In the same spirit I wouldn't even catch that exception and would let it propagate.

gastaldi:
https://github.com/jhalterman/failsafe is a great library to handling failures

gastaldi:
Looks like this PR and the original issue (#798) have gone stale. I'm closing it since it seems that it is no longer important

PR #806, opened 2017-07-05
Fix the log watcher

nicolaferraro:
Naming shadowing..

This should fix https://github.com/fabric8io/fabric8-maven-plugin/issues/964


jimmidyson:
[merge]

PR #808, opened 2017-07-06
Documents how to use the Watcher API

solsson:
Found this out only thanks to https://stackoverflow.com/questions/43623275/in-the-fabric8-kubernetes-client-events-api-what-does-watcher-action-indicate and https://stackoverflow.com/questions/43598572/in-a-watcher-in-the-fabric8-kubernetes-client-events-api-what-resources-can-i

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,26 @@ Service myservice = client.services().inNamespace("default").createNew()
                      .done();
 ```
 
+### Following events
+
+Use `io.fabric8.kubernetes.api.model.Event` as T for Watcher:
+
+```java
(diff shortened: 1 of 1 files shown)


iocanel:
ok to test

iocanel:
Thanks for taking the time and update docs. We desperately need help we can get in that front.

Could you please update your snippet so that we also show that watch returns a Watch object which acts as a handle, so that we can close the Watch when done?

Something like:

    Watch watch = client.events().inAnyNamespace().watch( ... );
    //do stuff
    watch.close();

or with `try with resources`

    try ( Watch watch = client.events().inAnyNamespace().watch( ... ) ) {
       //do stuff
    }

PR #820, opened 2017-07-14
fix null Object return true by using isReady() and add Node resource

VFT:
1. BaseOperation.isReady() return true when then resource object is null, I think it's a bug.
2. Add support for node when using isReady().
3. Fix conflicts in #792


iocanel:
ok to test

oscerd:
[merge]

PR #821, opened 2017-07-20
fix watching builds with openshift 3.6 and the API Groups REST API

jstrachan:


osiotest314:
ok to test

osiotest314:
ok to test

PR #832, opened 2017-08-02
Exec issues

iocanel:


oscerd:
[merge]

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
```diff
@@ -185,19 +197,25 @@ public void onOpen(WebSocket webSocket, Response response) {
 
   @Override
   public void onFailure(WebSocket webSocket, Throwable t, Response response) {
-      try {
-        //If we have closed the watch ignore everything
-        if (closed.get())  {
-          return;
+
+      //If we already called onClosed() or onFailed() before, we need to abort.
+      if (explicitlyClosed.get() || closed.get() || !failed.compareAndSet(false, true) ) {
```
jimmidyson: Calls to multiple atomics isn't in itself atomic... If you do need to do that, then consider AtomicReference to a holder object for the various states.

ckanner:
When I run the ExecExample，it occurs the error ```io.fabric8.kubernetes.client.KubernetesClientException: Upgrade request required```。I use kubernetes proxy, and set ```--disable-filter=true```。
I debug this problem, and find okhttpclient use ```HTTP 1.1``` to exec watch.
Is there any solution?

iocanel:
@ckanner: I think that as of https://github.com/kubernetes/kubernetes/pull/49534 the kubectl proxy should have no issues with upgrade requests. Never tried it myself though.

ckanner:
@iocanel I follow the issue kubernetes/kubernetes#49534 and set ```--disable-filter=true```, it doesn't solved. May be you can try it.

iocanel:
@ckanner: my understanding is that you also need a relatively new version of kubectl, no?

ckanner:
@iocanel my kubectl version is v1.7.6

PR #837, opened 2017-08-07
feat: Allow overriding the maximum concurrent requests per host.

iocanel:


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
```diff
@@ -101,6 +102,8 @@
   public static final Long DEFAULT_WEBSOCKET_TIMEOUT = 5 * 1000L;
   public static final Long DEFAULT_WEBSOCKET_PING_INTERVAL = 1 * 1000L;
 
+  public static final Integer DEFAULT_MAX_CONCURRENT_REQUESTS_PER_HOST = 10;
```
jimmidyson: The [okhttp default](https://github.com/square/okhttp/blob/c70e35802e984e8b9bf3fbf1479deb37c0bdfa43/okhttp/src/main/java/okhttp3/Dispatcher.java#L41) is 5. Should we stick with that or is 10 a better option for us for some reason?

jimmidyson: I guess in this case all requests are usually going to the same host, so in this case we're really setting the max concurrent requests.

iocanel: @jimmidyson: don't get what you want me to change.

iocanel: 5 seems too low to me and it can cause hard to diagnose issues. So 10 seems to be a better default.

iocanel:
I align with the okhttp defaults and merged

PR #844, opened 2017-08-17
Revert InputStreamPumper to a blocker read()

### LINKED ISSUE #843: InputStreamPumper should not use InputStream.available()
In version 2.6.1，InputStreamPumper used in.avaliable() at line 59. In the 'watchLog' processing, the InputStream is from okio.RealBufferedSource。According to okio.RealBufferedSouce's source code, available() returns buffer.size, the buffer never set until read(). 
But the InputStreamPumper doesn't read() before calling available(), so available() alwalys returns 0.
I cannot get logs when change to 2.6.1 from 2.4.1.

jwendell:
Kind of revert 61ae6ed33 (fix: InputStreamPumper now switched to non-blocking to prevent stuck threads.)

InputStream.available() is not reliable and can always return 0. In
particular, the RealBufferedSource implementation might have a race
condition where it first needs to read something before available()
can return anything other than zero.

Thus we get stuck on this statement "while (in.available() > 0..." and
never get anything actually read from the stream.

Closes #843.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/InputStreamPumper.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/InputStreamPumper.java
@@ -23,7 +23,6 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
-import java.io.InterruptedIOException;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 public class InputStreamPumper implements Runnable, Closeable {
@@ -53,19 +52,15 @@ public class InputStreamPumper implements Runnable, Closeable {
         synchronized (this) {
           thread = Thread.currentThread();
         }
+
         byte[] buffer = new byte[1024];
+        int length;
+
         try {
-            while (keepReading && !Thread.currentThread().isInterrupted()) {
-              while (in.available() > 0 && keepReading && !Thread.currentThread().isInterrupted()) {
-                int length = in.read(buffer);
-                if (length < 0) {
-                  throw new IOException("EOF has been reached!");
-                }
+            while (keepReading && !Thread.currentThread().isInterrupted() && (length = in.read(buffer)) != -1) {
                 byte[] actual = new byte[length];
                 System.arraycopy(buffer, 0, actual, 0, length);
                 callback.call(actual);
-              }
-              Thread.sleep(50); // Pause to avoid tight loop if external proc is too slow
             }
         } catch (IOException e) {
             if (!keepReading) {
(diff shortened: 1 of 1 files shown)


jwendell:
I don't know the motivation behind commit 61ae6ed33. This is just an attempt to get it working again.

oscerd:
@jimmidyson @iocanel wdyt?

jwendell:
@iocanel Like I said, this is just a trivial attempt to get it working again. If you have a proper solution for both issues (#843 and leaking due to the blocking approach), please, go ahead! I'll close this one.

Thanks!

iocanel:
@jwendell:

iocanel:
@jwendell: I don't have a solution in mind, but at the moment reverting will just cause equal (if not worse) problems than it solves

jwendell:
Ah ok. I thought you already had a proper fix in the works...

By the way, I'm working locally with this patch applied and cannot see any issue. What is specifically the problem you had with the blocking approach? is it reproducible?

iocanel:
The issue is that each exec operation leaves hunging threads. I think it should be visible if you run the exec example.

jwendell:
@iocanel I'm able to reproduce this hanging thread issue even on master. I think it's not a k8s-client issue, but a okhttp/okio one.

For example, running the ExecExample, I see 2 hanging threads at the end of execution:
```
[WARNING] thread Thread[OkHttp ConnectionPool,5,io.fabric8.kubernetes.examples.ExecExample] was interrupted but is still alive after waiting at least 15000msecs
[WARNING] thread Thread[OkHttp ConnectionPool,5,io.fabric8.kubernetes.examples.ExecExample] will linger despite being asked to die via interruption
[WARNING] thread Thread[Okio Watchdog,5,io.fabric8.kubernetes.examples.ExecExample] will linger despite being asked to die via interruption
[WARNING] NOTE: 2 thread(s) did not finish despite being asked to  via interruption. This is not a problem with exec:java, it is a problem with the running code. Although not serious, it should be remedied.
[WARNING] Couldn't destroy threadgroup org.codehaus.mojo.exec.ExecJavaMojo$IsolatedThreadGroup[name=io.fabric8.kubernetes.examples.ExecExample,maxpri=10]
...
```

For the hanging Okio thread, we could add a `return;` statement on this catch block: https://github.com/square/okio/blob/80f3aa2e7bb9ac9691e62ec619e6b53259d78958/okio/src/main/java/okio/AsyncTimeout.java#L328

For the hanging Okhttp thread, the solution seems to be harder. It uses a ThreadPool that keeps threads around with a keep alive setting, so, even adding a `return;` statement on this catch block doesn't fix the problem: https://github.com/square/okhttp/blob/77471444ec81ad9452ebde7ca2b58db58a1f77d1/okhttp/src/main/java/okhttp3/ConnectionPool.java#L68

We can discuss whether 2 points above are bugs or features, and whether they should be fixed on their upstream projects.

That said, I guess we can revert commit 61ae6ed once it doesn't fix the problem at all. What do you think?

PR #855, opened 2017-09-01
Fix watcher resourceVersion problem with 3.6

### LINKED ISSUE #852: Cannot watch build resources in Openshift 3.6
Related to [this issue](https://github.com/fabric8io/fabric8-maven-plugin/issues/1016) in the fabric8-maven-plugin.

The watch is always done with a resource version number that is fetched from the build list (not from the actual build) on the first attempt. Everything is looked up from the new API endpoint `apis/build.openshift.io`.

The error is like:
```
Caused by: io.fabric8.kubernetes.client.KubernetesClientException: Timeout: Too large resource version: 19393279, current: 19393278
	at io.fabric8.kubernetes.client.dsl.internal.WatchConnectionManager$2.onFailure(WatchConnectionManager.java:191)
	at okhttp3.internal.ws.RealWebSocket.failWebSocket(RealWebSocket.java:543)
	at okhttp3.internal.ws.RealWebSocket$2.onResponse(RealWebSocket.java:185)
	at okhttp3.RealCall$AsyncCall.execute(RealCall.java:141)
	at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:32)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

From my tests, it seems that the resource version number in the list object **is independent** from the resource version number of the actual build object (in the new API).

So the first revision number should not be taken from the list, but from the actual build. I don't know why it is taken from the list.

Using a empty revision number should be a viable option on the first attempt to create the watch. While on a subsequent reconnection, the client should use the last revision number received from the server.

nicolaferraro:
This should fix #852 by not using a resource version on the first call.

Since old versions of kubernetes, if you don't provide a resource version, [the watch starts from the latest observed state](https://github.com/kubernetes/kubernetes/blob/v1.4.2/pkg/storage/watch_cache.go#L292-L296), so there's no need to do a lookup before starting the watch.

Tested with Openshift origin 1.5 and 3.6.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
@@ -130,11 +130,11 @@ public class WatchConnectionManager<T extends HasMetadata, L extends KubernetesR
       }
     }
 
-    httpUrlBuilder.addQueryParameter("watch", "true");
-
     if (this.resourceVersion.get() != null) {
       httpUrlBuilder.addQueryParameter("resourceVersion", this.resourceVersion.get());
     }
+
+    httpUrlBuilder.addQueryParameter("watch", "true");
 
     Request request = new Request.Builder()
       .get()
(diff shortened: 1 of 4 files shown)


oscerd:
ok to test

nicolaferraro:
I've updated the endpoints used in mockwebserver to reflect the new behaviour of the watchers... crossing fingers :wink:

iocanel:
add to whitelist

jstrachan:
@iocanel LGTM - you happy with it?

oscerd:
[merge]

PR #879, opened 2017-10-13
OSFUSE-640: Fixed basic token overwriting with a stale bearer token

dhirajsb:


dhirajsb:
@iocanel what's causing the license-check status to be stuck? Is there something I need to do?

jimmidyson:
retest this please

PR #882, opened 2017-10-13
Allow to use non default context

fkorotkov:
In my case I'm trying to manage clusters in my Kubernetes Federation.  When you have a federation you end up with a context per each cluster as well as federation context to apply things across clusters. 

But some things like `StatefulSet` doesn't have a federation equivalent and therefore you can't apply them to the "global" context. For such things it will be useful to be able to configure each one of them per cluster in a federation. And in order to do this it will be nice to be able to configure context for a config.

This PR adds this ability to specify which context to try to create configuration from.

## Testing 

 Added a test. `mvn test` passes.


iocanel:
ok to test

oscerd:
[merge]

oscerd:
Thanks for the PR

fkorotkov:
Awesome! Thanks you all!

PR #890, opened 2017-10-23
first spike at a mechanism to support using CRDs

jstrachan:
this PR adds the ability to register new CRD clients providing a few DTOs and reusing the kubernetes-client DSL and underlying CustomResourceDefinition metadata to implement the REST API

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/CustomResourceDoneable.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/CustomResourceDoneable.java
@@ -1,12 +1,12 @@
 /**
  * Copyright (C) 2015 Red Hat, Inc.
- * <p>
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * <p>
- * http://www.apache.org/licenses/LICENSE-2.0
- * <p>
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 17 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/CustomResourceSupport.java
```diff
@@ -0,0 +1,68 @@
...
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.fabric8.kubernetes.api.model.ObjectMeta;
+
+/**
+ * A base class for implementing a custom resource kind
+ */
+@JsonDeserialize(using = com.fasterxml.jackson.databind.JsonDeserializer.None.class)
+public abstract class CustomResourceSupport implements HasMetadata {
```
iocanel: Maybe, CustomResource or BaseCustomResource?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/CustomResourceDoneable.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+package io.fabric8.kubernetes.client;
+
+import io.fabric8.kubernetes.api.builder.Function;
+import io.fabric8.kubernetes.api.model.Doneable;
+import io.fabric8.kubernetes.api.model.HasMetadata;
+
+/**
+ */
+public class CustomResourceDoneable<T extends HasMetadata> implements Doneable<T> {
```
iocanel: I am wondering if we can (and also if it makes sense to) make things more strict here and have T extends CustomResourceSupport (or whatever we end up naming this class).

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/CustomResourceSupport.java
jstrachan: yeah - wasn't totally sure TBH - lets go with CustomResource?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/CustomResourceDoneable.java
jstrachan: yeah - sounds good to me. I guess folks can always opt out of the CustomResource / CustomResourceDoneable classes anyway if they want to do things themselves

jstrachan:
@iocanel you cool with this PR now?

Issue #899, opened 2017-11-03
No logs for fabric8:run and fabric8:build


hrishin:
Logback seems to be not working. Need to get it fix.

https://github.com/fabric8io/fabric8-maven-plugin/issues/1049
https://github.com/fabric8io/fabric8-maven-plugin/issues/1048

PR #910, opened 2017-11-09
This will fix the issue of logs broken

piyush1594:
Implemented websockets for fetching logs.

Refactored code, removed comments

Refactored and Improvised Exception

#899


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java
```diff
@@ -15,6 +15,20 @@
  */
 package io.fabric8.kubernetes.client.dsl.internal;
 
+import io.fabric8.kubernetes.api.model.DoneablePod;
```
hrishin: can we remove the unused imports, if applicable?

piyush1594:
Hi @iocanel ,

Thanks for the reply. Meanwhile i was working, someone already submitted the patch. Just realized that this is already fixed. This time we can close the PR.

Thanks

PR #913, opened 2017-11-14
feat: Introduce crud mode in KubernetesServer.

iocanel:


iocanel:
[merge]

kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/KubernetesServer.java
```diff
@@ -16,25 +16,41 @@
...
+
+  public KubernetesServer(boolean https, boolean curdMode) {
     this.https = https;
+    this.curdMode = curdMode;
   }
 
   public void before() {
-    mock = new KubernetesMockServer(https);
+    mock = curdMode
+      ? new KubernetesMockServer(new Context(), new MockWebServer(), new HashMap<ServerRequest, Queue<ServerResponse>>(), new KubernetesCrudDispatcher(), true)
```
wasif-k: What is the reason behind always using `useHttps=true` during crud mode?

iocanel: @wasif-k: No reason. We should pass https instead of always true.

Issue #914, opened 2017-11-15
Add possibility to change MaxRequests on Dispatcher in HttpClientUtils


thorbenvh8:
Just an improvement, instead of being only able to update the MaxRequestsPerHost for the dispatcher. You can now update the MaxRequests as well (default otherwise always 64).

Thanks for merging this quickly :)

ckanner:
I also occurs the phenomenon.When there are many websocket's connections, some of them will stay in the executor pool and waiting for to schedule.The reason is the ```maxRequestsPerHost``` and ```maxRequests``` to limit them.

So I use the way to fix this problem
```java
final Config config = new ConfigBuilder()
            .withMasterUrl(k8sProperties.getRestUri())
            .withTrustCerts(true)
            .withNamespace(k8sProperties.getNamespace())
            .build();
        final Dispatcher dispatcher = new Dispatcher();
        dispatcher.setMaxRequestsPerHost(k8sProperties.getMaxRequests());
        dispatcher.setMaxRequests(k8sProperties.getMaxRequests());
        final OkHttpClient httpClient = HttpClientUtils.createHttpClient(config).newBuilder().dispatcher(dispatcher).build();
        KubernetesClient client = new AutoAdaptableKubernetesClient(httpClient, config);
```

May be the ```Config``` class can support defines the maxRequests.

thorbenvh8:
please see the pull request, I did exactly that :)

ckanner:
I know. The pull request doesn't merge to the master branch.

PR #915, opened 2017-11-15
#914: Add possibility to change MaxRequests on Dispatcher in HttpClie…

thorbenvh8:
…ntUtils
#914 
Thanks :)


iocanel:
Overall looks good, though I am trying to understand, how it plays with `maximum requests per host`

What happens when both are set?  It seems that out of the box we set 64 concurrent request but only 5 per host. Given that I can't think of a case where multiple hosts are involved does that mean that the limit is still 5?

thorbenvh8:
64 is the default value in the dispatcher, I dont know why its 64 i just used that. The default value for maxRequestsPerHost in the dispatcher is 5, what is the default value in the k8s client. I just used whats default.

iocanel:
ok to test

asacbp:
when do you think will we you have time to bake a release with this pull request? thanks!

iocanel:
[merge]

PR #925, opened 2017-12-01
[OSFUSE-667]: Regression between FIS 2 - R4 x R5: Revert [OSFUSE-640]

valdar:
Reverting OSFUSE-640: Fixed basic token overwriting with a stale bearer token"
because of https://issues.jboss.org/browse/OSFUSE-667 and was never asked to be baclported to this branch.

This reverts commit 464544d54a90637418b67c26e49952d0b5a3e2fb.

### DIFF
--- a/openshift-client/src/main/java/io/fabric8/openshift/client/internal/OpenShiftOAuthInterceptor.java
+++ b/openshift-client/src/main/java/io/fabric8/openshift/client/internal/OpenShiftOAuthInterceptor.java
@@ -21,7 +21,6 @@ import okhttp3.Interceptor;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
-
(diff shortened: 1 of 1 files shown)


valdar:
@iocanel I totally agree and understand your point. 
This is, though, the release branch for FIS 2; so in the narrower point of view of just FIS, those issues where not reported or being a problem in the FIS components that use {k8s,ocp}-client.

valdar:
@iocanel I totally agree and understand your point. 
This is, though, the release branch for FIS 2; so in the narrower point of view of just FIS, those issues where not reported or being a problem in the FIS components that use {k8s,ocp}-client.

I think that a decision needs to be made about braking or not the previous behaviour.

iocanel:
@valdar: Didn't realize that we were talking about the `fabric8io:v1.4.14.redhat` switched to approve.

valdar:
@iocanel cha we merge this then?

PR #926, opened 2017-12-01
waitUntilReady now first waits until resource exists

### LINKED ISSUE #921: awaiting readiness of a service fails because creating endpoints takes a long time
We're observing this issue in Arquillian Cube tests running against OpenShift Online Pro.

Arquillian Cube awaits readiness of all resources it creates (https://github.com/arquillian/arquillian-cube/blob/master/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionManager.java#L182-L183). One of such resources is often a service. For Fabric8 Kubernetes Client, awaiting readiness of a service means waiting until the corresponding endpoints are ready (https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ServiceOperationsImpl.java#L79).

If the service exists but no corresponding endpoint exists, Fabric8 Kubernetes Client will consider the service "not ready and never can be ready" (https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.java#L110) and throw a `KubernetesClientTimeoutException` immediately. This usually doesn't happen, because the endpoint gets created immediately. __BUT__, as I've observed on OpenShift Online Pro (the actual production cluster!), the time between creating a service and creating an endpoint can be very long -- in the order of minutes!

I believe that if a resource doesn't exist when the user asks for awaiting readiness, Fabric8 Kubernetes Client should (be able to) wait for the resource to appear. Say for example that the user asks "wait up to 5 minutes for readiness of these resources" -- Fabric8 Kubernetes Client should _not_ terminate the await after a few seconds just because the resource (or a "subordinate" resource) doesn't exist yet.

This affects RHOAR because it affects Arquillian Cube and Cube doesn't currently provide an option to skip this readiness await.

Ladicek:
I'm submitting this only to discuss the approach I've taken. The code contains a couple of TODOs that need to be resolved first.

Also the PR doesn't contain any test; could you please advise how to approach testing this change? I'm not at all familiar with tests in this codebase.

Fixes #921.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
@@ -954,8 +954,34 @@ public class BaseOperation<T, L extends KubernetesResourceList, D extends Doneab
     return i instanceof HasMetadata && Readiness.isReady((HasMetadata)i);
   }
 
+  protected T waitUntilExists(long timeoutInMillis) {
+    long end = System.currentTimeMillis() + timeoutInMillis;
+    while (System.currentTimeMillis() < end) {
+      T item = get();
+      if (item != null) {
+        return item;
+      }
+
+      // TODO is there a better way?
+      try {
+        Thread.sleep(1000);
+      } catch (InterruptedException e) {
+        // TODO
+      }
+    }
+
+    T item = get();
+    if (item != null) {
+      return item;
+    }
+
+    throw new IllegalArgumentException(type.getSimpleName() + " with name:[" + name + "] in namespace:[" + namespace + "] not found!");
+  }
+
+
   @Override
   public T waitUntilReady(long amount, TimeUnit timeUnit) throws InterruptedException {
-    return get();
+    // TODO do we really want this? pretty big behavioral change for resources that aren't readiness-capable
+    return waitUntilExists(timeUnit.toMillis(amount));
   }
 }
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
@@ -954,7 +954,7 @@ public class BaseOperation<T, L extends KubernetesResourceList, D extends Doneab
     return i instanceof HasMetadata && Readiness.isReady((HasMetadata)i);
   }
 
-  protected T waitUntilExists(long timeoutInMillis) {
+  protected T waitUntilExists(long timeoutInMillis) throws InterruptedException {
     long end = System.currentTimeMillis() + timeoutInMillis;
     while (System.currentTimeMillis() < end) {
       T item = get();
@@ -962,12 +962,8 @@ public class BaseOperation<T, L extends KubernetesResourceList, D extends Doneab
         return item;
       }
 
-      // TODO is there a better way?
-      try {
-        Thread.sleep(1000);
-      } catch (InterruptedException e) {
-        // TODO
-      }
+      // in the future, should probably combine this simple waiting loop with a watcher
+      Thread.sleep(1000);
     }
 
     T item = get();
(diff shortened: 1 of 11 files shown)


Ladicek:
Aha, there's a test failure that also points me to a place where things like this are tested. Nice :-)

Ladicek:
Fixed failing tests and added a new test.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java
```diff
@@ -335,22 +335,5 @@ public ExecWatch exec(String... command) {
...
-
-    if (Readiness.isReady(pod)) {
-      return pod;
-    }
-
-    ReadinessWatcher<Pod> watcher = new ReadinessWatcher<>(pod);
-    try (Watch watch = watch(watcher)) {
-      return watcher.await(amount, timeUnit);
-    }
-  }
```
Ladicek: Just thinking -- not sure why `Pod.waitUntilReady` was different than `<everything else>.waitUntilReady`, but the common implementation seems to work just fine.

Ladicek:
Ping @iocanel, did you maybe have time to take a look at this?

iocanel:
@Ladicek: no! I'll try to have a look at it during the weekend.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
```diff
@@ -954,8 +954,34 @@ public Boolean isReady() {
...
 
+  protected T waitUntilExists(long timeoutInMillis) {
+    long end = System.currentTimeMillis() + timeoutInMillis;
+    while (System.currentTimeMillis() < end) {
+      T item = get();
+      if (item != null) {
+        return item;
+      }
+
+      // TODO is there a better way?
```
iocanel: While implementing the `waitUntilReady` I found that the optimal approach is to go hybrid, by combining a while/wait loop and watchers. This limits the number of request send to the server, but also addresses potential races that can affect a purely watcher approach.

For me going with a pure while/wait just for the exists check is good enough at least for now.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/HasMetadataOperation.java
```diff
@@ -211,4 +211,20 @@ protected T periodicWatchUntilReady(int i, long started, long interval, long amo
...
+
+  @Override
+  public T waitUntilReady(long amount, TimeUnit timeUnit) throws InterruptedException {
+    if (Readiness.isReadinessApplicable(getType())) {
+      long started = System.currentTimeMillis();
+      waitUntilExists(timeUnit.toMillis(amount));
+      long alreadySpent = System.currentTimeMillis() - started;
+
+      // if awaiting existence took very long, let's give at least 10 seconds to awaiting readiness
+      long remaining = Math.max(10_000, timeUnit.toMillis(amount) - alreadySpent);
```
iocanel: I am not sure if I fancy giving it extra time. Some possible alterations:

- subtract that 10 seconds from the waitUntilExists time.
- check if remaining is less or equal to zero and then just check if resource is ready.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java
```diff
@@ -335,22 +335,5 @@ public ExecWatch exec(String... command) {
     public BytesLimitTerminateTimeTailPrettyLoggable<String, LogWatch> limitBytes(int limitBytes) {
         return new PodOperationsImpl(client, getConfig(), apiVersion, namespace, name, isCascading(), getItem(), getResourceVersion(), isReloadingFromServer(), getGracePeriodSeconds(), getLabels(), getLabelsNot(), getLabelsIn(), getLabelsNotIn(), getFields(), containerId, in, inPipe, out, outPipe, err, errPipe, withTTY, withTerminatedStatus, withTimestamps, sinceTimestamp, sinceSeconds, withTailingLines, withPrettyOutput, execListener, limitBytes);
     }
-
-  @Override
```
iocanel: This must have slipped. But its now removed so all fine.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java
iocanel: see my comment above.

Ladicek:
I've rebased on top of current `master` and added one commit to address the remaining items (removed the TODOs, better waiting for readiness when existence await took too long).

iocanel:
[merge]

PR #937, opened 2017-12-06
Support setting Impersonate-User header to perform operations using user impersonation

lulf:
I'm not very familiar with the code base, so I'm looking for comments if this is the correct way of implementing this feature and if it should be done differently.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/RequestConfig.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/RequestConfig.java
@@ -17,6 +17,10 @@ package io.fabric8.kubernetes.client;
 
 import io.sundr.builder.annotations.Buildable;
 
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
 import static io.fabric8.kubernetes.client.Config.DEFAULT_LOGGING_INTERVAL;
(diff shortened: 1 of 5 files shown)


iocanel:
@lulf: Thanks!

I am not very familiar with impersonate user, but my understanding is that its something you apply temporarily to one or maybe a few requests. I am not sure I see the point / value in configuring it upfront and make it applicable to all requests (but maybe I miss something, and please correct me if I do).

 The whole concept of the impersonate user, reminds me of the `.withRequestConfig()` we added a couple of months ago, which allowed you to override the request config for a single or a group of requests.

So, I am thinking that maybe impersonate user should be handled in a similar way. wdyt?

lulf:
@iocanel I didn't now about .withRequestConfig() but I agree it sounds like the appropriate way to use this feature! I will update the PR to use it, thanks!

lulf:
@iocanel I've removed the interception now, and setting the impersonated name was already on the request config.

iocanel:
@lulf: I think that we do need the interception, or we won't add the headers at all.

lulf:
@iocanel You're right, not sure what I was thinking. I'll add that back, but I guess what you really want is to have the stuff in Config removed?

lulf:
@iocanel I've updated the PR with re-adding the interceptor and it now supports all of the headers supported according to https://kubernetes.io/docs/admin/authentication/ under "User impersonation".

iocanel:
add to whitelist

iocanel:
[merge]

PR #940, opened 2017-12-06
Extended Loggable Interface to return java.io.Reader

### LINKED ISSUE #939: Extend Loggable interface to return java.io.Reader
Currently `getLog()` only returns `java.lang.String` which is unfortunate considering that logs can grow very large.

paoloantinori:
fixes #939


iocanel:
add to whitelist.

iocanel:
[merge]

Issue #963, opened 2017-12-22
Watcher onClose method is not called 


stheppi:
Here is the scenario to reproducer:
 We have a minikube instance running and we have the a watcher set for deployment and pod. We set the retry attempt to a value >= 0 (doesn't matter). Then we stop minikube and the watcher `onClose` is called. At this point we create a new Watcher; since the minikube is down the websocket connection fails and we never get the `onClose` method on this last instance.

I have looked at the code a bit to understand what's happening. I think the `WatchConnectionManager.java::runWatch`  code should do things differently:
``` webSocket = clonedClient.newWebSocket(request, new WebSocketListener() {
      @Override
      public void onOpen(final WebSocket webSocket, Response response) {
        if (response != null && response.body() != null) {
          response.body().close();
        }
        logger.debug("WebSocket successfully opened");
        webSocketRef.set(webSocket);
        currentReconnectAttempt.set(0);
        started.set(true);
        queue.clear();
        queue.add(true);
      }

      @Override
      public void onFailure(WebSocket webSocket, Throwable t, Response response) {
....
/*Line 179*/    if (response != null) {
          // We only need to queue startup failures.
          Status status = OperationSupport.createStatus(response);
          if (response.body() != null) {
            response.body().close();
          }
          logger.warn("Exec Failure: HTTP {}, Status: {} - {}", response.code(), status.getCode(), status.getMessage(),
              t);
/*Line 187*/          if (!started.get()) {
            queue.add(new KubernetesClientException(status));
            return; // no reconnect
          }
        } else {
          logger.warn("Exec Failure", t);
/*Line 193*/          if (!started.get()) {
            queue.add(new KubernetesClientException("Failed to start websocket", t));
            return; // no reconnect
          }
        }

        if (currentReconnectAttempt.get() >= reconnectLimit && reconnectLimit >= 0) {
          closeEvent(new KubernetesClientException("Connection failure", t));
          return;
        }

/*Line 204*        scheduleReconnect();
```

Both Lines 187 and 193 shouldn't return but let it flow to:
```
  if (currentReconnectAttempt.get() >= reconnectLimit && reconnectLimit >= 0) {
          closeEvent(new KubernetesClientException("Connection failure", t));
          return;
        }

/*Line 204*/        scheduleReconnect();
```

This way if the max reconnect has been achieved just call the close event (thus the watcher is notified); otherwise schedule a reconnect.

iocanel:
[...]

PR #964, opened 2017-12-22
Watcher onClose method is not called #963

stheppi:
Avoid returning from the WebSocketListener#onFailure if the websocket connection failed (even if the `start`  flag hasn't been set)
Let the code flow to the check if max retries has been reached. 
This way the Watcher#onClose will eventually be called.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
@@ -186,13 +186,11 @@ public class WatchConnectionManager<T extends HasMetadata, L extends KubernetesR
               t);
           if (!started.get()) {
             queue.add(new KubernetesClientException(status));
-            return; // no reconnect
           }
         } else {
           logger.warn("Exec Failure", t);
           if (!started.get()) {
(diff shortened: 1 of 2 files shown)


oscerd:
Can you take a look at

io.fabric8.kubernetes.client.mock.WatchTest.testHttpErrorReconnect(WatchTest.java:141)

The test is failing.

stheppi:
WatchOverHTTP.testHttpErrorReconnect() fails but i don't understand why since it is going over http get.
I would need a bit of help here.

iocanel:
@stheppi: Let me have a look!

iocanel:
@stheppi: There is something I miss! In the original issue you raised, you pointed out that some return statements should be removed and let the code flow through the end which is something that makes sense to me. This pull request though, takes an other route. 

Why do u remove the started check?
And why do u add the reconnect check in each case and don't just let it flow through?

Just removing the return statements from the designated places and also adding the Exception to the queue, whenever we send it to closeEvent, seems to just work fine.

Do I miss something?

stheppi:
@iocanel 
  It was a mistake indeed to remove the `started` check. i have pushed what i think is the correct code. tests are saying green. please have a look

iocanel:
add to whitelist

iocanel:
[merge]

PR #974, opened 2018-01-03
Avoid WatchConnectionManager thread leak

stheppi:
Avoid the thread leaking when `watch.waitUntilReady();` throws an exception when it can't establish a connection to k8


stheppi:
During testing we stopped minikube and saw 4k threads left around. See the image below.

thread leak

When the watch is created there is the call to `waitUntilReady`. This can fail if the connection to k8 fails and therefore will leave the inner threadpool thread around.

iocanel:
[merge]

oscerd:
ok to test

iocanel:
add to whitelist

piyush1594:
ok to test

piyush1594:
[merge]

PR #988, opened 2018-01-15
Remove kubernetes-api from kubernetes-itests.

rohanKanojia:
kubernetes-api is fat and ugly, plus it introduces  a dependency cycle and
it can be a source of issues.


piyush1594:
Thanks, @rohanKanojia. I am merging this.

Issue #1002, opened 2018-02-05
Update PodSecurityPolicy and SecurityContextConstraint Regression test


piyush1594:
In place of deleting all the pod security policies and SecurityContextConstraints , just delete the one that was created in the test.

piyush1594:
Since #1003 is merged, I am closing this.

PR #1003, opened 2018-02-05
Update PodSecurityPolicyTest and SecurityContextConstraintsTest

piyush1594:
Update PodSecurityPolicyTest and SecurityContextConstraintsTest
to delete only the one created in Regression Test
#1002


oscerd:
ok to test

PR #1010, opened 2018-02-15
Log usage of unstable api

### LINKED ISSUE #983: Alert users when using a beta feature (OSFUSE-661)
Resources in Kubernetes move from `beta` to GA in different releases. Kubernetes usually maintains old resource endpoints for few releases before removing them.

If a user wants to use e.g. a `StatefulSet` now, this client library will use the `v1beta1` endpoint that works with Kubernetes from 1.6 to 1.9. But in 1.9 `StatefulSets` are upgraded to `v1`. Endpoint `v1beta1` still works (I guess), but it will be removed in later versions.

When a resource is alpha/beta, model and behavior may also change between releases (`StatefulSet` are a lot different from their ancestors `PetSet`).

That said, I think we should alert users when they are using a beta resource, because such resources will certainly change in future releases of Kubernetes (so users may need to upgrade this library in the best case if they upgrade Kubernetes).

I was thinking to two ways to alert them:

1. Deprecate beta features in the DSL methods, explaining in the deprecation message that they may be removed. This may seem counter-intuitive, since deprecation is usually used for old stuff, but helps the user to understand the risks of using them...
2. Another option is to display a warning log the first time a user try to access a resource with a `alpha` or `beta` version (it should be easier to maintain than 1)

Currently the set of beta features supported contains StatefulSet, ThirdPartyResource, CustomResourceDefinitions, DaemonSet and others...

I would like to know your opinion before going for one of the two..

@oscerd, @iocanel

nicolaferraro:
Fixes #983.
This is a simple impl that logs usage of alpha/beta versions only the first time each type is used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
   New Feature
    * Added support for StorageClass - https://github.com/fabric8io/kubernetes-client/pull/978
    * Added support for PodSecurityPolicy - https://github.com/fabric8io/kubernetes-client/pull/992
+   * The client now warns when using Kubernetes alpha or beta resources - https://github.com/fabric8io/kubernetes-client/pull/1010
   
   Improvements
    * Fixed issue of SecurityContextConstraints not working - https://github.com/fabric8io/kubernetes-client/pull/982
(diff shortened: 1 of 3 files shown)


nicolaferraro:
Sure!

oscerd:
I'll add a line to the changelog

oscerd:
The line is there already, sorry

oscerd:
[merge]

PR #997, opened 2018-01-30
Test showing pauses in pod waitUntilReady

carlossg:
Minimal test from https://github.com/jenkinsci/kubernetes-plugin/pull/271

Running this test shows 30 second waits when running multiple concurrent threads using waitUntilReady

```
mvn clean test -Dtest=WaitTest
...
Running io.fabric8.kubernetes.client.WaitTest
Created pod: test-waiting-2g74l
2018/01/30 20:54:49 Thread 1 finished
2018/01/30 20:54:49 Thread 8 finished
2018/01/30 20:54:49 Thread 19 finished
2018/01/30 20:54:49 Thread 4 finished
2018/01/30 20:54:49 Thread 7 finished
2018/01/30 20:55:19 Thread 10 finished
2018/01/30 20:55:19 Thread 17 finished
2018/01/30 20:55:19 Thread 12 finished
2018/01/30 20:55:19 Thread 0 finished
2018/01/30 20:55:19 Thread 18 finished
2018/01/30 20:55:28 Thread 14 finished
2018/01/30 20:55:28 Thread 11 finished
2018/01/30 20:55:28 Thread 6 finished
2018/01/30 20:55:28 Thread 9 finished
2018/01/30 20:55:28 Thread 3 finished
2018/01/30 20:55:28 Thread 2 finished
2018/01/30 20:55:28 Thread 13 finished
2018/01/30 20:55:28 Thread 16 finished
2018/01/30 20:55:28 Thread 5 finished
2018/01/30 20:55:28 Thread 15 finished
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 44.234 sec - in io.fabric8.kubernetes.client.WaitTest
```

PR #1011, opened 2018-02-15
Optimize watch to avoid missing events

nicolaferraro:
This Fixes #997 (that is a PR). Watchers did not specify the starting `resourceVersion`, so they started from "latest" and they could miss the readiness event.. relying on a timeout and a subsequent retry..

With this change, the watcher starts from latest observed "resourceVersion", so the readiness event is not missed and it is much faster.


oscerd:
[merge]

Issue #987, opened 2018-01-14
Error when reading PEM certificate with empty lines


greenboxal:
I was setting up a Spinnaker cluster and when the API client was being instantiated, an exception was being thrown inside `CertUtils.createTrustStore`.

After some debugging I found out that my Kubernetes cluster had a CA PEM file with an empty line at the end, and I tracked the bug here: https://github.com/fabric8io/kubernetes-client/blob/5e85f3e8727cdd0f6b47b2e1d416ce6610f715dd/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/CertUtils.java#L91

I removed the last LF in my CA file and it worked.

Looks like it reads through the first PEM armor and, as it has some bytes left in the end of the file (a LF in my case), it attempts to read another PEM armor. As the rest of the file contains no PEM armor, it throws an EOF exception.

PR #1021, opened 2018-03-02
Prevent IOException if certificate file has trailing empty lines

georgeharley:
As described in issue #987 it is possible for certificate files that include trailing empty lines to trigger IOExceptions because the CertUtils createTrustStore() method continues to work with the pemInputStream even though all of its certificates have been processed. It looks like this scenario is not being caught in the CertUtilsTest tests because they do not actually use the CertUtils.getInputStreamFromDataOrFile() method to get an input stream from the test PEM files but instead used getClass().getResourceAsStream().


piyush1594:
ok to test

oscerd:
ok to test

oscerd:
[merge]

PR #1044, opened 2018-03-23
Fix #1004: Multiple document handling breaks if "---" found anywhere in the document

### LINKED ISSUE #1004: Multiple document handling breaks if --- found anywhere in the document.
So, the idea was to support yaml files that contain multiple documents. This works, in most cases, by detecting using `---` as a delimiter to split documents. Problem arises though when `---` is not used as a delimiter but its part of string.

For example: https://github.com/syndesisio/syndesis/blob/master/app/deploy/syndesis-restricted.yml#L355

rohanKanojia:
Don't use --- as a delimiter to split documents. Instead, split document by '\n', process
line by line and only split when a single '---' found in the document. #1004

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/Serialization.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/Serialization.java
@@ -241,7 +241,7 @@ public class Serialization {
 
   private static List<KubernetesResource> getKubernetesResourceList(Map<String, String> parameters, String specFile) {
     List<KubernetesResource> documentList = new ArrayList<>();
-    String[] documents = specFile.split(DOCUMENT_DELIMITER);
+    String[] documents = splitSpecFile(specFile);
     for (String document : documents) {
       if (validate(document)) {
         ByteArrayInputStream documentInputStream = new ByteArrayInputStream(document.getBytes());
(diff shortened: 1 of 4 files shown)


piyush1594:
ok to test

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/Serialization.java
```diff
@@ -253,17 +253,32 @@ public Type getType() {
...
+
+    return nValidDocuments > 1;
+  }
+
+  private static String[] splitSpecFile(String aSpecFile) {
+    List<String> documents = new ArrayList<>();
+    String[] lines = aSpecFile.split(System.lineSeparator());
+    int nLine = 0;
+
+    do {
```
piyush1594: This thing can be reduced to a single loop in place of a loop in the loop. This is just decreasing the code readability. You can just use a single loop which will iterate over all the line and whenever you see a `DOCUMENT_DELIMETER` add the previous string to documents List and start a new string. The complexity of both is same but why to increase the reading complexity if we can achieve it in a single loop.

piyush1594:
You forgot to add this to CHANGELOG.md

piyush1594:
[merge]

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/Serialization.java
```diff
@@ -253,17 +253,37 @@ public Type getType() {
...
+  }
+
+  private static String[] splitSpecFile(String aSpecFile) {
+    List<String> documents = new ArrayList<>();
+    String[] lines = aSpecFile.split(System.lineSeparator());
+    int nLine = 0;
+    StringBuilder builder = new StringBuilder();
+
+    while(nLine < lines.length) {
+      if(lines[nLine].length() >= DOCUMENT_DELIMITER.length()
```
dennis-benzinger-hybris: @piyush1594, doesn't this split the `aSpecFile` on every line which is shorter than the delimiter?

Especially empty lines which you might use for logical separation.

piyush1594: This is separating all the new lines, adding them to an array which is `lines` and then iterating over the `lines` array.

You might be facing some issue because of this if a line's length is less then delimiter length that is already fixed by #1070

piyush1594: Kubernetes Client 3.1.12 released with the fix

PR #1045, opened 2018-03-25
Support `error` websocket channel

giannello:
This PR stems from jenkins-kubernetes not being able to properly deal with the return code when `exec`ing into a container. The kubernetes-client doesn't provide a separate pipe for the error channel exposed over the websocker, merging it with `stderr` instead.

This changeset exposes the `error` channel as a separate stream, and renames the others to keep a consistent naming strategy. The `resize` channel is deliberately ignored.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
@@ -63,15 +63,15 @@ public class ExecWebSocketListener extends WebSocketListener implements ExecWatc
     private static final Logger LOGGER = LoggerFactory.getLogger(ExecWebSocketListener.class);
 
     private final Config config;
-    private final InputStream stdin;
-    private final OutputStream stdout;
-    private final OutputStream stderr;
-    private final OutputStream error;
+    private final InputStream in;
+    private final OutputStream out;
+    private final OutputStream err;
+    private final OutputStream errChannel;
 
-    private final PipedOutputStream stdinPipe;
-    private final PipedInputStream stdoutPipe;
-    private final PipedInputStream stderrPipe;
-    private final PipedInputStream errorPipe;
+    private final PipedOutputStream input;
+    private final PipedInputStream output;
+    private final PipedInputStream error;
+    private final PipedInputStream errorChannel;
 
     private final AtomicReference<WebSocket> webSocketRef = new AtomicReference<>();
     private final ExecutorService executorService = Executors.newSingleThreadExecutor();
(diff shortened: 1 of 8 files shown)


giannello:
@carlossg  https://issues.jenkins-ci.org/browse/JENKINS-49762

piyush1594:
ok to test

piyush1594:
retest

carlossg:
Thanks @giannello ! 
I gave it a try but I think it's still missing the methods in `PodOperationsImpl` in order to pass the stream where you would want errors to be written.

It can get confusing given that there's already 
* writingError
* readingError
* redirectingError

`Errorable` interface also needs to be changed or a new one created.

And is the header `Sec-WebSocket-Protocol: v4.channel.k8s.io` needed too? https://github.com/kubernetes-client/python/issues/297

giannello:
@carlossg added you as a collaborator in case you want to submit patches. Will take a look at the classes you mentioned tomorrow.
Not sure about the header, `v4` should be the default if nothing else is specified - I don't really have an opinion and I'm around here just because I stumbled on that python client PR that mentioned the existence of a dedicated signaling channel, so please don't rely on my java skills.
Thank you for quickly jumping in, very appreciated.

carlossg:
In the meantime and given how many name changes are involved I'd love to hear from the maintainers before implementing the next changes

carlossg:
I have added the needed methods, and the implementation on https://github.com/jenkinsci/kubernetes-plugin/pull/300 to use them

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
```diff
@@ -232,18 +244,18 @@ public void onMessage(WebSocket webSocket, ByteString bytes) {
...
-                            err.write(byteString.toByteArray());
+                        if (stderr != null) {
+                            stderr.write(byteString.toByteArray());
                         }
                         break;
                     case 3:
-                        if (err != null) {
-                            err.write(byteString.toByteArray());
+                        if (error != null) {
+                            error.write(byteString.toByteArray());
```
carlossg: this is the interesting part, not sending error stream to stderr stream

carlossg:
This is ready, tests in kubernetes-plugin are passing using the new methods, would love to hear about the API changes and method naming from a maintainer
https://github.com/jenkinsci/kubernetes-plugin/pull/300

@iocanel ?

oscerd:
ping @iocanel

calebdelnay:
I'm wondering if there is anything a random outsider could do to help push this forward?

We use Jenkins to build Docker images using workers allocated via the Kubernetes plugin. However due to https://github.com/jenkinsci/kubernetes-plugin/pull/300, which in turn depends on this issue, we're doing a manual build of the Kubernetes plugin with these fixes applied. It would be great to have these merged and included in upcoming releases.

iocanel:
Don't know how I missed this.
Will have a look shortly.

carlossg:
Ping @bparees

bparees:
@piyush1594 @iocanel bump, we definitely want to help enable the kubernetes pipeline plugin.

carlossg:
I wonder if it would be useful to avoid all the renaming for the sake of simplicity first so you can better see the actual changes needed

giannello:
I can push a less intrusive change, will find some time this afternoon to find a better naming.

carlossg:
any ETA for merging and releasing?

iocanel:
@giannello @carlossg: Apologies for the long delay. Since my main concern was naming and its addressed I'll merge asap.

carlossg:
@iocanel how long until it makes it into a release ?

piyush1594:
Hey, @iocanel Thinking of releasing the client. We need to bump the minor version or it is not required for the time being

piyush1594:
Hey @carlossg Kubernetes Client 3.2.0 released which have this feature.

PR #1054, opened 2018-04-11
Fix #1048 : Add support for the k8s/openshift version

### LINKED ISSUE #1048: Add support for retuning the k8s/openshift version
If would be nice if the API could provide a method to return the k8s/openshift version running in the server. 

As of https://github.com/openshift/origin/issues/19182#issuecomment-378233606
>If user needs to, he can easily request that version directly accessing /version or /version/openshift which will return json output the reporter asks. The first endpoint returns k8s version, the latter openshift.

rohanKanojia:


kubernetes-tests/src/test/java/io/fabric8/openshift/client/server/mock/DeploymentConfigTest.java
```diff
@@ -180,4 +182,11 @@ public void testDeployingLatestHandlesMissingLatestVersion() {
     assertNotNull(deploymentConfig);
     assertEquals(new Long(1), deploymentConfig.getStatus().getLatestVersion());
   }
+
+  @Test
```
piyush1594: I did not get the point of this. I think we have already added a test for this. Secondly, if this is a test that is required, then no assert is there.

piyush1594:
Please add changelog.md. Also, add about this new feature in docs.

kubernetes-tests/src/test/java/io/fabric8/openshift/client/server/mock/DeploymentConfigTest.java
rohanKanojia: Oops, This got added by accident.sorry

piyush1594:
@iocanel @oscerd I liked the naming convention `client.cluster().version()` What are your thoughts on this.

iocanel:
I would rather not have the `cluster()` bit as its somehow confusing.

So, I'd prefer to have getVersion() object directly in the root of the KubernetesClient interface.

The returned object could provide info about clietnt, kubernetes and openshift versions.

piyush1594:
retest this please

rohanKanojia:
ok to test

piyush1594:
ok to test

rohanKanojia:
ok to test

kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/ClusterVersionTest.java
```diff
@@ -0,0 +1,49 @@
...
+import io.fabric8.kubernetes.client.KubernetesClient;
+import io.fabric8.kubernetes.client.server.mock.KubernetesServer;
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+
+public class ClusterVersionTest {
```
iocanel: I guess that this should be called VersionInfoTest or something like this. 
Not a blocker just nitpicking.

iocanel:
@rohanKanojia: There is a failure on 3.6.0. Once we fix that we can merge.

geoand:
Tests seem to be passing :)

oscerd:
[merge]

oscerd:
When will we release?

piyush1594:
Is 3.1.12 version fine? then i can release

oscerd:
I think so.

piyush1594:
Ok cool, i will do the release today. Thanks

Issue #1046, opened 2018-03-27
retrieval of imagestreamtags do not appear to work with the v3.x client


gabemontero:
see see https://github.com/fabric8io/jenkins-sync-plugin/pull/12#issuecomment-366088791

and most recently https://github.com/fabric8io/jenkins-sync-plugin/pull/12#issuecomment-366337554

@hrishin @rupalibehera FYI

hrishin:
@gabemontero Thanks for raising this issue. Which OpenShift version you have used to get `imagestreamtags`?

cc: @piyush1594

piyush1594:
@hrishin This seems like an issue which affected FMP too

hrishin:
@piyush1594 could you refer an issue from F8MP for this?

piyush1594:
Hey @hrishin i am talking about this https://github.com/fabric8io/fabric8-maven-plugin/pull/1225

piyush1594:
@gabemontero From the issue comments you mentioned, it seems like an error with client 3.0.3 onwards.

gabemontero:
that is correct @piyush1594 ... 3.0.x onward 

it works with v2.3.1 of the fabric client

@hrishin - answers to your questions:

1) we were testing against v3.9 of openshift ... the GA of that version just landed earlier this week

2) we produced it running the plugin extended tests we've discussed previously
   a) long term, you guys really need to get to a point where you can run this, if your are still having problems after the various instruction details I sent, the only other thing I can think of is to have a bluejeans session and try to sort it out live 

    or you can get PR https://github.com/fabric8io/jenkins-sync-plugin/pull/12 to a point where you guys have our latest stuff, and then we can start submitting your new features against upstream 

   b) short term, if you can at least get a 3.9 openshift environment up in however way you get it up, to execute the failing test, run
     i) oc create -f https://raw.githubusercontent.com/openshift/origin/master/test/extended/testdata/imagestreamtag-jenkins-slave-pods.yaml
     ii) oc new-app -f https://raw.githubusercontent.com/openshift/origin/master/test/extended/testdata/jenkins-slave-template.yaml
     iii) oc start-build openshift-jee-sample
     
    the corresponding jenkins job run failed for us because the imagestreamtag could not be retrieved 

    See https://github.com/fabric8io/jenkins-sync-plugin/pull/12#issuecomment-366088791 on the type of debug I added to confirm that we could not retrieve the imagestreamtag

piyush1594:
[...]

PR #1060, opened 2018-04-17
Added ImageStreamTagTest both unit and regression

piyush1594:
modified ImageStreamTagExample to work properly
Removed unused variable from test
#1046


piyush1594:
This will not work until the new version of model gets reflected in the client. Added test to get the bug on time. Thanks

piyush1594:
New model of model has been reflected and tests seems to be working now

oscerd:
[merge]

PR #1074, opened 2018-05-10
Fixes premature call to watcher onClose

nicolaferraro:
This is related to https://github.com/fabric8io/fabric8-maven-plugin/issues/1263

The client tries to create a standard watch, then falls back to a http watch in some cases (to be understood why it always do so in f-m-p). The problem is that when it cleans up the standard watch, it closes the websocket but it also calls the "onClose" method on the user watcher, and it shouldn't, because the user watcher is going to be used by the fallback http watch.

This avoids the call to the user onClose callback if the watcher is going to be reused.


oscerd:
ok to test

rohanKanojia:
[merge]

PR #1086, opened 2018-05-30
Determine kubernetes service host and port from environment if available

Vlatombe:
This can differ from `kubernetes.default.svc` in some occasions.

Uses the same logic as the go client (https://github.com/kubernetes/client-go/blob/master/rest/config.go#L313-L336)

PR #1097, opened 2018-06-12
Remove Test

piyush1594:
If we keep that test, it will return different URl for
every other CI. If we do some changes in test, it will
be similar to other tests like setting masterUrl and all.
This started happening after PR #1086
Asserting not null and is URL is the only solution
but does not make any difference as we have another tests for this func.

### DIFF
--- a/kubernetes-client/src/test/java/io/fabric8/kubernetes/client/ConfigTest.java
+++ b/kubernetes-client/src/test/java/io/fabric8/kubernetes/client/ConfigTest.java
@@ -199,14 +199,6 @@ public class ConfigTest {
   }
 
   @Test
-  public void testWithServiceAccount() {
-    System.setProperty(Config.KUBERNETES_KUBECONFIG_FILE, "/dev/null");
-    Config config = Config.autoConfigure(null);
-    assertNotNull(config);
-    assertEquals("https://kubernetes.default.svc/", config.getMasterUrl());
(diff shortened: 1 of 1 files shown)


piyush1594:
This may be because of #1086

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/ConfigTest.java
```diff
@@ -201,6 +201,7 @@ private static String decodeUrl(String url) {
   @Test
   public void testWithServiceAccount() {
     System.setProperty(Config.KUBERNETES_KUBECONFIG_FILE, "/dev/null");
+    System.setProperty(Config.KUBERNETES_MASTER_SYSTEM_PROPERTY, "https://kubernetes.default.svc/");
```
hrishin: Here service account could return the different value hence assertion could vary from environment to environment. 

To fix it, Its better way to have deterministic master URL value for SA or let be as this PR to temp fix.

hrishin: Since its blocking the release. Can we let it go and create another issue to fix this?

cc: @iocanel

piyush1594:
@Vlatombe Would you like to review and provide feedback for this?

Issue #888, opened 2017-10-19
Bearer token not passed for SubjectAccessReviews


jcantrill:
@iocanel  Following is the result of my investigation.  I can't seem to find where I am not configuring something correctly.  Once case the token is provided in the header, the other it is not:
```
public class Test {

    public static void main(String[] args) throws Exception{
        ConfigBuilder builder = new ConfigBuilder()
                .withOauthToken("t4yZV4PsA2KbQ6ziuVLm6Hwr3GcIe0LYS9n_S4TCpzo")
                .withMasterUrl("https://192.168.122.89:8443");
        try(NamespacedOpenShiftClient osClient = new DefaultOpenShiftClient(builder.build())){
            SubjectAccessReviewResponse response = osClient
                    .inAnyNamespace()
                    .subjectAccessReviews()
                    .createNew()
                    .withApiVersion("authorization.openshift.io/v1")
                    .withVerb("get")
                    .withResource("pods/log")
                    .done();
            
                List<Project> projects = osClient.projects().list().getItems();
        }
    }

}
```
Note the SAR includes the token where the listing of projects does not

```
15:01:04.191 [main] DEBUG io.fabric8.kubernetes.client.Config - Trying to configure client from Kubernetes config...
15:01:04.200 [main] DEBUG io.fabric8.kubernetes.client.Config - Found for Kubernetes config at: [/home/jeff.cantrill/.kube/config].
Oct 19, 2017 3:01:04 PM okhttp3.internal.platform.Platform log
INFO: --> GET https://192.168.122.89:8443/ http/1.1
Oct 19, 2017 3:01:04 PM okhttp3.internal.platform.Platform log
INFO: Host: 192.168.122.89:8443
Oct 19, 2017 3:01:04 PM okhttp3.internal.platform.Platform log
INFO: Connection: Keep-Alive
Oct 19, 2017 3:01:04 PM okhttp3.internal.platform.Platform log
INFO: Accept-Encoding: gzip
Oct 19, 2017 3:01:04 PM okhttp3.internal.platform.Platform log
INFO: User-Agent: okhttp/3.8.1
Oct 19, 2017 3:01:04 PM okhttp3.internal.platform.Platform log
INFO: --> END GET
Oct 19, 2017 3:01:04 PM okhttp3.internal.platform.Platform log
INFO: <-- 200 OK https://192.168.122.89:8443/ (5ms)
[...]

PR #1101, opened 2018-06-19
fix (#888): OpenShiftOAuthInterceptor now always tries to set a token.

iocanel:


oscerd:
[merge]

jcantrill:
@iocanel @hrishin   Do you know when this will be available on Maven and what version?  @ewolinetz this should allow us to enable the integration tests for the openshift-elasticsearch-plugin

PR #1107, opened 2018-06-27
Fix typo in private var name.

### LINKED ISSUE #1106: what is 'curdMode' in mock/KubernetesServer
See:
https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/KubernetesServer.java

There is a constructor arg `boolean curdMode`, which maybe is `crudMode` (create/replace/update/delete). 
Anyway it would be awesome to have at least a one-line comment there what it is and when to use which mode.

I am happy to send a PR (also renaming curd to crud if that is what it is).

ensonic:
Rename 'curdMode' to 'crudMode' and add a comment what it does.
Fixes #1106.


oscerd:
ok to test

piyush1594:
[merge]

Issue #1089, opened 2018-06-01
Query parameters are not correctly processed if set in `masterUrl`


ibuziuk:
```
          String masterUrl = "https://oso-master-url?param=value";

          ConfigBuilder configBuilder =
                  new ConfigBuilder()
                      .withMasterUrl(masterUrl)
                      .withNamespace("namespace)
                      .withTrustCerts(true)
                      .withOauthToken("token");

          KubernetesClient client = new DefaultOpenShiftClient(configBuilder.build());
          List<ServiceAccount> serviceAccounts = client.serviceAccounts().list().getItems();
          for (ServiceAccount sa : serviceAccounts) {
              System.out.println(sa);
          }
```

The sample above fails with the following trace:

> Exception in thread "main" io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://oso-master-url/?param=value/api/v1/namespaces/namespace/serviceaccounts. 

Basically, query parameter becomes a url segment and instead of expected `https://oso-master-url/api/v1/namespaces/namespace/serviceaccounts?param=value`, the following url is used while performing requests to oso `https://oso-master-url/?param=value/api/v1/namespaces/namespace/serviceaccounts`. Currently, query parameters are not handled at all while generating `OpenShiftUrl` in KubernetesClient and it simply becomes a url segment.

[1] https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/URLUtils.java#L34

hrishin:
@ibuziuk its limitation to add query string params to the client request. Possible option to send query params is using OKHttpClient request interceptor. (though its kind of overkill to add query parameters)

```java
  @Test
  public void queryParamTest() {
    String masterUrl = "https://MASTER_URL";


    ConfigBuilder configBuilder =
      new ConfigBuilder()
        .withMasterUrl(masterUrl)
        .withNamespace("default")
          .withToken("token");

    Map<String, String> params = new HashMap<String, String>();
    params.put("param1", "value1");

    OkHttpClient httpClient= new OkHttpClient.Builder().addInterceptor(new QueryStringParamInterceptor(params )).build();
    KubernetesClient client = new DefaultKubernetesClient(httpClient, configBuilder.build());

    client.serviceAccounts().list().getItems();


  }

  private class QueryStringParamInterceptor implements Interceptor {

    private final Map<String, String> params;

[...]

PR #1108, opened 2018-06-27
fixes(#1089): Appends query params at the end of the URL.

lordofthejars:
This PR appends query params at the end of the URL so in case of master URL has parameters they are appended correctly at the end.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 #### 3.2.1 (To be released)
 
+  Bugs
+    * Fix #1089 : Query parameters are not correctly processed if set in `masterUrl`
+
 #### 3.2.0
   Bugs
    * Fix #1083 : Mock Kubernetes server only handles core and extensions API groups
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/URLUtils.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/URLUtils.java
@@ -15,17 +15,39 @@
  */
 package io.fabric8.kubernetes.client.utils;
 
+import java.net.URI;
+import java.net.URISyntaxException;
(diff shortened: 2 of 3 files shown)


piyush1594:
ok to test

piyush1594:
@lordofthejars You need to squash the commits. Also, you need to make an an entry to changelog.md

lordofthejars:
They can be squashed when PR merging.

@bartoszmajsak could we add an ike plugin for checking changelog?

lordofthejars:
@piyush1594 Before sending changelog, build was green in OS3.7 and red in OS3.6, now it is green in 3.6 and red in 3.7, (flaky build?)

bartoszmajsak:
> @bartoszmajsak could we add an ike plugin for checking changelog?

this would mean new plugin, but sure we can hack it together

>  Although the ideal scenario would be that changelog was created automatically.

yeah. using properly prefixed commits landing to master that should be achievable

hrishin:
@lordofthejars 

Do we have to address following one case for handling URL?

```java
  // could name if better. 
  public void shouldHandleMultiParamsURI() {

    // Given
    String masterUrl = "https://oso-master-url:8888/kubernetes?param1=value";

    // When

    final String fullUrl = URLUtils.join(masterUrl, "port?port=port");

    // Then
    assertThat(fullUrl, is("https://oso-master-url:8888/kubernetes/port?param1=value&port=port"));
  }
```

Unfortunately, there are few [places](https://github.com/fabric8io/kubernetes-client/blob/master/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/BuildOperationsImpl.java#L166) where URL passed in different way.

lordofthejars:
@hrishin humm true, not easy at all all the possible use cases, but  now PR is merged :( So new PR then.

Issue #1066, opened 2018-04-26
Add support for missing RBAC model and operation classes


iocanel:

iocanel:
Yeah, I think that there are missing stuff.

piyush1594:
@iocanel @hrishin Adding the support of Role and RoleBinding then

piyush1594:
Closed by #1111

PR #1111, opened 2018-06-28
Adds support for Kubernetes RBAC Role and Role Binding

piyush1594:
#1066

### DIFF
--- a/kubernetes-itests/src/test/java/io/fabric8/kubernetes/KubernetesRoleBindingIT.java
+++ b/kubernetes-itests/src/test/java/io/fabric8/kubernetes/KubernetesRoleBindingIT.java
@@ -1,0 +1,186 @@
+/**
+ * Copyright (C) 2015 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.fabric8.kubernetes;
+
+import io.fabric8.kubernetes.api.model.rbac.KubernetesRoleBinding;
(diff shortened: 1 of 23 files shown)


kubernetes-itests/src/test/java/io/fabric8/kubernetes/KubernetesRoleIT.java
```diff
@@ -0,0 +1,207 @@
...
+      .build();
+
+    client.rbac().kubernetesRoles().inNamespace(currentNamespace).createOrReplace(kubernetesRole);
+  }
+
+  @Test
+  public void get() {
+    kubernetesRole = client.rbac().kubernetesRoles().inNamespace(currentNamespace).withName("job-reader").get();
+
+    assertNotNull(kubernetesRole);
```
hrishin: Better to have serializer and deserializer tests than too many assertions.

piyush1594: @hrishin Here serializer and deserializer that does not make sense as these are real time tests and when you create real time object, lots of fields get added like creationTimestamp and there are more.

hrishin: @piyush1594 make sense then.

kubernetes-itests/src/test/java/io/fabric8/kubernetes/KubernetesRoleBindingIT.java
```diff
@@ -0,0 +1,201 @@
...
+      .build();
+
+    client.rbac().kubernetesRoleBindings().inNamespace(currentNamespace).createOrReplace(kubernetesRoleBinding);
+  }
+
+  @Test
+  public void get() {
+
+    // Do not run tests on opeshift 3.6 and 3.6.1
+    assumeFalse(checkClusterVersion());
```
hrishin: Instead of repeating for every case, it's better to put into init() function.

piyush1594:
ok to test

piyush1594:
@hrishin Updated the PR as suggested

iocanel:
@piyush1594: I'd ditch 3.6.0 altogether but keep 3.6.1 around.

rohanKanojia:
Thanks @piyush1594 @iocanel .

rohanKanojia:
[merge]

piyush1594:
@iocanel Its the same case for both 3.6.0 and 3.6.1, that's why I skipped on both

Issue #1112, opened 2018-06-28
Query parameters are not correctly processed if set in `masterUrl` with endpoints with QP


lordofthejars:
See https://github.com/fabric8io/kubernetes-client/pull/1108#issuecomment-400949245 where it is explained this error case.

piyush1594:
Closed in #1114

PR #1114, opened 2018-06-28
fix(#1112): Append multiple query params

lordofthejars:


kubernetes-client/src/test/java/io/fabric8/kubernetes/client/utils/URLUtilsTest.java
```diff
@@ -40,6 +40,32 @@ public void shouldJoinUrlWithoutQueryParams() {
 
   }
 
+  @Test
+  public void shouldAppendUrlWithQueryParams() {
+    // Given
+    String masterUrl = "https://oso-master-url:8888/kubernetes?param1=value";
+
+    // When
+
```
hrishin: please drop the blank line

lordofthejars:
@hrishin if this class was better designed then it would be possible (this class should be a URLBuilder) not an appender (the problem is that making this change implies a deep refactor on callers).

I have tried to do the best as I could the problem is that I need to deal with strings all the time, and all possible problems with special characters. I can try to create some methods, but ay idea would be welcomed as well.

hrishin:
@rohanKanojia @piyush1594 review, please.

rohanKanojia:
Please squash your commits

oscerd:
We ask to contributor to squash commit. Usually when I contribute to a project I adapt myself to the rules I find.

lordofthejars:
I try to improve any project in any sense, not only code but also in the mechanics, but if the project is fine in losing the history of why a PR looks like it looks, then it is fine for me. I was just trying to raise my hand about this possibility that's all.

oscerd:
We can discuss about that. But it wasn't sounding like "raise my hand", it was looking more like "you're doing things wrong, listen to me".

lordofthejars:
This is what happens with text messages, that you cannot give feeling hehehe

oscerd:
Sure :-) Lets start a discussion about that in an issue :-)

piyush1594:
@lordofthejars So the sentiment behind this is, when you merge the things in master, they will show in history, when the commit was done, not when the PR is merged, afaik. So When PR with multiple commits gets merged, then commits shows up at different places in history and that confuses, so making a single commit with a message like `fix #this` and that commit message will contain the messages of all the commits in one combined, and the user can check git diff for all the relevant changes. Feel free if it does not look good. So we follow, 1 issue, 1 pr, 1 commit policy.

lordofthejars:
@piyush1594 Basically the idea is the PR contains everything. Then when everyone is ok with the changes you merge the PR to master using the button Squash & Merge. When you push this button by default the first commit will be the title of the commit and the other commits will be part of the description. At this point, you can either delete these comments from description, change the title, or just merge. Notice that this is pretty much the same as when you squash a PR on your machine.

Since it is an squash the commits does not show in different places, there is one commit squashed when you are merging, pretty much the same as when you do it locally.

The advantages of this method:

* It shows when the PR is merged, since squash rewrites, which is fine because this is what you want in master history, when it was effective.
* You maintain the PR all the evolution of code, so if you need to know why a PR looks like it is, you only need to check the history and read the code review and see how the commits are evolving
* If you need to squash locally you can do it wrong (you squash more than you need), and then if you don't notice and do a push force in the PR then everything is screwed, in this way GH takes care of this.

bartoszmajsak:
> This is the policy adopted in this project, bad practice or not.

So it's written in stone and can't be improved? :)

> We ask to contributor to squash commit. 

Just wondering why asking to do manual work if this can be automated by a simple click on GH UI called "squash and merge"? On top of that all what @lordofthejars said about losing history/evolution of the PR is valid here too.

oscerd:
@bartoszmajsak maybe it's better if you read the whole discussion. It's not written in stone, but it wasn't sounding like a suggestion or an improvement. We can discuss about this and it's not a problem. But if someone contribute to a project, open a PR and start criticizing how things are done, well it's not good for people maintaining this stuff.

lordofthejars:
I think I have not criticized anything, as I said maybe at first it sounds like, but I think I clarified in the next comment so I am not sure why are you saying this now.

oscerd:
This project is still alive because of these guys working hard on maintaining and improving it. If someone arrives in a project you're maintaining and tell you "you're doing things wrong" your reaction will be exactly the same.

bartoszmajsak:
@oscerd If the critique is constructive I don't mind it really. At the end of the day it's about the code / practice around, not people. Also keep in mind that not all people are native english speaker, so it could be a good approach to look at what they write through "take it easy" lenses. Just my .2$

oscerd:
@lordofthejars I was illustrating what happened to @bartoszmajsak . You already clarified and I believe we are fine.

oscerd:
@bartoszmajsak @lordofthejars no problem from my side, I was just explaining why my reaction was like that.

bartoszmajsak:
@oscerd I read the whole thread and I think people got a bit too emotional here. But hey, it's Monday. Grab your coffee and happy hacking.

oscerd:
I'm already at my third coffee. It's better to stop.

lordofthejars:
@oscerd I am part of this team, currently, I am in the same team of all of them for this reason I said that I am not criticizing just trying to improve work to work day, because I am working with the project.

And as I said my intention was not to criticize and I have clarified ASAP.

bartoszmajsak:
> I'm already at my third coffee. It's better to stop.

@oscerd lol, same here :)

oscerd:
I didn't know about that. By the way, I explained why I reacted in that way. Probably I have interpreted the comment the wrong way. That's the problem, as you said, of text messages.

rohanKanojia:
@lordofthejars : Have you tried building these tests locally on these openshift versions?

lordofthejars:
@rohanKanojia not in all versions but I have noticed that all builds were successful, then I push the changelog file and then I get for example two failures and one success so the problem is not in tests at all but in the environment. (Or said in another way, the tests works always well when run locally)

Issue #1085, opened 2018-05-30
Impersonate-User / Impersonate-Group / Impersonate-Extra-* are not configurable for DefaultOpenShiftClient (works fine with DefaultKubernetesClient)


ibuziuk:
`Impersonate-User / Impersonate-Group / Impersonate-Extra-` are meant to be supported by both OpenShift & Kubernetes clients via `ImpersonatorInterceptor` [1]. However, the following sample with setting `Impersonate-User` header works only for Kubernetes client:

```
       ConfigBuilder configBuilder =
                  new ConfigBuilder()
                      .withMasterUrl(masterUrl)
                      .withNamespace("user-namespace")
                      .withTrustCerts(true)
                      .withOauthToken(cheServiceAccountToken);
                   

          Config config = configBuilder.build();
          config.getRequestConfig().setImpersonateUsername("username");

          KubernetesClient client = new DefaultKubernetesClient(config);
          List<Pod> pods = client.pods().list().getItems();
          for (Pod pod : pods) {
              System.out.println(pod.getMetadata().getName());
          }
```
If  `KubernetesClient client = new DefaultKubernetesClient(config);` from the sample would be changed to `DefaultOpenShiftClient` the Impersonate header would not be set by `ImpersonatorInterceptor`[1] during request processing. The reason why  `Impersonate-*` headers are not set for OpenShiftClient is because `Config` [2] constructor that is used internally during transformation to `OpenShiftConfig` does not honour `Impersonate` settings during building `RequestConfig` [3].  RequestConfig is later used in `ImpersonatorInterceptor` for setting headers.

[1] https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/ImpersonatorInterceptor.java#L36
[2] https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java#L213
[3] https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/RequestConfig.java#L56

ibuziuk:
[...]

PR #1117, opened 2018-07-02
fix(#1085): Propagate Impersonate attributes

lordofthejars:
#1085

### DIFF
--- a/kubernetes-client/src/test/java/io/fabric8/kubernetes/client/ConfigTest.java
+++ b/kubernetes-client/src/test/java/io/fabric8/kubernetes/client/ConfigTest.java
@@ -17,15 +17,12 @@
 package io.fabric8.kubernetes.client;
 
 import io.fabric8.kubernetes.client.utils.Serialization;
-import io.fabric8.openshift.clnt.v3_1.DefaultOpenShiftClient;
-import io.fabric8.openshift.clnt.v3_1.OpenShiftClient;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
 import java.net.MalformedURLException;
 import java.net.URLDecoder;
-
 import java.util.HashMap;
 import java.util.Map;
 import okhttp3.OkHttpClient;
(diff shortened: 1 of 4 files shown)


piyush1594:
ok to test

piyush1594:
@lulf @ibuziuk Can you please take a look?

lulf:
LGTM. Consider adding a test to ConfigTest.

lordofthejars:
@lulf Do you think it is worth to implement these parameters to be set as system properties as well? Or just propagating?

lulf:
@lordofthejars Agreed, it would make it consistent with how the other settings are handled.

ibuziuk:
LGTM

lordofthejars:
@lulf Added tests and sys prop support. I have not implemented system properties for extra parameters. My point is that extra parameters field is a map, and usually you have two options, the first one is creating a prefixed system property where last part of key is the key and that value the map:

i.e:

```
kubernetes.impersonate.extras.key1=value
kubernetes.impersonate.extras.key2=vale
```

But you need to scan for all preferences and env vars and iterate.

The other option is to parse a string: `kubernetes.impersonate.extras=key1=value, key2=value`

Based on my experience is that if it is not totally required, the best way is to avoid configuring maps but what do you think about this concrete case?

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/ConfigTest.java
```diff
@@ -17,13 +17,17 @@
 package io.fabric8.kubernetes.client;
 
 import io.fabric8.kubernetes.client.utils.Serialization;
+import io.fabric8.openshift.clnt.v3_1.DefaultOpenShiftClient;
```
piyush1594: You need to fix this import @lordofthejars

piyush1594:
@lordofthejars Please add changelog.md also

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/ConfigTest.java
lordofthejars: Done

lordofthejars:
@piyush1594 Done, when everyone approves the PR I will squash everything.

lulf:
@lordofthejars LGTM

piyush1594:
Thanks @lordofthejars

Issue #1119, opened 2018-07-04
Regression Tests Breaking


piyush1594:
Regression Tests are breaking more often because of these 3 tests and are kind of blocking for all the PR's. We should make it work more reliably.

```
Errors: 
[ERROR]   DeploymentIT.waitTest:130 » KubernetesClientTimeout Timed out waiting for [0] ...
[ERROR]   PodIT.exec:139 » KubernetesClientTimeout Timed out waiting for [0] millisecond...
[ERROR]   PodIT.log:132 » KubernetesClientTimeout Timed out waiting for [0] milliseconds...
```

The error is something like this 
```
[ERROR] Tests run: 7, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 1,411.297 s <<< FAILURE! - in io.fabric8.kubernetes.PodIT
[ERROR] log(io.fabric8.kubernetes.PodIT)  Time elapsed: 630.268 s  <<< ERROR!
io.fabric8.kubernetes.client.KubernetesClientTimeoutException: Timed out waiting for [0] milliseconds for [Pod] with name:[pod1-ztkffc] in namespace [itest-7f7810e0].
	at io.fabric8.kubernetes.PodIT.log(PodIT.java:132)

[ERROR] exec(io.fabric8.kubernetes.PodIT)  Time elapsed: 630.28 s  <<< ERROR!
io.fabric8.kubernetes.client.KubernetesClientTimeoutException: Timed out waiting for [0] milliseconds for [Pod] with name:[pod1-w6pfcl] in namespace [itest-7f7810e0].
	at io.fabric8.kubernetes.PodIT.exec(PodIT.java:139)
```

```
[ERROR] Tests run: 6, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 828.331 s <<< FAILURE! - in io.fabric8.kubernetes.DeploymentIT
[ERROR] waitTest(io.fabric8.kubernetes.DeploymentIT)  Time elapsed: 637.4 s  <<< ERROR!
io.fabric8.kubernetes.client.KubernetesClientTimeoutException: Timed out waiting for [0] milliseconds for [Deployment] with name:[deployment1] in namespace [itest-7f7810e0].
	at io.fabric8.kubernetes.DeploymentIT.waitTest(DeploymentIT.java:130)
```

piyush1594:
Closed in #1121

PR #1121, opened 2018-07-04
Improved Pod Exec, Pod Log and Deployment wait Test

piyush1594:
#1119

### DIFF
--- a/kubernetes-itests/src/test/java/io/fabric8/kubernetes/DeploymentIT.java
+++ b/kubernetes-itests/src/test/java/io/fabric8/kubernetes/DeploymentIT.java
@@ -23,7 +23,6 @@ import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.kubernetes.client.internal.readiness.Readiness;
 import org.arquillian.cube.kubernetes.api.Session;
 import org.arquillian.cube.kubernetes.impl.requirement.RequiresKubernetes;
-import org.arquillian.cube.openshift.api.OpenShiftResource;
 import org.arquillian.cube.requirement.ArquillianConditionalRunner;
 import org.jboss.arquillian.test.api.ArquillianResource;
 import org.junit.*;
@@ -36,6 +35,7 @@ import java.util.concurrent.TimeUnit;
 import static junit.framework.TestCase.assertNotNull;
 import static junit.framework.TestCase.assertTrue;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
 import static org.junit.Assert.assertEquals;
 
 @RunWith(ArquillianConditionalRunner.class)
@@ -78,12 +78,12 @@ public class DeploymentIT {
       .endSpec()
       .build();
 
-    client.extensions().deployments().inNamespace(currentNamespace).createOrReplace(deployment1);
+    client.extensions().deployments().inNamespace(currentNamespace).create(deployment1);
   }
 
   @Test
   public void load() {
-    Deployment aDeployment = client.extensions().deployments().inNamespace(currentNamespace).load(getClass().getResourceAsStream("/test-deployments.yml")).get();
+    Deployment aDeployment = client.extensions().deployments().load(getClass().getResourceAsStream("/test-deployments.yml")).get();
     assertThat(aDeployment).isNotNull();
     assertEquals("nginx-deployment", aDeployment.getMetadata().getName());
   }
@@ -105,7 +105,7 @@ public class DeploymentIT {
   @Test
   public void update() {
     deployment1 = client.extensions().deployments().inNamespace(currentNamespace).withName("deployment1").edit()
-      .editSpec().withReplicas(2).and().done();
+      .editSpec().withReplicas(2).endSpec().done();
     assertThat(deployment1).isNotNull();
     assertEquals(2, deployment1.getSpec().getReplicas().intValue());
   }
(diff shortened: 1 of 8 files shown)


rohanKanojia:
Hm, let's test this change a few times and see how it behaves; whether we get intermittent failures or not.

piyush1594:
Tests for this PR went green in one go, I think we can merge this and  let's check on PR's about the behaviour

piyush1594:
@oscerd @hrishin @lordofthejars WDYT? Doing force git push and blocking CI will not be a good thing as we can run only 4 jobs at a time. And it's a very small change.

kubernetes-itests/src/test/java/io/fabric8/kubernetes/DeploymentIT.java
```diff
@@ -128,6 +128,6 @@ public void cleanup() throws InterruptedException {
     Thread.sleep(6000);
     client.extensions().deployments().inNamespace(currentNamespace).delete();
     // Wait for resources to get destroyed
-    Thread.sleep(30000);
+    Thread.sleep(60000);
```
hrishin: How come 60000?

piyush1594: Just a guess that 1 minute will be good, and also CI went green. So

lordofthejars: @piyush1594 Do not merge yet. Is there a way to query if namespace is deleted? Something like `client.extensions.inNamespace.status`

piyush1594: Hey @lordofthejars 

You can try

```
client.namespaces().list();
```
and
```
client.namespaces().withName("").get();
```

lordofthejars: Ok, then look we can use awaitility instead of a thread.sleep so you can do:
https://github.com/awaitility/awaitility

```java
await().atMost(2, MINUTE).until(namespaceDeleted());
```

So this internally implements a poling so it will not take all the time one minute but just the amount of time we need. So we can put 2 minutes for just in case there are some slow machines.

piyush1594: so here we are deleting the deployment and pod, so we need to put that condition @lordofthejars

lordofthejars: You need to replace the `Thread.sleep` for the `await()` call

hrishin: agreed with @lordofthejars.

piyush1594:
@hrishin @lordofthejars I have addressed your concerns. Please review and feel free to comment.

kubernetes-itests/src/test/java/io/fabric8/kubernetes/PodIT.java
```diff
@@ -168,6 +170,24 @@ public void onClose(int i, String s) {
...
+
+  private Callable<Boolean> podDeleted() {
+    return new Callable<Boolean>() {
+      public Boolean call() throws Exception {
+        return client.pods().inNamespace(currentNamespace).list().getItems().size() == 0; // The condition that must be fulfilled
+      }
+    };
+  }
+
+  private Callable<Boolean> podReady() {
```
lordofthejars: I will create a class called PodReady that implements `Callable`  and the constructor will receive a string that will be the pod name so they check if it ready. So instead of defining this logic in all tests, you can reuse it.

kubernetes-itests/src/test/java/io/fabric8/kubernetes/PodIT.java
```diff
@@ -168,6 +170,24 @@ public void onClose(int i, String s) {
   public void cleanup() throws InterruptedException {
     client.pods().inNamespace(currentNamespace).withName(pod1.getMetadata().getName()).delete();
     // Wait for resources to get destroyed
-    Thread.sleep(30000);
+    await().atMost(2, TimeUnit.MINUTES).until(podDeleted());
+  }
+
+
+  private Callable<Boolean> podDeleted() {
```
lordofthejars: I will create a `PodDelete` interface implementing `Callable` which checks if pods are deleted.

kubernetes-itests/src/test/java/io/fabric8/kubernetes/DeploymentIT.java
```diff
@@ -113,21 +114,41 @@ public void update() {
...
+    await().atMost(15, TimeUnit.MINUTES).until(deploymentReady());
+    Deployment deploymentOne = client.extensions().deployments()
+      .inNamespace(currentNamespace).withName("deployment1").get();
     assertTrue(Readiness.isDeploymentReady(deploymentOne));
   }
 
   @After
   public void cleanup() throws InterruptedException {
-    Thread.sleep(6000);
+    await().atMost(15, TimeUnit.MINUTES).until(deploymentReady());
```
hrishin: 15 minutes would be long, isn't it? what was the previous time?

piyush1594: Previously it was 10 Minutes, but it was failing, that's why increased it to 15 minutes but that is the max limit. If the condition met earlier it will proceed.

lordofthejars:
Now looks awesome, let's see if the timeout works.

lordofthejars:
@piyush1594 now it seems it is working. Have you tried 3 or 4 times to be sure we have not been lucky at first try?

piyush1594:
I tried locally around 20 times and here twice.

lordofthejars:
Go, Go, Go

oscerd:
[merge]

PR #1123, opened 2018-07-06
Improve regression test

piyush1594:
Improve regression test to only build the client and not run unit test
As we have another job for this, this will decrease the rt execution time


pom.xml
```diff
@@ -393,6 +401,14 @@
       </dependencies>
       <build>
         <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
```
rohanKanojia: Nitpick, it would be nice if you define this value in `<properties>` on top of the file

piyush1594:
This will save approx 10 minutes.

piyush1594:
@oscerd Can you merge this, please ?

oscerd:
[merge]

Issue #1099, opened 2018-06-15
CustomResourceDefinitions: withResourceVersion() causes NoSuchMethodError


ljnelson:
I'll work over the next few days to get a reproducer, but I'm following the general example code in [`CRDExample.java`](https://github.com/fabric8io/kubernetes-client/blob/3b5c121c95725abfb6c4150c1e60515e2f8c3d1c/kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/CRDExample.java#L78).

I'm using `kubernetes-client` 3.2.0 and `kubernetes-model` 2.1.1.

I have a custom resource definition and a `CustomResource` of type `Foo`.  I have a `MixedOperation<Foo, FooList, DoneableFoo, Resource<Foo, DoneableFoo>>` instance in my hand that resulted from invoking `client.customResources(fooCrd, Foo.class, FooList.class, DoneableFoo.class)`.

Now instead of calling `list()` right away on it, I call `withResourceVersion(0)` in my particular use case (which for whatever it's worth is building out the Java equivalent of the Kubernetes-native `tools/cache` package).  (Note that the [Go code does the equivalent](https://github.com/kubernetes/client-go/blob/8d6e3480fc03b7337a24f349d35733190655e2ad/tools/cache/listwatch.go#L176).)

With non-custom resources this `withResourceVersion(0)` invocation followed by a `.list()` invocation works just fine.

With custom resources, though, I see this (edited) stack:

```
io.fabric8.kubernetes.client.KubernetesClientException: An error has occurred.
	at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:62)
	at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:53)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.withResourceVersion(BaseOperation.java:685)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.withResourceVersion(BaseOperation.java:49)
[snip]
[...]

PR #1124, opened 2018-07-06
fix(#1099): CustomResourceDefinitions: withResourceVersion() causes NoSuchMethodError

mgorniew:
Fixes issue [#1099](https://github.com/fabric8io/kubernetes-client/issues/1099).

+ changes how file paths to resources are resolved in tests (to work also Windows).

### DIFF
--- a/kubernetes-client/src/test/java/io/fabric8/kubernetes/client/ConfigTest.java
+++ b/kubernetes-client/src/test/java/io/fabric8/kubernetes/client/ConfigTest.java
@@ -17,20 +17,20 @@
 package io.fabric8.kubernetes.client;
 
 import io.fabric8.kubernetes.client.utils.Serialization;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.InputStream;
-import java.io.UnsupportedEncodingException;
-import java.net.MalformedURLException;
-import java.net.URLDecoder;
-import java.util.HashMap;
-import java.util.Map;
+import io.fabric8.kubernetes.client.utils.Utils;
 import okhttp3.OkHttpClient;
 import okhttp3.TlsVersion;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.util.HashMap;
+import java.util.Map;
 
 import static okhttp3.TlsVersion.TLS_1_1;
 import static okhttp3.TlsVersion.TLS_1_2;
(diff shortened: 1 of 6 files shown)


piyush1594:
ok to test

piyush1594:
Thanks @mgorniew for the PR. Can you please add changelog.md also

ljnelson:
Are you sure this fixes #1099?  I haven't tested the fix and I'm not an expert in the codebase but is this really the root cause?  Maybe there was a typo on the issue number?

kubernetes-client/src/test/java/io/fabric8/kubernetes/TestUtils.java
```diff
@@ -0,0 +1,20 @@
+package io.fabric8.kubernetes;
```
rohanKanojia: license header missing here

rohanKanojia:
@mgorniew : Please squash your commits too ;-) . Also run `mvn -N license:format` to update all license headers :)

kubernetes-client/src/test/java/io/fabric8/kubernetes/TestUtils.java
rohanKanojia: After looking at this patch closely, I think it would be better to move this method to `Utils` instead of creating a separate `TestUtils`

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/ConfigTest.java
```diff
@@ -16,13 +16,17 @@
...
 package io.fabric8.kubernetes.client;
 
+import io.fabric8.kubernetes.TestUtils;
 import io.fabric8.kubernetes.client.utils.Serialization;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
 import java.net.MalformedURLException;
+import java.net.URISyntaxException;
```
rohanKanojia: I don't think we need these imports ;-)

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/internal/CertUtilsTest.java
```diff
@@ -15,14 +15,18 @@
...
 import io.fabric8.kubernetes.client.DefaultKubernetesClient;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.kubernetes.client.utils.Utils;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
+import java.net.URISyntaxException;
+import java.net.URL;
 import java.net.URLDecoder;
+import java.nio.file.Paths;
```
rohanKanojia: Same here.. some unused imports

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/ConfigTest.java
```diff
@@ -234,7 +230,7 @@ public void testWithKubeConfig() {
     assertEquals("https://172.28.128.4:8443/", config.getMasterUrl());
     assertEquals("testns", config.getNamespace());
     assertEquals("token", config.getOauthToken());
-    assertTrue(config.getCaCertFile().endsWith("testns/ca.pem"));
```
rohanKanojia: @mgorniew : How does this test the change you made in `CustomResourceOperationsImpl`?

rohanKanojia:
@mgorniew : Hi, How did you test this change? Were you able to run `CRDExample` with this change? Perhaps I might be missing something here ;-)

mgorniew:
I didn't use `CRDExample`, just tested this in my code. But I've run `CRDExample` today and it works. I can add this `withResourceVersion` call to example.

rohanKanojia:
@mgorniew : Yeah, would be nice if you could add `withResourceVersion` call to that example. Also could you please address my comment regarding TestUtil ;-)

mgorniew:
Isn't `filePath` used only in test code? Why to put this method in `Utils` class?

rohanKanojia:
@mgorniew : In my opinion utils is a general purpose utility class independent of test/src, so it makes sense to move it to Util instead of creating a separate class; in case we need to use this method we can use it :-)

rohanKanojia:
[merge]

PR #1127, opened 2018-07-07
Add new option to disable hostname verification - Closes #1126

### LINKED ISSUE #1126: Add option to disable hostname verification without disabling the certificate validation
The configuration options `kubernetes.trust.certificates` / `KUBERNETES_TRUST_CERTIFICATES` makes it possible to disable the validation of server certificates and it also disables the validation of the hostname where the client is connecting against the hostnames listed in the certificate.

However in some cases it might be useful to disable the only hostname verification. For example because the user has valid certificates but is just using an incorrect hostname because of some network tunnelling etc. In such case disabling the validation completely would be too strong and unnecessary and all what would be needed is to disable the hostname validation.

scholzj:
This PR adds new option `kubernetes.disable.hostname.verification` / `KUBERNETES_DISABLE_HOSTNAME_VERIFICATION` which when set to `true` will disable the hostname verification in the OkHttp client. That should help when the client is connecting to Kubernetes cluster over an hostname which is not included in the API server certificates.


piyush1594:
ok to test

piyush1594:
@scholzj Thanks for the contribution. Can you please update CHANGELOG.md also. Thanks

scholzj:
@piyush1594 Sorry, I read it in the contributing guide but forgot it :-/. It should be fixed now. Thanks.

oscerd:
Ok to test

hrishin:
@scholzj could you please rebase the PR? _regret for the inconvenience caused._

scholzj:
@hrishin No problem - rebased.

PR #1130, opened 2018-07-09
Fix Format of CHANGELOG

piyush1594:


piyush1594:
Merged this as it was a small doc change

PR #1132, opened 2018-07-11
Fix the ability to load Deployment that uses extensions/v1beta1

geoand:


oscerd:
Ok to test

iocanel:
... also it might make sense to evaluate the possible breakage on downstream projects if we actually go for a proper fix.

geoand:
@iocanel I am all for a proper fix, but I can't really comment on the effects downstream.
That said, should we decide to pursue such a fix, I am willing to work on it

geoand:
@iocanel could you point me to the "problematic" code you mentioned in your comment?

iocanel:
@geoand: https://github.com/fabric8io/kubernetes-client/blob/df459cb490cb918500a7e92542ea57c03ff3d0c5/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java#L889

geoand:
@iocanel Thanks, I'll have a look!

geoand:
I updated the test to be more accurate in testing the fix

geoand:
@iocanel FWIW, After debugging the code I found that it's not https://github.com/fabric8io/kubernetes-client/blob/df459cb490cb918500a7e92542ea57c03ff3d0c5/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java#L889, that changes apiGroup (the method is executed  does not make any changes the check at: https://github.com/fabric8io/kubernetes-client/blob/df459cb490cb918500a7e92542ea57c03ff3d0c5/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java#L923 does  not pass - since `apiGroupVersion` is null).

What happens is that the apiGroup is hardcoded at: https://github.com/fabric8io/kubernetes-client/blob/e8e171d82d7b974a8c3356c1061da5277717444b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/DeploymentOperationsImpl.java#L54

iocanel:
Damn you @geoand! It was too convinient for me to blame someone else for it, why did u have to dig around ?

geoand:
Regardless, I will keep digging  into this and see if we can improve the situation :)

iocanel:
Well, even if its not the root cause of the issue, IMHO the fact that it proactively mutates resources complicates things. 

However, its a fight that we shall give another day. 

So, I'll approve this.

geoand:
Thanks @iocanel  for your assistance with this!

PR #1158, opened 2018-07-27
Add support for label selectors in the mock server

### LINKED ISSUE #1125: ConfigMap labels are ignored when using mock KubernetesServer
Here's the simple method that fetches list of configmaps with a particular label value:

```
public List<ConfigMap> fetchConfigMaps(KubernetesClient client, String namespace, String labelKey, String labelValue) {
    ConfigMapList configMapList = client
            .configMaps()
            .inNamespace(namespace)
            .withLabel(labelKey, labelValue)
            .list();

    return configMapList.getItems();
}
```

In the unit test below one configmap has proper label and the other one has not. So I expect that the method returns 1:

```
    @Rule
    public KubernetesServer kubernetesServerSbx = new KubernetesServer(true, true);

    @Test
    public void testFetchConfigMaps() {
        KubernetesClient client = kubernetesServerSbx.getClient();

        createNamespace(client, "ns1");
        createCM(client, "ns1", "cm1", Map.of("label1", "value"));
        createCM(client, "ns1", "cm2", Map.of("label2", "value"));

        ConfigMapFetcher fetcher = new ConfigMapFetcher();

        List<ConfigMap> configMaps = fetcher.fetchConfigMaps(client, "ns1","label1", "value");

        assertEquals(1, configMaps.size());
    }

    private static void createNamespace(KubernetesClient client, String namespace) {
        client.namespaces().create(
                new NamespaceBuilder()
                        .withNewMetadata()
                        .withName(namespace)
                        .endMetadata()
                        .build());
    }

    private static void createCM(KubernetesClient client,
                                 String namespace,
                                 String cmName,
                                 Map<String, String> labels) {

        client.configMaps().inNamespace(namespace).create(
                new ConfigMapBuilder()
                        .withNewMetadata()
                        .withName(cmName)
                        .withLabels(labels)
                        .endMetadata()
                        .build());
    }
```

However the unit test fails because the method return 2 instead of 1.

I'm using `kubernetes-client` and `kubernetes-server-mock` version 3.2.0

Here's the complete source code: [my-app.zip](https://github.com/fabric8io/kubernetes-client/files/2171556/my-app.zip)

drigz:
This supports simple label selectors (eg foo=bar). The != operator would
go beyond the limits of the AttributeSet-based matching in
CrudDispatcher.

There was a test using labels in ConfigMapCrudTest, but it was only
passing because, at the point of execution, all resources match the
label selector. This change moves changes the order of assertions so
that the effect of the label selector is observable.

This should fix #1125, although I haven't tested that code directly.

### DIFF
--- a/kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/KubernetesAttributesExtractor.java
+++ b/kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/KubernetesAttributesExtractor.java
@@ -58,8 +58,16 @@ public class KubernetesAttributesExtractor implements AttributeExtractor<HasMeta
   private static final String VALUE_GROUP = "(?<value>[a-zA-Z0-9-_.]+)";
   private static final Pattern LABEL_REQUIREMENT_EQUALITY = Pattern.compile(KEY_GROUP + EQUALITY_GROUP + VALUE_GROUP);
 
+  // These elements are added to the path/query fragment so that it can be parsed by HttpUrl. Their
+  // values are not important, HttpUrl expects the scheme to be http or https.
+  private static final String SCHEME = "http";
+  private static final String HOST = "localhost";
+
   private HttpUrl parseUrlFromPathAndQuery(String s) {
-    return HttpUrl.parse("http://localhost" + s);
+    if (!s.startsWith("/")) {
+      s = "/" + s;
+    }
+    return HttpUrl.parse(String.format("%s://%s%s", SCHEME, HOST, s));
   }
 
   @Override
--- a/kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/KubernetesAttributesExtractor.java
+++ b/kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/KubernetesAttributesExtractor.java
@@ -21,6 +21,8 @@ import io.fabric8.mockwebserver.crud.Attribute;
 import io.fabric8.mockwebserver.crud.AttributeExtractor;
 import io.fabric8.mockwebserver.crud.AttributeSet;
 import io.fabric8.zjsonpatch.internal.guava.Strings;
+import java.util.Map;
+import okhttp3.HttpUrl;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -34,9 +36,11 @@ public class KubernetesAttributesExtractor implements AttributeExtractor<HasMeta
 
   private static final Logger LOGGER = LoggerFactory.getLogger(KubernetesAttributesExtractor.class);
 
+  public static final String KEY = "key";
   public static final String KIND = "kind";
   public static final String NAME = "name";
   public static final String NAMESPACE = "namespace";
+  public static final String VALUE = "value";
 
   private static final String API_GROUP = "/o?api(s/[a-zA-Z0-9-_.]+)?";
   private static final String VERSION_GROUP = "(/(?<version>[a-zA-Z0-9-_]+))?";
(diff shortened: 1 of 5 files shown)


rohanKanojia:
ok to test

kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/KubernetesAttributesExtractor.java
```diff
@@ -47,16 +51,29 @@
...
   protected static final Pattern PATTERN = Pattern.compile(API_GROUP + VERSION_GROUP + NAMESPACE_GROUP + KIND_GROUP + NAME_GROUP + END_GROUP);
 
+  private static final String LABEL_KEY_PREFIX = "labels:";
+  private static final String KEY_GROUP = "(?<key>[a-zA-Z0-9-_./]+)";
+  // Matches a==b and a=b but not a!=b.
+  private static final String EQUALITY_GROUP = "(==|(?<!!)=)";
+  private static final String VALUE_GROUP = "(?<value>[a-zA-Z0-9-_.]+)";
+  private static final Pattern LABEL_REQUIREMENT_EQUALITY = Pattern.compile(KEY_GROUP + EQUALITY_GROUP + VALUE_GROUP);
+
+  private HttpUrl parseUrlFromPathAndQuery(String s) {
```
rohanKanojia: Is it possible to get rid of this hard coded string?

lordofthejars: Apart from what @rohanKanojia suggests to have this `localhost` as constant I will do next things in this case. I know they might sounds weird but I have done this in several projects and works pretty well in terms of readable code.

. Create a `private static final String PROTOCOL = http`
. Create a `private static final String HOST = localhost`
. Check if `s` starts with `/` or not and in case of not, append it.
. Do `HttpUrl.parse(String.format("%s://%s%s", PROTOCOL, HOST, s))`

if `s` can contain the port as well, split the method between `int port, String path` it is always a better practice to not merge concepts.

kubernetes-server-mock/src/test/java/io/fabric8/kubernetes/client/server/mock/KubernetesAttributesExtractorTest.java
```diff
@@ -84,6 +86,20 @@ public void shouldHandleResource() {
...
+    KubernetesAttributesExtractor extractor = new KubernetesAttributesExtractor();
+    Map<String, String> labels = new HashMap<>();
+    labels.put("name", "myname");
+    Pod pod = new PodBuilder().withNewMetadata().withLabels(labels).endMetadata().build();
+
+    AttributeSet attributes = extractor.extract(pod);
+
+    AttributeSet expected = new AttributeSet();
+    expected = expected.add(new Attribute("labels:name", "myname"));
+    Assert.assertTrue("Expected " + attributes + " to match " + expected, attributes.matches(expected));
```
lordofthejars: Just a comment: 
The project already imports AssertJ, why use JUnit assertions which are not readable at all?

drigz: JUnit assertions are used through the rest of the file. I'm new to the project so I couldn't say why - I just aim for consistency.

kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/KubernetesAttributesExtractor.java
drigz: Done, thanks!

PR #1170, opened 2018-08-03
Fix #1140: Provide User Utilities

### LINKED ISSUE #1140: Provide user utilites
We have lots of utilites that are mostly meant for internal consumption.
The lack of a non-internal utility for doing simple things with Kubernetes resource has led a lot of people using KubernetesHelper which was always meant to be replaced by the client.

We should create a utility class that provides at least the following:

- Sanitize strings based on the Kubernetes regex for:
  - resource names
  - resource labesl
  - resource annotations

- Null safe operations for:
  - name
  - namespace
  - labels
  - annotations

Maybe more need to be added, but for now, these would suffice.

rohanKanojia:
#1140


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
```diff
@@ -397,7 +397,7 @@ public T createOrReplace(T... items) {
       throw new IllegalArgumentException("Nothing to create.");
     }
 
-    if (Utils.isNullOrEmpty(name) && item instanceof HasMetadata) {
+    if (Utils.StringUtils.isNullOrEmpty(name) && item instanceof HasMetadata) {
```
lordofthejars: If you use the full name, you don't need to import statically. Same for the other cases in the source code.
It should be enough to do `impor io.fabric8.kubernetes.client.utils.Utils.StringUtils;`

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/osgi/ManagedKubernetesClient.java
```diff
@@ -16,54 +16,7 @@
...
-import io.fabric8.kubernetes.api.model.DoneablePersistentVolume;
-import io.fabric8.kubernetes.api.model.PersistentVolumeClaim;
-import io.fabric8.kubernetes.api.model.PersistentVolumeClaimList;
-import io.fabric8.kubernetes.api.model.DoneablePersistentVolumeClaim;
-import io.fabric8.kubernetes.api.model.ServiceAccount;
-import io.fabric8.kubernetes.api.model.ServiceAccountList;
-import io.fabric8.kubernetes.api.model.DoneableServiceAccount;
-import io.fabric8.kubernetes.api.model.ServiceList;
-import io.fabric8.kubernetes.api.model.DoneableService;
+import io.fabric8.kubernetes.api.model.*;
```
lordofthejars: Is the project allow star operation or it has been an automatic reformat? Usually using star operation is not allowed in projects.

rohanKanojia: ah, Yeah. It's wrong setting in my IntelliJ. Will fix it in next update :)

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesResourceUtil.java
```diff
@@ -0,0 +1,285 @@
...
+  /**
+   * Null safe set operation of setting Kubernetes resource's name/namespace
+   *
+   * @param entity Kubernetes
+   * @param namespace
+   * @param name
+   */
+  public static void setName(HasMetadata entity, String namespace, String name) {
+    ObjectMeta metadata = getOrCreateMetadata(entity);
+    metadata.setNamespace(namespace);
```
iocanel: I'd like to avoid bringing in utils that mutate resources. IMHO resources should be immutable and all manipulations should happen via builders. It may be too late for that now, as it's too disruptive, but at least we should avoid doing that internally.

So either have setXXX use builders and return a new object or remove them all together.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
```diff
@@ -480,7 +480,7 @@ private static boolean loadFromKubeconfig(Config config, String context, String
           config.setUsername(currentAuthInfo.getUsername());
           config.setPassword(currentAuthInfo.getPassword());
 
-          if (Utils.isNullOrEmpty(config.getOauthToken()) && currentAuthInfo.getAuthProvider() != null && !Utils.isNullOrEmpty(currentAuthInfo.getAuthProvider().getConfig().get(ACCESS_TOKEN))) {
+          if (Utils.StringUtils.isNullOrEmpty(config.getOauthToken()) && currentAuthInfo.getAuthProvider() != null && !Utils.StringUtils.isNullOrEmpty(currentAuthInfo.getAuthProvider().getConfig().get(ACCESS_TOKEN))) {
```
iocanel: I am not a big fan of this nesting.

We should either leave string utilities under utils, it have StringUtils a separate file.

rohanKanojia:
@iocanel : Hey, I've incorporated your comments in this PR. Could you please take a look again whenever you get time :-)?

rohanKanojia:
@iocanel : polite ping, could you please take a look again?

rohanKanojia:
[merge]

PR #1171, opened 2018-08-03
Fix #1139 : Make it easy to get the URL of a service.

### LINKED ISSUE #1139: Make it easy to get the URL of a service.
We need something like:

    client.services().inNamespace("ns").withName("myservice").getURL();

The implementation of the `getURL()` should be similar with what is provided by `KubernetesHelper.getServiceURL()`, but it should be implemented cleanly. 

It might make sense to have different strategies for doing service -> url. For example when using openshift-client it might make sense to take Routes into consideration. So, I would use java.util.ServiceLoader to dynamically load the providers of this functionality. It also might make sense each provider to have a toggle and a priority assigned to it.

DSL wise we just need to introduce ServiceResource that will work in the same spririt as PodResource etc.

rohanKanojia:
#1139 
+ Implemented getUrl() method for service
+ Fixed an arquillian dependency error that comes up while
  running tests in IntelliJ


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/URLFromServiceImplUtil.java
```diff
@@ -0,0 +1,194 @@
...
+import io.fabric8.kubernetes.api.model.extensions.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+public class URLFromServiceImplUtil {
```
lordofthejars: I will name it `URLFromServiceUtil`

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/URLFromServiceImplUtil.java
```diff
@@ -0,0 +1,194 @@
...
+  public static Logger logger = LoggerFactory.getLogger(URLFromServiceImplUtil.class);
+  public static final String DEFAULT_PROTO = "tcp";
+  private static final String HOST_SUFFIX = "_SERVICE_HOST";
+  private static final String PORT_SUFFIX = "_SERVICE_PORT";
+  private static final String PROTO_SUFFIX = "_TCP_PROTO";
+
+  public static String serviceToHostOrBlank(String serviceName) {
```
lordofthejars: I will call it `resolveHostFromEnvVarOrSystemProperty` and the same for port, proto, .... because when I read the code that calls this method I was not sure if it were resolving the env varor not.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/URLFromNodePortImpl.java
```diff
@@ -0,0 +1,68 @@
...
+    Integer nodePort = port.getNodePort();
+    if(nodePort != null) {
+      try {
+        NodeList nodeList = client.nodes().list();
+        if(nodeList != null && nodeList.getItems() != null) {
+          for(Node item : nodeList.getItems()) {
+            NodeStatus status = item.getStatus();
+            if(!bFound && status != null) {
+              List<NodeAddress> addresses = status.getAddresses();
+              for(NodeAddress address : addresses) {
```
lordofthejars: Can you put this second for inside a method? Just do not create two nested `for` directly. I usually try to avoid this since for a reader is easier to check what this method exactly do. Since Java does not have multi return like go, this new method could return a simple private class with required arguments.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/URLFromNodePortImpl.java
```diff
@@ -0,0 +1,68 @@
...
+                String ip = address.getAddress();
+                if (!ip.isEmpty()) {
+                  clusterIP = ip;
+                  portNumber = nodePort;
+                  bFound = true;
```
lordofthejars: Following my previous comment `bFound` is not required anymore and you can just check for null.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ServiceOperationsImpl.java
```diff
@@ -83,4 +79,37 @@ public Service waitUntilReady(long amount, TimeUnit timeUnit) throws Interrupted
...
+    ServiceLoader<ServiceToURLProvider> urlProvider = ServiceLoader.load(ServiceToURLProvider.class, Thread.currentThread().getContextClassLoader());
+    Iterator<ServiceToURLProvider> iterator = urlProvider.iterator();
+    List<ServiceToURLProvider> servicesList = new ArrayList<>();
+
+    while(iterator.hasNext()) {
+      servicesList.add(iterator.next());
+    }
+
+    // Sort all loaded implementations according to priority
+    Collections.sort(servicesList, new ServiceToUrlSortComparator());
```
lordofthejars: I will add from line 90 to 99 into a method that gets and order services, so the method `getURL` looks more compact.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/ServiceToURLProvider.java
```diff
@@ -0,0 +1,37 @@
...
+
+    ServiceToUrlImplPriority(final int newVal) {
+      value = newVal;
+    }
+
+    public int getValue() { return value; }
+  }
+
+  ServiceToUrlImplPriority getPriority();
+
```
iocanel: I'd use an int instead of an enum.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/URLFromNodePortImpl.java
rohanKanojia: ah, I wish we had already upgraded this project to Java 8, I could have used ` javafx.util.Pair ` :smile:

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/ServiceToURLProvider.java
lordofthejars: oh true well the enum can still be used but this method should return an int, so we can use our enum internally so calling `.getValue()` in implementations but it is true that the return type should be an int.

rohanKanojia:
@iocanel : Hey, I updated according to your comments. Could you please take a look again whenever you get time :-)?

rohanKanojia:
@iocanel : polite ping, could you please take a look again?

rohanKanojia:
[merge]

rohanKanojia:
retest this please

rohanKanojia:
[merge]

rohanKanojia:
retest this please

rohanKanojia:
@iocanel @oscerd : Are builds cached in test job and not in merge job? Somehow it seems to be failing on fresh builds.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/URLUtils.java
```diff
@@ -106,4 +108,42 @@ private static StringBuilder getQueryParams(URI firstPart, String[] parts, int i
...
+    return buffer.toString();
+  }
+
+  public static boolean isValidURL(String url) {
+    try {
+      URI u = new URI(url);
+    } catch (URISyntaxException exception) {
+      return false;
+    }
+    return url.contains("null") ? false : true;
```
gastaldi: url may be null at this point?

rohanKanojia: ah, yeah. Thanks for pointing out

rohanKanojia:
[merge]

PR #1174, opened 2018-08-08
Fix #1172 : Use v1beta1 for CronJob

### LINKED ISSUE #1172: Use v1beta1 for CronJob
Currently `v2alpha1` is being used & deprecated since k8s 1.8. The recommended version is `v1beta1`.

This is a followup issue to the conversation starting at https://github.com/fabric8io/kubernetes-client/issues/770#issuecomment-411082595

@rohanKanojia - would be great if this made it into next weeks release, thanks!

rohanKanojia:
#1172

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@
 
     * Fix #1173 : Send response to Callback for exec commands that have no textual feedback
 
+    * Fix #1172 : Use v1beta1 for CronJob
+
   Improvements
   
     * Added Kubernetes/Openshift examples for client.getVersion()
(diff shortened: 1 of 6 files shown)


rohanKanojia:
@iocanel @oscerd @piyush1594 @hrishin : Hm, tests are failing on older Openshift versions because they don't recognize `v1beta1` apiVersion. Shall we care about cronjob support on these versions? Any ideas if we can make setting `apiVersion` somehow configurable in the dsl itself?

piyush1594:
@rohanKanojia You need to add the entry here for CronJob also https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/BackwardsCompatibilityInterceptor.java#L109

piyush1594:
@rohanKanojia Here tests are failing may be because you forgot to change model version also.

iocanel:
We recently added an interceptor that is meant to change the spirit version
if an operation fails with a specific error code.

I think that if just need to make the interceptor aware of CronJob.

rohanKanojia:
[merge]

Issue #1090, opened 2018-06-04
Connections are leaked when java.net.ProtocolException occurs while a WebSocket establishing


sleshchenko:
Hi folks, 

I'm from Eclipse Che Team, we use fabric8-kubernetes client to work with Kubernetes/OpenShift clusters.

I discovered an issue with connections leaks when java.net.ProtocolException occurs while a WebSocket establishing. An example of it is trying to execute a command in a non-existing pod.

It can be stably reproduced on Mock KubernetesServer or real K8s/OS cluster. I have prepared a test for that in my fork https://github.com/sleshchenko/kubernetes-client/blob/connectionLeaksTest/kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/ExecTest.java

`#testConnectionLeaksAfterExecutingCommandInNonExistingPodOnMockServer` for reproducing on mock kubernetes server.

`#testConnectionLeaksAfterExecutingCommandInNonExistingPodOnRealServer` can be used for reproducing on real Kubernetes/OpenShift server. Note that `MASTER_URL` and `OAUTH_TOKEN` should be configured.

Note that `System.gc();` just accelerates appearance `connection was leaked` messages and this bug can be reproduced without invocation of garbage collector.

So, **HOW TO REPRODUCE IT**:
1. Clone fork 
2. Run test, for example, `#testConnectionLeaksAfterExecutingCommandInNonExistingPodOnMockServer`.
3. Wait ~10 minutes. For me after 5th iteration(after 5 minutes) logs appears.

```
...
10:57:11.354 [main] INFO  io.fabric8.kubernetes.client.mock.WatchTest - #######################################################
10:57:11.354 [main] INFO  io.fabric8.kubernetes.client.mock.WatchTest - ####                                               ####
10:57:11.354 [main] INFO  io.fabric8.kubernetes.client.mock.WatchTest - ####                   Iteration #  5              ####
10:57:11.354 [main] INFO  io.fabric8.kubernetes.client.mock.WatchTest - ####                                               ####
10:57:11.354 [main] INFO  io.fabric8.kubernetes.client.mock.WatchTest - #######################################################
10:57:11.354 [main] INFO  io.fabric8.kubernetes.client.mock.WatchTest - Try to execute 5 echo command in non-existing pod
Jun 04, 2018 10:57:11 AM okhttp3.mockwebserver.MockWebServer$3 processOneRequest
[...]

PR #586, opened 2016-12-13
OSFUSE-390: Eliminating leaked connections

nicolaferraro:
The okhttp response was never closed so a warning was being logged in all Kubernetes tests.

"Did you forget to close a response body?"

PR #1198, opened 2018-09-07
Attempting to fix connection leaks by closing ResponseBody objects where necessary #1090

ljnelson:
Signed-off-by: Laird Nelson <ljnelson@gmail.com>

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
@@ -27,6 +27,7 @@ import io.fabric8.kubernetes.client.utils.InputStreamPumper;
 import io.fabric8.kubernetes.client.utils.NonBlockingInputStreamPumper;
 import io.fabric8.kubernetes.client.utils.Utils;
 import okhttp3.Response;
+import okhttp3.ResponseBody;
 import okhttp3.WebSocket;
 import okhttp3.WebSocketListener;
 import okio.ByteString;
@@ -205,6 +206,12 @@ public class ExecWebSocketListener extends WebSocketListener implements ExecWatc
             if (listener != null) {
                 listener.onOpen(response);
             }
+            if (response != null) {
+                ResponseBody body = response.body();
+                if (body != null) {
+                    body.close(); // TODO: if this fails...?
+                }
+            }
         }
     }
 
@@ -213,7 +220,13 @@ public class ExecWebSocketListener extends WebSocketListener implements ExecWatc
 
       //If we already called onClosed() or onFailed() before, we need to abort.
       if (explicitlyClosed.get() || closed.get() || !failed.compareAndSet(false, true) ) {
-        //We are not going to notify the listener, sicne we've already called onClose(), so let's log a debug/warning.
+        if (response != null) {
+          ResponseBody body = response.body();
+          if (body != null) {
+            body.close(); // TODO: if this fails...?
+          }
+        }
+        //We are not going to notify the listener, since we've already called onClose(), so let's log a debug/warning.
         if (LOGGER.isDebugEnabled()) {
           LOGGER.warn("Received [" + t.getClass().getCanonicalName() + "], with message:[" + t.getMessage() + "] after ExecWebSocketListener is closed, Ignoring.");
         }
@@ -232,6 +245,12 @@ public class ExecWebSocketListener extends WebSocketListener implements ExecWatc
       } finally {
         if (listener != null) {
           listener.onFailure(t, response);
+        }
+        if (response != null) {
+          ResponseBody body = response.body();
+          if (body != null) {
+            body.close(); // TODO: if this fails...?
+          }
         }
       }
     }


rohanKanojia:
@ljnelson : Thanks, Have you tested it with test case provided in #1090 ??

rohanKanojia:
@ljnelson : I tested your change with test case provided by OP but unfortunately I still get leaks :-(

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
```diff
@@ -213,7 +220,13 @@ public void onFailure(WebSocket webSocket, Throwable t, Response response) {
 
       //If we already called onClosed() or onFailed() before, we need to abort.
```
rohanKanojia: One more thing I noticed while debugging is that `webSocketRef` is set only in case of `onOpen`. Shall it be set just after opening the WebSocket or every time in each callback? In the testcase control gets directly to onFailure (because it's a 404). Since this socket gets closed in `closeWebSocketOnce` method .....

rohanKanojia:
I have one more suspicion of leak on `NonBlockingInputStreamPumper`, since we're not telling it what to do on close. We had one fix in the past to eliminate leaks by Nicola https://github.com/fabric8io/kubernetes-client/pull/586

ljnelson:
@rohanKanojia I've added a slight refactoring of the `OperationSupport#createStatus(Response)` method.  See if this helps.

ljnelson:
Additionally, I'm not sure that #586 solves the body-close problem in the case of exceptions:

https://github.com/fabric8io/kubernetes-client/blob/f7843cff15b239e4466167f63c599fc9ca4e4146/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/InputStreamPumper.java#L86

This will not be reached if there's an exception beforehand, like, say, here:

https://github.com/fabric8io/kubernetes-client/blob/f7843cff15b239e4466167f63c599fc9ca4e4146/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/InputStreamPumper.java#L67

I'll continue updating this PR to fix leaks.

ljnelson:
Taking a slightly different approach.  A `grep` over the sourcebase reveals that there is only one usage of `BlockingInputStreamPumper` and one usage of `NonBlockingInputStreamPumper`:

One is here:

https://github.com/fabric8io/kubernetes-client/blob/59e5b9d0f318f4f78f417efaec34050ed8122b8f/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/LogWatchCallback.java#L147

…and the other is here:

https://github.com/fabric8io/kubernetes-client/blob/7ea9ce885f9fad72949db6e3f5cc8dceb8ce8766/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java#L114

That last one is irrelevant as far as closing `ResponseBody` instances, so I'm ignoring it.

So I'm going to fix `LogWatchCallback.java` in this PR; stay tuned.  The approach will be to make sure that we retain the `Future` from the submission to the executor service (we'll call it `pumperTask`), change the executor service to have exactly two threads instead of one, and submit another sibling task that calls (and blocks forever on) `pumperTask`'s `get()` method.  If we get an `ExecutionException` on this watchdog thread (e.g. an unhandled runtime exception from the `InputStreamPumper` which incidentally does not properly deal with such exceptions, but that's another issue) or if the `get()` returns normally (indicating that the `pumper` is done and/or closed), then it is safe to close the associated response body.

ljnelson:
@rohanKanojia ^^ see what you think.

rohanKanojia:
ok to test

rohanKanojia:
@ljnelson : Thanks for the update. Unfortunately, leaks don't seem to be reduced even from this refactoring. I tested it again and I could still see leak messages in logs: https://pastebin.com/HssZXJbB

ljnelson:
Where can I find `ExecTest.java`?  I don't see it in this repository but that is the class that is present in your pastebin excerpt.  I'd like to turn its logging up to `FINE` and see where it thinks the leak is coming from.

rohanKanojia:
@ljnelson : It's being provided in #1090 issue [description](https://github.com/sleshchenko/kubernetes-client/blob/connectionLeaksTest/kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/ExecTest.java). When logging set to `FINE` it points to:
```

Sep 10, 2018 10:09:40 PM okhttp3.internal.platform.Platform log
WARNING: A connection to http://localhost:60435/ was leaked. Did you forget to close a response body?
java.lang.Throwable: response.body().close()
	at okhttp3.internal.platform.Platform.getStackTraceForCloseable(Platform.java:144)
	at okhttp3.RealCall.captureCallStackTrace(RealCall.java:89)
	at okhttp3.RealCall.enqueue(RealCall.java:98)
	at okhttp3.internal.ws.RealWebSocket.connect(RealWebSocket.java:183)
	at okhttp3.OkHttpClient.newWebSocket(OkHttpClient.java:436)
	at io.fabric8.kubernetes.client.dsl.internal.PodOperationsImpl.exec(PodOperationsImpl.java:267)
	at io.fabric8.kubernetes.client.dsl.internal.PodOperationsImpl.exec(PodOperationsImpl.java:61)
	at io.fabric8.kubernetes.client.mock.ExecTest.doExec(ExecTest.java:99)
	at io.fabric8.kubernetes.client.mock.ExecTest.doIteration(ExecTest.java:84)
	at io.fabric8.kubernetes.client.mock.ExecTest.testConnectionLeaks(ExecTest.java:73)
	at io.fabric8.kubernetes.client.mock.ExecTest.testMockServer(ExecTest.java:50)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:48)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:369)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeWithRerun(JUnit4Provider.java:275)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:239)
	at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:160)
	at org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader(ForkedBooter.java:373)
	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:334)
	at org.apache.maven.surefire.booter.ForkedBooter.execute(ForkedBooter.java:119)
	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:407)

```

ljnelson:
I'm beginning to think that the problem may lie in the fact that when the websocket is created the [associated `StreamAllocation` is explicitly divorced from connection pooling](https://github.com/square/okhttp/blob/5b37cda9e00626f43acf354df145fd452c3031f1/okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java#L204).  I don't understand this area of the OkHttp code enough to say why, exactly, but I don't see any place where this "real" connection is explicitly deallocated.  Maybe it is.  At any rate, we can't really do anything about this, I don't think, although there is a [way to be notified when a connection is released](https://square.github.io/okhttp/3.x/okhttp/okhttp3/EventListener.html#connectionReleased-okhttp3.Call-okhttp3.Connection-); maybe we should be doing something there.

I say all this because with my changes there are no further areas in the kubernetes-client codebase where `ResponseBody` is left open.  So closing response bodies is not sufficient to prevent leak messages.

Another possibility is the following: I don't think [the `clonedClient` in watches is ever closed](https://github.com/fabric8io/kubernetes-client/search?q=clonedclient&unscoped_q=clonedclient).  I don't think it _has_ to be closed, but maybe that would forcibly clean the associated connection pool before those log messages get generated?

ljnelson:
@rohanKanojia Can you tell me why this reference is retained:

https://github.com/fabric8io/kubernetes-client/blob/79132f3348d24acb677de3ee20a96da8d1c67c84/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java#L72

It is set here:

https://github.com/fabric8io/kubernetes-client/blob/79132f3348d24acb677de3ee20a96da8d1c67c84/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java#L153

It is not referenced anywhere else in the code.

There is, of course, an `AtomicReference` that holds the current websocket:

https://github.com/fabric8io/kubernetes-client/blob/79132f3348d24acb677de3ee20a96da8d1c67c84/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java#L62

It is set here:

https://github.com/fabric8io/kubernetes-client/blob/79132f3348d24acb677de3ee20a96da8d1c67c84/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java#L160

Is there a reason I'm not seeing for the duplication?

rohanKanojia:
@ljnelson : Sorry for the delayed response. Actually, we are in opposite timezones ;-) . Yeah, this duplication could be due to lots of people coming and modifying WatchConnectionManager accordingly over time. Not sure whether there is some history associated with it or not.

ping @iocanel

ljnelson:
Related: https://github.com/square/okhttp/issues/4258

rohanKanojia:
@ljnelson : okhttp 3.12 has been released. Would really appreciate if you could find time to revisit this.

iocanel:
Test

Issue #1201, opened 2018-09-08
Various improper InterruptedException catch blocks


ljnelson:
There are several cases in the codebase where `InterruptedException` is caught and then wrapped inside a `KubernetesClientException`.  Here is one arbitrary example:

https://github.com/fabric8io/kubernetes-client/blob/f7843cff15b239e4466167f63c599fc9ca4e4146/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java#L220-L222

As [Brian Goetz writes](https://www.ibm.com/developerworks/library/j-jtp05236/index.html#2.1), the proper way to deal with a caught `InterruptedException` that you're not going to rethrow is to ensure that the thread's interrupted status is properly set.

I would propose that catch blocks of this form contain a `Thread.currentThread().interrupt();` call before throwing the `KubernetesClientException`.

I'll try to submit a PR that addresses this.

ljnelson:
See PR #1202.

rohanKanojia:
Closed via #1202

PR #1202, opened 2018-09-08
Fixes problems with swallowed InterruptedExceptions #1201

### LINKED ISSUE #1199: Improperly handled InterruptedException in NonBlockingInputStreamPumper?
I am not sure, but this:

https://github.com/fabric8io/kubernetes-client/blob/f7843cff15b239e4466167f63c599fc9ca4e4146/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/NonBlockingInputStreamPumper.java#L68-L70

...effectively does nothing.  The presence of an `InterruptedException` means the interrupted status has been cleared, so `Thread#interrupted()` will redundantly clear the interrupted status as well as return false, and its return value is ignored here.

Is it possible that the author actually meant something quite different here, namely `Thread#interrupt()`?

ljnelson:
Signed-off-by: Laird Nelson <ljnelson@gmail.com>

This PR should fix issue #1201.  It deliberately does not fix #1199 as it is not immediately clear to me what the intent of that particular catch block was.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ### CHANGELOG
 
-#### 4.0.5
+#### 4.0.6
   Bugs
   
     * Fix #1214 : Watch resource versions not correctly tracked resulting in 410 errors on reconnect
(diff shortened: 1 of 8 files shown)


rohanKanojia:
ok to test

ljnelson:
Done.

rohanKanojia:
@oscerd :Shall we cherry-pick this to 3.x and 3.1.x also?

rohanKanojia:
[merge]

PR #1183, opened 2018-08-17
Move client CD job to CircleCI

kishansagathiya:
Fixes #1217

PR #1207, opened 2018-09-19
Move CD to CICO

kishansagathiya:
This PR is to test and demonstrate that https://github.com/fabric8io/kubernetes-client/pull/1183 works.
@rohanKanojia @hrishin


kishansagathiya:
I think it will be better if I use a new branch since I will be adding many empty commits

PR #1209, opened 2018-09-19
IpAddressMatcher should not throw an exception on an unresolved hostname

rohanKanojia:
It should not throw exception, but the matcher should just return false.
Related to https://github.com/fabric8io/fabric8-maven-plugin/issues/1389


oscerd:
@rhuss can you please approve?

PR #1212, opened 2018-09-20
Fixes flaky integration tests.

### LINKED ISSUE #1182: Explicit sleep on the thread can make tests flaky
I see there are tests which wait until the resource is destroyed by explicitly waiting 30s. This has two problems at least:

  * it doesn't guarantee that condition is met 
  * if it's met faster we waste time sitting idle and the whole build is slower for no reason

https://github.com/fabric8io/kubernetes-client/blob/7ea9ce885f9fad72949db6e3f5cc8dceb8ce8766/kubernetes-itests/src/test/java/io/fabric8/openshift/RouteIT.java#L106

If there is a way in the client to check if the resource does not exist we could think about using polling instead? For example by using this little library https://github.com/awaitility/awaitility/

dev-gaur:
Fixes #1182.


kubernetes-itests/src/test/java/io/fabric8/kubernetes/ConfigMapIT.java
```diff
@@ -101,14 +114,26 @@ public void update() {
...
-    Thread.sleep(30000);
+    try {
+      if (client.configMaps().inNamespace(currentNamespace).list().getItems().size()!= 0) {
+        client.configMaps().inNamespace(currentNamespace).delete();
+      }
+
+      Delete<ConfigMap> configMapDelete = new Delete<ConfigMap>(ConfigMap.class, client, "configmap1", currentNamespace);
+      await().atMost(30, TimeUnit.SECONDS).until(configMapDelete);
+    } catch (NullPointerException exception) {
+      cleanup();
```
bartoszmajsak: Isn't it a potentially endless recursion here? Why are we handling `NPE` here?

kubernetes-itests/src/test/java/io/fabric8/kubernetes/NetworkPolicyIT.java
```diff
@@ -158,16 +165,25 @@ public void update(){
...
     assertEquals(0,networkPolicyList.getItems().size());
   }
 
   @After
   public void cleanup() throws InterruptedException {
-    client.network().networkPolicies().delete();
-    // wait for graceful deletion
-    Thread.sleep(6 * 1000);
+    try {
+      if (client.network().networkPolicies().list().getItems().size()!= 0) {
```
bartoszmajsak: `client.network().networkPolicies()` can be a variable to start with. Then perhaps `!var.list().isEmpty()` might improve readability a bit?

kubernetes-itests/src/test/java/io/fabric8/kubernetes/NetworkPolicyIT.java
```diff
@@ -158,16 +165,25 @@ public void update(){
...
+        client.network().networkPolicies().delete();
+      }
+      Delete<NetworkPolicy> networkPolicyDelete = new Delete<NetworkPolicy>(NetworkPolicy.class, client, "networkpolicy", currentNamespace);
+      await().atMost(30, TimeUnit.SECONDS).until(networkPolicyDelete);
+    }catch (NullPointerException exception) {
+      cleanup();
```
bartoszmajsak: Formatting + the same question about an endless call as in the other test.

kubernetes-itests/src/test/java/io/fabric8/kubernetes/ServiceAccountIT.java
```diff
@@ -85,23 +91,35 @@ public void list() {
...
-    // Wait for resources to get destroyed
-    Thread.sleep(30000);
+    try {
+      client.serviceAccounts().inNamespace(currentNamespace).delete();
+      // Wait for resources to get destroyed
+      Delete<ServiceAccount> serviceAccount1Delete = new Delete<ServiceAccount>(ServiceAccount.class, client, "serviceaccount1", currentNamespace);
+      Delete<ServiceAccount> serviceAccount2Delete = new Delete<ServiceAccount>(ServiceAccount.class, client, "serviceaccount2", currentNamespace);
+
+      await().atMost(30, TimeUnit.SECONDS).until(serviceAccount1Delete);
+      await().atMost(30, TimeUnit.SECONDS).until(serviceAccount2Delete);
```
bartoszmajsak: Can this be combined to one `&&` condition ?

kubernetes-itests/src/test/java/io/fabric8/commons/Delete.java
```diff
@@ -0,0 +1,57 @@
...
+        return ((OpenShiftClient)this.client).buildConfigs().inNamespace(this.namespace).list().getItems().size() == 0;
+      case "ConfigMap":
+        return this.client.configMaps().inNamespace(this.namespace).list().getItems().size() == 0;
+      case "StatefulSet":
+        return this.client.apps().statefulSets().inNamespace(this.namespace).list().getItems().size() == 0;
+      case "Deployment":
+      return this.client.apps().deployments().inNamespace(this.namespace).list().getItems().size() == 0;
+      case "Service":
+        return this.client.services().inNamespace(this.namespace).list().getItems().size() == 0;
+      case "ServiceAccount":
```
rohanKanojia: Why is service account name hardcoded here?

kubernetes-itests/src/test/java/io/fabric8/commons/Delete.java
```diff
@@ -0,0 +1,57 @@
...
+
+  public Delete(Class<T> typeParameterClass, KubernetesClient client, String name, String namespace) {
+    this.typeParameterClass = typeParameterClass;
+    this.client = client;
+    this.name = name;
+    this.namespace = namespace;
+  }
+
+  @Override
+  public Boolean call() throws Exception {
```
rohanKanojia: Are you sure that Exception is thrown from this method? if not, please remove this.

kubernetes-itests/src/test/java/io/fabric8/commons/Ready.java
```diff
@@ -0,0 +1,58 @@
...
+  public Boolean call() throws Exception {
+    switch (this.typeParameterClass.getSimpleName()) {
+      case "Route":
+        return ((OpenShiftClient)this.client).routes().inNamespace(this.namespace).withName(this.name).get() != null;
+      case "Template":
+        return ((OpenShiftClient)this.client).templates().inNamespace(this.namespace).withName(this.name).get() != null;
+      case "ImageStream":
+        return ((OpenShiftClient)this.client).imageStreams().inNamespace(this.namespace).withName(this.name).get()!= null;
+      case "DeploymentConfig":
+        System.out.println("In deployment config ready");
```
rohanKanojia: You might wanna remove this ;-)

kubernetes-itests/src/test/java/io/fabric8/kubernetes/ConfigMapIT.java
```diff
@@ -16,7 +16,10 @@
 
 package io.fabric8.kubernetes;
 
+import io.fabric8.commons.Delete;
+import io.fabric8.commons.Ready;
 import io.fabric8.kubernetes.api.model.*;
+import io.fabric8.kubernetes.client.Config;
```
rohanKanojia: Do we need this import?

kubernetes-itests/src/test/java/io/fabric8/kubernetes/ConfigMapIT.java
```diff
@@ -89,10 +95,17 @@ public void list() {
 
   @Test
   public void update() {
+    Ready<ConfigMap> configMap1Ready = new Ready<ConfigMap>(ConfigMap.class, client, "configmap1", currentNamespace);
```
rohanKanojia: Explicit type argument can be replaced with <>

kubernetes-itests/src/test/java/io/fabric8/commons/Ready.java
dev-gaur: oops

kubernetes-itests/src/test/java/io/fabric8/commons/Delete.java
dev-gaur: forgot to omit this

kubernetes-itests/src/test/java/io/fabric8/kubernetes/ConfigMapIT.java
dev-gaur: noted. thanks!

kubernetes-itests/src/test/java/io/fabric8/commons/Delete.java
dev-gaur: oops

kubernetes-itests/src/test/java/io/fabric8/kubernetes/ConfigMapIT.java
dev-gaur: forgot to omit

kubernetes-itests/src/test/java/io/fabric8/kubernetes/ConfigMapIT.java
dev-gaur: I followed this up from [here](https://github.com/fabric8io/kubernetes-client/pull/1121/files#diff-e892e898797ea5223e863136b7348d7cR134) .

If the 'entity/resource' is not responding yet, we can catch the exception and try cleaning up again. I think, practically, the recursive loop here being endless is unlikely. Therefore, I let it there as good measure.

@rohanKanojia what do you think?

kubernetes-itests/src/test/java/io/fabric8/kubernetes/ServiceAccountIT.java
dev-gaur: As in this ?
```await().atMost(30, TimeUnit.SECONDS).until(serviceAccount1Delete && serviceAccount2Delete);```

This doesn't work. data types of serviceAccountDelete instances aren't compatible for this boolean specific operation.

kubernetes-itests/src/test/java/io/fabric8/kubernetes/ConfigMapIT.java
rohanKanojia: Even if ending up in a recursive loop is unlikely; I believe we should limit this to some finite number of retry attempts, rather than having no control over the number of retries encountered

kubernetes-itests/src/test/java/io/fabric8/kubernetes/NetworkPolicyIT.java
dev-gaur: Hey, 

Attempting this would look like [this](https://gist.github.com/rohanKanojia/f49dd6ce84b9317fb21f8653e92e8c81#file-networkpolicyit-java-L47) and i think this line looks really long and highly unreadable. I would be happy to take any suggestions from you on workarounds.

rohanKanojia:
Please ignore this centos CI build, I'm currently working on fixing this.

rohanKanojia:
retest this please

rohanKanojia:
[merge]

PR #1215, opened 2018-09-25
Fix tracking of watched resource versions

### LINKED ISSUE #1214: Watch resource versions not correctly tracked resulting in 410 errors on reconnect
The WatchConnectionManager incorrectly compares new and current resource versions, leading to it losing track of the resource versions. If it's disconnected and attempt to reconnect this will often lead to a 410 due to a too old resource version being sent on the new watch request.

This happens because resource versions are currently only being updated in the WatchConnectionManager if `currentResourceVersion.compareTo(newResourceVersion) < 0`. Since these variables are Strings this comparison results in in e.g. an update from "9" to "10" being discarded since "9".compareTo("10") > 0.

According to https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#concurrency-control-and-consistency the resourceVersion should be treated as an opaque value, so this comparison could probably be considered incorrect behaviour even if it worked as intended. There's no need to do the check (or the compareAndSet) since messages are guaranteed to be processed in order.

JonathanO:
Fixes #1214

resourceVersions are Strings, so comparing them with compareTo leads to
unexpected results ("9".compareTo("10") is > 0.) Also since the versions
are supposed to be opaque we should not be trying to read any meaning
or ordering into them.

Since messages are processed in order we should just always store the
last processed resourceVersion.

I'm not sure this even needs to use an Atomic* type, but it seems safer
to just leave that in place.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,11 +16,6 @@
 
 version: 2
 jobs:
-  TEST:
-    machine: true
-    steps:
-      - checkout
-      - run: mvn clean install
   OPENSHIFT_3.6.0:
     machine: true
     steps:
@@ -221,7 +216,6 @@ workflows:
   version: 2
   build_and_test:
     jobs:
-      - TEST
       - OPENSHIFT_3.6.0
       - OPENSHIFT_3.6.1
       - OPENSHIFT_3.7.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 #### 4.0.5
   Bugs
+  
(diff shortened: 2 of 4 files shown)


rohanKanojia:
ok to test

JonathanO:
The tests are failing with a "java.lang.RuntimeException: Container never came up" in FeatureInstallationTest, is this likely to be a result of my change or is a test being flaky?

rohanKanojia:
@JonathanO : ah, This is due to overloaded CI server. Let me restart them.

rohanKanojia:
[merge]

ljnelson:
@rohanKanojia See #1221.  Hope I pushed the right buttons and chanted the right `git` spells!

Issue #1200, opened 2018-09-08
Does this client library support credential plugin (like aws-iam-authenticator) in kube config file?


lilida:
I did a quick look at the code and tested in my environment (a bit older version though) It looks like aws-iam-authenticator is not supported yet. Is this true?

jglick:
Confirmed. After configuring `~/.kube/config` [like this](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/README.md#4-set-up-kubectl-to-use-authentication-tokens-provided-by-aws-iam-authenticator-for-kubernetes), which follows [this spec](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#configuration), e.g. `kubectl get pods` on the command line works but [this code](https://github.com/fabric8io/kubernetes-client/blob/cee8550053e671bb0b3489e37196924f4c949a3f/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/HttpClientUtils.java#L102-L108) fails to add a `Authorization: Bearer …` header from the `.status.token` output of my `aws-iam-authenticator` command, resulting in an error such as

```
io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://….sk1.us-east-1.eks.amazonaws.com/api/v1/namespaces/…/pods. Message: Forbidden! User aws doesn't have permission. pods is forbidden: User "system:anonymous" cannot list pods in the namespace "…".
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.requestFailure(OperationSupport.java:476)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.assertResponseCode(OperationSupport.java:413)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:381)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:344)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:328)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.list(BaseOperation.java:598)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.list(BaseOperation.java:63)
```

rohanKanojia:
@jglick : Keeping it open till someone implements the rest of the parts of spec. Could you please specify what else parts need to be implemented for it to get closed?

[...]

PR #1224, opened 2018-10-05
#1200: Support client authenticator commands like aws-iam-authenticator

jglick:
Rudimentary implementation of #1200. Enough to get my authentication working at least; does not implement some parts of the documented spec.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
     * Chore #1168 : Upgrade to Java 8
     * Fix #1179 : Impersonate-Extra keys may be specified multiple times
     * Fix #1182 : Fixes flaky integration tests.
+    * Fix #1200 : Support client authenticator commands like `aws-iam-authenticator`.
 
   Dependency Upgrade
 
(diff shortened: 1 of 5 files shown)


rohanKanojia:
ok to test

jglick:
This probably deserves an entry in `ConfigTest`. Would the tests always be run on Linux?

rohanKanojia:
@jglick : As per current setup, tests are running on ubuntu based environments.

rohanKanojia:
ok to test

rhuss:
Any idea why the test on ci.centos.org fails with 

```
+ /usr/bin/timeout 20m ssh -t -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -l root 172.19.2.86 -t 'cd payload && /bin/bash cico_build.sh'
Warning: Permanently added '172.19.2.86' (ECDSA) to the list of known hosts.
/bin/bash: cico_build.sh: No such file or directory
```

rohanKanojia:
@rhuss : I'm working on moving cd pipiline on centos ci, that branch hasn't been merged into master yet, so it's failing because it's not configured. I enabled webhook because I needed it for testing cd pipeline.

rhuss:
@rohanKanojia thanks for the explanation, makes sense.

rohanKanojia:
[merge]

roldevg:
@jglick Thanks for the feature. I tried now with version 4.1.2 and it works as expected.

PR #1231, opened 2018-10-12
Fix #1226: Extend and move integration tests

### LINKED ISSUE #1226: Extend and move integrations tests
Currently, we have integration checks for OpenShift 3.6.0, 3.6.1, 3.7.0 and 3.9.0 running on CircleCI.

Since these tests occupy most of our CircleCI queue (they also take a bit to run) and also the latest versions (3.10 and 3.11alpha are missing), I recommend:

* Moving out these test to the CentOS CI
* Disable tests on 3.6.0 and 3.6.1 (I don't think that they are used significantly) and increase the minimum requirement on OpenShift for kubernetes-client
* Add tests for 3.10 and 3.11alpha

rohanKanojia:
Fixes #1226

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,123 +16,6 @@
 
 version: 2
 jobs:
-  OPENSHIFT_3.6.0:
-    machine: true
-    steps:
-      - checkout
-      - run:
-          command: |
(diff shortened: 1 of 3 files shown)


rohanKanojia:
retest this please

PR #1232, opened 2018-10-12
Make PatchUtils thread safe

Vlatombe:
The previous initialization was not thread safe. I have seen occurrences
of patches generated with SerializationFeature.WRITE_EMPTY_JSON_ARRAYS
option.

I have been able to reproduce a race condition in debug mode only, I
haven't been able to build a multithread unit test to exhibit the
behaviour.


rohanKanojia:
ok to test

oscerd:
[merge]

PR #1239, opened 2018-10-22
Close the initial request before discarding it.

msilpala:
Plugs a connection leak. 

This is the same symptom as #1198 addresses in several other places. In this case, once you get to line 121, you _know_ you're not going to need to use `response` or `request` for anything other than getting the response status code. The 'response` needs to be closed in order not to leak a connection. Getting its status code still works after closure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 #### 4.1.1
   Bugs
+  
+    * Fix #1239: Fix one case of OkHttp connection leaks
 
   Improvements
 
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/BackwardsCompatibilityInterceptor.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/BackwardsCompatibilityInterceptor.java
@@ -123,6 +123,7 @@ public class BackwardsCompatibilityInterceptor implements Interceptor {
       ResourceKey key = getKey(matcher);
       ResourceKey target = responseCodeToTransformations.get(response.code()).get(key);
       if (target != null) {
+        response.close(); // At this point, we know we won't reuse or return the response; so close it to avoid a connection leak.
         String newUrl = new StringBuilder(url)
             .replace(matcher.start(API_VERSION), matcher.end(API_VERSION), target.version) // Order matters: We need to substitute right to left, so that former substitution don't affect the indexes of later.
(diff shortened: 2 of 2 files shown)


msilpala:
@dbyron0 Updated for clarity.

msilpala:
FWIW, I have confirmed via debugger-driven patch that in our particular use case (running https://github.com/spinnaker/clouddriver's v1 kubernetes provider), this code change does prevent the leaking-connection issue.

rohanKanojia:
ok to test

rohanKanojia:
ok to test

msilpala:
I'm now working on getting the whole thing to build locally, so I can figure out the issue.

msilpala:
Looks like this change will require potentially a bunch of mock-training changes. Digging and learning...

msilpala:
That's reasonable. I looked a little to see whether these are legit breaks, or whether the tests are just fragile/need setup adjustments. I suspect it's fragile tests. Other things pulled me away from digging deeper right now, and I'd greatly appreciate if someone who's more familiar with the code could take a _quick_ look to see whether I'm right.

Else I'll get to it when I get to it. Possibly this week; possibly next week. In the meantime, we're rolling forward by including this change as a patch in our production deployments of [clouddriver](https://github/spinnaker/clouddriver).

Thanks, all.

msilpala:
Update: we did run into issues testing with this patch. Looks like those failing tests are failing legitimately. Looking at it later today.

msilpala:
This update runs all the tests successfully on my machine. Fingers crossed that it's the same in the Circle.

msilpala:
Well, at least that's a lot closer. Guess I need to install OpenShift 3.9 locally to figure out what's still not working.

msilpala:
@rohanKanojia Looks like this might be good to go now. In the week since my last comment, the one failing OpenShift test has gone from red to green. I can't explain that; I didn't change anything. But maybe something else changed on master which corrected that particular case?

In any event, I hope this can be merged now. Any other changes to request?

rohanKanojia:
I just restarted it ;-), we're running openshift clusters in free tier of circleci that's why server gets overloaded sometimes.

rohanKanojia:
@msilpala : Could you please add a line to CHANGELOG.md also about this change? It looks good to me :+1:

msilpala:
@rohanKanojia CHANGELOG updated.

rohanKanojia:
retest this please

rohanKanojia:
[merge]

Issue #1241, opened 2018-10-23
OpenShiftOAuthInterceptor gets new token each time


Exosik:
After this change:
https://github.com/fabric8io/kubernetes-client/commit/c1a6b10c799d5bef6198f37d331fd344950855a3#diff-1f50b87d0d0cacc403fa370cc659ace2

token is not persisted and each request generates new one.

Possible solution:
move/duplicate "synchronized" section (lines 80-87) to line 67

rohanKanojia:
@Exosik :Would really appreciate if you could raise a PR ;-) .

PR #1242, opened 2018-10-25
ISSUE-1241: do not recreate token each time

Exosik:
Token should be persisted each time it's acquired so it will be not acquired each time it's needed

issue #1241


rohanKanojia:
ok to test

oscerd:
LGTM

oscerd:
[merge]

Issue #1260, opened 2018-11-09
Unable to build repo


traviswinter:
Tried on both Windows 10 and Ubuntu 18.0.4.  Getting issue running mvn install on the model project.  Is there an official repo setup guide?  Or are there some common known quirks contributors are using to get past this?

Running on OpenJDK 8 (tried 7 as well but there were complaints about improper version)

[INFO] Reactor Summary:
[INFO]
[INFO] Fabric8 :: Kubernetes :: Project 4.1-SNAPSHOT ...... SUCCESS [  3.933 s]
[INFO] Fabric8 :: Kubernetes Model Parent ................. SUCCESS [  0.404 s]
[INFO] Fabric8 :: Kubernetes Model Annotator .............. SUCCESS [  0.622 s]
[INFO] Fabric8 :: Kubernetes Model ........................ FAILURE [ 41.456 s]
[INFO] Fabric8 :: Kubernetes :: Java Client ............... SKIPPED
[INFO] Fabric8 :: Kubernetes :: Server Mock ............... SKIPPED
[INFO] Fabric8 :: Openshift :: Java Client ................ SKIPPED
[INFO] Fabric8 :: Openshift :: Server Mock ................ SKIPPED
[INFO] Fabric8 :: Kubernetes :: Client Examples ........... SKIPPED
[INFO] Fabric8 :: Kubernetes :: Platforms ................. SKIPPED
[INFO] Fabric8 :: Kubernetes :: Platforms :: Karaf ........ SKIPPED
[INFO] Fabric8 :: Kubernetes :: Platforms :: Karaf :: Features SKIPPED
[INFO] Fabric8 :: Kubernetes :: Platforms :: Karaf :: Tests SKIPPED
[INFO] Fabric8 :: Kubernetes :: Tests ..................... SKIPPED
[INFO] Fabric8 :: Kubernetes and Openshift :: UberJar 4.1-SNAPSHOT SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 47.616 s
[INFO] Finished at: 2018-11-09T11:10:46-05:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.5.1:compile (default-compile) on project kubernetes-model: Compilation failure: Compilation failure:
[...]

PR #1262, opened 2018-11-09
Chore: Upgrade to surefire 3.0.0-M1

rohanKanojia:
Contains fix : https://issues.apache.org/jira/browse/SUREFIRE-1588
which was causing problems like #1260


rohanKanojia:
[merge]

PR #1268, opened 2018-11-16
Fix #1246: Separate stderr and stdout streams in ExecWebSocketListener

### LINKED ISSUE #1246: ExecWebSocketListener always combines stdout and stderr
Combining stdout and stderr into a single stream makes it impossible for clients to discern stdout from stderr. It's critical for an exec implementation to treat these as separate streams. In unix programming, stderr is meant for logging (not just errors) and stdout is meant for output.

For example if you were to send the command:
```
tar -cvf - /some/directory
```
the contents of the tar file generated from the directory is written to the stdout stream and the logging of all of the files in the tar file is written to the stderr stream, so combining these streams results in a corrupt tar file and no log messages.

This was broken in https://github.com/fabric8io/kubernetes-client/pull/1176 which was intended to fix @emilojkovic's issue https://github.com/fabric8io/kubernetes-client/issues/1173.

In 1173, the problem was that the [ExecPipesExample](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/ExecPipesExample.java) was only pumping the stdout stream and not doing anything with the stderr stream. To solve this, I believe @emilojkovic could have simply made their code add another InputStreamPumper for stderr. If that wasn't working, then there is another bug we need to track down.

For convenience in cases like @emilojkovic's, it may make sense to add a `combineOutput` option, however.

My teammate @daniel-chu and I are happy to make a PR to fix this, but we're curious how we should proceed. @emilojkovic would your original issue have been solved if you had added an InputStreamPumper for the error stream?

cc: @rohanKanojia who reviewed the original PR

daniel-chu:
Fixes #1246 - brief summary of issue: the exec listener was changed to combine stderr with stdout

This PR just reverts those changes so that stderr is properly written to the stderr stream and kept separate from stdout. 

If adding a simple way for people to use the exec api through the client would be useful, @PaulFurtado described what we've written to do this in [this comment](https://github.com/fabric8io/kubernetes-client/issues/1246#issuecomment-436899940). We'd be happy to take a look at adding something similar to the client.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
     * Fix #1266: Not setting optional Impersonate-Group results in NPE in ImpersonatorInterceptor
     * Fix #1238: Renamed files with invalid Windows characters
 	  * Fix #1260: Added Windows support in ConfigTest.honorClientAuthenticatorCommands
+    * Fix #1246: Fix bug causing stderr to be written to the stdout stream in ExecWebSocketListener
 
   Improvements
     * Fix #1226 : Extend and move integrations tests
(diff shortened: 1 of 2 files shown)


rohanKanojia:
ok to test

oscerd:
Can you please rebase?

Issue #1228, opened 2018-10-11
OkHttpClient reports connection leak in OperationSupport.handleResponse


uce:
The following happens with version `4.0.5` using `KubernetesClient`.

`OkHttpClient` reports the following connection leak on log level `FINE`. 

```
WARNING: A connection to https://[...] was leaked. Did you forget to close a response body?
java.lang.Throwable: response.body().close()
  at okhttp3.internal.platform.Platform.getStackTraceForCloseable(Platform.java:144)
  at okhttp3.RealCall.captureCallStackTrace(RealCall.java:89)
  at okhttp3.RealCall.execute(RealCall.java:73)
  at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:379)
  at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:344)
  at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleGet(OperationSupport.java:313)
  at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleGet(OperationSupport.java:296)
  at io.fabric8.kubernetes.client.dsl.base.BaseOperation.handleGet(BaseOperation.java:794)
  at io.fabric8.kubernetes.client.dsl.base.BaseOperation.getMandatory(BaseOperation.java:210)
  at io.fabric8.kubernetes.client.dsl.base.HasMetadataOperation.edit(HasMetadataOperation.java:68)
  at io.fabric8.kubernetes.client.dsl.internal.DeploymentOperationsImpl.edit(DeploymentOperationsImpl.java:78)
  at io.fabric8.kubernetes.client.dsl.internal.DeploymentOperationsImpl.edit(DeploymentOperationsImpl.java:44)
  at io.fabric8.kubernetes.client.dsl.internal.DeploymentOperationsImpl$DeploymentReaper.reap(DeploymentOperationsImpl.java:163)
  at io.fabric8.kubernetes.client.dsl.base.BaseOperation.delete(BaseOperation.java:614)
  at io.fabric8.kubernetes.client.dsl.base.BaseOperation.delete(BaseOperation.java:63)
  at io.fabric8.kubernetes.client.dsl.base.BaseOperation.delete(BaseOperation.java:655)
  at io.fabric8.kubernetes.client.dsl.base.BaseOperation.delete(BaseOperation.java:643)
[...]

PR #1275, opened 2018-11-22
Issue 1228: InputStream not closing correctly during unmarshal action

traviswinter:
Looks like the InputStream not being closed was the source of the leakage for me.  Referencing https://github.com/fabric8io/kubernetes-client/issues/1228 since that was the base of discussion around the issue.

Mixed with the okhttp WebSocket fix, this should get most of the leakages.


rohanKanojia:
ok to test

rohanKanojia:
[merge]

PR #1286, opened 2018-12-08
PodPreset Support

rohanKanojia:
Added support for PodPreset in client

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
   Bugs
 
     * Fix #1271: Issue deploying ReplicaSet to extensions/v1beta1
+    
     * Fix #1152: Renamed Kubernetes RBAC resources to use non-prefixed names and renamed Openshift RBAC resources to prefix with Openshift
 
   Improvements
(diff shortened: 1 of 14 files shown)


rohanKanojia:
@oscerd : Somehow I was wasn't able to set up `PodPreset` on my minikube. Would it be ok to let this in without being actually tested on real cluster? Wdyt?

oscerd:
What minikube version do you have? I'm not sure if it can be a problem related to this.

rohanKanojia:
I'm using v0.30.0

oscerd:
Did you try this?
https://github.com/kubernetes/minikube/issues/2209
In particular this?
https://github.com/kubernetes/minikube/issues/2209#issuecomment-380089059
and here too:
https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#is-there-a-recommended-set-of-admission-controllers-to-use

oscerd:
What's the status on this?

rohanKanojia:
@oscerd : ah, sorry. Almost forgot this, Will test it and provide an update in 1-2 working days

rohanKanojia:
@oscerd : Looks like I was having issues due to slow internet connection ;-) . Just tested it and it seems to work.

oscerd:
[merge]

PR #1306, opened 2018-12-28
Support KUBECONFIG env var with multiple paths

carlossg:
Pick the first kubeconfig file when env var has multiple ones

Example

```
/home/jdoe/.kube/config:/home/jdoe/.kube/config-something-else.yml
```

See https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#the-kubeconfig-environment-variable

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
   Improvements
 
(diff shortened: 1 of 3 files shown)


rohanKanojia:
[merge]

PR #1322, opened 2019-01-22
Codegen

iocanel:
This pull request, uses code generation based on velocity transformation, in order to generate boilerplate code. So, it removes all handlers and many of the operation implementations (this can be further improved).


iocanel:
[merge]

kubernetes-client/src/test/resources/test-kubeconfig-exec
```diff
@@ -1,4 +1,4 @@
-apiVersion: v1
+apiGroupVersion: v1
```
jglick: #1347: this does not look right. An actual `~/.kube/config` uses `apiVersion`, not `apiGroupVersion`. So the test is being updated to match the new broken behavior?

iocanel: Right, this is wrong.

PR #1339, opened 2019-02-01
oAuth token rotation support

peterjurkovic:
Added support for an oAuth token roration. 

### Usage
A user can implement `OAuthTokenProvider` interface, method getToken.  
In `HttpClientUtils` will be called `config.getOauthToken()` which will delegate a call to the provider. 

```java
class RotatingOAuthTokenProvider implements OAuthTokenProvider {
    
    public String getToken() {
        // ...
    }
}
```

The implementation can be then supplied to a `ConfigBuilder` as bellow.

```java
new ConfigBuilder().withOauthTokenProvider(rotatingOAuthTokenProvider)
                   .build()
```

If the `OAuthTokenProvider` is supplied, takes precedence over token autoconfiguration, or static token.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/OAuthTokenProvider.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/OAuthTokenProvider.java
@@ -1,0 +1,13 @@
+package io.fabric8.kubernetes.client;
+
+public interface OAuthTokenProvider {
+
+  /**
+   * Returns a Bearer token used for authorization between a client
+   * and a kubernetes cluster. The token will be injected into an Authorization header.
+   *
+   * @return oauth token
+   */
+  String getToken();
+
+}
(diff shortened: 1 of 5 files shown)


rohanKanojia:
ok to test

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/OAuthTokenProvider.java
```diff
@@ -0,0 +1,13 @@
+package io.fabric8.kubernetes.client;
```
rohanKanojia: missed license check here, you can add it via command: `mvn -N license:format`

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ### CHANGELOG
 
+#### 4.2.0
+
```
rohanKanojia: You need to add this in `4.1.2` yet to be released version, Sorry for the confusion. Ideally, it should be 4.1-SNAPSHOT

peterjurkovic: Ah, sorry about this, I should have asked. Updating..

peterjurkovic:
Could someone re-run the checks. I see
```
Failure to transfer org.apache.sshd:sshd-core:jar:1.7.0 from https://repo.maven.apache.org
```

oscerd:
I think it's fine if we merge now

oscerd:
[merge]

PR #1343, opened 2019-02-02
Feature 213: Added require( ) method to Resource object class.

### LINKED ISSUE #213: provide an equilvalent to .get() which fails if the resource is missing with a meaningful error message describing the namespace/name/labels being filtered

dev-gaur:
Fix #213

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Requirable.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Requirable.java
@@ -1,3 +1,18 @@
+/**
+ * Copyright (C) 2015 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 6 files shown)


rohanKanojia:
ok to test

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
```diff
@@ -199,6 +195,29 @@ public T get() {
...
+      if (answer instanceof HasMetadata) {
+        HasMetadata hasMetadata = (HasMetadata) answer;
+        updateApiVersion(hasMetadata);
+      } else if (answer instanceof KubernetesResourceList) {
+        KubernetesResourceList list = (KubernetesResourceList) answer;
+        updateApiVersion(list);
+      }
+      return answer;
+    } catch (KubernetesClientException e) {
+      if (e.getCode() != 404) {
```
rohanKanojia: just a suggestion, better replace this by `HttpURLConnection.HTTP_NOT_FOUND` here and at some other places in this file

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Requirable.java
```diff
@@ -0,0 +1,14 @@
+package io.fabric8.kubernetes.client.dsl;
+
```
rohanKanojia: missing license header here, run `mvn -N license:format` to resolve

rohanKanojia:
[merge]

PR #1346, opened 2019-02-05
Fix #1326: Make CustomResource buildable so that CustomResourceFluent…

### LINKED ISSUE #1326: Make CustomResource @Buildable
I believe making `CustomResource` `@Buildable` would fix the problem reported here: https://github.com/sundrio/sundrio/issues/104

Is there a reason why `CustomResource` is not `@Buildable` already?

bfink13:
… is generated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@
 
     * Fix #1348: support `v1beta1` version for `ExecCredentials`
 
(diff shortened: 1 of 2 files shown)


bfink13:
Rebased on master and squashed

rohanKanojia:
[merge]

Issue #1347, opened 2019-02-06
Rename apiVersion → apiGroupVersion breaks client


jglick:
What is the rename by @iocanel in #1322 about? It matches neither [documentation](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#resources) nor what I observe from `kubectl get -o json` in a `v1.11.5-gke.5` cluster. In fact using the `master` version of this library, I get errors such as

```
com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "apiVersion" (class io.jenkins.x.client.kube.PipelineActivityList), not marked as ignorable (4 known properties: "items", "apiGroupVersion", "kind", "metadata"])
 at [Source: java.io.BufferedInputStream@73f07257; line: 1, column: 16] (through reference chain: io.jenkins.x.client.kube.PipelineActivityList["apiVersion"])
	at com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException.from(UnrecognizedPropertyException.java:62)
	at com.fasterxml.jackson.databind.DeserializationContext.handleUnknownProperty(DeserializationContext.java:834)
	at com.fasterxml.jackson.databind.deser.std.StdDeserializer.handleUnknownProperty(StdDeserializer.java:1093)
	at com.fasterxml.jackson.databind.deser.BeanDeserializerBase.handleUnknownProperty(BeanDeserializerBase.java:1478)
	at com.fasterxml.jackson.databind.deser.BeanDeserializerBase.handleUnknownVanilla(BeanDeserializerBase.java:1456)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.vanillaDeserialize(BeanDeserializer.java:282)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:140)
	at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:3814)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:2931)
	at io.fabric8.kubernetes.client.utils.Serialization.unmarshal(Serialization.java:235)
Caused: io.fabric8.kubernetes.client.KubernetesClientException: An error has occurred.
	at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:64)
[...]

PR #1348, opened 2019-02-06
Support client.authentication.k8s.io/v1beta1

jglick:
Amends #1224 to support the current `apiVersion` documented [here](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins). With #1347 worked around, this has been successfully tested on a GKE cluster using

```yaml
users:
- name: gcp
  user:
    exec:
      apiVersion: "client.authentication.k8s.io/v1beta1"
      command: "sh"
      args:
        - "-c"
        - |
            gcloud config config-helper --format=json | jq '{"apiVersion": "client.authentication.k8s.io/v1beta1", "kind": "ExecCredential", "status": {"token": .credential.access_token, "expirationTimestamp": .credential.token_expiry}}'
```

(The `gcloud container clusters get-credentials` command recommended by Google rewrites your `~/.kube/config` to use

```yaml
users:
- name: …
  user:
   auth-provider:
     name: gcp
     config:
       access-token: …
       cmd-args: config config-helper --format=json
       cmd-path: gcloud
       expiry: …
       expiry-key: '{.credential.token_expiry}'
       token-key: '{.credential.access_token}'
```

which `Config` does not support, and which I think is semi-deprecated.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,9 @@
     * Updated compatability matrix after model upgrade in README.md 
 
     * Fix #1306: Support `KUBECONFIG` env var with multiple paths
- 
+
+    * Fix #1348: support `v1beta1` version for `ExecCredentials`
+
   Dependency Upgrade
 
   New Feature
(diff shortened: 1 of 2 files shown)


rohanKanojia:
ok to test

PR #1354, opened 2019-02-07
Suppress warning about CRDs being beta

jglick:
If you are using custom resources, there is nothing you can do about the fact that [this system is still officially in beta](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#customresourcedefinition-v1beta1-apiextensions-k8s-io) so it is just annoying to get a warning about it in every JVM session. This is particularly noxious in the case of [Jenkins X serverless builds](https://jenkins-x.io/architecture/cloud-native-jenkins/) which print a message like

```
  18.333 [id=39]	WARNING	i.f.k.c.i.VersionUsageUtils#alert: The client is using resource type 'customresourcedefinitions' with unstable version 'v1beta1'
```

near the start of every build, as the `jx-resources` plugin (which uses `kubernetes-client` via `jx-java-client`) is initialized and begins updating a `PipelineActivity` CR.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/VersionUsageUtils.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/VersionUsageUtils.java
@@ -39,6 +39,10 @@ public final class VersionUsageUtils {
       return;
     }
 
+    if (type.equals("customresourcedefinitions") && version.equals("v1beta1")) {
+      return;
+    }
+
     if (isUnstable(version)) {
       if (LOG_EACH_USAGE || UNSTABLE_TYPES.putIfAbsent(type + "-" + version, true) == null) {
         alert(type, version);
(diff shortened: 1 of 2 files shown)


rohanKanojia:
ok to test

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/VersionUsageUtils.java
```diff
@@ -39,6 +39,10 @@ public static void log(String type, String version) {
       return;
     }
 
+    if (type.equals("customresourcedefinitions") && version.equals("v1beta1")) {
```
rohanKanojia: In my opinion, we should not add checks like these here. Maybe print that alert in debug mode rather than warn?

jglick: Yes, something I thought about. Can do it that way if you prefer.

rohanKanojia:
[merge]

PR #1355, opened 2019-02-13
Fix #1008 Use a reasonable buffer size for exec stdin

### LINKED ISSUE #1008: use exec copy file to pod too slow
My client version is 3.1.8
When i use k8s web socket api directly, every thing is ok.
But when i use your client, the upload speed is only about 1Kb per second.

my code is:

    public void uploadFile(KubernetesClient client,
                           String namespace,
                           String podName,
                           String containerName,
                           String desFilePath,
                           InputStream inputStream) throws IOException {
        ExecWatch watch = client
                .pods()
                .inNamespace(namespace)
                .withName(podName)
                .inContainer(containerName)
                .redirectingInput()
                .redirectingError()
                .exec("tar", "xf", "-", "-C", "/");
        TarArchiveEntry entry = new TarArchiveEntry(desFilePath);

        try (TarArchiveOutputStream tarOut = new TarArchiveOutputStream(watch.getInput());
             ByteArrayOutputStream byteOut = new ByteArrayOutputStream();) {

            byte[] bytes = new byte[4096];
            int count;
            while ((count = inputStream.read(bytes)) > 0) {
                byteOut.write(bytes, 0, count);
            }

            entry.setSize(byteOut.size());
            tarOut.putArchiveEntry(entry);
            byteOut.writeTo(tarOut);
            tarOut.closeArchiveEntry();
        }
    }

carlossg:
#1008 

We have fixed this issue passing our own `PipedInputStream`
https://github.com/jenkinsci/kubernetes-plugin/pull/427

But the default size makes any exec command really slow. Although it should be configurable somewhere

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 #### 4.1-SNAPSHOT
   Bugs
+   
+   * Fix #1008: Use a reasonable buffer size for exec stdin
   
   Improvements
   
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Inputable.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Inputable.java
@@ -23,4 +23,6 @@ public interface Inputable<I, P, T> {
     T writingInput(P in);
 
     T redirectingInput();
+
+    T redirectingInput(Integer bufferSize);
 }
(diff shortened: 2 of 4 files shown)


rohanKanojia:
@carlossg : Thanks, Could you please add an example/test of exec copy too which you used to verify this?

carlossg:
we have a workaround in https://github.com/jenkinsci/kubernetes-plugin/pull/427 but I'm not sure this is the right fix. What should be the default buffer size ? how to change it ? 
maybe add a `.redirectingInput(int bufferSize)` ?

rohanKanojia:
Hmm, yeah makes sense to me. @iocanel @oscerd thoughts??

rohanKanojia:
[merge]

PR #1370, opened 2019-02-26
Fix #1005: Loading a template from file and processing it locally by passing parameters map is broken.

### LINKED ISSUE #1005: Loading a template from file and processing it locally by passing a parameters map is broken

rohanKanojia:
Fix #1005

Actually, I'm not sure how it used to work before. Right now I've added a small fix so this operation works:
```
    KubernetesList list = client.templates().withParameters(map).load(getClass().getResourceAsStream("/template-with-number-params.yml")).processLocally(map);
```
So basically we're passing parameters in `withParameters` method, hashmap passed in `processLocally` is of no use. Was client able to parse yaml correctly with parameters earlier?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
   Bugs
    
    * Fix #1008: Use a reasonable buffer size for exec stdin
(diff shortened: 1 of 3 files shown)


rohanKanojia:
[merge]

Issue #1374, opened 2019-02-27
Add support for resizeChannel in ExecWebSocketListener


juju812:
After walk through the `ExecWebSocketListener` I found that currently `resizeChannel` is not supported. 
Is there any plan to support it? Or any other work around here to resize tty while bash execution?

Thx.

juju812:
seems like we can construct command with env to initialize terminal size like this:

```
exec("env", "TERM%3Dxterm", "COLUMNS%3D" + columns, "LINES%3D" + lines, "bash")
```

but without `resizeChannel` support we can't adjust the size online

rohanKanojia:
@juju812 : Hmm, Could you please create a PR for this?

juju812:
@rohanKanojia the workaround is just used for execute bash with tty. If a PR is need, in my view it can be added as an exec example.

Or I need to implement the `resizeChannel` in `ExecWebSocketListener`?

rohanKanojia:
@juju812 : if you have time to create PR with full implementation, then please do. Otherwise, we can add workaround as an example too to resolve confusion if someone else faces the same issue.

juju812:
@rohanKanojia A quick PR is created for the workaround. 

And about the `resizeChannel` implementation, unfortunately I don't have time now, but I'll inform you if I start handling this.

rohanKanojia:
@juju812 : Much appreciated, Thanks :+1:

4qing:
Hello, I now need this function to implement a web terminal that can dynamically change the size. I want to know if k8s supports this function. Which api is specific, I don’t seem to find this api on the k8s offical document.Thanks! @juju812

rohanKanojia:
@4qing : Maybe you can pick it up where @juju812 left.... We still need to implement resizeChannel in ExecWebSocketListener.

4qing:
Excuse me, I am bothered again, I looked at the python client implementation,[https://github.com/kubernetes-client/python/pull/515/](https://github.com/kubernetes-client/python/pull/515/),it send message by this way : 
```python
self.k8s_stream.write_channel(4,
                                      json.dumps({"Height": rows,
                                                  "Width": cols}))
```

In our project, it seems that the message is sent in the following way:
```java
private void send(byte[] bytes) throws IOException {
        if (bytes.length > 0) {
            WebSocket ws = webSocketRef.get();
            if (ws != null) {
                byte[] toSend = new byte[bytes.length + 1];
                toSend[0] = 0;
[...]

PR #1386, opened 2019-03-05
Add workaround to initialize terminal size while execute bash with tty.

juju812:
Related to issue https://github.com/fabric8io/kubernetes-client/issues/1374.

It's just a workaround, and if we want adjust the terminal size online,  resizeChannel support is needed.


rohanKanojia:
ok to test

rohanKanojia:
[merge]

PR #1453, opened 2019-03-23
feat: First draft of exposing port-forward via service DSL.

### LINKED ISSUE #1447: Portforward should also be accessible through the Service DSL
`kubectl` allows you to portforward to  a service. This is done by looking up for the pods that provide the service and then delagating to pod port-forward.

We should do the same.

iocanel:
Fixes #1447

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
   Improvements
 
     * Fix #1425: Preserve labels and fields when using CRD's withResourceVersion()
+    * Service DSL now includes methods for port forwarding
   
   Dependency Upgrade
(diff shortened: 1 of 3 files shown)


rohanKanojia:
@iocanel : Could you please add an example/test for this?

iocanel:
> @iocanel : Could you please add an example/test for this?

Port forwarding is an area where we are lacking tests, even in the pod dsl. 
Its no surprise as its pretty complicated and I am not sure if we can even mock it.

We should find a solution, but I wouldn't hold the PR for that.

rohanKanojia:
@iocanel :pray:

rohanKanojia:
[merge]

Issue #1455, opened 2019-03-25
How to use subject access review in kubernetes-client?


raiRaiyan:
I want to use the kubernetes client to figure out if the client has the required permission to the resource. I can do this in kubectl using the following command:  

`kubectl auth can-i create deployments --namespace dev`

This api uses `SelfSubjectAccessReview` according to the [documentation](https://kubernetes.io/docs/reference/access-authn-authz/authorization/). Does the kubernetes-client support this? If yes, how do I use it?

Going through the source code, I was able to create the API object from `io.fabric8.kubernetes.api.model.authorization` like this:
```
LocalSubjectAccessReview review = new LocalSubjectAccessReviewBuilder()
                        .withNewSpec()
                        .withNewResourceAttributes()
                        .withResource("namespace")
                        .withVerb("create")
                        .endResourceAttributes()
                        .endSpec().build();
```

But i cannot use this either as there is no `Handler` for this type of a resource. Is there any way that I can use this API object for the time being?

rohanKanojia:
Does it work when you use resource DSL for creation? Something like this:
```
KubernetesClient client = new DefaultKubernetesClient();
HasMetadata response = client.resource(review)..inNamespace("dev").createOrReplace();
```

raiRaiyan:
No, it does not work. I get this error:
```
No handler found for object:LocalSubjectAccessReview(apiVersion=authorization.k8s.io/v1, kind=LocalSubjectAccessReview, metadata=null, spec=SubjectAccessReviewSpec(extra={}, groups=[], nonResourceAttributes=null, resourceAttributes=ResourceAttributes(group=null, name=null, namespace=null, resource=namespace, subresource=null, verb=create, version=null, additionalProperties={}), uid=null, user=null, additionalProperties={}), status=null, additionalProperties={})
```

I tried converting the review object to json and using load:
```
String json = mapper.writeValueAsString(review);
List<HasMetadata> result = ((DefaultKubernetesClient) client).load(new ByteArrayInputStream(json.getBytes())).fromServer().get();
```

I get the same error. It looks for a handler for this object type and fails. Following is the stack trace:
```
java.lang.NullPointerException: null
	at io.fabric8.kubernetes.client.dsl.internal.NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.acceptVisitors(NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.java:289)
[...]

PR #1466, opened 2019-03-29
Adding support for using SubjectAccessReview 

raiRaiyan:
This commit adds support for SubjectAccessReview and LocalSubjectAccessReview. 

The openshift client already has support for `subjectAccessReviews`, so I had named it `subjectAccessReviewAuth` in `KubernetesClient`.


raiRaiyan:
This is pr for #1455

rohanKanojia:
ok to test

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/osgi/ManagedKubernetesClient.java
```diff
@@ -79,27 +79,7 @@
...
-import io.fabric8.kubernetes.client.dsl.PodResource;
-import io.fabric8.kubernetes.client.dsl.PolicyAPIGroupDSL;
-import io.fabric8.kubernetes.client.dsl.RbacAPIGroupDSL;
-import io.fabric8.kubernetes.client.dsl.Resource;
-import io.fabric8.kubernetes.client.dsl.RollableScalableResource;
-import io.fabric8.kubernetes.client.dsl.SchedulingAPIGroupDSL;
-import io.fabric8.kubernetes.client.dsl.ServiceResource;
-import io.fabric8.kubernetes.client.dsl.SettingsAPIGroupDSL;
-import io.fabric8.kubernetes.client.dsl.StorageAPIGroupDSL;
+import io.fabric8.kubernetes.client.dsl.*;
```
rohanKanojia: Please avoid using wildcard imports

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/SubjectAccessReviewDSLImpl.java
```diff
@@ -0,0 +1,154 @@
...
+    return isNamespaced;
+  }
+
+  @Override
+  public SubjectAccessReview create(SubjectAccessReview... resources) {
+    return new CreatableSubjectAccessReview().create(resources);
+  }
+
+  @Override
+  public io.fabric8.kubernetes.api.model.authorization.DoneableSubjectAccessReview createNew() {
```
rohanKanojia: Why do you need to specify the full class with package when you're having it in import?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/SubjectAccessReviewDSLImpl.java
```diff
@@ -0,0 +1,154 @@
...
+    return new CreatableSubjectAccessReview().createNew();
+  }
+
+  @Override
+  public Createable<LocalSubjectAccessReview, LocalSubjectAccessReview, DoneableLocalSubjectAccessReview> inNamespace(String namespace) {
+    //Should we use this namespace to set the metadata namespace and the spec namespace of the LocalSubjectAccessReview object?
+    return new SubjectAccessReviewDSLImpl(this.context, namespace, LSAR_PLURAL, true).local(namespace);
```
rohanKanojia: In my opinion, this namespace should be used for setting metadata namespace only.

raiRaiyan: That would mean the user has to provide the namepsace again in the spec, in addition to the `SubjectAccessReviewDSL`  when using a `LocalSubjectAccessReview`. This is redundant, don't you think so?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/SubjectAccessReviewDSLImpl.java
raiRaiyan: This was the IDE's doing. I will change this.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/SubjectAccessReviewDSLImpl.java
rohanKanojia: Yeah, but this aligns with standard kubernetes flow. I think in kubernetes also it's done like this. Could you please check?

raiRaiyan: Yes. I checked this. You are required to give the same namespace in the metadata and the spec if you are writing a yaml. But kubectl client has an easier syntax
```
kubectl auth can-i create deployments --namespace dev
```
This looks like a cleaner way to do this. I thought setting the namespace if the user has not set it, will make it easier to use. Thought?

rohanKanojia: Hm, well we can then make an adjustment in its *OperationsImpl class that overrides default implementation.

raiRaiyan: I have done it n `setNamespace()` method in the same class.

rohanKanojia:
@raiRaiyan : Could you please update CHANGELOG again resolving conflicts?

CHANGELOG.md
```diff
@@ -9,14 +9,14 @@
...
   Bugs
-  
     * Fix #1387: ValidatingWebhookConfigurationOperationsImpl should be a NonNamespaceOperation
     * Fix #1429: Fixes JsonMappingException: No resource type found for:v1#List when reading a Kubernetes List YAML
     * Fix #760: Api get pod from yaml issue
     * Fix #807: Loading a deployment from server with a config file throws exception
    
   Improvements
 
+    * Fix #1455: Use SubjectAccessReview and LocalSubjectAccessReview in kubernetes client using subjectAccessReviewAuth()
```
rohanKanojia: we released 4.2.0 last week, you have to move it above in 4.2-SNAPSHOT section.

Issue #1297, opened 2018-12-19
I get a wrong result when use exec  and params contains '&'


zxbu:
When the cmd like this 
```
            String cmd = "test -f /data/config.bak && mv -f /data/config.bak /data/config"
            
            kubernetesClient.pods()
                    .inNamespace(namespace)
                    .withName(podName)
                    .writingOutput(pipedOutputStream)
                    .writingError(System.err)
                    .exec("bash", "-c", cmd);
```

but  `mv -f /data/config.bak /data/config` will not run .

I think the reason is   url encoder/decoder, So I try like this, and it worked:

```
           String cmd = "test -f /data/config.bak && mv -f /data/config.bak /data/config"

            kubernetesClient.pods()
                    .inNamespace(namespace)
                    .withName(podName)
                    .writingOutput(pipedOutputStream)
                    .writingError(System.err)
                    .exec("bash", "-c", URLEncoder.encode(cmd, "UTF-8"));
```

rohanKanojia:
I think we need to encode command here as well: 
https://github.com/fabric8io/kubernetes-client/blob/31ddcea6d95cd9a12be6f5ea0aa6285e7ee0d3bf/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java#L235-L240

karthikmg:
Hi @rohanKanojia 

I would like to fix this issue.
This is my first time contributing to open source.

I will use the same **URLEncoder.encode(cmd, "UTF-8")** fix and send a pull request.

PR #1469, opened 2019-03-30
Encoding command in UTF-8

karthikmg:
Fix for issue [1297](https://github.com/fabric8io/kubernetes-client/issues/1297)

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java
@@ -24,6 +24,7 @@ import java.io.PipedInputStream;
 import java.io.PipedOutputStream;
 import java.io.Reader;
 import java.net.URL;
+import java.net.URLEncoder;
 import java.nio.channels.ReadableByteChannel;
 import java.nio.channels.WritableByteChannel;
 import java.nio.file.Path;
@@ -256,6 +257,7 @@ public class PodOperationsImpl extends HasMetadataOperation<Pod, PodList, Doneab
             } else {
                 sb.append("&command=");
             }
+            cmd = URLEncoder.encode(cmd, "UTF-8");
             sb.append(cmd);
         }
 
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java
@@ -257,7 +257,13 @@ public class PodOperationsImpl extends HasMetadataOperation<Pod, PodList, Doneab
             } else {
                 sb.append("&command=");
             }
-            cmd = URLEncoder.encode(cmd, "UTF-8");
+          
+            try {
+              cmd = URLEncoder.encode(cmd, "UTF-8");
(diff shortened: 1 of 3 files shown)


oscerd:
Ok to test

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java
```diff
@@ -256,6 +258,13 @@ public ExecWatch exec(String... command) {
             } else {
                 sb.append("&command=");
             }
+          
+            try {
+              cmd = URLEncoder.encode(cmd, "UTF-8");
+            } catch(UnsupportedEncodingException encodEx) {
```
rohanKanojia: would be nice if you could move this method to URLUtils...

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java
```diff
@@ -256,6 +258,13 @@ public ExecWatch exec(String... command) {
             } else {
                 sb.append("&command=");
             }
+          
+            try {
```
dev-gaur: Can you use try-with-resources here ? It would be better
```
try ( cmd = URLEncoder.encode(cmd, "UTF-8") ) {
   sb.append(cmd);
} catch (ex........) { ... }
```

Also, You can create a URL utility method to wrap  ```URLEncoder.encode(cmd, "UTF-8")``` , like @rohanKanojia suggested.

karthikmg: @dev-gaur 
URLEncoder.encode(cmd, "UTF-8") returns a String. This cannot be used in Try-With-Resources as it expects an instance that implements AutoCloseable.
Please let me know if my understanding is different.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/URLUtils.java
```diff
@@ -150,4 +152,7 @@ public static boolean isValidURL(String url) {
     }
   }
 
+  public static String encodeCommand(String cmd) throws UnsupportedEncodingException {
```
dev-gaur: @karthikmg A utility method should be usable anywhere required. The name `encodeCommand` suggests that its sole purpose is to encode commands. But its actually a general utilitilty function which can be used anywhere suitable.

Also, mind the fact that this function encodes a Url here, so, any string which resembles a Url can be an argument for his function. I understand that in this particular use case, we're encoding a command string, but the function isn't restricted to this use case. It can accept any string here.

The name `encodeToUTF` would make much more sense.

``` encodeToUTF(String url) { ... }```

and then calling it like 

```
URLUtils.encodeToUTF(cmd)
```

@rohanKanojia  correct me if I'm wrong

karthikmg: @dev-gaur 
Pushed the refactoring changes as per your review.
Please let me know where to add the test case. Should I add test case for URLUtils.encodeToUTF() or exec() in the PodOperationsImpl?

Also, I made changes to Changelog.md file and pushed them before the above refactoring of method name. But CI build is failing. Could you please let me know what the issue is so that I can fix it?

rohanKanojia:
[merge]

PR #1472, opened 2019-04-01
First draft of custom resources improvements

rohanKanojia:
Introduce a rawCustomResource() in client dsl 

rawCustomResource() would allow client to create, delete, list, update custom
resources without having to require Pojos of custom resources. Now the client would only
deal with JSONObjects.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/DefaultKubernetesClient.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/DefaultKubernetesClient.java
@@ -69,6 +69,7 @@ import io.fabric8.kubernetes.api.model.ServiceAccount;
 import io.fabric8.kubernetes.api.model.ServiceAccountList;
 import io.fabric8.kubernetes.api.model.DoneableServiceAccount;
 import io.fabric8.kubernetes.client.dsl.*;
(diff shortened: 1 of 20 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/DefaultKubernetesClient.java
```diff
@@ -244,6 +245,11 @@ public KubernetesListMixedOperation lists() {
       .withDoneableType(doneClass));
   }
 
+  @Override
+  public RawCustomResourceOperationsImpl rawCustomResource(CustomResourceDefinition customResourceDefinition) {
```
iocanel: I am wondering if this can be named `customResource` instead of `rawCustomResource`?

iocanel: Also it would probably make sense to be able to have the same functionality with just specifying:

- apiGroup
- apiVersion
- kind
- plural  (optionally).

rohanKanojia:
retest this please

rohanKanojia:
Support to watch custom resources via this new dsl endpoint still needs to be implemented, but I think we can merge this as well.

rohanKanojia:
[merge]

PR #1487, opened 2019-04-10
Issue 1468: Taking resource labels into consideration when comparing resources

brokenjpl:
Resource only equal if labels are also a match.

This restrictions prevented us from updating only the labels on the resources when calling `kubeClient.resource(resourceObj).createOrReplace();`

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/ResourceCompare.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/ResourceCompare.java
@@ -18,6 +18,7 @@ package io.fabric8.kubernetes.client.utils;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.commons.collections.MapUtils;
 
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


rohanKanojia:
ok to test

brokenjpl:
I'll add an entry to the CHANGELOG.MD shortly.

Consider adding this requirement to the CONTRIBUTING.md for future contributors :).

brokenjpl:
> Consider adding this requirement to the CONTRIBUTING.md for future contributors :).

Just saw that it's there. I take it back!

rohanKanojia:
@brokenjpl : Would appreciate if you could append this to our CONTRIBUTING.md ;-) in this PR itself

rohanKanojia:
oh, cool :smile:

brokenjpl:
Not clear as to what failed. Could I get a hint?

oscerd:
[merge]

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/ResourceCompare.java
```diff
@@ -43,4 +52,11 @@
         result.remove(METADATA);
         return result;
     }
+
+    private static Map<String, Object> fetchLabels(Map<String, Object> map){
+        if (!map.containsKey(METADATA) || !((Map<Object, Object>)map.get(METADATA)).containsKey(LABELS)){
+            return MapUtils.EMPTY_MAP;
```
Ladicek: Why not just `Collections.emptyMap()`? This adds a hard dependency on Commons Collections, which is useless (perfectly acceptable alternative directly in the JDK) and actually causes `java.lang.NoClassDefFoundError: org/apache/commons/collections/MapUtils` at runtime.

brokenjpl: Oups, good catch, thanks.

rohanKanojia: np, 4.2.2 is released with the fix

PR #1492, opened 2019-04-15
Chore #1470: Fix Javadoc issues

### LINKED ISSUE #1470: Fix JavaDoc issues across project
After upgrading javadoc maven plugin, we're facing build failures during release process. I had to revert to previous version in order to release 4.2.0. But we should get all docs fixed before next release.

### How to reproduce:
```
mvn javadoc:jar
```

You would be able to see lots of warnings, that it's complaining about. I fixed errors but it's failing even on warnings.

rohanKanojia:
Fix #1470


12692d:
Secure my accout

PR #1498, opened 2019-04-21
fixing nullPointer for "generateName" 

### LINKED ISSUE #1373: client.load(new FileInputStream(args[0])).createOrReplace(); throws nullPointer for "generateName"
If the pod-spec.yaml contains `generateName` tag then createOrReplace() is throwing the below exception.

java.lang.NullPointerException
    at java.net.URI$Parser.parse(URI.java:3042)
    at java.net.URI.<init>(URI.java:588)
    at io.fabric8.kubernetes.client.utils.URLUtils.join(URLUtils.java:48)
    at io.fabric8.kubernetes.client.dsl.base.BaseOperation.getMandatory(BaseOperation.java:208)
    at io.fabric8.kubernetes.client.dsl.base.BaseOperation.get(BaseOperation.java:177)
    at io.fabric8.kubernetes.client.handlers.PodHandler.reload(PodHandler.java:53)
    at io.fabric8.kubernetes.client.handlers.PodHandler.reload(PodHandler.java:32)
    at io.fabric8.kubernetes.client.dsl.internal.NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.createOrReplace(NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.java:202)
    at io.fabric8.kubernetes.client.dsl.internal.NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.createOrReplace(NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.java:62)
    at com.nokia.k8s.InterpreterLanuch.main(InterpreterLanuch.java:66)

kunal-kushwaha:
Fixes #1373
Generating a random name incase the name provided is null instead of throwing an exception.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
@@ -58,6 +58,7 @@ import java.util.Objects;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Predicate;
+import java.util.Random;
 import okhttp3.HttpUrl;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
@@ -153,6 +154,16 @@ public class BaseOperation<T, L extends KubernetesResourceList, D extends Doneab
     if(Utils.isNotNullOrEmpty(value)) {
       requestUrlBuilder.addQueryParameter(name, value);
     }
+  }
+
+  private static String generateRandomString(int size) {
+    StringBuffer stringBuffer = new StringBuffer(size);
+    Random random = new Random();
+    for (int i = 0; i < size; i++) {
+      int randomChar = 97 + (int)(random.nextFloat() * 26);
+      stringBuffer.append((char)randomChar);
+    }
+    return stringBuffer.toString();
   }
 
   @Override
@@ -252,7 +263,7 @@ public class BaseOperation<T, L extends KubernetesResourceList, D extends Doneab
   @Override
   public R withName(String name) {
     if (name == null || name.length() == 0) {
-      throw new IllegalArgumentException("Name must be provided.");
+      name = generateRandomString(3) + "-" + generateRandomString(3) + "-pod";
     }
     return (R) newInstance(context.withName(name));
   }
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
@@ -251,6 +251,14 @@ public class BaseOperation<T, L extends KubernetesResourceList, D extends Doneab
 
   @Override
   public R withName(String name) {
+    HasMetadata h = (HasMetadata) item;
+    String value = h.getMetadata() != null ? h.getMetadata().getName() : null;
+    if(value != null || !value.isEmpty()){
+      return (R) newInstance(context.withName(name));
+    }
+    if (name == null || name.isEmpty()) {
+      throw new IllegalArgumentException("Name must be provided.");
+    }
     return (R) newInstance(context.withName(name));
   }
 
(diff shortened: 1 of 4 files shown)


rohanKanojia:
ok to test

kunal-kushwaha:
> In my opinion, we should simply leave random name generation up to Kubernetes API server. We should just allow the name to be null if generateName is present. That should do the trick!

That seems the right way to do it, on it.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
```diff
@@ -251,9 +251,6 @@ public D edit() throws KubernetesClientException {
 
   @Override
   public R withName(String name) {
-    if (name == null || name.length() == 0) {
```
rohanKanojia: we should skip this check only when generateName != null

dev-gaur: instead of name.length( ) == 0, use name.isEmpty( )

rohanKanojia:
I think we should implement one more method to this interface names `withName(String name, String generateName)` which should be used when passing name internally from handlers

kunal-kushwaha:
> I think we should implement one more method to this interface names `withName(String name, String generateName)` which should be used when passing name internally from handlers

I just have this question, how do i get information from yaml files like generateName in my Java class?

rohanKanojia:
You don't have to get information from yaml as it's already done. This yaml is deserialized into HasMetadata object so it should be there.

kunal-kushwaha:
Using name.isEmpty() instead of name.length() == 0
Would like to know if I have used HasMetadata the right way and if any changes are required.

kunal-kushwaha:
@dev-gaur @rohanKanojia  
Using name.isEmpty() instead of name.length() == 0
Would like to know if any changes are to be made.

rohanKanojia:
@kunal-kushwaha : I'm not sure whether it still solves the problem. Did you test this on your machine?

rohanKanojia:
well, we can fetch generateName like you're fetching name from that `name` method. that should make things easy. We won't need to add a new method to this interface. 

```
  private static <T> String generateName(T item) {
    if(item instanceof HasMetadata) {
      HasMetadata h = (HasMetadata) item;
      return h.getMetadata() != null ? h.getMetadata().getGenerateName() : null;
    }
    return null;
  }
```
then we can use it like this, I haven't tested this but I think it can work:
```
  @Override
  public R withName(String name) {
    String generateName = generateName(item);

    if (generateName == null && name == null || name.isEmpty()) {
      throw new IllegalArgumentException("Name must be provided.");
    }
    return (R) newInstance(context.withName(name));
  }
```

kunal-kushwaha:
@rohanKanojia
I get it now,I tried using HasMetadata h = (HasMetadata) item; but I wasn't calling h.getMetadata() the right way thats why it failed all the test cases in the previous commit. 
Working on it.

kunal-kushwaha:
@rohanKanojia
I get it now, I tried using HasMetadata h = (HasMetadata) item; but I wasn't calling h.getMetadata() the right way (fead985).
Working on it.

kunal-kushwaha:
@rohanKanojia 
Thanks
I get it now, I did something similar in fead985 but wasn't calling h.getMetadata() the right way.
Working on it.

kunal-kushwaha:
Can one of the admins verify it?

oscerd:
Ok to test

kubernetes-tests/src/test/resources/test-pod.yml
```diff
@@ -17,6 +17,7 @@
 apiVersion: v1
 kind: Pod
 metadata:
+  generateName: zep-ag-pod
```
rohanKanojia: you're providing `name` and `generateName` both here, you should use a yaml which has only `generateName` in it's metadata to test your fix.

rohanKanojia:
@kunal-kushwaha : Could you please check why the added test is failing?

kunal-kushwaha:
> @kunal-kushwaha : Could you please check why the added test is failing?

Yes working on it

rohanKanojia:
Hmm, I pulled this and tested it. It doesn't look like this approach works. But on revisiting this issue I'm even more unsure about it's use case.

PR #1506, opened 2019-04-29
Add example for custom resource new dsl

rohanKanojia:
Minor improvements in RawCustomResourceOperationsImpl like `load()`, edit with Maps and an example


rohanKanojia:
[merge]

PR #1521, opened 2019-05-14
Fix #758: Deleting Deployments with `.cascading(true)` creates a new Replica Set

### LINKED ISSUE #758: Deleting Deployments with `.cascading(true)` creates a new replica set
When I invoke 
```groovy
client.extensions()
  .deployments()
  .inNamespace(namespace)
  .withName(name)
  .cascading(true)
  .delete()
```

On some deployment `abc`, the client library deletes the deployment `abc` as well as the current replica set it's managing, but leaves behind a new replicaset `abc-<random number>` with 0 pods.

My client library is at version 2.2.14

rohanKanojia:
Fixes #758

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/ResourceHandler.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/ResourceHandler.java
@@ -120,7 +120,7 @@ public interface ResourceHandler<T, V extends VisitableBuilder<T, V>> {
    * @param item          The resource to delete.
    * @return              The true if the resource was successfully deleted.
    */
-  Boolean delete(OkHttpClient client, Config config, String namespace, T item);
+  Boolean delete(OkHttpClient client, Config config, String namespace, Boolean cascading, T item);
 
 
     /**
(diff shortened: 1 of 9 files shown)


rohanKanojia:
[merge]

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/KubernetesListOperationsImpl.java
```diff
@@ -152,7 +152,7 @@ public Boolean delete(List<KubernetesList> lists) {
     for (KubernetesList list : lists) {
       for (HasMetadata item : list.getItems()) {
         ResourceHandler<HasMetadata, HasMetadataVisitiableBuilder> handler = Handlers.get(item.getKind(), item.getApiVersion());
-        if (!handler.delete(client, config, namespace, item)) {
+        if (!handler.delete(client, config, namespace, false, item)) {
```
deiwin: I don't have a good understanding of this code, but is it possible that this PR (and possibly this line) changed the default value for `cascading` from `true` to `false` in some scenarios? I just upgraded from 4.1.3 to 4.3.0 and some things that were previously garbage collected are left orphaned now.

rohanKanojia: Yes, it might be possible. Let me check whenever I get time. Sorry for the inconvenience!

jglick: Speculated to have caused [JENKINS-58306](https://issues.jenkins-ci.org/browse/JENKINS-58306).

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/ResourceHandler.java
```diff
@@ -120,7 +120,7 @@ public boolean equals(Object obj) {
    * @param item          The resource to delete.
    * @return              The true if the resource was successfully deleted.
    */
-  Boolean delete(OkHttpClient client, Config config, String namespace, T item);
+  Boolean delete(OkHttpClient client, Config config, String namespace, Boolean cascading, T item);
```
jglick: missing `@param`

PR #1555, opened 2019-05-28
Fix #1553: Add API for field selectors not matching a value.

### LINKED ISSUE #1553: Allow setting also field selectors for non-matching fields
The Kubernetes API spec mentions for field-selectors they support two operators;
- key `=` value
- key `!=` value
https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/

Currently the Kubernetes client provides only the option to specify matching key-value pairs using `withField(s)`. And with a hack its possible to do `withField("key!", "value")` which would result in `key!=value` (notice the exclamation mark after "key!").
But this might stop working when internals change.

Hence, to keep aligned with 'label' selectors, it would be nice to have a `withFieldsNotMatching` extension to the API for the Kubernetes client.

mdonkers:
The Kubernetes API allows specifying field selectors where key-value
pair match, but also specifying non-matching key-value pairs using the
operator `!=`: https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/

Currently the Kubernetes Client allows specifying matching field
selectors using the `withField(key, value)` method. And by appending
a `!` (exclamation mark) after the key it's possible to 'hack' also the
non-matching pairs. But should internals change, this will quickly
break.

Instead this PR provides the necessary endpoint to specify non-matching
key-value pairs in field selectors by providing the method
`withFieldNot(key, value)`.

Some decisions to further discuss:

- Non-matching pairs can be specified multiple times with the same key.
Therefore a multi-value map is needed. To keep with the other filters
(labels), I used `Map<String, String[]>`. The alternative was to use
a list.
The downside of this decision is that a new String array needs
to be created when multiple values with the same key are added. But the
likeliness is quite small this will get called very often. And because
of the `merge` function on `Map`, for the first call it will directly
use the provided value, so no real overhead.

- The API is kept the same as `withField` however, and not like
`withLabelIn` which allows varargs. I figured this would be
non-intuitive. Downside here is that you cannot 'override' a key-value
pair. It would just add more values to the same key. But for me the
intuitiveness weighted stronger than being able to override pairs.

Fixes #1553

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
   Improvements
    * Fix #1589: Move HorizontalPodAutoscaler to autoscaling/v1
+   * Fix #1553: Allow to explicitly set non-matching field selectors using `withoutField`
 
   Dependency Upgrade
 
@@ -21,7 +22,7 @@
     * Fix #1550: MutatingWebhookConfigurationOperationsImpl should be a NonNamespaceOperation
 
   Improvements
-  
+
     * Added example for raw custom resources.
 
   Dependency Upgrade
@@ -47,22 +48,22 @@
 #### 4.2.1 (15-04-2019)
 
   Bugs
-    
+
     * Fix #1297: wrong result produced when exec in used and params contains '&'. Url string not encoded properly.
     * Fix #1473: Use correct plural form in OpenshiftRole
     * Fix #1480: The kubernetes-client is not optionally depending on bouncycastle.
     * Fix #1490: Resource could not be loaded
     * Fix #1468: Taking labels into consideration when comparing resources for equality.
-  
+
   Improvements
-  
+
     * Fix #1455: Use SubjectAccessReview and LocalSubjectAccessReview in kubernetes client using subjectAccessReviewAuth()
-  
+
   Dependency Upgrade
-  
+
 
   New Feature
-  
+
     * First Draft of Custom Resource Improvements (#1472)
 
 #### 4.2.0 (29-03-2019)
@@ -276,7 +277,7 @@
 
   New Feature
 
-    * Fix #1020 : Support for Kubernetes/Openshift v1.9/v3.9 respectively    
+    * Fix #1020 : Support for Kubernetes/Openshift v1.9/v3.9 respectively
 
   Improvements
 
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Filterable.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Filterable.java
@@ -43,6 +43,10 @@ public interface Filterable<T> {
 
   T withField(String key, String value);
 
+  T withoutFields(Map<String, String> fields);
+
+  T withoutField(String key, String value);
+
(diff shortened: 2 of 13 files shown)


rohanKanojia:
ok to test

mdonkers:
Apparently copying the license from another file introduced unwanted `<p>` markers and spaces.
Updated the license and squashed the commits.

rohanKanojia:
[merge]

iocanel:
It seems that @rohanKanojia merged, while I was requesting changes.
@mdonkers can you please check my review and do a followup PR?

Thanks!

rohanKanojia:
@iocanel : oh, sorry

iocanel:
> @iocanel : oh, sorry

No  worries. I just stoped the merge build, so @mdonkers can apply the changes in this PR.

mdonkers:
Thanks for the feedback so far.

The difference between Field selectors and Label selectors, is that the latter also supports "set-based requirements". I believe these map to the `withLabelsIn` and `withLabelsNotIn` functions. Field selectors only support equality requirements (like labels), matching to the `withField`, `withLabel` etc.
So in that regard `withFieldNotIn` does not make sense to me. Neither does keeping the parameters equal to `withLabelsNotIn` as it shouldn't be like a set-based requirement.

I did miss the `withoutLabel` functions, which I believe match more the similar functionality (so at least it should be renamed to `withoutField`). But then the current `withoutLabel` implementation misses the logic where you can negate multiple values for the same key. I could try changing the backing map to `Map<String, String[]>` as well, without changing the API. Only the `withoutLabels(Map<String, String> labels)` would not totally match the backing. And `withoutLabel(key)` (so just taking the key, with `null` value) doesn't make sense for fields, only for labels.

That, to me, feels like the best we could do without creating any breaking API changes. If there's a policy in place, might then be nice to deprecate `withoutLabels` or change it so it takes a multi-value map as well. But with generics, there's no easy migration path.

iocanel:
My comment refers to `withFieldsNot` that accepts as an argument a `Map<String, String[]>`.

Regardless of how it has been implemented underneath or what means the api server provides, if logically they are doing the same thing, they should be named similarly.

mdonkers:
I've renamed `withFieldsNot` to `withoutField` and updated `withoutLabel` correspondingly. I believe this is the better way forward and consistent (also with the Kubernetes API). And except for the deprecation, I believe it doesn't break any existing API calls.
Please let me know if this would be a way forward, or whether it makes it worse :-/

rohanKanojia:
@mdonkers : We just made a release of 4.3.0 yesterday. Could you please update changelog.md? I'm okay with your change, do you think we should also add `withoutFieldsIn(Map<>)` or `withoutFields(Map<>)` like Ioannis mentioned in his comment too?

mdonkers:
Hi,

Updated the `CHANGELOG.md` to the correct state.
Also added a `withoutFields(Map<>)` function as suggested, but added the `@Deprecated` flag as with the `withoutLabels(Map<>)` since we need to move this to something better (supporting a multi-value map to support multiple values for the same key).

rohanKanojia:
cool, Thanks @mdonkers

rohanKanojia:
[merge]

PR #1594, opened 2019-06-18
Parsing kube config file without reflection. Fixes #1591

### LINKED ISSUE #1591: Parse kube config file without reflection
As discussed in https://github.com/fabric8io/kubernetes-client/issues/1427#issuecomment-502155141, to increase Quarkus/Native Image compatibility and reducing reflection usage, we're proposing a change in how the kube config file is being parsed.

The suggested change is to use [SnakeYAML](https://bitbucket.org/asomov/snakeyaml/) (library [used by Jackson](https://github.com/FasterXML/jackson-dataformats-text/tree/master/yaml) to work with YAML files) to map the config files to a map and then manually populate the `io.fabric8.kubernetes.api.model.Config` model. Unfortunately mapper libraries (like Dozer and Apache Commons Beanutils) relies on heavy reflection usage to map POJOs.

This change won't add any other dependency that's already in the project's dependency tree.

ricardozanini:
Fixes #1591 

**Work in Progress: this PR has been opened to discuss the implementation details with manteiners**

Hi @geoand and @iocanel! As promised, here's the PR with an implementation suggestion to remove reflection completely from the configuration use case. 

I've tested this feature across your unit tests, IT and inside the native image. Seems solid, but I intend to write some more tests. 

Before keep going, I'd like to hear from you if this approach seems fine or if you'd like to take another one. With this new approach I was able to run the authentication and config features from kubernetes client inside a native image without registering any class for reflection.

I understand that's not the ideal world and in the future we could auto-generate those "mappers" with a maven plugin during build time.

Let me know your thoughts.

Cheers!

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/KubeConfigUtils.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/KubeConfigUtils.java
@@ -15,8 +15,6 @@
  */
 package io.fabric8.kubernetes.client.internal;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import io.fabric8.kubernetes.api.model.AuthInfo;
 import io.fabric8.kubernetes.api.model.Cluster;
 import io.fabric8.kubernetes.api.model.Config;
@@ -24,25 +22,31 @@ import io.fabric8.kubernetes.api.model.Context;
 import io.fabric8.kubernetes.api.model.NamedAuthInfo;
 import io.fabric8.kubernetes.api.model.NamedCluster;
 import io.fabric8.kubernetes.api.model.NamedContext;
+import io.fabric8.kubernetes.client.internal.mappers.ConfigMapper;
+import org.yaml.snakeyaml.Yaml;
 
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.IOException;
 import java.util.List;
 
 /**
  * Helper class for working with the YAML config file thats located in
  * <code>~/.kube/config</code> which is updated when you use commands
- * like <code>osc login</code> and <code>osc project myproject</code>
+ * like <code>oc login</code> and <code>oc project myproject</code>
  */
 public class KubeConfigUtils {
+    
   public static Config parseConfig(File file) throws IOException {
-    ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
-    return mapper.readValue(file, Config.class);
+    final Yaml yamlMapper = new Yaml();
+    final ConfigMapper configMapper = new ConfigMapper();  
+    return configMapper.map(yamlMapper.load(new FileInputStream(file)));
   }
 
   public static Config parseConfigFromString(String contents) throws IOException {
-    ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
-    return mapper.readValue(contents, Config.class);
+    final Yaml yamlMapper = new Yaml();
+    final ConfigMapper configMapper = new ConfigMapper();
+    return configMapper.map(yamlMapper.load(contents));
   }
 
   /**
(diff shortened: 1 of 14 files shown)


rohanKanojia:
ok to test

geoand:
Thanks @ricardozanini !

I'll let @iocanel comment since he is the expert here

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/KubeConfigUtils.java
```diff
@@ -15,34 +15,38 @@
  */
 package io.fabric8.kubernetes.client.internal;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
```
iocanel: Why do you replace jackson with a custom snakeyaml + custom deserializers?
Is this really needed? 
Why is jackson an issue only on `Config` and not when desrializing responses from the api server?

ricardozanini: Hi @iocanel! I replaced only the `Config` model because this way we can leverage from the configuration and authentication features from the client without using reflection. The idea is to replace with a custom mapper the responses as well.

In our project, having the client handle the authentication features, we can leverage from the `OkHttpClient` to query the API and deserialize using another options (like json path for example).

I see three options to handle this serialization issue when working with Quarkus in a native image:

1. Register the entire model using [GraalVM SDK](https://github.com/oracle/graal/blob/master/substratevm/REFLECTION.md)
2. [Register with Quarkus](https://quarkus.io/guides/building-native-image-guide.html) ([could be an extension](https://quarkus.io/guides/extension-authors-guide))
3. Use a custom mapper

This PR tackles the #3 option since the use case is only to parse the kube yaml files and it seemed a simple approach at first. Those mappers could be generated during build time, like the way you guys are doing with the domain model.

iocanel: So, my undersntanding is that long term we will go with option `3` for everything.
Untill, then can't we handle `Config` using `1`?

I want to avoid introducing a mapping mechanism just for handling a single class.

geoand: Can we perhaps just add a GraalVM json file for registering reflection with GraalVM as @iocanel mentions? Wouldn't that be an easier first step?
Also, that file could contain other GraalVM specific configurations, could it not?

ricardozanini: @iocanel a long term option should be `3` if we decide to drop reflection usage at all. You're right to deny introducing a custom mapper for a single class, I'd say the same. This is just a proposal for a mapper that we could build for the entire client. I could work on it.

@geoand manually registering the entire domain model is the easiest solution and it's the approach that you took for the https://github.com/quarkusio/quarkus/pull/2910 PR, right? Yeah, I guess that this option is a great first step and we can go ahead with it! For this PR, though, I'd say to introduce the option `3` for the entire client.

So, the question is: **should we create our custom (generated) mapper for the long term or we stick with manual reflection registration provided by Quarkus/SubtrateVM?** Having the custom mapper won't exclude the Quarkus extension, since there's still the unsupported feature in subtrateVM regarding the EC Keys with Bouncy Castle.

geoand: I would say that making incremental improvements on reducing reflection in this project will be great down the line. That said, I think @iocanel has some thoughts on how to make that easier based on some other he has done.

The Quarkus PR does register everything in the Kubernetes Model for reflection, but with a twist :). It doesn't force everything to end up in the native binary, but instructs GraalVM to include all methods and fields for the model classes that are actually used in the application that includes the it.

ricardozanini: So I guess that we can close this PR since @iocanel is already working on something? I worked around the `Config` parsing stuff just setting `KUBERNETES_AUTH_TRYKUBECONFIG` to `false`. And then mapped the JSON response as simple `Map`s to avoid reflection of the domain classes. Worked, but now I guess with the extension I'll drop this work around.

geoand: Let's what and see what the man himself will say :). 
I know the work in question was done on another project so I am not sure if it can be of immediate use here.

In any case, thanks a lot of looking into this!

ricardozanini:
@iocanel I agree with you, I'm closing this PR for now. Many thanks!

PR #1599, opened 2019-06-22
Corrected type returned by Config.builder() 

### LINKED ISSUE #1592: io.fabric8.kubernetes.client.Config.builder().build creates the wrong type of Config object
- io.fabric8.kubernetes.client.Config.builder() gives you a io.fabric8.kubernetes.api.model.ConfigBuilder object.
- io.fabric8.kubernetes.api.model.ConfigBuilder.build() gives you a io.fabric8.kubernetes.api.model.Config, which is not a io.fabric8.kubernetes.client.Config like you would expect.

Because of this, the following code won't compile because DefaultKubernetesClient expects io.fabric8.kubernetes.client.Config:

```
ConfigBuilder configBuilder = Config.builder();
// Set some properties on the ConfigBuilder
Config config = configBuilder.build();
final DefaultKubernetesClient client = new DefaultKubernetesClient(config);
```

If someone can verify that this is an actual bug and I'm not just missing something, I would be happy to fix it up.

AnthonyWeston:
Closes #1592


rohanKanojia:
ok to test

oscerd:
[merge]

PR #1616, opened 2019-07-03
add type parameter to custom resource list

fstab:
Without the type parameter, `list.getItems()` returns a `List<Object>`. It should return `List<T>`. Built-in types implement this correctly, for example `PodList` extends `KubernetesResourceList<Pod>`, so `PodList.getItems()` returns a `List<Pod>`. Adding the type parameter makes this work for custom resources.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
     * assertNotNull replaced with assertTrue for boolean statements in unit tests
     * Test coverage for PodPreset
     * Added test coverage for PersistentVolume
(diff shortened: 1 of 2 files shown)


rohanKanojia:
ok to test

rohanKanojia:
@fstab : Could you also please add a line to CHANGELOG.md regarding this change?

fstab:
@rohanKanojia: done adding a line to `CHANGELOG.md`.

rohanKanojia:
[merge]

PR #1638, opened 2019-07-18
Fix #1607: WaitUntilReady for lists

### LINKED ISSUE #1607: WaitUntilReady is currently broken.
Using something like:

      client.resourceList(list).waitUntilReady(config.getReadinessTimeout(), TimeUnit.MILLISECONDS);

This worked great until `4.1.0` but seems broken in `4.2.0` as it seems to return immediately.

rohanKanojia:
Fix #1607

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
@@ -22,7 +22,13 @@ import io.fabric8.kubernetes.api.model.KubernetesResourceList;
 import io.fabric8.kubernetes.api.model.LabelSelector;
 import io.fabric8.kubernetes.api.model.LabelSelectorRequirement;
 import io.fabric8.kubernetes.api.model.RootPaths;
-import io.fabric8.kubernetes.client.*;
+import io.fabric8.kubernetes.client.Config;
+import io.fabric8.kubernetes.client.ConfigBuilder;
+import io.fabric8.kubernetes.client.KubernetesClientException;
+import io.fabric8.kubernetes.client.OperationInfo;
+import io.fabric8.kubernetes.client.ResourceNotFoundException;
+import io.fabric8.kubernetes.client.Watch;
+import io.fabric8.kubernetes.client.Watcher;
 import io.fabric8.kubernetes.client.dsl.Deletable;
 import io.fabric8.kubernetes.client.dsl.EditReplacePatchDeletable;
 import io.fabric8.kubernetes.client.dsl.FilterWatchListDeletable;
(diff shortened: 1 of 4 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
```diff
@@ -989,7 +995,26 @@ protected T waitUntilExists(long amount, TimeUnit timeUnit) throws InterruptedEx
...
   public T waitUntilReady(long amount, TimeUnit timeUnit) throws InterruptedException {
-    return waitUntilExists(amount, timeUnit);
+
+    long timeoutInNanos = timeUnit.toNanos(amount);
+    long end = System.nanoTime() + timeoutInNanos;
+
+    while (System.nanoTime() < end) {
+      T item = fromServer().get();
+
+      if (Readiness.isReady((HasMetadata) item)) {
```
iocanel: This call may through an Exception if `waiting` makes no sense for the resource.

When used for a single resource it's ok, but if batch waiting delegates to this call, this may be problematic.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.java
```diff
@@ -82,7 +82,42 @@
...
+
+    try {
+      final CountDownLatch latch = new CountDownLatch(size);
+      for (final HasMetadata meta : items) {
+        final ResourceHandler<HasMetadata, HasMetadataVisitiableBuilder> h = handlerOf(meta);
+        executor.submit(() -> {
+          try {
+            result.add(h.waitUntilReady(client, config, meta.getMetadata().getNamespace(), meta, amount, timeUnit));
+          } catch (Throwable t) {
+            //consider all errors as not ready.
```
iocanel: This is tricky.

If a resource doesn't support the notion of waiting `Readiness.isReady` will throw an Exception, so with the current approach, you will wait until timeout. For that particular case, we should do exactly the opposite `consider resource ready`.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.java
```diff
@@ -82,7 +82,42 @@
...
+            LOGGER.warn("Error while waiting for: [" + meta.getKind() + "] with name: [" + meta.getMetadata().getName() + "] in namespace: [" + meta.getMetadata().getNamespace() + "]: " + t.getMessage()+ ". The resource will be considered not ready.");
+            LOGGER.debug("The error stack trace:", t);
+          } finally {
+            //We don't want to wait for items that will never become ready
```
iocanel: This comment conflicts with the previous comment.

So, how does it work when an error happens? do we consider it ready or not?

kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/WaitUntilReadyExample.java
```diff
@@ -48,14 +50,33 @@ public static void main(String args[]) throws IOException, InterruptedException
         .endSpec()
         .build();
 
+      Pod secondPod = new PodBuilder()
```
iocanel: It's not a second pod that is missing from this test, but more like a resource like a Secret or something, which is not waitable.

oscerd:
Is this merge able?

rohanKanojia:
yeah, Let's merge this.

rohanKanojia:
@iocanel : Please see if you can review this again whenever you have some time.

rohanKanojia:
[merge]

Issue #1505, opened 2019-04-29
Unexpected response(403 Forbidden), to the authorization request. Missing header:[Location]!


mchoma:
With OCP 4.0 / 4.1 (Kubernetes Master Version v1.13.4+81fc896) we are facing problem with fabric8 client authenticating using username/password - `IOException: Unexpected response(403 Forbidden), to the authorization request. Missing header:[Location]!` We have workarounded it by using token. Same code works ok with OCP 3.11 

```
Caused by: java.io.IOException: Unexpected response(403 Forbidden), to the authorization request. Missing header:[Location]!
	at io.fabric8.openshift.client.internal.OpenShiftOAuthInterceptor.authorize(OpenShiftOAuthInterceptor.java:133)
	at io.fabric8.openshift.client.internal.OpenShiftOAuthInterceptor.acquireNewToken(OpenShiftOAuthInterceptor.java:111)
	at io.fabric8.openshift.client.internal.OpenShiftOAuthInterceptor.intercept(OpenShiftOAuthInterceptor.java:63)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
	at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:200)
	at okhttp3.RealCall.execute(RealCall.java:77)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:379)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:344)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleGet(OperationSupport.java:313)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleGet(OperationSupport.java:296)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.handleGet(BaseOperation.java:826)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.getMandatory(BaseOperation.java:243)
	... 47 more
```

mnovak1:
The issue is that:
[...]

PR #1641, opened 2019-07-18
Openshift 4.1 + FMP: Unexpected response (403 Forbidden), to the authorization request. Missing header:[Location]!

rnetuka:
From downstream. Support for authorization in OpenShift 4. Has backwards compatibility with OpenShift 3.

### DIFF
--- a/openshift-client/pom.xml
+++ b/openshift-client/pom.xml
@@ -144,6 +144,12 @@
       <version>${project.version}</version>
     </dependency>
 
+    <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
+      <version>20180813</version>
+    </dependency>
+
     <!-- Compile Only Dependencies -->
     <dependency>
       <groupId>io.sundr</groupId>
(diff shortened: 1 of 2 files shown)


openshift-client/pom.xml
```diff
@@ -144,6 +144,12 @@
       <version>${project.version}</version>
     </dependency>
 
+    <dependency>
```
rohanKanojia: Is it really necessary ? Would be nice if you could do this without introducing this dependency ;-)

rohanKanojia:
@rnetuka : Please rebase your branch also with latest master

llowinge:
Is related to https://github.com/fabric8io/kubernetes-client/issues/1505

openshift-client/pom.xml
rohanKanojia: @rnetuka : We use jackson.  Could you please update it? I'm planning to do a release today so it would be nice if it could go in.

rohanKanojia:
closing for #1659

PR #1662, opened 2019-07-26
Fix #1661: HasMetadata interface should be generic

### LINKED ISSUE #1661: HasMetadata interface should be generic
HasMetadata interface extends KuberenetesResource which is generic interface. It should be possible to provide generic type for HasMetadata also.

mgorniew:


rohanKanojia:
ok to test

mgorniew:
I'm trying to update generate tool for this change, but not sure how to do it.
Using checking _generate_ executable will produce different JSON, than currently present in _kube-schema.json_.
Building _generate_ from source give me tool which outputs schema with even more differences.
Is there any procedure how to build this to get same schema as current one?

PR #1666, opened 2019-07-26
Fix #1661: Remove generic parameter from KubernetesResource

### LINKED ISSUE #1661: HasMetadata interface should be generic

mgorniew:


rohanKanojia:
ok to test

gastaldi:
[merge]

Issue #1667, opened 2019-07-26
WatchConnectionManager's websocket Origin header uses port of -1 when no port specified


kjcjohnson:
For the WatchConnectionManager, the websocket request's Origin header has the port set to -1 if no port was explicitly specified in the URL.

So for a URL of https://example.com/k8s/clusters/etc, the origin header ends up as 

`Origin: https://example.com:-1` 

instead of just 

`Origin: https://example.com` 

like would be expected.

This seems to be causing an issue when connecting to a Kubernetes cluster running under Rancher 2.2.5 or higher, which is now checking the Origin header against the Host header. See the related issue rancher/rancher#21777 where this is failing.

The problem is here: 
https://github.com/fabric8io/kubernetes-client/blob/c175e1e7a1463cf3c13f71af4593efe6e683b58b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java#L147-L151
Calling `getPort()` on a `java.net.URL` object will return -1 if the port is not set, and thus we end up with a port of -1 in the origin header.

I added some extra logging statements to print out the actual request headers and caught this "live":
```
Jul 26, 2019 4:27:54 PM FINE io.fabric8.kubernetes.client.dsl.internal.WatchConnectionManager runWatch

Connecting websocket ... io.fabric8.kubernetes.client.dsl.internal.WatchConnectionManager@297470a6

Jul 26, 2019 4:27:54 PM FINE io.fabric8.kubernetes.client.dsl.internal.WatchConnectionManager runWatch

Watch request to https://rancher.example.com/k8s/clusters/c-scffx/api/v1/namespaces/test-ns/pods?fieldSelector=metadata.name%3Dtestpod-zrmh2&watch=true {

Jul 26, 2019 4:27:54 PM FINE io.fabric8.kubernetes.client.dsl.internal.WatchConnectionManager runWatch

Origin: https://rancher.example.com:-1

Jul 26, 2019 4:27:54 PM FINE io.fabric8.kubernetes.client.dsl.internal.WatchConnectionManager runWatch

} End watch request.
```

PR #1669, opened 2019-07-27
Check for port before setting Origin header for watch requests.

kjcjohnson:
When the Kubernetes URL used doesn't specify a port
(e.g., https://example.com/api/v1/...), the origin header for
watch requests ends up with a port of -1 (e.g., https://example.com:-1).
This happens because calling `getPort()` on a java.net.URL object that
does not have a port explicitly specified will always return -1. The
return value was always just inserted into the origin header.

Now, we check for this and only append a port to the origin header if
`getPort()` returns something other than -1. We make this change in both
the WatchConnectionManager and WatchHTTPManager classes.

I wasn't able to come up with a way to make a test for this, because the 
mock Kubernetes server support specifying the port, and especially not 
the defaults of 80/443. 

Fixes issue #1667

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ### CHANGELOG
 
 ### 4.3-SNAPSHOT
-  Bugs  
+  Bugs
   * Fixed user/password authentication bug in OpenShift 4
   * Fix #1667: Origin header for watch requests had a port of -1 when no port specified
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### 4.3-SNAPSHOT
   Bugs
+   * Fix #1667: Origin header for watch requests had a port of -1 when no port specified
 
   Improvements
    * Test coverage for PersistentVolumeClaim
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
@@ -144,10 +144,15 @@ public class WatchConnectionManager<T extends HasMetadata, L extends KubernetesR
 
     httpUrlBuilder.addQueryParameter("watch", "true");
 
+    String origin = requestUrl.getProtocol() + "://" + requestUrl.getHost();
+    if (requestUrl.getPort() != -1) {
+        origin += ":" + requestUrl.getPort();
+    }
+
     Request request = new Request.Builder()
       .get()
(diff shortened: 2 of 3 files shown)


gastaldi:
Ok to test

rohanKanojia:
ok to test

cjellick:
Do we need an origin header at all? That’s really just useful for protecting browsers from cross site scripting.

Could we simply drop the header all together?

rohanKanojia:
@kjcjohnson: Could you please test your patch without header? If it's not breaking anything then let's drop it.

kjcjohnson:
It looks like dropping the origin header won't work. I took some Wireshark captures of the packets, and they're identical other than the Origin header, but only the request with the header succeeds:

Without Origin header:
```
GET /k8s/clusters/c-scffx/api/v1/namespaces/test-ns/pods?fieldSelector=metadata.name%3Dtestpod-1sntr&watch=true HTTP/1.1
Upgrade: websocket
Connection: Upgrade
Sec-WebSocket-Key: 1Ass6sr+wPqYyEddMR7FPQ==
Sec-WebSocket-Version: 13
Authorization: Bearer <token>
Host: rancher.example.com
Accept-Encoding: gzip
User-Agent: okhttp/3.12.0

HTTP/1.1 403 Forbidden
X-Powered-By: ARR/3.0
X-Powered-By: ASP.NET
Date: Thu, 01 Aug 2019 19:00:43 GMT
Content-Length: 0
```

With Origin header:
```
GET /k8s/clusters/c-scffx/api/v1/namespaces/test-ns/pods?fieldSelector=metadata.name%3Dtestpod-rfw80&watch=true HTTP/1.1
Origin: http://rancher.example.com
Upgrade: websocket
Connection: Upgrade
Sec-WebSocket-Key: zmA6ry8uoV5uumoYhh8rCg==
Sec-WebSocket-Version: 13
Authorization: Bearer <token>
Host: rancher.example.com
Accept-Encoding: gzip
User-Agent: okhttp/3.12.0

HTTP/1.1 101 Switching Protocols
Upgrade: websocket
Sec-WebSocket-Accept: WcnYoYGI9IQ0piiOUMVaDwgV1OU=
X-Powered-By: ARR/3.0
Connection: Upgrade
X-Powered-By: ASP.NET
Date: Thu, 01 Aug 2019 19:25:45 GMT
```

CHANGELOG.md
```diff
@@ -1,8 +1,9 @@
 ### CHANGELOG
 
 ### 4.3-SNAPSHOT
-  Bugs
+  Bugs
```
gastaldi:
```suggestion
  Bugs
```

gastaldi:
[merge]

PR #1687, opened 2019-08-06
fix: Fixed URL resolution algorithm

### LINKED ISSUE #1684: REGRESSION: Processing of OpenShift Templates with apiVersion = v1 no longer works in kubernetes-client 4.4.0
Eg. the library is requesting the following URL when processing templates: https://dev.rdu2c.fabric8.io:8443/api/v1/namespaces/test-project-1565023015447/processedtemplates. 

In 4.3.1, it was 
https://dev.rdu2c.fabric8.io:8443/oapi/v1/namespaces/test-project-1565023127814/processedtemplates (notice the oapi in the path)

gastaldi:
Template and other openshift-related resource operations may use the /apis endpoint if and only if the API Group name is specified. If for some reason the API Group Name is not declared (for example, a Template object with apiVersion:v1), then the /oapi endpoint is used.

Fixes #1684

### DIFF
--- a/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/OpenShiftOperation.java
+++ b/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/OpenShiftOperation.java
@@ -63,18 +63,15 @@ public class OpenShiftOperation<T extends HasMetadata, L extends KubernetesResou
 
   @Override
   public URL getRootUrl() {
-    OpenShiftConfig config = OpenShiftConfig.wrap(context.getConfig());
-    OpenShiftClient oc = new DefaultOpenShiftClient(context.getClient(), config);
-    if (config.isOpenShiftAPIGroups(oc)) {
-      oc.close();
-      return super.getRootUrl();
-    } else {
-      oc.close();
+    // This is an OpenShift resource. If no API Group Name is specified, use /oapi endpoint
+    if (Utils.isNullOrEmpty(context.getApiGroupName())) {
       try {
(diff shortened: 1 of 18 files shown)


gastaldi:
Ok to test

rohanKanojia:
@gastaldi : Some tests seem to be failing. You might want to fix those

openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/OpenShiftOperation.java
```diff
@@ -63,18 +63,15 @@ public OpenShiftConfig getConfig() {
...
   public URL getRootUrl() {
-    OpenShiftConfig config = OpenShiftConfig.wrap(context.getConfig());
-    OpenShiftClient oc = new DefaultOpenShiftClient(context.getClient(), config);
-    if (config.isOpenShiftAPIGroups(oc)) {
-      oc.close();
-      return super.getRootUrl();
-    } else {
-      oc.close();
+    // This is an OpenShift resource. If no API Group Name is specified, use /oapi endpoint
+    if (Utils.isNullOrEmpty(context.getApiGroupName())) {
```
rohanKanojia: Thanks, seems fine to me. Have you tested this on both OS3 and OS4?

gastaldi: @rohanKanojia Yes, I tested in both OS3 and OS4 and it works great! :smile:

gastaldi:
[merge]

PR #1715, opened 2019-08-19
Fixed deserialize of `IntOrString` with correct `Kind` instead of `null`

ilylia:
when I use `kubernetes-client` to fetch some resource in kubernetes, especially for `Ingress` or `Service`, the port in that is with type `IntOrString`, and with field `Kind` value `null`. This bring on many troubles.


rohanKanojia:
ok to test

rohanKanojia:
[merge]

PR #1722, opened 2019-08-21
fixes portforwarding when there are more services with the same name in multiple namespaces

BJMg:
I encountered an incorrect port forwarding (on service) when I had more than one namespace. In each namespace there were services with the same name. When I called portForward on service it randomly chose one of these services instead of the service with current namespace.
This pull request should remedy this behaviour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### 4.4-SNAPSHOT
 #### Bugs
   * Fix #1706: admissionregistration resources are now parsed correctly
+  * Fix #1722: Service port forward are now done in the correct namespace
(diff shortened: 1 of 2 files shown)


rohanKanojia:
ok to test

rohanKanojia:
@BJMg : Could you please add a line to CHANGELOG regarding this?

BJMg:
> @BJMg : Could you please add a line to CHANGELOG regarding this?

Thanks for accepting. It is done.

rohanKanojia:
There is a minor conflict, Could you please resolve it? I think we are good to merge after that.

PR #1734, opened 2019-08-29
fix: Remove @Valid annotation from generated classes fields.

iocanel:
This is meant to address: https://github.com/quarkusio/quarkus/issues/3593

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 #### Improvements
   * Added in kubernetes-server-mock (CRUD) the withoutLabel filter and respective tests
+  * Removed @Valid annotation from all model class fields (improves quarkus integration).
 
 #### Dependency Upgrade
 
--- a/extensions/knative/model/pom.xml
+++ b/extensions/knative/model/pom.xml
@@ -117,7 +117,7 @@
                     <sourceDirectory>${basedir}/src/main/resources/schema</sourceDirectory>
                     <targetPackage>io.fabric8.knative.api.model</targetPackage>
                     <includeConstructors>true</includeConstructors>
-                    <includeJsr303Annotations>true</includeJsr303Annotations>
+                    <includeJsr303Annotations>false</includeJsr303Annotations>
                     <includeToString>false</includeToString>
                     <includeHashcodeAndEquals>false</includeHashcodeAndEquals>
                     <outputDirectory>${basedir}/target/generated-sources</outputDirectory>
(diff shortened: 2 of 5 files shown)


iocanel:
Just to say a few words about this change.
The idea is to limit the use of the `@Valid` inside the generated classes.
I am by no means expert on jsr303 but I feel that we don't need the annotation, since we maunually validate each Object (including nested) via builders.

So, I'd say that its pretty safe to remove the annotations added by the `jsonschema2pojo` plugin and just keep the onces added by our annoator.

gastaldi:
[test]

gastaldi:
Ok to test

gastaldi:
@iocanel could you please add an entry in the CHANGELOG? Thanks!

gsmet:
You need the annotations if you want Hibernate Validator to cascade automatically the validation to the properties being beans.

If you are doing that manually, you certainly don't need them.

gsmet:
Note that I'm not yet sure it will be sufficient to fix the Quarkus issue we have but it will certainly help. I'll try to follow up on this once this is out of the way.

iocanel:
@gsmet: I am not sure if we need to `automatically cascade the validation to nested beans`. The reason being that the builders will build/validate the whole object tree recursively each time the `build` method is called. Of course, this means that validation will only work when Objects are manipulated via builders, but TBH I don't think it was ever intended these objects to be manipulated otherwise. WDYT?

gsmet:
Typically, if you have:
```
public class MyFirstBean {

    @Valid
    private MySecondBean bean;
}

public class MySecondBean {

    @NotNull
    private String property;
}
```
and you validate an instance of `MyFirstBean`, the `bean` property will also be validated, the validation will cascade and `MySecondBean#property` will be validated and the `@NotNull` constraint be enforced.

If you remove the `@Valid`, when you validate `MyFirstBean`, it won't cascade to the `MySecondBean` object.

Obviously, if you validate a `MySecondBean` instance directly, it will cascade.

iocanel:
@gsmet: I get what you mean.

In our case if you have classes like those:

    @Buildable
    public class MyFirstBean {
        private MySecondBean bean;
    }

    @Buildable
    public class MySecondBean {
        @NotNull
        private String property;
    }

An oversimplified view of the generated builder will look like:

    public class MyFristBeanBuilder {
        private MySeconBuilderBuilder seconBeanBuilder; 

        public MyFirstBean build()  {
              MyFristBean bean = new MyFirstBean(secondBeanBuilder.build());
              validate(bean);
              return bean;

        }
    }
 
So during the `build()` all nested objects will also get built and validated.

Again, this only covers the builder path, but I think we can live with that.

geoand:
Is there anything that you would like me to test/check @iocanel , @gsmet ?

rohanKanojia:
[merge]

gsmet:
Just a heads up that we will be releasing a new Quarkus version on Thursday so if you can get this released by tomorrow evening, we will be able to include it. If not, it will be in a later one.

/cc @gastaldi

PR #1736, opened 2019-08-30
Fix #1674: Cascade delete can't be overriden

### LINKED ISSUE #1674: Cascade delete can't be overriden
Version: `4.3.1`

@rohanKanojia
It seems that since this commit https://github.com/fabric8io/kubernetes-client/pull/1620/commits/008d3789d6020b032038e2db415bc334ab83dace was merged and released in `4.3.1`. The following code is always performing cascade delete, even though it shouldn't as I'm overriding the property.

I haven't found the way not to perform cascade delete as I wanted to workaround [my previous issue](https://github.com/fabric8io/kubernetes-client/issues/1673).

```java
for (HasMetadata hasMetadata : listRemovableResources()) {
  resource(hasMetadata).withGracePeriod(0).cascading(false).delete();
}

// resource method is from DefaultOpenShiftClient::resource
@Override
public NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicable<HasMetadata, Boolean> resource(HasMetadata item) {
  return delegate.resource(item);
}
```

rohanKanojia:
Fix #1674 

@dsimansk : I've tested this on the following scenarios

1.
```
      client.resource(aResource).withGracePeriod(0).cascading(true).delete();
```
Cascade deletion is performed since it's explicitly specified

2.
```
      client.resource(nginxDeployment).withGracePeriod(0).cascading(false).delete();
```
Cascade deletion is not performed 
3.
```
      client.resource(nginxDeployment).withGracePeriod(0).delete();
```
Cascade deletion is performed since it's default behavior in `kubectl ` also

4.
```
      client.apps().deployments().inNamespace("default").withName("nginx").cascading(false).delete();
```
Cascade deletion is not performed since it's overriden by dsl

5.
```
      client.apps().deployments().inNamespace("default").withName("nginx").delete();
```
Cascade deletion is performed since it's like that in `kubectl` also

6.
```
      client.apps().deployments().inNamespace("default").withName("nginx").cascading(true).delete();
```
Cascade deletion is perfomed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### 4.4-SNAPSHOT
 #### Bugs
   * Fix #1730: Fix failing build on jdk11
+  * Fix #1634: Cascade delete can't be overriden
 
 #### Improvements
 
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/DefaultKubernetesClient.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/DefaultKubernetesClient.java
@@ -112,13 +112,13 @@ public class DefaultKubernetesClient extends BaseClient implements NamespacedKub
 
   @Override
   public ParameterNamespaceListVisitFromServerGetDeleteRecreateWaitApplicable<HasMetadata, Boolean> load(InputStream is) {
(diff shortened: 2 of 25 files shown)


rohanKanojia:
@dsimansk : Would be awesome if you could find time to test this(just to stay on the safe side so that I don't introduce yet another regression)

rohanKanojia:
[merge]

Issue #1708, opened 2019-08-15
Websocket connection issue


stanlyDoge:
I have upgraded from 4.3.0 to 4.4.1 but my operator is failing wit these exceptions: https://pastebin.com/7qAwaavp

I have laso tried 4.4.0 with the same result.
Any idea what went wrong?

```
oc version
oc v3.11.82
kubernetes v1.11.0+d4cacc0
features: Basic-Auth GSSAPI Kerberos SPNEGO

Server https://127.0.0.1:8443
kubernetes v1.11.0+d4cacc0
```

rohanKanojia:
which openshift version are you using?

rohanKanojia:
Could you please provide some test code which seems to be failing on this new release?

stanlyDoge:
I will take a look whether I can, but I may be too complicated to get the malfunctioning part.

stanlyDoge:
I did take a look again at this and there is little bit more of desciption.

So when my operator starts, fabric8 does open a websocket
```
WatchConnectionManager:158 - Request{method=GET, url=https://172.30.0.1/apis/mycrd/v1beta1/namespaces/myproject/crdss?watch=true , tags={}}
WatchConnectionManager:166 - WebSocket successfully opened
```
but each time I want to pull data about crd (e.g. `client.services().inNamespace(namespace).withName("service").get()`)
the socket connection fails with  
```
WatchConnectionManager:304 - Submitting reconnect task to the executor
WatchConnectionManager:320 - Scheduling reconnect task
WatchConnectionManager:114 - Connecting websocket ... io.fabric8.kubernetes.client.dsl.internal.WatchConnectionManager@63ca5d90
WatchConnectionManager:158 -  Request{method=GET, url=https://172.30.0.1/apis/mycrd/v1beta1/namespaces/myproject/crdss?watch=true , tags={}}
WatchConnectionManager:176 - onFailure - response is null
WatchConnectionManager:114 - Connecting websocket ... io.fabric8.kubernetes.client.dsl.internal.WatchConnectionManager@33603116
2019-09-05 12:07:06 WARN  WatchConnectionManager:209 - Exec Failure
java.io.InterruptedIOException: executor rejected
```

Might this be an issue with reusing a websocket? Not much changed there since 4.3.0 where everything wokrs perfect.

rohanKanojia:
Hmm, I see. Would try to reproduce your error using this.

rohanKanojia:
[...]

PR #1761, opened 2019-09-14
fix (#1759): Avoid closing the client on each openshift request.

### LINKED ISSUE #1759: Portforwarding is broken
It fails consistnently in openshift 3.x, with an error like:

    Caused by: java.util.concurrent.RejectedExecutionException: Task io.dekorate.deps.okhttp3.RealCall$AsyncCall@5e20ab3f rejected from java.util.concurrent.ThreadPoolExecutor@4846507f[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 0]

Noticed that when I tried to upgrade decorate to 4.5.1.

Used git bisect and found that the offending commit is: 6c20171e4b4ba21d8aa30fb711490dffe207c682

Need to fix the issue introduced in this commit and add test to prevent this from happening again.

iocanel:
Fix #1759 #1762 #1708

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/KubernetesClient.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/KubernetesClient.java
@@ -166,4 +166,5 @@ public interface KubernetesClient extends Client {
 
   SubjectAccessReviewDSL subjectAccessReviewAuth();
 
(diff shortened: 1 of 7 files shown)


walnut-tom:
Tested on Openshift 3.9, 3.11.16 and 4.0

PR #1784, opened 2019-09-26
Fix #1724: createOrReplace function does not work properly for custom defined resources

### LINKED ISSUE #1724: createOrReplace function does not work properly for custom defined resources
I'm trying to use createOrReplace function for some CDR but it returns HTTP 422 (Unprocessed entity) on second call. This happens because resource version is not provided in metadata: "mycrd.charts.helm.k8s.io "test" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update". 

Probably it is releated to https://github.com/kubernetes/kubernetes/issues/70674

rohanKanojia:
Fix #1724

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### 4.6-SNAPSHOT
 #### Bugs
+* Fix #1724: createOrReplace function does not work properly for Custom defined resources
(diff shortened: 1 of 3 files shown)


rohanKanojia:
[merge]

PR #1788, opened 2019-09-27
feat: Bump knative and add support for multiple serving values.

iocanel:
It seems that there are multiple knative version and we only support one.


rohanKanojia:
[merge]

rohanKanojia:
[merge]

Issue #1776, opened 2019-09-20
Tekton Extension "nodeSelector" 


stephen-harlow:
I'm currently trying to use the Tekton extension code examples [(something like this)](https://github.com/fabric8io/kubernetes-client/blob/master/extensions/tekton/examples/src/main/java/io/fabric8/tekton/api/examples/PipelineResourceCreate.java), but I'm running into some issues. When I try to recreate the pipeline example from [here](https://github.com/tektoncd/pipeline/blob/master/docs/tutorial.md#task-inputs-and-outputs), all of the code works as expected, **except** for the last TaskRun. I can create the PipelineResources and Tasks. But when I try to execute that section that creates the TaskRuns, I get this as a response:

 ```Failure executing: POST at: https://kubernetes/apis/tekton.dev/v1alpha1/namespaces/default/taskruns. Message: Internal error occurred: admission webhook "webhook.tekton.dev" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "nodeSelector". Received status: Status(apiVersion=v1, code=500, details=StatusDetails(causes=[StatusCause(field=null, message=admission webhook "webhook.tekton.dev" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "nodeSelector", reason=null, additionalProperties={})], group=null, kind=null, name=null, retryAfterSeconds=null, uid=null, additionalProperties={}), kind=Status, message=Internal error occurred: admission webhook "webhook.tekton.dev" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "nodeSelector", metadata=ListMeta(_continue=null, resourceVersion=null, selfLink=null, additionalProperties={}), reason=InternalError, status=Failure, additionalProperties={}).```

I know that the other sections are being created as designed, since when I manually do a ```kubectl apply -f <YAML.yaml>``` (where YAML.yaml is a yaml file with just the TaskRun), the pod appears in the Kubernetes Service I'm using. 

**Dependency Section of pom.xml Involved**
```xml
<dependency>
	<groupId>io.fabric8</groupId>
	<artifactId>tekton-client</artifactId>
	<version>4.4.0</version>
</dependency>
<dependency>
	<groupId>io.fabric8</groupId>
	<artifactId>kubernetes-client</artifactId>
	<version>4.4.0</version>
</dependency>
<dependency>
	<groupId>io.fabric8</groupId>
	<artifactId>kubernetes-model</artifactId>
	<version>4.4.0</version>
[...]

PR #1791, opened 2019-09-30
chore (#1776): Bump tekton to v0.7.0.

iocanel:
Let's bump tekton to v0.7.0 so that we can address this: #1776


rohanKanojia:
[merge]

PR #1798, opened 2019-10-05
Check if BouncyCastle provider is set

torrescd:
bcpkix-jdk15on have only 400kb instead of 4.5MB from bcprov-ext-jdk15on

With this check we can take advantage of a BouncyCastleProvider that is already set (for example by the jenkins plugin bouncycastle-api)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### 4.6-SNAPSHOT
 #### Bugs
+* Fix #1796: Check if BouncyCastle provider is set
 * Fix #1724: createOrReplace function does not work properly for Custom defined resources
 * Fix #1789: Create or replace on operation seems broken
 * Fix #1782: Informer Deadlock; Fix lock typo in SharedProcessor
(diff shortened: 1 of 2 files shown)


rohanKanojia:
ok to test

kubernetes-client/pom.xml
```diff
@@ -209,7 +209,6 @@
     </dependency>
     <dependency>
       <groupId>org.bouncycastle</groupId>
-      <optional>true</optional>
```
gastaldi: Not sure we want to make this a required dependency. `bcprov-jdk15on` is resolved as a transitive dependency through this too

torrescd: You are right. I have revert both pom modifications.

torrescd:
> LGTM, can you squash both commits into a single one? Thanks!

Done!

gastaldi:
Ah almost forgot, one last thing: Can you add an entry to the CHANGELOG.md file?

torrescd:
> Ah almost forgot, one last thing: Can you add an entry to the CHANGELOG.md file with this change?

Done!

rohanKanojia:
[merge]

PR #1799, opened 2019-10-07
delete duplicated readme.md

honnix:
There is another file named as `README.md` and that seems to be more up-to-date. Having both files in a case insensitive file system confuses git.


rohanKanojia:
ok to test

oscerd:
[merge]

Issue #1529, opened 2019-05-16
watch should reconnect on status code 410 with new resource version


Udbhav30:
kubernetes-client/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
`if (status.getCode() == HTTP_GONE) { 
   webSocketRef.set(null); // lose the ref: closing in close() would only generate a Broken pipe 
                           // exception 
   // shut down executor, etc. 
   closeEvent(new KubernetesClientException(status)); 
   close(); 
   return; 
} `

In above particular scenario we should re-trigger the watcher which i cross checked is done in go-client.

https://github.com/fabric8io/kubernetes-client/blob/12fa0993793baa9ba1835c36f2fa696bd739a1b9/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java#L681

This issue is also observed in :
https://issues.apache.org/jira/browse/SPARK-24266

Udbhav30:
cc @foxish

PR #1800, opened 2019-10-07
schedule reconnect in case of HTTP_GONE

honnix:
Reset resource version to null and schedule reconnect in case of HTTP_GHONE

This tries to address https://github.com/fabric8io/kubernetes-client/issues/1529

### DIFF
--- a/kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/WatchOverHTTP.java
+++ b/kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/WatchOverHTTP.java
@@ -16,6 +16,8 @@
 
 package io.fabric8.kubernetes.client.mock;
 
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
 import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.api.model.PodBuilder;
 import io.fabric8.kubernetes.api.model.Status;
@@ -27,18 +29,15 @@ import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.Watch;
 import io.fabric8.kubernetes.client.Watcher;
 import io.fabric8.kubernetes.client.server.mock.KubernetesServer;
+import java.net.HttpURLConnection;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import junit.framework.AssertionFailedError;
 import org.junit.Rule;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.migrationsupport.rules.EnableRuleMigrationSupport;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import java.net.HttpURLConnection;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.TimeUnit;
-
-import static org.junit.jupiter.api.Assertions.assertTrue;
 
 @EnableRuleMigrationSupport
 public class WatchOverHTTP {
@@ -50,6 +49,7 @@ public class WatchOverHTTP {
     .build();
   static final WatchEvent outdatedEvent = new WatchEventBuilder().withStatusObject(outdatedStatus).build();
   final String path = "/api/v1/namespaces/test/pods?fieldSelector=metadata.name%3Dpod1&resourceVersion=1&watch=true";
+  final String pathWithoutResourceVersion = "/api/v1/namespaces/test/pods?fieldSelector=metadata.name%3Dpod1&watch=true";
   @Rule
   public KubernetesServer server = new KubernetesServer(false);
   Logger logger = LoggerFactory.getLogger(WatchTest.class);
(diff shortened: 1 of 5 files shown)


honnix:
Couldn't figure out how this could fail `PATCH` request. 🤔

rohanKanojia:
ok to test

rohanKanojia:
@honnix : Not your fault. Our tests are flaky.

honnix:
I seems I should fix these test cases as they look very relevant.

rohanKanojia:
I meant integration tests, not unit tests

honnix:
@rohanKanojia Yeah I'm working on it. Not very familiar with the code base, so I didn't find the test case.

kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/WatchTest.java
```diff
@@ -58,38 +58,92 @@
...
+
+    final String path = "/api/v1/namespaces/test/pods?fieldSelector=metadata.name%3Dpod1&resourceVersion=1&watch=true";
+    final String pathWithoutResourceVersion = "/api/v1/namespaces/test/pods?fieldSelector=metadata.name%3Dpod1&watch=true";
+
+    server.expect()
+        .withPath(path)
+        .andReturn(200, new StatusBuilder().withCode(200).build()).once();
+    server.expect()
+        .withPath(path)
+        .andReturn(200, mapper.writeValueAsString(outdatedEvent) + '\n').once();
```
honnix: I couldn't figure out a better way to do this. The mock server requires a strict line with line feed.

honnix: Also I'm not entirely sure HTTP code 200 with 410 in the body is really what k8s returns.

honnix:
Now this makes me wondering whether watching from the beginning of history is a bad idea. Can someone point me how `resourceVersion` is set initially before creating a `WatchConnectionManager`?

kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/WatchOverHTTP.java
```diff
@@ -92,55 +92,38 @@ public void onClose(KubernetesClientException cause) {}
   }
 
   @Test
-  public void testOutdated() throws InterruptedException {
-    logger.info("testOutdated");
+  public void testHttpErrorReconnectOutdatedAndModified() throws InterruptedException {
```
honnix: This file is not named properly and doesn't seem to be picked up by junit.

honnix:
Maybe it is ok because when this happens, old resource versions got wiped out from history.

honnix:
@rohanKanojia PTAL again. Thanks.

rohanKanojia:
@honnix : Could you please add a line to CHANGELOG regarding this change?

honnix:
@rohanKanojia Sure. I saw that label but didn't realise I should update it. Will do.

rohanKanojia:
[merge]

wiggzz:
This feels like it breaks the contract with the watcher. 

If we get an HTTP_GONE from Kubernetes, then we need to notify the watcher we are going to start again from the beginning - otherwise deletion events can be missed. 

Think about the case where the user of this library has some internal cache of the resources being watched. If we start again from the beginning, this means k8s will send us everything that currently exists, but may not send us deletion events from things that used to exist. 

For instance: we perform a list on all the resources in the cluster and get some current resource version "1234". Then we wait some period of time (say, an hour). In this time, some resources are deleted. Then we ask to see any changes since that point in time by specifying a resource version "1234" with our watch request. If this is too old (and k8s cannot track back that far), then this change means that transparently the watcher will just send us ADDED events for all the resources currently in the cluster, but the deletion events that happened since resource version "1234" will not be sent. This will mean if we were trying to maintain some local cache, the resources that were deleted will be leaked.

If you are going to do this, you'll need to notify the caller to delete everything in their cache because k8s can no longer send all the changes from the resource version they specified.

honnix:
@wiggzz Thanks for the explanation and it is a very good point. I had similar worries when working on this PR because as you pointed out it does break the contract.

I'm fine reverting this change.

honnix:
@wiggzz PTAL https://github.com/fabric8io/kubernetes-client/pull/1853

wiggzz:
Thanks - appreciate the quick turn around.

honnix:
Likewise, thanks for pointing out the defect.

rohanKanojia:
@honnix @wiggzz : Thanks guys, I will cut a release soon this week and let your guys know

xiateapu:
I'm a novice, so.... could you please tell me when or why kubernetes api returns HTTP_GONE?

wiggzz:
> I'm a novice, so.... could you please tell me when or why kubernetes api returns HTTP_GONE?

In the context of a watch, it will return `HTTP_GONE` when you ask to see changes for a `resourceVersion` that is too old - i.e. when it can no longer tell you what has changed since that version, since too many things have changed. In that case, you'll need to start again, by not specifying a `resourceVersion` in which case the watch will send you the current state of the thing you are watching and then send updates from that point.

rohanKanojia:
@xiateapu : ^^

xiateapu:
Actually i had this problem as shown below:  
image

It seems that `resourceVersion` has a big gap between last watched and current ?   
So i'm very confused that why dose the `too many things have changed` happened？Can‘t kubernetes tell me these change in time?

xiateapu:
> > I'm a novice, so.... could you please tell me when or why kubernetes api returns HTTP_GONE?
> 
> In the context of a watch, it will return `HTTP_GONE` when you ask to see changes for a `resourceVersion` that is too old - i.e. when it can no longer tell you what has changed since that version, since too many things have changed. In that case, you'll need to start again, by not specifying a `resourceVersion` in which case the watch will send you the current state of the thing you are watching and then send updates from that point.

Thanks for your answer ^ ^

xiateapu:
> @xiateapu : ^^

^ ^

wind57:
> > I'm a novice, so.... could you please tell me when or why kubernetes api returns HTTP_GONE?
> 
> In the context of a watch, it will return `HTTP_GONE` when you ask to see changes for a `resourceVersion` that is too old - i.e. when it can no longer tell you what has changed since that version, since too many things have changed. In that case, you'll need to start again, by not specifying a `resourceVersion` in which case the watch will send you the current state of the thing you are watching and then send updates from that point.

what about an informer? `ResourceEventHandler` has a default method `onNothing`. Its documentation says the same thing : 

> Called after an empty list is retrieved on start or after an HTTP GONE when the Store is empty

I am asking this in the context of spring-cloud-kubernetes project. Previously, we had used Watches for detecting changes in config maps/secrets. I am trying to move to informers now. So our code is like this at the moment:

```
					@Override
					public void onClose(WatcherException exception) {
						log.warn("ConfigMaps watch closed", exception);
						Optional.ofNullable(exception).map(e -> {
							log.debug("Exception received during watch", e);
							return exception.asClientException();
						}).map(KubernetesClientException::getStatus).map(Status::getCode)
								.filter(c -> c.equals(HttpURLConnection.HTTP_GONE)).ifPresent(c -> watch());
					}
``` 

If I move this to informers, it seems that I should handle this in the same manner, something like:

```
				@Override
				public void onNothing() {
					boolean isStoreEmpty = informer.getStore().list().isEmpty();
					if(!isStoreEmpty) {
						// HTTP_GONE, thus re-inform
						inform();
					}
				}
```

I am a bit lost (due to my limited knowledge) between this should be handled, or an informer (like a watcher) will know what to do on its own.

Can some of the pros comment on this please? thank you.

manusa:
Hi @wind57 
There have been a few issues on this repo about Spring Cloud Kubernetes and the HTTP Gone problem. With more recent versions of the client, this should be less of a problem since watchers are started using [bookmarks](https://kubernetes.io/docs/reference/using-api/api-concepts/#watch-bookmarks) (if the cluster supports it).

In any case, [as suggested](https://github.com/fabric8io/kubernetes-client/issues/2787#issuecomment-1046768554) in some of the reported issues, Spring Cloud should migrate from using raw Watchers to Informers.

Informers are self-healing, so as a client consumer you should not worry about reconnecting, and so on. You can use the `onNothing` event to log or monitor these incidents, but you shouldn't do any error-recovery stuff.

For your use-case, you can consider Informers as improved Watchers which take care of all the stuff for you.

wind57:
thank you @manusa. this is exactly what I am trying to do currently, move to informers, it's [almost ready](https://github.com/spring-cloud/spring-cloud-kubernetes/pull/1036/files#diff-8127d11528a34e5857472226e2ea61b5f90eb00b4038247cf428c30d429ae691) For the time being I will comment `onNothing` and see if we get any reports in the future.

As usual, much appreciate your answer.

PR #1802, opened 2019-10-09
kubernetes-client

cmadunezim:
cisco-test / master


rohanKanojia:
@cmadunezim : Could you please elaborate the use case you're trying to solve with this PR?

PR #1807, opened 2019-10-11
Fix #1775: KubernetesList.list().delete(resources) orphanDependents semantics change between 4.1.3 and 4.5.2

### LINKED ISSUE #1775: KubernetesList.list().delete(resources) orphanDependents semantics change between 4.1.3 and 4.5.2
We upgraded our application from 4.1.3 to 4.5.2 and notice a behavioural change in the API.

In 4.1.3 `KubernetesList.list().delete(resources)` sent the following for each resource:

`{"apiVersion":"v1","kind":"DeleteOptions","orphanDependents":false} `

whereas in 4.5.2 the same code results in:

`{"apiVersion":"v1","kind":"DeleteOptions","gracePeriodSeconds":0,"orphanDependents":true}`

so the semantics of the delete are now different.

We restored the old behaviour by changing application as follows:

`resources.getItems().forEach(resource -> client.resource(resource).cascading(true).delete());`

I see the note in the CHANGELOG about the propagation policy feature, but nothing alerts a user to the API change.  Was this change intentional?

rohanKanojia:
Fix #1775

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 #### Bugs
 * Fix #1796: Check if BouncyCastle provider is set
 * Fix #1724: createOrReplace function does not work properly for Custom defined resources
+* Fix #1775: KubernetesList.list().delete(resources) orphanDependents semantics change between 4.1.3 and 4.5.2
 * Fix #1789: Create or replace on operation seems broken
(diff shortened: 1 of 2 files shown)


rohanKanojia:
[merge]

PR #1814, opened 2019-10-16
chore(deps): use Jackson BOM 2.9.10

jorsol:
Allow the use of the Jackson BOM to define the versions of the dependencies of Jackson and clean up the definitions of the dependencies in child poms.

This also updates Jackson to 2.9.10 which is backward compatible.


rohanKanojia:
ok to test

rohanKanojia:
[merge]

PR #1817, opened 2019-10-16
Avoid spurious print statement before invoking ResourceEventHandler.onAdd()

lalithsuresh:
Users of the informer API shouldn't be seeing these debug statements. Since it isn't added to any other type of notification, I've just removed it for now.


rohanKanojia:
ok to test

rohanKanojia:
[merge]

PR #1827, opened 2019-10-23
Fix fluent dsl

andreaTP:
The return type of `withGracePeriod` and of `withPropagationPolicy` prevent additional composition (and make the two settings mutually exclusive ATM).

The `gracePeriod` default to `0L` using the empty constructor of `OperationContext` which is possibly harmful (e.g. it will trigger a force kill).


rohanKanojia:
ok to test

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
```diff
@@ -922,13 +922,13 @@ public OperationInfo forOperationType(String type) {
   }
 
   @Override
-  public Deletable<Boolean> withGracePeriod(long gracePeriodSeconds)
+  public FilterWatchListDeletable<T, L, Boolean, Watch, Watcher<T>> withGracePeriod(long gracePeriodSeconds)
```
iocanel: Since, it works without breaking anything I am fine with it.

I don't see however, how it solves the two methods being mutually exclusive.

andreaTP: `withGracePeriod` now return a [Deletable](https://static.javadoc.io/io.fabric8/kubernetes-client/4.6.1/io/fabric8/kubernetes/client/dsl/Deletable.html) and on it you can only call `delete`, you need to do an unsafe cast to be able to chain further methods (i.e. such as `withPropagationPolicy`)

andreaTP: added a "demonstration" test for it

andreaTP:
added changelog and tests for the changes.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationContext.java
```diff
@@ -39,7 +39,8 @@
   protected boolean cascading;
   protected boolean reloadingFromServer;
 
-  protected long gracePeriodSeconds;
+  // Default to k8s 30s value: https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods
+  protected long gracePeriodSeconds = 30L;
```
scholzj: I run into this issue today when I found out that when deleting pods the termination grace period in my pods is always overwritten to `0` and the pods are basically force deleted. 

However, I'm wondering if this is the right fix. It still overwrites whatever the user configured in his pod which is IMHO wrong. I think the default should be -1 which will mean that the gracePeriod is not set in the request and the default from the resource will be used by Kubernetes => that will be either 30 seconds if the user didn't specified anything or whatever the user specified in his pods.

rohanKanojia:
[merge]

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationContext.java
andreaTP: haven't had time to test the `grace period` to `-1`, but, if it works, I see your concern and I agree on the approach.

andreaTP: this PR got already merged, do you mind to follow up?

scholzj: I opened #1833 to change the default grace period value.

PR #1830, opened 2019-10-25
Pass Locale.ROOT to String.toLowerCase/toUpperCase

jglick:
…rather than the no-argument overloads. Otherwise the novice Turkish/Azeri/etc. Java developer who has not yet set their system locale to English will be plagued with RBAC rejections of the attempt to list `ıngresses` and the like, due to the terrible behavior of these Java Platform methods which should have been deprecated long ago. (On the rare occasions that you wanted the no-arg behavior, be explicit and pass `Locale.getDefault()`.)


rohanKanojia:
ok to test

rohanKanojia:
@jglick: Build seems to be failing. Could you please address these failures whenever you get time?

rohanKanojia:
[merge]

PR #1831, opened 2019-10-25
Cleanup log4j dependency from project

jorsol:
The SLF4J is used as logging dependency, log4j is used only on tests, so this replaces the log4j declaration for slf4j-simple in the test scope.

This cleanup helps to reduce the dependencies used in the project and keep a single implementation in place.

This also removes the unnecessary dependency on jul-to-slf4j, which ends up being transitively included in projects that use kubernetes-client.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ buildNumber.properties
 .project
 .settings/
 .classpath
+.factorypath
 
 ### Java template
 *.class
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Fix #1828: VersionInfo date parsing of year
 
 #### Improvements
+* Cleanup log4j dependency from project
 
 #### Dependency Upgrade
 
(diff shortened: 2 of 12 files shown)


rohanKanojia:
ok to test

gastaldi:
Good job in removing the dependency on Log4J. IMHO the `slf4j-simple` dependency should be marked as test scope and it should be up to the consumer to decide which implementation to use or you may end up with warnings when using it (or being forced to use exclusions).

jorsol:
> Good job in removing the dependency on Log4J. IMHO the `slf4j-simple` dependency should be marked as test scope and it should be up to the consumer to decide which implementation to use or you may end up with warnings when using it (or being forced to use exclusions).

Yes, it's correct  `slf4j-simple` dependency is marked as test scope, I completely agree that the consumer should decide the implementation to use.

gastaldi:
Goticha, I haven't noticed that the dependencies without test scope are test projects :)

Can you add an entry in `CHANGELOG.md` with this change and rebase to a single commit?

jorsol:
Done @gastaldi, include `CHANGELOG.md` entry and rebased.

gastaldi:
Perfect, +1 to merge this after CI is green. Good job @jorsol !

gastaldi:
[merge]

PR #1832, opened 2019-10-27
chore: refactor poms and update some dependencies

jorsol:
This started as a quick cleanup of some `pom.xml` that have duplicate declarations and update some dependencies and plugins across the whole project, but it got out of control when started to refactor to the recommended order of maven declarations that end up formatting the style of the poms.

This can be really hard to review since many changes are simply related to reordering items and change indentation.

I hopefully believe that the end result is cleaner, but fell free to close and ignore this if there is no time to review. (Crossing fingers to not break anything)


rohanKanojia:
ok to test

rohanKanojia:
[merge]

kubernetes-itests/pom.xml
```diff
@@ -17,44 +17,37 @@
...
-  <url>http://maven.apache.org</url>
-
-  <properties>
-    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <junit.version>4.12</junit.version>
-  </properties>
+  <name>Fabric8 :: Kubernetes :: Regression :: Tests</name>
 
   <dependencies>
     <dependency>
```
rohanKanojia: I know this change was not intentional but it has disabled integration tests running on CircleCI. Arquillian is not compatible with Junit 5 yet.

jorsol: Actually it was intentional, the `junit-jupiter-migrationsupport` transitively has the Junit 4.12 dependency, so probably I broke something else since Junit 4 is still there.

I will try to find some time to look at this, but I have no experience with arquillian.

rohanKanojia: no problem. I've reverted this change in a separate PR for this module.

jorsol: I see the issue now, yes this change effectively disabled integration test from Arquillian :disappointed:

PR #1833, opened 2019-10-27
Respect the termination grace period from the Kubernetes resource as a default value

scholzj:
As discussed in #1827, this PR sets the default grace period to `-1` seconds as default. That should make the client not specify any grace period when deleting resources and this respect the grace period configured in the resource it self (e.g. `terminationGracePeriod` field in PodSpec) or the Kubernetes default which is IMHO the most natural default behavior. When you test it for example with Pod, you should see that:
* If the Pod specifies its own grace period it is respected by default
* If the Pod doesn't specify any grace period, the Kubernetes default 30 seconds are used
* If the client specifies the grace period using `withGracePeriod(...)`, the value from client is used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### 4.7-SNAPSHOT
 #### Bugs
+* Fix #1833: Respect the termination grace period from the Kubernetes resource by default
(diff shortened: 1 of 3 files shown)


rohanKanojia:
ok to test

rohanKanojia:
[merge]

PR #1863, opened 2019-11-09
Fixes #744

### LINKED ISSUE #744: Minor API issue: Watcher<T> should probably be Watcher<T extends HasMetadata>
Hello; thanks for your work on this project.

It seems to me that the `Watcher` class should not take an undifferentiated type parameter but should instead use `T extends HasMetadata`.  (Creating, say, a `Watcher<Integer>` is currently permitted but we all know such a thing would never be called.)

I base this off the [only location I could find in the source base where the `eventReceived` method is called](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java#L207-L217), and the payload is forced into being a subclass of `HasMetadata`.

kunal-kushwaha:
Fixes #744

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/ResourceHandler.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/ResourceHandler.java
@@ -20,10 +20,12 @@ package io.fabric8.kubernetes.client;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Predicate;
+
+import io.fabric8.kubernetes.api.model.HasMetadata;
 import okhttp3.OkHttpClient;
 import io.fabric8.kubernetes.api.builder.VisitableBuilder;
 
-public interface ResourceHandler<T, V extends VisitableBuilder<T, V>> {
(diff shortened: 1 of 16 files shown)


rohanKanojia:
@kunal-kushwaha : build seems to be failing, Could you please address those build failures whenever you get time?

kunal-kushwaha:
There have been uses of Watcher<String>, String is an an undifferentiated type parameter. How do I replace it when it doesn't implement HasMetadata?

rohanKanojia:
Apologies for the late reply, I think uses of Watcher with String parameters were added by me in order to handle raw CustomResource events where we don't have POJOs to deserialize. 

Maybe we can introduce another interface similar to the current `Watcher` called `RawWatcher` which would handle those raw String-based events??

kunal-kushwaha:
@rohanKanojia: Ah been a long time indeed. Will work on it when I get some time free from current schedule, with college and internships 😅

rohanKanojia:
No rush, take your time :-)

kunal-kushwaha:
Didn't get time to look into this :(

PR #1866, opened 2019-11-14
Fix Disabled integration tests.

rohanKanojia:
During his testing @manusa found out that integration tests are
not getting executed during their run on CircleCI. The reason was due
to incompatibility of Junit 5 with Arquillian. I've reverted to Junit 4
in kubernetes-itests/ module for now.

We can update this whenever this gets resolved:
    https://github.com/arquillian/arquillian-core/issues/137

### DIFF
--- a/kubernetes-itests/pom.xml
+++ b/kubernetes-itests/pom.xml
@@ -31,14 +31,10 @@
 
   <dependencies>
     <dependency>
-      <groupId>org.junit.jupiter</groupId>
-      <artifactId>junit-jupiter-engine</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.junit.jupiter</groupId>
-      <artifactId>junit-jupiter-migrationsupport</artifactId>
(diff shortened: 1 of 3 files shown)


kubernetes-itests/pom.xml
```diff
@@ -31,14 +31,10 @@
...
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.junit.jupiter</groupId>
-      <artifactId>junit-jupiter-migrationsupport</artifactId>
-      <scope>test</scope>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <!-- Arquillian is still not compatible with Junit 5, so using Junit 4 here -->
+      <version>4.12</version>
```
manusa: Could we use a property in the root pom.xml instead of a "hardcoded" value instead? e.g. `${junit.compatible-with-arquillian.version}`

rohanKanojia:
[merge]

PR #1867, opened 2019-11-14
Fix #1041: Support cascading delete on custom resources

### LINKED ISSUE #1041: Support cascading delete on custom resources
The interface returned by `KubernetesClient.customResources(...)` doesn't provide a method to set `cascading=true`.

The object itself (an instance of `CustomResourcesOperationsImpl`) does have a `cascading` flag, and a constructor that allows it to be specified, but it's defaulted to `false` in the only constructor that's actually used.

As a work around, you can construct an instance yourself using the constructor that does have the relevant parameter, but it would be helpful if the interface had a method for setting it, as other resource types do.  It's worth noting that while technically the object does have a method for setting `cascading=true` that it inherits from the base class, casting to the base class and calling the method explodes because it uses reflection to find a suitable constructor and no such constructor exists in this case.

In particular, this makes it difficult to delete [Argo](https://github.com/argoproj/argo) workflows and their underlying pods.

rohanKanojia:
Fix #1041

### DIFF
--- a/kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/CustomResourceTest.java
+++ b/kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/CustomResourceTest.java
@@ -187,8 +187,6 @@ public class CustomResourceTest {
     Map<String, Object> result = client.customResource(customResourceDefinitionContext)
       .delete("ns1", "example-hello", true);
     assertEquals("Success", result.get("status"));
-    DeleteOptions expectedDeleteOptions = new DeleteOptions();
-    expectedDeleteOptions.setPropagationPolicy("Orphan");
 
     RecordedRequest request = server.getLastRequest();
     assertEquals("DELETE", request.getMethod());
--- a/kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/CustomResourceTest.java
+++ b/kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/CustomResourceTest.java
@@ -25,6 +25,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import io.fabric8.kubernetes.api.model.DeleteOptions;
+import okhttp3.mockwebserver.RecordedRequest;
 import org.junit.Rule;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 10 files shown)


kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/CustomResourceTest.java
```diff
@@ -177,6 +179,58 @@ public void testDelete() {
...
 
+  @Test
+  public void testCascadingDeletion() throws IOException, InterruptedException {
+    server.expect().delete().withPath("/apis/test.fabric8.io/v1alpha1/namespaces/ns1/hellos/example-hello").andReturn(HttpURLConnection.HTTP_OK, "{\"metadata\":{},\"apiVersion\":\"v1\",\"kind\":\"Status\",\"details\":{\"name\":\"prometheus-example-rules\",\"group\":\"monitoring.coreos.com\",\"kind\":\"prometheusrules\",\"uid\":\"b3d085bd-6a5c-11e9-8787-525400b18c1d\"},\"status\":\"Success\"}").once();
+
+    KubernetesClient client = server.getClient();
+    Map<String, Object> result = client.customResource(customResourceDefinitionContext)
+      .delete("ns1", "example-hello", true);
+    assertEquals("Success", result.get("status"));
+    DeleteOptions expectedDeleteOptions = new DeleteOptions();
```
manusa: Can't find where this is used.

kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/TypedCustomResourceApiTest.java
```diff
@@ -0,0 +1,140 @@
...
+import static org.junit.Assert.assertTrue;
+
+@EnableRuleMigrationSupport
+public class TypedCustomResourceApiTest {
+  @Rule
+  public KubernetesServer server = new KubernetesServer();
+
+  private MixedOperation<PodSet, PodSetList, DoneablePodSet, Resource<PodSet, DoneablePodSet>> podSetClient;
+
+  private CustomResourceDefinition podSetCrd = new CustomResourceDefinitionBuilder()
```
manusa: Variable initiation should be performed in a setUp() (@Before) method to avoid mutability problems when the variable is reused across tests.

rohanKanojia:
[merge]

Issue #1879, opened 2019-11-26
Selecting custom kube config for the library in springboot


contactreji:
Hello guys

We have a requirement where we run a springboot app in multiple config profiles. We use ```fabric8io/kubernetes-client``` library to communicate with remote k8s cluster to launch deployments of certain resources.

Our requirement is to point to different k8s cluster. (having the springboot jar run 3 instances with 3 config profiles in the same vm) 

Kube config file is at ```$HOME/.kube/config ```. We wanna have like ```$HOME/.kube/config _dev```, ```$HOME/.kube/config_stage```, and ```$HOME/.kube/config_prod```

How do I force my app to pick up custom kube config files instead of pointing to the default ``````$HOME/.kube/config ``` alone?

rohanKanojia:
I think there is a system property to override this behavior: `KUBERNETES_KUBECONFIG_FILE` . Could you please try it out and see if it helps your use case?

contactreji:
Thanks @rohanKanojia . I overrid the property called ```kubeconfig``` and it did the magic. passed the location/name of the new file as jvm argument and it worked.

PR #1882, opened 2019-11-28
Document kubeconfig property usage

manusa:
Relates to https://github.com/fabric8io/kubernetes-client/issues/1879

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,33 +52,47 @@ This will use settings from different sources in the following order of priority
 
 System properties are preferred over environment variables. The following system properties & environment variables can be used for configuration:
 
-* `kubeconfig` / `KUBECONFIG`: Name of the kubernetes config file to read (defaults to `~/.kube/config`)
-* `kubernetes.master` / `KUBERNETES_MASTER`
-* `kubernetes.api.version` / `KUBERNETES_API_VERSION`
-* `kubernetes.oapi.version` / `KUBERNETES_OAPI_VERSION`
-* `kubernetes.trust.certificates` / `KUBERNETES_TRUST_CERTIFICATES`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,6 +52,7 @@ This will use settings from different sources in the following order of priority
 
 System properties are preferred over environment variables. The following system properties & environment variables can be used for configuration:
 
+* `kubeconfig` / `KUBECONFIG`: Name of the kubernetes config file to read (defaults to `~/.kube/config`)
```
rohanKanojia: Sorry, I'm not sure but isn't the property name `KUBERNETES_KUBECONFIG_FILE`? Have you tried configuring `kubeconfig` using this property?

manusa: Haven't tried it, but checked the code. This looks like a legacy property that doesn't match the other naming conventions:
https://github.com/fabric8io/kubernetes-client/blob/6a47d262e37bee71b9579551a9a56288761f02c2/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java#L110

So yes, the Java variable has `KUBERNETES_KUBECONFIG_FILE` this name, but not its  value neither the conversion to environment variables.

rohanKanojia: oh, yeah. You're right.

manusa:
> Thanks a lot! Looks good with minor comment 👍
> 
> Thanks for adding a comment also to property. Could you please do the same for other properties also if you have time?

Yes, that's a good idea, will do.

Are these properties documented elsewhere (asciidoc, etc)?

rohanKanojia:
I'm not sure whether they are documented anywhere :( . You can try looking into code and try getting hint via comments :-( .

rohanKanojia:
[merge]

rohanKanojia:
Thanks, I think it's a good start.

PR #1887, opened 2019-12-03
Fixes #1880: Remove Reapers from kubernetes-client and leave it upto Kubernetes API server for cascade deletion.

### LINKED ISSUE #1880: Deleting statefulset fails with "Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden"
Testing 4.6.2 in OCP 3.11, we have discovered that using the following API to delete a statefulset with persistent volume claim ends with the following error:

` client.resource(statefulSet).cascading(true).delete();`

`Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden`

Turning up logging you see that it is the attempt to scale the statefulset prior to the delete that is actually failing.   The patch generated internally is actually invalid and includes a spurious mutation of the template namespaces, which is rejected by the server.

`INFO: [{"op":"replace","path":"/spec/replicas","value":0},{"op":"add","path":"/spec/template/metadata/namespace","value":"test"},{"op":"add","path":"/spec/volumeClaimTemplates/0/metadata/namespace","value":"test"}]`

There is a reproducer for this issue here:

https://github.com/k-wall/kc-del-statefulset

I dug into the issue a little, and can see that internally the `NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableImpl$ChangeNamespace` are actually erroneously mutating the namespaces of the `template` and `volumeClaimTemplates`.  This is coming from a call chain start at:  `NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableImpl#delete`.  This causes the patch sent to be server to be incorrect.

I don't understand the code well, but this looks fishy.  I suspect those visitors shouldn't be touching those.

Now with 4.6.3 (or 4.6.4), the issue actually disappears.   This is because #1856 prevents the consideration of the reaper (`reloadingFromServer` is false).  However, I think this is merely side stepping this defect.  I worry is that it will show up elsewhere, so I make the report.

rohanKanojia:
Fix #1880 

This PR seems to remove Reapers from codebase which seemed to be performing deletion of child resources of a specific Kubernetes resource since we now have moved on to [DeleteOptions](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#deleteoptions-v1-meta) `orphanDependencts` or `propagationPolicy`.

I've tested this on the following scenarios and it seems to be working fine(since deletion is independent of these reapers now).
```
// Deployments
client.apps().deployments().inNamespace("default").withName("nginx").delete();
client.apps().deployments().inNamespace("default").withName("nginx").cascading(false).delete();

// ReplicaSets
client.apps().replicaSets().inNamespace("default").withName("nginx").delete();
client.apps().replicaSets().inNamespace("default").withName("nginx").cascading(false).delete();

// ReplicationControllers
client.replicationControllers().inNamespace("default").withName("nginx")delete();
client.replicationControllers().inNamespace("default").withName("nginx").cascading(false).delete();

// DaemonSets
client.apps().daemonSets().inNamespace("default").withName("fluentd-elasticsearch").delete();
client.apps().daemonSets().inNamespace("default").withName("fluentd-elasticsearch").cascading(false).delete();

// Jobs
client.batch().jobs().inNamespace("default").withName("hello").delete();
client.batch().jobs().inNamespace("default").withName("hello").cascading(false).delete();

// StatefulSets
client.apps().statefulSets().inNamespace("default").withName("web").cascading(false).delete();
client.apps().statefulSets().inNamespace("default").withName("web").delete();

// Generic Resource API
client.resource(deployment).inNamespace("default").delete();
client.resource(deployment).inNamespace("default").cascading(false).delete();

// ResourceList API
client.resourceList(kubernetesResourceList).inNamespace("default").delete();
client.resourceList(kubernetesResourceList).inNamespace("default").cascading(false).delete();
```


oscerd:
ok to test

pom.xml
```diff
@@ -427,7 +427,7 @@
       <plugin>
         <groupId>org.jacoco</groupId>
         <artifactId>jacoco-maven-plugin</artifactId>
-        <version>0.8.5</version>
+        <version>0.8.4</version>
```
manusa: Why this downgrade?

rohanKanojia: I was seeing an `Error while creating report: null: EOFException`, so I thought recent upgrade could be a reason.

rohanKanojia:
[merge]

pom.xml
rohanKanojia: Seems like I was wrong in suspecting upgrade could be the cause of the exception. It has started coming up again :confused: https://ci.fabric8.io/job/kubernetes-client-pullreq/2382/console

PR #1892, opened 2019-12-05
ServiceCatalog crud test added

Skyterix1991:


kubernetes-tests/pom.xml
```diff
@@ -76,6 +76,18 @@
...
+      <groupId>me.snowdrop</groupId>
+      <artifactId>servicecatalog-client</artifactId>
+      <version>1.1.1</version>
+    </dependency>
+    <dependency>
+      <groupId>me.snowdrop</groupId>
+      <artifactId>servicecatalog-mock</artifactId>
+      <version>1.1.1</version>
+      <scope>test</scope>
+    </dependency>
```
kunal-kushwaha: As servicecatalog is a part of client now, you don't need to import it from snowdrop. Instead use : 
```
<groupId>io.fabric8</groupId>
<artifactId>servicecatalog-client</artifactId>
<version>${project.version}</version>
```

kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/ServiceCatalogCrudTest.java
```diff
@@ -0,0 +1,135 @@
+package io.fabric8.kubernetes.client.mock;
+
+import me.snowdrop.servicecatalog.api.client.ServiceCatalogClient;
+import me.snowdrop.servicecatalog.api.model.ClusterServiceBroker;
+import me.snowdrop.servicecatalog.api.model.ClusterServiceBrokerBuilder;
+import me.snowdrop.servicecatalog.api.model.ClusterServiceBrokerList;
+import me.snowdrop.servicecatalog.mock.ServiceCatalogServer;
```
kunal-kushwaha: Same thing as previous, do not use snowdrop.

kubernetes-tests/pom.xml
kunal-kushwaha: Also, don't explicitly type version numbers. Use tags instead.

PR #1901, opened 2019-12-09
Fix #1850: Build logs are always populated by timestamps

### LINKED ISSUE #1850: Build log is populated by timestamps
After upgrading openshift client we start to experience on OpenShift 4 that timestamps are injected also into middle of line, which breaks looking for specific pattern 

Log in openshift console:
```
[INFO] BUILD FAILURE
```

Log in test:
```
2019-10-31T19:47:58.096107367Z [IN2019-10-31T19:47:58.096124533Z FO] BUIL2019-10-31T19:47:58.096141464Z D FAIL2019-10-31T19:47:58.09615766Z URE
```

I believe it has something to do with https://github.com/fabric8io/kubernetes-client/issues/579. How can we turn off this behaviour? 

This is how we get log
```
	public String getBuildLog(Build build) {
		return builds().withName(build.getMetadata().getName()).getLog();
	}
```

rohanKanojia:
Fix #1850 

+ Added an option `withoutUsingTimestamp()` in Build dsl to set `timestamps=false`

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Timestampable.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Timestampable.java
@@ -20,6 +20,4 @@ public interface Timestampable<T> {
 
     T usingTimestamps();
 
-    T withoutUsingTimestamps();
-
 }
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Timestampable.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Timestampable.java
@@ -20,4 +20,6 @@ public interface Timestampable<T> {
 
     T usingTimestamps();
 
+    T withoutUsingTimestamps();
+
 }
--- a/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/BuildOperationContext.java
+++ b/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/BuildOperationContext.java
@@ -48,7 +48,7 @@ public class BuildOperationContext extends OperationContext {
   public BuildOperationContext() {
   }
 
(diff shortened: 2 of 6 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Timestampable.java
```diff
@@ -20,4 +20,6 @@
 
     T usingTimestamps();
 
+    T withoutUsingTimestamps();
```
manusa: Isn't this redundant and confusing?

I think that:

https://github.com/fabric8io/kubernetes-client/blob/503fd099714e151211f9b3681cd686bdfbee3e33/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/BuildOperationsImpl.java#L133-L135

will be invoked even if this configuration flag is omitted.

We could do the same for:

https://github.com/fabric8io/kubernetes-client/blob/503fd099714e151211f9b3681cd686bdfbee3e33/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java#L160-L162

and completely remove this option.


Is this possible? what do you think?

openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/BuildOperationContext.java
```diff
@@ -58,7 +58,7 @@ public BuildOperationContext(OkHttpClient client, Config config, String plural,
     this.errPipe = errPipe;
     this.tty = tty;
     this.terminatedStatus = terminatedStatus;
-    this.timestamps = true;
+    this.timestamps = timestampes;
```
manusa: nit: There's a typo in the constructor variable, a good time to fix it.

rohanKanojia:
I've modified `usingTimestamps()` to only set `timestamps=true` when invoked. Otherwise it would be false by default

rohanKanojia:
[merge]

PR #1920, opened 2019-12-30
Fix #417: Support Subresources

### LINKED ISSUE #417: Support subresources, e.g. scale, deploymentrollback, etc
Certain operations are applicable via subresources rather than updating the resource. We need a general way to support this as it's important for authorization policies that allow updating via subresources but not the resource itself.

rohanKanojia:
Fix #417 

Adds support for Scale and DeploymentRollback in DSL

+ Changed extensions/Scale to autoscaling/Scale
+ Added tests

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
@@ -22,6 +22,9 @@ import io.fabric8.kubernetes.api.model.KubernetesResourceList;
 import io.fabric8.kubernetes.api.model.LabelSelector;
 import io.fabric8.kubernetes.api.model.LabelSelectorRequirement;
 import io.fabric8.kubernetes.api.model.RootPaths;
+import io.fabric8.kubernetes.api.model.Status;
+import io.fabric8.kubernetes.api.model.extensions.DeploymentRollback;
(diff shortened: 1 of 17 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
```diff
@@ -824,12 +817,51 @@ protected T handlePatch(T current, T updated) throws ExecutionException, Interru
...
+    }
+  }
+
   protected T handleGet(URL resourceUrl) throws InterruptedException, ExecutionException, IOException {
     T answer = handleGet(resourceUrl, getType());
     updateApiVersionResource(answer);
     return answer;
   }
 
+  protected URL getCompleteResourceUrl() throws MalformedURLException {
```
manusa: I think we can make this private

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
```diff
@@ -824,12 +817,51 @@ protected T handlePatch(T current, T updated) throws ExecutionException, Interru
     return handlePatch(current, updated, getType());
   }
 
+  protected Scale handleScale() {
+    try {
```
manusa: can we replace the code block by `return handleScale(null)` so we don't have duplicate code?

rohanKanojia:
[merge]

Issue #1895, opened 2019-12-06
Parsing different quantity formats


avmohan:
Is there an equivalent of https://github.com/kubernetes/apimachinery/blob/adc6f4cd9e7d28fc1c1e5efc658d940d55c0f356/pkg/api/resource/quantity.go in java?

rohanKanojia:
I think we have a class, try searching `Quantity` in codebase.

avmohan:
The class is there - https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-model/kubernetes-model/src/main/java/io/fabric8/kubernetes/api/model/Quantity.java -  but it doesn't actually "parse" it into a proper typed value.

avmohan:
To clarify I was looking for something equivalent to https://github.com/kubernetes-client/java/blob/master/kubernetes/src/main/java/io/kubernetes/client/custom/Quantity.java

Do you think it's a good addition?

rohanKanojia:
Yes, I do think it's a valuable addition.

Our Quanity.java has been deprecated as per latest kubernetes releases. We need to modify it. Actually it's already there in model for parsing go structs, but https://github.com/kubernetes/apimachinery/blob/adc6f4cd9e7d28fc1c1e5efc658d940d55c0f356/pkg/api/resource/quantity.go doesn't follow standard kubernetes types format. Our go script assumes all structs in types.go, I think we need to manually generate Quantity.java for this to work.

rohanKanojia:
seemd related to #885

rohanKanojia:
@avmohan : Hi, I've created a PR to support this use case of comparing different Quantity objects of different formats. Could you please see if this would fulfil your use case or you are looking for something more flexible(like converting Quantities from one to other as implemented in official client)?

avmohan:
Thanks @rohanKanojia . This would work.

PR #1921, opened 2019-12-31
Refactor Quantity Class to handle format

### LINKED ISSUE #885: Quantity doesn't honor the unit
When using the two-argument constructor of Quantity like in `new Quantitity("256","Mi")`, then the unit is ignored when creating the descriptors. Workaround: `new Quantity("256 Mi")`

See https://github.com/syndesisio/syndesis-rest/issues/713 where this happened in the wild.

rohanKanojia:
Fix #885 #1895 

+ Added parsing logic to parse `format` from `amount` string
+ Added equality operator for comparing Quantity objects regardless of their formats

### DIFF
--- a/kubernetes-model/kubernetes-model/src/main/java/io/fabric8/kubernetes/api/model/Quantity.java
+++ b/kubernetes-model/kubernetes-model/src/main/java/io/fabric8/kubernetes/api/model/Quantity.java
@@ -108,8 +108,15 @@ public class Quantity  implements Serializable {
 
         @Override
         public void serialize(Quantity value, JsonGenerator jgen, SerializerProvider provider) throws IOException, JsonProcessingException {
-            if (value != null && value.getAmount() != null) {
-                jgen.writeString(value.getAmount());
+            if (value != null) {
+              StringBuilder objAsStringBuilder = new StringBuilder();
+              if (value.getAmount() != null) {
+                objAsStringBuilder.append(value.getAmount());
+              }
+              if (value.getFormat() != null) {
+                objAsStringBuilder.append(value.getFormat());
+              }
+              jgen.writeString(objAsStringBuilder.toString());
             } else {
                 jgen.writeNull();
             }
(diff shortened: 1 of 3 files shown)


kubernetes-model/kubernetes-model/src/main/java/io/fabric8/kubernetes/api/model/Quantity.java
```diff
@@ -46,86 +48,201 @@
...
-     */
-    public Quantity(String amount) {
-        this.amount = amount;
-    }
+  /**
+   * Single argument constructor for setting amount.
+   *
+   * @param amount amount of quantity specified.
+   */
+  public Quantity(String amount) {
```
manusa: If this is only used to deserialize it would probably be better to create a static factory/parser method and use that plus the use of `AbstractMap.SimpleEntry` is not really appropriate here.
```
* static Quantity parse(String formattedValue) {
    //... do your parsing stuff
    return new Quantity(amountStr, formatStr);
}
```

If the "intelligent" constructor is really necessary, then ditch the AbstractMap.SimpleEntry and use the parsed instance to prefill the new instance (kind of ugly too):
```
final Quantity parsedQuantity = parse(amount);
this.amount = parsedQuantity.getAmount();
this.format = parsedQuantity.getFormat();
```

kubernetes-model/kubernetes-model/src/main/java/io/fabric8/kubernetes/api/model/Quantity.java
```diff
@@ -46,86 +48,201 @@
...
+
+  public String getFormat() {
+    return format;
+  }
+
+  public void setFormat(String format) {
+    this.format = format;
+  }
+
+  public static BigDecimal getNormalizedValue(Quantity quantity) {
```
manusa: Is this returning the quantity in bytes? why not just call it `getAsBytes` or  `getAmountInBytes`?

kubernetes-model/kubernetes-model/src/main/java/io/fabric8/kubernetes/api/model/Quantity.java
```diff
@@ -46,86 +48,201 @@
...
     }
 
+    return digit.multiply(multiple);
+  }
 
-    @JsonAnyGetter
-    public Map<String, Object> getAdditionalProperties() {
-        return this.additionalProperties;
+  @Override
+  public boolean equals(Object o) {
```
manusa: There's no hashCode implementation!

rohanKanojia:
[merge]

rohanKanojia:
[merge]

PR #1931, opened 2020-01-06
Fix #1925: Kubernetes Service should always be loaded from server while patching it.

### LINKED ISSUE #1925: Can't replace Service when upgraded | spec.clusterIP: Invalid value: "": field is immutable.
Following is the code snippet used to create Service resource. It is failing when I upgraded Fabric8 from 4.1.1 to 4.6.4.

```
ParameterNamespaceListVisitFromServerGetDeleteRecreateWaitApplicable<HasMetadata, Boolean> load =
                kubernetesClient.resourceList(resourceDescriptor)
 final List<HasMetadata> res = load.createOrReplace()
```
resourceDescriptor is a String holding the Service manifest file.

> Failed to apply : 
> apiVersion: v1
> kind: Service
> metadata:
>   name: httpbin
>   labels:
>     app: httpbin
> spec:
>   ports:
>     - name: http
>       port: 5511
>       targetPort: 8080
>   selector:
>     deploymentconfig: httpbin
>  : io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: PUT at: https://xxx:8443/api/v1/namespaces/xxx/services/httpbin. Message: Service "httpbin" is invalid: spec.clusterIP: Invalid value: "": field is immutable. Received status: Status(apiVersion=v1, code=422, details=StatusDetails(causes=[StatusCause(field=spec.clusterIP, message=Invalid value: "": field is immutable, reason=FieldValueInvalid, additionalProperties={})], group=null, kind=Service, name=httpbin, retryAfterSeconds=null, uid=null, additionalProperties={}), kind=Status, message=Service "httpbin" is invalid: spec.clusterIP: Invalid value: "": field is immutable, metadata=ListMeta(_continue=null, remainingItemCount=null, resourceVersion=null, selfLink=null, additionalProperties={}), reason=Invalid, status=Failure, additionalProperties={}).

rohanKanojia:
Fix #1925 
While replace method, client was showing variable behaviors when done through different
ways in dsl. For example,

   client.services().inNamespace("test").createOrReplace(item);

Used to load Service from server since reloadingFromServer was set to true in this case but
if we use a generic handler reloadingFromServer was set to false. I think in case of replace
operations we should always load object from server since Kubernetes API doesn't accept
patching Service object wihout ClusterIP.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 #### Bugs
 * Fix #1850: Add option to disable timestamps in build logs on Openshift
 * Fix #1902: Fix the usage of reflection, so that `getMetadata` is detected properly
+* Fix #1925: Client should always read services from server during replace
 * Fix #1486: Creating CRDs with schema validation is broken
 * Fix #1707: HorizontalPodAutoscalerSpecBuilder found no metric method
 * Fix #885: Quantity doesn't honour the unit
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ServiceOperationsImpl.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ServiceOperationsImpl.java
@@ -62,7 +62,7 @@ public class ServiceOperationsImpl extends HasMetadataOperation<Service, Service
   @Override
   public Service replace(Service item) {
       try {
-        Service old = getMandatory();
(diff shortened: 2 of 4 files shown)


rohanKanojia:
@cedar715: Could you please test this patch if it's working for your use case? It was working for me when I tested using your approach.

rohanKanojia:
retest this please

rohanKanojia:
[merge]

Issue #1948, opened 2020-01-14
Do we have a plan to support LeaderElection?


wangyang0918:
Just like the official kubernetes client[1], does fabric8io kubernetes-client could support leader election? It is a very useful feature.

[1]. https://github.com/kubernetes-client/java/blob/master/examples/src/main/java/io/kubernetes/client/examples/LeaderElectionExample.java

rohanKanojia:
@wangyang0918 : Thanks for pointing this out. I think we can consider supporting this feature. I'm not familiar with this. Could you please elaborate what use cases it is designed to solve?

rohanKanojia:
I think this also is a utility in client-go https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/leaderelection.go

wangyang0918:
@rohanKanojia Thanks for your quick response.
I am from Flink community and focused on running [Flink on kubernetes](https://docs.google.com/document/d/1-jNzqGF6NfZuwVaFICoFQ5HFFXzF5NVIagUZByFMfBY/edit?usp=sharing) recently. I choose `fabric8io/kubernetes-client` sdk, because it used more widely than the official client. Unfortunately, it does not support leader election. In Flink, also other big data framework, we have internal high availability implementation. Multiple masters will participate in the election and the leader publish the address information to config map. All the workers retrieve the master information and connect to the master. 

So if the kubernetes-client could provide the ability of leader election, it will be very useful for us.

rohanKanojia:
@wangyang0918 : Thanks for your feedback. I think it makes sense for us to support it.

wangyang0918:
@rohanKanojia Glad to hear that. It is my pleasure to help with the new feature test once you have an implementation.

manusa:
> ... It is my pleasure to help with the new feature test once you have an implementation.

@wangyang0918 we've just open a PR with an implementation #1955

manusa:
Kubernetes Client [4.8.0](https://github.com/fabric8io/kubernetes-client/releases/tag/v4.8.0) containing a LeaderElection implementation was released 3 days ago.

I'm closing this issue as the initial request is fulfilled.

If there is any problem or some feature missing from our implementation (e.g. an additional lock implementation), please open a new issue.

PR #1954, opened 2020-01-16
WIP: model build with leadeelection works

manusa:


manusa:
If the _"snapshoted"_ version of `WatchEvent` is not removed, there is a collision in `WatchEventFluent` as there are more than one `KubernetesResources` implementing classes named `EndpointPort` (One in the base api apackage and the other in the discovery).

The first commit, just removing the snapshot of WatchEvent, compiles the project but will cause failures on dependent modules.

The second commit, modifying the snapshoted WatchEvent to support raw Object as the field values [seems the way to go](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#watchevent-v1-meta) (to avoid sundr.io problems) but is a strong change in the API.

The third option would be to improve github.com/sundrio/sundrio to allow having several fields with different classes that share the same name (e.g. `EndpointPort`)

Any advice on this matter would be really appreciated in order to continue with the implementation of #1948

cc: @rohanKanojia @iocanel

rohanKanojia:
@manusa : Hey, sorry I didn't look into this. Why did you close this?

manusa:
> @manusa : Hey, sorry I didn't look into this. Why did you close this?

I saw there are already some workarounds in place for my same problem here:
https://github.com/fabric8io/kubernetes-client/blob/af637660b423f1e858335824ef6b783f2cf47261/kubernetes-model/pkg/schemagen/generate.go#L145-L172

I'm going to try to reimplement this in a more sustainable way, working on it now.

PR #1963, opened 2020-01-23
Watch: Use Serialization not a default ObjectMapper

zklapow:
The `WatchHTTPManager` currently uses a default jackson `ObjectMapper` but this prevents the user from reconfiguring the object mapper with custom modules, which can in turn break deserialization into objects which use types handled by those modules. This PR updates the `WatchHTTPManager` to use the `Serialization` util and its `ObjectMapper`


rohanKanojia:
ok to test

rohanKanojia:
[merge]

zklapow:
thanks for the review and the super-speedy turn around, much appreciated!

rohanKanojia:
@zklapow : Your fix is available in [v4.7.1](https://repo1.maven.org/maven2/io/fabric8/kubernetes-client/4.7.1/)

PR #1976, opened 2020-01-30
fix: remove useless resource-*.vm from jar packages

### LINKED ISSUE #1847: Duplicate resources on classpath
The `kubernetes-client-*.jar` and `openshift-client-*.jar` both ship with `resource-*.vm` within them. I don't know if these files are loaded at runtime, they seem to be a part of code generation, so perhaps not.

If they're loaded at runtime there is no guarantee on what file will be read, it might depend on the order of jars on the classpath or on some heuristic of the class loader being used. In that case it might be better to name them differently for each jar to guarantee that the correct version is loaded.

If they're not needed at runtime it would be best that they're not shipped in jars.

manusa:
Fix #1847

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### 4.7-SNAPSHOT
 #### Bugs
+* Fix #1847: Remove resource-\*.vm files from \*-client.jar
 
 #### Improvements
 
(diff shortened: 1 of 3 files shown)


oscerd:
[merge]

rohanKanojia:
[merge]

rohanKanojia:
[merge]

Issue #1874, opened 2019-11-22
Line separation chars bug


nbn:
Hi. Sorry for not having the time to look into this on your behalf, but I had a problem with line endings using kubernetes-client version 4.6.3.

Loading a yaml file with a list of objects on a windows machine only contained 1 element, which puzzled me, so I did some tests and determined the problem being newline characters.

The following example yielded  6 and 1 objects respectively when using the  multiple-document-template.yml from the examples resources , and should be easy to create a unit test from.

` 
      OpenShiftClient client = new DefaultOpenShiftClient();

        String unixText = StringUtils.join(Files.readAllLines(new File(file).toPath()), "\n");
        String winText = StringUtils.join(Files.readAllLines(new File(file).toPath()), "\r\n");

        List<HasMetadata> unixList = client.load(new ByteArrayInputStream(unixText.getBytes("UTF-8"))).get();
        List<HasMetadata> winList = client.load(new ByteArrayInputStream(winText.getBytes("UTF-8"))).get();

        log.debug("Sizes {},{}", winList.size(), unixList.size());`

rohanKanojia:
@nbn : seems like a bug we can fix ;-)

manusa:
Hi @nbn 
Could you please provide the content of a file that will reproduce the **exact** issue.

I **can** reproduce some problems, but non specific to line endings.

BTW, direct deserialization/unmarshaling works OK:
```
 final List<String> fileLines = Files.readAllLines(
   new File(SerializationTest.class.getResource("/test-template.yml").getFile()).toPath(), StandardCharsets.UTF_8);
 final String crlfFile = String.join(" \r\n", fileLines);
 final KubernetesResource stringUnmarshall = Serialization.unmarshal(crlfFile, KubernetesResource.class);
```

nbn:
Hello,
As I wrote I was using the multiple documents from the examples project. I tried to re-create the problem (at home on a mac), but could not. I then sent the code to my work pc (win10) and ran the same example with the wrong result.. Have embedded the crude example and the file.

[openshiftclientbug.tar.gz](https://github.com/fabric8io/kubernetes-client/files/4154909/openshiftclientbug.tar.gz)

nbn:
Just running mvn package followed by java -jar target/openshiftclientbug-1.0-SNAPSHOT.jar

manusa:
Hi, thnx for your input. We'll try with your provided files.

manusa:
[...]

PR #1982, opened 2020-02-03
test: tests verifying CRLF works + refactored failing and disabled activated tests

manusa:
Added tests verifying reported misbehavior in #1874.

All tests pass, #1874 is probably caused by something else. Additional feedback is required from @nbn for viable fix.


rohanKanojia:
[merge]

PR #1985, opened 2020-02-04
Fix: authenticator uses id-token from ~/.kube/config

### LINKED ISSUE #1177: Bug: access-token vs. id-token
I believe that I've found a bug in the fabric8 Kubernetes client. We were testing it with a service (VMware Kubernetes Engine) that uses OIDC and it failed to authenticate. It gave an error like:

```
Cannot access cluster for detecting mode: Failure executing: GET at: https://HOSTNAME/. Message: Unauthorized! Token may have expired! Please log-in again. Unauthorized
```

The token is stored in kubectl as an OIDC auth-provider, and it's called "id-token". An edited version from my kubectl file looks like this:

```
users:
- name: me
  user:
    auth-provider:
      config:
        client-id: 86a...
        client-secret: 0df...
        id-token: eyJhbGciOiJSU...
        idp-certificate-authority: /var/folders/...
        idp-issuer-url: https://XXX
        refresh-token: ""
```

I believe that `id-token` is correct, based on my reading of the go-client, which is used by kubectl. From https://github.com/kubernetes/client-go/blob/master/plugin/pkg/client/auth/oidc/oidc.go#L43 

```
cfgIDToken                  = "id-token"
```

If I look in `kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java`, I see:

```
  private static final String ACCESS_TOKEN = "access-token";
...
          if (Utils.isNullOrEmpty(config.getOauthToken()) && currentAuthInfo.getAuthProvider() != null && !Utils.isNullOrEmpty(currentAuthInfo.getAuthProvider().getConfig().get(ACCESS_TOKEN))) {
            config.setOauthToken(currentAuthInfo.getAuthProvider().getConfig().get(ACCESS_TOKEN));
          }
```

I manually edited my kubectl file to rename `id-token` to `access-token`, and I re-ran `mvn fabric8 deploy` and it worked. 

I appreciate why you use ACCESS_TOKEN, because it is used for other auth providers. Probably this code needs to be adapted to use either access-token or id-token, depending on what is available.

I'd offer a patch, but I probably won't get to it soon. Hopefully this bug description is sufficient for you to understand and address the issue. (And if you really think hard about it, you probably want to be able to use refresh-tokens, but that starts to be more work...)

matthyx:
Fixes #1177 

According to the official [documentation](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#openid-connect-tokens):
> the authenticator uses the id_token (not the access_token) from the OAuth2 token response as a bearer token

This explains why there is no `access-token` in the file.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
@@ -136,6 +136,7 @@ public class Config {
   public static final String HTTPS_PROTOCOL_PREFIX = "https://";
 
   private static final String ACCESS_TOKEN = "access-token";
+  private static final String ID_TOKEN = "id-token";
 
   private boolean trustCerts;
   private boolean disableHostnameVerification;
@@ -537,8 +538,14 @@ public class Config {
           config.setUsername(currentAuthInfo.getUsername());
           config.setPassword(currentAuthInfo.getPassword());
 
-          if (Utils.isNullOrEmpty(config.getOauthToken()) && currentAuthInfo.getAuthProvider() != null && !Utils.isNullOrEmpty(currentAuthInfo.getAuthProvider().getConfig().get(ACCESS_TOKEN))) {
-            config.setOauthToken(currentAuthInfo.getAuthProvider().getConfig().get(ACCESS_TOKEN));
+          if (Utils.isNullOrEmpty(config.getOauthToken()) && currentAuthInfo.getAuthProvider() != null) {
+            if (!Utils.isNullOrEmpty(currentAuthInfo.getAuthProvider().getConfig().get(ACCESS_TOKEN))) {
+              // GKE token
+              config.setOauthToken(currentAuthInfo.getAuthProvider().getConfig().get(ACCESS_TOKEN));
+            } else if (!Utils.isNullOrEmpty(currentAuthInfo.getAuthProvider().getConfig().get(ID_TOKEN))) {
+              // OpenID Connect token
+              config.setOauthToken(currentAuthInfo.getAuthProvider().getConfig().get(ID_TOKEN));
+            }
           } else if (config.getOauthTokenProvider() == null) {  // https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins
             ExecConfig exec = currentAuthInfo.getExec();
             if (exec != null) {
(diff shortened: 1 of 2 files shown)


rohanKanojia:
ok to test

rohanKanojia:
@matthyx : Hi, Thanks  a lot for quick fix. Could you also please add a line to CHANGELOG regarding this change? Some tests validating desired behaviour would be awesome too.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
```diff
@@ -537,8 +537,9 @@ private static boolean loadFromKubeconfig(Config config, String context, String
           config.setUsername(currentAuthInfo.getUsername());
           config.setPassword(currentAuthInfo.getPassword());
 
-          if (Utils.isNullOrEmpty(config.getOauthToken()) && currentAuthInfo.getAuthProvider() != null && !Utils.isNullOrEmpty(currentAuthInfo.getAuthProvider().getConfig().get(ACCESS_TOKEN))) {
-            config.setOauthToken(currentAuthInfo.getAuthProvider().getConfig().get(ACCESS_TOKEN));
+          if (Utils.isNullOrEmpty(config.getOauthToken()) && currentAuthInfo.getAuthProvider() != null && !Utils.isNullOrEmpty(currentAuthInfo.getAuthProvider().getConfig().get(
```
manusa: I'm not really sure if this will introduce any regression as we are completely ditching `ACCES_TOKEN` which was added in #636 > #650

I would probably do as the issue suggests ("...Probably this code needs to be adapted to use either access-token or id-token, depending on what is available.."): so try to use ID_TOKEN or try to fallback to ACCESS_TOKEN

matthyx: @manusa good point... reading the `kubectl` source code I thought it cannot use `access-token` directly, but in fact it's used on gcloud.
I will refactor the code to support both cases.

matthyx:
> Some tests validating desired behaviour would be awesome too.

I am not sure how to test this behaviour since we require a GKE cluster and an OIDC compatible cluster.

rohanKanojia:
@matthyx : oh, I see. Ignore my comment then

rohanKanojia:
[merge]

codythe1st:
will this fix be added to version 4.6.4? have same issue. changing 'id-token' to 'access-token' fixes the issue.

manusa:
@codythe1st Is it possible for you to upgrade to latest Kubernetes-Client version, or  do you really need this fix to get backported to 4.6?

PR #1988, opened 2020-02-05
fix: YAML with multiple documents parsed to lists

manusa:
When loading a YAML file containing multiple documents, not all documents
are unmarshalled when loading a file with LF line endings in a windows
machine. Current fix considers any line ending in a platform independent
way.

Fixes: #1874


rohanKanojia:
[merge]

PR #1994, opened 2020-02-11
Chore: Update Openshift model to latest version

rohanKanojia:
Updated Openshift model to latest revisions. There were not any tags
pushed since v4.1.0 so I just updated the revision


manusa:
Maybe we should start thinking in an alternative repo to generate our OpenShift API schema.

Would this be a valid alternative https://github.com/code-ready/crc ?

rohanKanojia:
Umm, I don't think so. I think they rely on these too. They ship `oc` binary (github.com/openshift/origin) and just create a vm with single node openshift cluster running inside it.

rohanKanojia:
[merge]

PR #2011, opened 2020-02-19
fix: apiVersion and kind from CRD in registration

zregvart:
In `CustomResourceOperationsImpl` construction
`KubernetesDeserializer::registerCustomKind` is invoked without
`apiVersion` and with `kind` set to the class name. This could lead to
issues when deserializing custom resources, in my example:

```
com.fasterxml.jackson.databind.JsonMappingException: No resource type found for:camel.apache.org/v1alpha1#Integration
 at [Source: (String)"{"type":"DELETED","object":{"apiVersion":"camel.apache.org/v1alpha1","kind":"Integration","metadata":{"annotations":{"prometheus.io/port":"9779","prometheus.io/scrape":"true","syndesis.io/deploy-id":"i-M0SZstrdU_fCuujYkqGz:2","syndesis.io/deployment-version":"2","syndesis.io/integration-id":"i-M0SZstrdU_fCuujYkqGz","syndesis.io/integration-name":"t2l"},"creationTimestamp":"2020-02-19T13:14:10Z","generation":1,"labels":{"syndesis.io/app":"syndesis","syndesis.io/component":"integration","syndesis."[truncated 3644 chars]; line: 1, column: 4142] (through reference chain: io.fabric8.kubernetes.api.model.WatchEvent["object"])
	at com.fasterxml.jackson.databind.JsonMappingException.from(JsonMappingException.java:271) ~[jackson-databind-2.10.2.jar:2.10.2]
	at io.fabric8.kubernetes.internal.KubernetesDeserializer.deserialize(KubernetesDeserializer.java:78) ~[kubernetes-model-4.6.1.jar:4.6.1]
	at io.fabric8.kubernetes.internal.KubernetesDeserializer.deserialize(KubernetesDeserializer.java:33) ~[kubernetes-model-4.6.1.jar:4.6.1]
```

Because only `"Integration"` would be registered, whereas type lookup is
using `apiVersion` and `kind.

This changes that to specify the `apiVersion` and `kind` based on the
definition given in the `CustomResource`.

### DIFF
--- a/kubernetes-client/src/test/java/io/fabric8/kubernetes/client/dsl/internal/CustomResourceOperationsImplTest.java
+++ b/kubernetes-client/src/test/java/io/fabric8/kubernetes/client/dsl/internal/CustomResourceOperationsImplTest.java
@@ -1,0 +1,86 @@
+/**
+ * Copyright (C) 2015 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.fabric8.kubernetes.client.dsl.internal;
+
+import java.io.IOException;
+
+import io.fabric8.kubernetes.api.model.KubernetesResource;
+import io.fabric8.kubernetes.api.model.apiextensions.CustomResourceDefinition;
+import io.fabric8.kubernetes.api.model.apiextensions.CustomResourceDefinitionBuilder;
+import io.fabric8.kubernetes.client.CustomResource;
+import io.fabric8.kubernetes.client.CustomResourceList;
+import io.fabric8.kubernetes.internal.KubernetesDeserializer;
+
+import org.junit.jupiter.api.Test;
+
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.MappingJsonFactory;
+
+import static org.hamcrest.core.IsInstanceOf.instanceOf;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+
+public class CustomResourceOperationsImplTest {
+
+  public static class MyCustomResource extends CustomResource {
+  }
+
+  public static class MyCustomResourceList extends CustomResourceList<MyCustomResource> {
+  }
+
+  private final CustomResourceDefinition crd = new CustomResourceDefinitionBuilder()
+    .withApiVersion("custom.group/v1alpha1")
+    .withKind("MyCustomResource")
+    .withNewMetadata()
+      .withName("custom.name")
+    .endMetadata()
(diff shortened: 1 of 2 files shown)


rohanKanojia:
ok to test

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/dsl/internal/CustomResourceOperationsImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+		JsonFactory factory = new MappingJsonFactory();
+		JsonParser parser = factory.createParser("{\n" + 
+		    "    \"apiVersion\": \"custom.group/v1alpha1\",\n" + 
+		    "    \"kind\": \"MyCustomResource\"\n" + 
+		    "}");
+
+		KubernetesDeserializer deserializer = new KubernetesDeserializer();
+		KubernetesResource resource = deserializer.deserialize(parser, null);
+
+		assertThat(resource, instanceOf(MyCustomResource.class));
```
manusa: I'd probably add an assertion validation the content of a field in order to verify deserialization worked.

(e.g. `assertThat(((MyCustomResource)resource).getApiVersion(), equalTo("custom.group/v1alpha1"))`)

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/dsl/internal/CustomResourceOperationsImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+public class CustomResourceOperationsImplTest {
+
+	public static class MyCustomResource extends CustomResource {
+    }
+
+	public static class MyCustomResourceList extends CustomResourceList<MyCustomResource> {
+    }
+
+    CustomResourceDefinition crd = new CustomResourceDefinitionBuilder()
```
manusa: Please add access modifiers to these field (private)

Ideally field instantiation should happen inside a test setup method (`@BeforeEach`) in order to avoid side effects when new tests are added to the suite.

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/dsl/internal/CustomResourceOperationsImplTest.java
zregvart: Added https://github.com/fabric8io/kubernetes-client/pull/2011/files#diff-1500b8672a6f6391abfa76c29b8128cfR83

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/dsl/internal/CustomResourceOperationsImplTest.java
zregvart: Changed to `private final` this should cause no side effects that I can think of.

zregvart:
@manusa thanks, can you have another look?

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/dsl/internal/CustomResourceOperationsImplTest.java
manusa: Making it final doesn't make it immutable which may lead to some side-effects:

So if one of the test cases (not at the moment, and probably not ever) does something like `crd.setApiVersion("bar")` in order to do some corner-case test and some other test asserts on this field, it will fail. It will also fail in a non-deterministic way depending on how the system computes the test execution order.

That's why it's always recommended to use setup and tear down methods in tests.

rohanKanojia:
[merge]

rohanKanojia:
[merge]

rohanKanojia:
[merge]

Issue #1984, opened 2020-02-04
There is unknown field "container" in Tekton


JINwonLEE:
#### Issue description
I was trying to make Tekton `Task` with java fabric8.tekton apis, however I got some issues.

However I got an error message in runtime. 

![image](https://user-images.githubusercontent.com/18036257/73721122-e0161000-4766-11ea-9735-65862031d554.png)

#### Steps to reproduce the issue

I just compile with javac, and run the jar file. 

It is just sample test code, and there are already information about environments in link below. 

[issue in stackoverflow](https://stackoverflow.com/questions/60049533/there-is-unknown-field-container-in-tekton?noredirect=1#comment106205962_60049533)

#### What's the expected result?

- Make `Task` in tekton

#### What's the actual result?

- Got Error message as a screenshot provided.

rohanKanojia:
@JINwonLEE : Hi Jin, Thanks for you bug report. We'll try to get this fixed and update to latest tekton versions in coming release

JINwonLEE:
@rohanKanojia : Thanks for reply. Can I get some information or opinions about this issue? I have no idea how to figure it out. Thanks!!

rohanKanojia:
@Fabian-K  : polite ping, Do you have any idea about this? I haven't worked on tekton extension so can't give much feedback. Would appreciate if you could help out here

Fabian-K:
This seems to be a duplicate of https://github.com/fabric8io/kubernetes-client/issues/1900 

As far as I can see this is caused by the handling of embedded structs in go. A step in Tekton is basically a vanilla k8s container definition with (currently) one additional field for scripts. (https://github.com/tektoncd/pipeline/blob/v0.9.0/pkg/apis/pipeline/v1alpha1/task_types.go#L66-L73)

Serialization to JSON currently does not "union" all fields into one object. Instead, it creates an intermediate object for the vanilla k8s container definition. It should be like `{"image": "ubuntu","name":"test"}`, however its currently serialized as `{"container": {"image": "ubuntu","name":"test"}}`.

I don´t know why, but deserialization seems to handle this correctly -> it is possible to list tasks via the tekton client.

JINwonLEE:
@Fabian-K  Thanks for reply. 

As I understand, currently Tekton supports Script not Container. 

It means if I use `Script` instead of Container, then I could implement Tekton `Task` properly. 

[...]

PR #2020, opened 2020-02-24
Updated Tekton model to v0.11.0-rc2

rohanKanojia:
Kudos to @pradeepitm12 for helping me out on upgrading this.


extensions/tekton/generator/cmd/generate/generate.go
```diff
@@ -16,43 +16,69 @@
...
+  V1alpha1TaskRunList                     v1alpha1.TaskRunList
+  V1alpha1ClusterTask                     v1alpha1.ClusterTask
+  V1alpha1ClusterTaskList                 v1alpha1.ClusterTaskList
+  V1alpha1Condition                       v1alpha1.Condition
+  V1alpha1ConditionList                   v1alpha1.ConditionList
+  V1alpha1PipelineTask                    v1alpha1.PipelineTask
+  V1alpha1PipelineTaskResource            v1alpha1.PipelineTaskResources
+  V1alpha1ResourceDeclaration             v1alpha1.ResourceDeclaration
+
+  //V1alpha2Pipeline                        v1alpha2.Pipeline
```
manusa: Is there any reason to keep these commented-out lines?

jeffmaury:
As I'm working on the same kind of changes (Triggers), I noticed that this PR will bring a breaking change as few classes have moved to a different packages and even some have their name changed. It is a issue ?

rohanKanojia:
@jeffmaury : Hi, Could you please tell me classes which I moved to different package? I'm not so familiar with tekton so I could have done some mistake. Happy to modify my PR as per your suggestions :-) .

Or are you saying that you're also working on updating model and would like to update it yourself? I know you've already sent a patch to fix 2 issues on our tekton extensions so it would be better if you contribute more :smile: (whether it's code contribution or review)

jeffmaury:
For what I recall (but that would need a need to diff comparison):
- Step has been moved to v1alpha2 package
- Template (pod) has been moved to a different pod package (was v1alpha1)
- Task params is now a list of v1alpha2.ParmSpec (was v1alpha1.Parm) see updated PipeLineCrudTest
- pipeline.v1alpha1.PipelineResource moved to resource.v1alpha1.PipelineResource

jeffmaury:
Also these changes could be outlined with:
`mvn clean clirr:check`

rohanKanojia:
@jeffmaury : I understand that this PR would break backward compatibility with previous versions, but I think it's due to breaking changes in Tekton and we would need to adapt accordingly? If you need release without this patch, we can cut a release without merging this? WDYT?

jeffmaury:
I don't need that PR to be in for my purpose. However, I have a question regarding API versioning as Tekton is also v1beta1 now so what is the policy regarding API versions:
- expose all versions ports (not done by Knative for example)
- uses the latest version (but will break backward compatibility)

rohanKanojia:
@jeffmaury : Sorry for the late reply, I just came back from PTO. In my opinion we should use the latest version since Knative and Tekton are moving very fast and it would be nice to stay updated to the latest versions.

rohanKanojia:
Tekton folks have released another version: https://github.com/tektoncd/pipeline/releases . I will update this to that

rohanKanojia:
@jeffmaury : Hi Jeff, I've updated model to `v0.11.0-rc2` and modified DSL to use `v1beta1` resources. Could you please take a look at this PR and see if this is the change you wanted?

Fabian-K:
Hi,
great that you are already working on updating tekton to the latest version! I had a look at this pr and noticed that it has the same issue with `TaskResourceBinding` as we faced with the `Step`class.

Instead of having the two properties `paths` and `pipelineResourceBinding` it needs to have all properties of the `PipelineResourceBinding`(e.g. name, resourceRef, resourceSpec) and in addition paths.

For `Step` that was fixed by @jeffmaury with https://github.com/fabric8io/kubernetes-client/pull/2033.

As already mentioned, unfortunately I don´t know the generation logic. Is it possible to find a universal fix instead of workarounds for individual classes? Root cause seems to be the handling of embedded structs in go (see my comment in https://github.com/fabric8io/kubernetes-client/issues/1984#issuecomment-582415107).

rohanKanojia:
@Fabian-K : Unfortunately, I'm also no so familiar with Golang either. But looking at code maybe `field.Anonymous && field.Type.Kind() == reflect.Struct` seems to be checking for Embedded Structs. I tried removing the check for length but that started breaking model generation process. In the end, I had to hardcode `TaskResourceBinding` like `Container` to get expected results.

https://github.com/fabric8io/kubernetes-client/blob/8321522e75d0e6f5b5ef0aadea3d82f07beff758/extensions/tekton/generator/pkg/schemagen/generate.go#L424

@iocanel : Do you have some idea on how to make handling of Embedded Structs a bit smarter?

extensions/tekton/generator/cmd/generate/generate.go
```diff
@@ -16,43 +16,65 @@
...
 
 func main() {
 	packages := []schemagen.PackageDescriptor{
 		//"$ref": "#/definitions/github_com_tektoncd_pipeline_vendor_k8s_io_api_core_v1_Probe",
 		{"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1", "tekton.dev", "io.fabric8.tekton.pipeline.v1alpha1", "tekton_v1alpha1_"},
-		{"github.com/knative/pkg/apis", "v1", "io.fabric8.knative.v1", "knative_"},
+    {"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1", "tekton.dev", "io.fabric8.tekton.pipeline.v1beta1", "tekton_v1alpha2_"},
+    {"github.com/tektoncd/pipeline/pkg/apis/resource/v1alpha1", "tekton.dev", "io.fabric8.tekton.resource.v1alpha1", "tekton_resource_v1alpha1_"},
+    {"github.com/tektoncd/pipeline/pkg/apis/pipeline/pod", "tekton.dev", "io.fabric8.tekton.pipeline.pod", "tekton_pod_"},
+    {"github.com/knative/pkg/apis", "v1", "io.fabric8.knative.v1", "knative_"},
```
hrishin: Why do we need?

rohanKanojia: Umm, not sure. I didn't touch it while upgrading model. Let me try removing it.

rohanKanojia: Looks like some structs in tekton model depend on these packages(one of these is this https://github.com/tektoncd/pipeline/blob/master/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L28 ), when I removed them I started getting errors.

hrishin: ahh, no worries

rohanKanojia:
[merge]

PR #2021, opened 2020-02-25
Generating toString for custom resource classes

fernanfs:
Most of the api model already has toString implementations providing sufficient insight during runtime. The CustomResource and associated classes was missing a toString implementation.

Similar to the many parts of the kubernetes client codebase, this has been implemented using lombok.

### DIFF
--- a/kubernetes-client/pom.xml
+++ b/kubernetes-client/pom.xml
@@ -143,6 +143,11 @@
       <scope>compile</scope>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>org.projectlombok</groupId>
+      <artifactId>lombok</artifactId>
+      <scope>provided</scope>
+    </dependency>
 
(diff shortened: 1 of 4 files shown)


rohanKanojia:
ok to test

kubernetes-client/pom.xml
```diff
@@ -143,6 +143,12 @@
       <scope>compile</scope>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>org.projectlombok</groupId>
+      <artifactId>lombok</artifactId>
+      <scope>compile</scope>
```
manusa: I think the appropriate scope here should be provided here. Lombok should not be propagated transitively

oscerd: +1

fernanfs: done.

rohanKanojia:
[merge]

Issue #1981, opened 2020-02-03
Enable Sonarcube code analysis on project


rohanKanojia:
We're already using sonar cloud for analyzing PRs and project in Fabric8 Maven Plugin and Docker Maven Plugin. It seems to be working well lately. 

I think we should enable it here as well. We just need to enable project on sonarcloud and add circleci jobs for analyzing PRs https://github.com/fabric8io/fabric8-maven-plugin/blob/master/.circleci/config.yml

manusa:
Maybe it's a good time to introduce GitHub Actions in the project and leverage a little the CircleCI build resources.

Right now we're running actions for IT both in [JKube](https://github.com/eclipse/jkube/actions) and unofficially for [FMP](https://github.com/manusa/fmp-integration-tests/actions) and they are working quite well.

rohanKanojia:
@manusa : Yeah, that's a good idea. Right now we're running tests on Openshift instance only but in case of Jkube we're also running on Kubernetes. We should be testing on both platforms.

manusa:
Sonar scanning is working:
- https://sonarcloud.io/dashboard?id=kubernetes_client
- https://github.com/fabric8io/kubernetes-client/actions/runs/48034811
- https://github.com/fabric8io/kubernetes-client/actions/runs/48207189

PR #2024, opened 2020-02-25
fix: Do cleanup after LogWatchCallback finished

jzwlqx:
I wrote some code to watch logs from a container in a pod, but found that when the pod was deleted, log inputstream that got from PodOperationsImpl.watchLog().getOutput() didn't close and read would hang forever. 

To reproduce the issue, following the steps below :

Step 1. create a simple pod, any image will be ok as long as it can long run and give some output continuously . (Infact, if the pod is not exists, the java code here would hang in the same way)

```
apiVersion: v1
kind: Pod
metadata:
  name: t01
  namespace: test
spec:
  containers:
  - image: any-image
    name: default

```
Step 2: Run the code 
```
public static void main(String[] args) throws IOException {
        KubernetesClient client = new DefaultKubernetesClient();
        InputStream in = client.pods()
                .inNamespace("test")
                .withName("t01")
                .inContainer("default")
                .watchLog()
                .getOutput();

        byte[] buf = new byte[1024];
        int len = -1;
        while ((len = in.read(buf)) > -1) {  // hang here
            System.out.println("read len " + len);
        }
        System.out.println("finished");
    }
```
Step 3:  delete the pod `kubectl -n test delete pod t01`, After the pod is deleted, Java code still hang at the line in.read(buf)

The reson is that LogWatchCallback not close the newly created piped stream, It has close and cleanUp methods but nowhere would call that.

 This fix had been tested.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/BlockingInputStreamPumper.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/BlockingInputStreamPumper.java
@@ -60,6 +60,8 @@ public class BlockingInputStreamPumper extends InputStreamPumper {
           } else {
             LOGGER.debug("Interrupted while pumping stream.");
           }
+        } finally {
+          close();
         }
     }
 }
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/LogWatchCallback.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/LogWatchCallback.java
@@ -136,6 +136,7 @@ public class LogWatchCallback implements LogWatch, Callback, AutoCloseable {
         }
 
         LOGGER.error("Log Callback Failure.", ioe);
+        cleanUp();
         //We only need to queue startup failures.
         if (!started.get()) {
             queue.add(ioe);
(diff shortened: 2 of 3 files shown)


rohanKanojia:
ok to test

jzwlqx:
ping @iocanel

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/BlockingInputStreamPumper.java
```diff
@@ -60,6 +60,8 @@ public void run() {
           } else {
             LOGGER.debug("Interrupted while pumping stream.");
           }
+        }finally {
```
rohanKanojia: nit: add a space before `finally {`

jzwlqx: Thanks for your comment. Had corrected

jzwlqx:
Sonar Scanner failed strangely after add one space. Do I have to solve that?

rohanKanojia:
@jzwlqx : No, it's a fault on our end. You don't have to worry about it.

manusa:
We're still working on Sonar integration (#1981)
It's a very complex project and needs to be fine tuned.
GitHub Actions (and GitHub in general) infrastructure [was failing](https://www.githubstatus.com/incidents/q07bfjh7jf1t) yesterday too.

rohanKanojia:
[merge]

PR #2025, opened 2020-02-26
Fix #2023

### LINKED ISSUE #2023: CRD Delete not working | Jackson deserialization of empty response not working
Hi,
We've a strange behaviuor. We're deleting a CRD in our tests, and the mocked server (and also the official kubernetes server) returns 200 with an empty body. 

At this point the client throws an exception, due to the fact that the empty body can't be deserialized to an hashmap. The problem arises in the following function of the class "RawCustomResourceOperationsImpl": 

`private Map<String, Object> makeCall(String url, String body, RawCustomResourceOperationsImpl.HttpCallMethod callMethod) {
        Request request = body == null ? this.getRequest(url, callMethod) : this.getRequest(url, body, callMethod);

        try {
            Response response = this.client.newCall(request).execute();
            Throwable var6 = null;

            Map var7;
            try {
                if (!response.isSuccessful()) {
                    throw requestFailure(request, createStatus(response));
                }

                var7 = (Map)this.objectMapper.readValue(response.body().string(), HashMap.class);
            } catch (Throwable var17) {
                var6 = var17;
                throw var17;
            } finally {
                if (response != null) {
                    if (var6 != null) {
                        try {
                            response.close();
                        } catch (Throwable var16) {
                            var6.addSuppressed(var16);
                        }
                    } else {
                        response.close();
                    }
                }

            }

            return var7;
        } catch (Exception var19) {
            throw KubernetesClientException.launderThrowable(var19);
        }
    }`

`var7 = (Map)this.objectMapper.readValue(response.body().string(), HashMap.class);` is the one that is giving problems to us, deserializing an empty string ("") to hashmap.

We're using fabric8 kubernetes client version 4.8.0 & kubernetes-server-mock 4.8.0

reste85:
Jackson is unable to deserialize an empty string (coming from an empty response body, which tipically occurs when you delete a CustomResource with the makeCall method defined in RawCustomResourceOperationsImpl (which is used when you use the RawCustomResource dsl).

Solution: Returning empty map when response body is an empty string

Fixes https://github.com/fabric8io/kubernetes-client/issues/2023

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/RawCustomResourceOperationsImpl.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/RawCustomResourceOperationsImpl.java
@@ -745,7 +745,11 @@ public class RawCustomResourceOperationsImpl extends OperationSupport {
     Request request = (body == null) ? getRequest(url, callMethod) : getRequest(url, body, callMethod);
     try (Response response = client.newCall(request).execute()) {
       if (response.isSuccessful()) {
-        return objectMapper.readValue(response.body().string(), HashMap.class);
+        String respBody = response.body().string();
+        if(Utils.isNullOrEmpty(respBody))
+          return new HashMap<>();
+        else
+          return objectMapper.readValue(respBody, HashMap.class);
       } else {
         throw requestFailure(request, createStatus(response));
       }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 ### 4.8-SNAPSHOT
 #### Bugs
 
+* Fix #2023: Class RawCustomResourceOperationsImpl can't handle HTTP responses with empty body coming from the k8s 
+cluster (Jackson deserialization error was throwed). This kind of response can be returned after executing operations 
+like the delete of a custom resource. 
+
 #### Improvements
 
 * Fix #2012: osgi: Allow the ManagedKubernetesClient to consume an available OAuthTokenProvider


rohanKanojia:
ok to test

reste85:
i see the IT tests are failing, any suggestion on that?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/RawCustomResourceOperationsImpl.java
```diff
@@ -745,7 +746,10 @@ private String getLabelsQueryParam(Map<String, String> labels) {
     Request request = (body == null) ? getRequest(url, callMethod) : getRequest(url, body, callMethod);
     try (Response response = client.newCall(request).execute()) {
       if (response.isSuccessful()) {
-        return objectMapper.readValue(response.body().string(), HashMap.class);
+        if(StringUtils.isBlank(response.body().string()))
```
rohanKanojia: Umm, won't `response.body().string()` close the response body?

reste85: correct, fixing it in few minutes ;)

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/RawCustomResourceOperationsImpl.java
```diff
@@ -34,6 +34,7 @@
 import okhttp3.Request;
 import okhttp3.RequestBody;
 import okhttp3.Response;
+import org.apache.commons.lang.StringUtils;
```
iocanel: commons-lang is an optional dependency.
We shouldn't use it unless we handle the case that its not present.

In any case we have our internal StringUtils called Strings and should always use that.

reste85: Ok, fixing it right now

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/RawCustomResourceOperationsImpl.java
```diff
@@ -28,6 +28,7 @@
 import io.fabric8.kubernetes.client.utils.Serialization;
 import io.fabric8.kubernetes.client.utils.Utils;
 import io.fabric8.kubernetes.client.utils.WatcherToggle;
+import io.fabric8.zjsonpatch.internal.guava.Strings;
```
iocanel: Please use this: https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/Utils.java#L289

manusa:
Please squash you commits too

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/RawCustomResourceOperationsImpl.java
reste85: Ok 👍

reste85:
> Please squash you commits too

done

reste85:
Ok all seems working... Sorry for the delay (was not so into in committing on an open source project 👍 )
Just a question: what about your release process? When usually those kind of fixes are going to be released?

Thank you again and have a nice day!

rohanKanojia:
@reste85 : We generally release after every 3-4 weeks.

reste85:
> @reste85 : We generally release after every 3-4 weeks.

Ok thank you, so i'll watch your release process to get the latest version ;) Thank you!

rohanKanojia:
@reste85 : It looks good to be merged Thanks a lot :+1: . Could you please add a line regarding the change to `CHANGELOG.md`?

reste85:
> @reste85 : It looks good to be merged Thanks a lot 👍 . Could you please add a line regarding the change to `CHANGELOG.md`?

Done ;)

rohanKanojia:
[merge]

PR #2033, opened 2020-02-27
Update model so that Step does not contain a Container but has the sa…

### LINKED ISSUE #1900: Cannot create a Task using the Tekton extension
I am trying to create a Task from this payload:

```
apiVersion: tekton.dev/v1alpha1
kind: Task
metadata:
  name: echo-hello-world1
spec:
  steps:
    - name: echo
      image: ubuntu
      command:
        - echo
      args:
        - "hello world"

```

I always go rejected by the server and noticed that various attributes for the step are stored as additionalProperties and are lost when editing the resource.
I looked at the code and noticed that the Step model class is defined as containing a Container where it is an extension of Container (from the Go code).

### LINKED ISSUE #1984: There is unknown field "container" in Tekton

jeffmaury:
…me attributes

Fixes #1900
Fixes #1984 

- Optimal solution would be to have Step inherits from Container but I'm not a generator expert

Signed-off-by: Jeff MAURY <jmaury@redhat.com>

### DIFF
--- a/extensions/tekton/generator/pkg/schemagen/generate.go
+++ b/extensions/tekton/generator/pkg/schemagen/generate.go
@@ -421,7 +421,7 @@ func (g *schemaGenerator) getStructProperties(t reflect.Type) map[string]JSONPro
 		desc := getFieldDescription(field)
 		omitEmpty := isOmitEmpty(field)
 		prop := g.getPropertyDescriptor(field.Type, desc, omitEmpty)
-		if field.Anonymous && field.Type.Kind() == reflect.Struct && (len(name) == 0 || strings.HasPrefix(name, "Common") || name == "PlanReference" || name == "TypeMeta") {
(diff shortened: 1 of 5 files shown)


rohanKanojia:
ok to test

manusa:
License is missing for task.json file, please run `mvn -N license:format`

rohanKanojia:
retest this please

rohanKanojia:
@Fabian-K: Hi, Could you please take a look at this PR? It's related to #1984

Fabian-K:
I don´t know much about the generator logic. The resulting tekton-schema.json looks good!

rohanKanojia:
[merge]

PR #2039, opened 2020-03-03
Fix #2009: Regression `withoutLabel()` stopped to work

### LINKED ISSUE #2009: Regression `withoutLabel()` stopped to work
Regression between `4.1.1` to `4.3.1`

```java
List<ConfigMap> configMapList = Stream.of(
		new ConfigMapBuilder("keep-label-map").addLabel(KEEP_LABEL, null).build(),
		new ConfigMapBuilder("null-label-map").build(),
		new ConfigMapBuilder("arbitrary-label-map").build()
).collect(Collectors.toList());

openshift.configMaps().withoutLabel(KEEP_LABEL).list().getItems().size() == 3

openshift.configMaps().withLabel(KEEP_LABEL).list().getItems().size() == 1

openshift.configMaps().withLabelNotIn(KEEP_LABEL).list().getItems().size() = 2
```

Same goes also for other resources:
```java
openshift.pods().list().getItems().size() == 2

openshift.pods().withoutLabel(OpenShift.KEEP_LABEL).list().getItems().size() == 2

openshift.pods().withLabel(OpenShift.KEEP_LABEL).list().getItems().size() == 1

openshift.pods().withLabelNotIn(OpenShift.KEEP_LABEL).list().getItems().size() == 1
```
Additional info
[1] https://github.com/xtf-cz/xtf/issues/306
[2] https://github.com/xtf-cz/xtf/pull/309

rohanKanojia:
Fix #2009

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
@@ -553,7 +553,7 @@ public class BaseOperation<T, L extends KubernetesResourceList, D extends Doneab
           sb.append(",");
         }
         Map.Entry<String, String[]> entry = iter.next();
-        if (entry.getValue() != null) {
+        if (Utils.isNotNull(entry.getValue())) {
           for (int i = 0; i < entry.getValue().length; i++) {
             if (i > 0) {
               sb.append(",");
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/Utils.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/Utils.java
@@ -302,6 +302,12 @@ public class Utils {
     return !(array == null || array.length == 0);
   }
 
+  public static <T> boolean isNotNull(T ...refList) {
+    return Optional.ofNullable(refList)
+      .map(refs -> Stream.of(refs).allMatch(Objects::nonNull))
+      .orElse(false);
+  }
+
   public static String getProperty(Map<String, Object> properties, String propertyName, String defaultValue) {
     String answer = (String) properties.get(propertyName);
     if (!isNullOrEmpty(answer)) {
--- a/kubernetes-client/src/test/java/io/fabric8/kubernetes/client/internal/UtilsTest.java
+++ b/kubernetes-client/src/test/java/io/fabric8/kubernetes/client/internal/UtilsTest.java
@@ -72,6 +72,8 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
 
@@ -162,6 +164,7 @@ public class UtilsTest {
 
   @Test
   public void testGetPluralFromKind() {
+    // Given
     Map<String, Class> pluralToKubernetesResourceMap = new HashMap<>();
     pluralToKubernetesResourceMap.put("bindings", Binding.class);
     pluralToKubernetesResourceMap.put("componentstatuses", ComponentStatus.class);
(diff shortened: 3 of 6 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
```diff
@@ -540,7 +540,7 @@ public String getLabelQueryParam() {
           sb.append(",");
         }
         Map.Entry<String, String[]> entry = iter.next();
-        if (entry.getValue() != null) {
+        if (entry.getValue() != null && Utils.isNotNull(entry.getValue())) {
```
manusa: First null check is redundant, `Utils#isNotNull`already handles that case
```suggestion
        if (Utils.isNotNull(entry.getValue())) {
```

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/Utils.java
```diff
@@ -60,6 +60,17 @@ private Utils() {
...
+  public static <T> boolean isNotNull(T[] refList) {
+    if (refList != null) {
+      for (T ref : refList) {
+        if (ref != null) {
+          return true;
+        }
+      }
+    }
+    return false;
+  }
```
manusa: Nit: Can be replaced by a less complex functional and declarative approach
```suggestion
  public static <T> boolean isNotNull(T... refList) {
    return Optional.ofNullable(refList)
      .map(refs -> Stream.of(refs).allMatch(Objects::nonNull))
      .orElse(false);
  }
```

It would also be good to move this function near the rest of `isNull*`methods

manusa: Add some Unit Tests to verify and document the behavior of this function:
e.g.
```
  @Test
  @DisplayName("isNotNull, null, should return false")
  public void isNotNullNullTest() {
    // When
    final boolean result = Utils.isNotNull((Object)null);
    // Then
    assertFalse(result);
  }

  @Test
  @DisplayName("isNotNull, some null values, should return false")
  public void isNotNullSomeAreNullTest() {
    // When
    final boolean result = Utils.isNotNull("Not Null", null);
    // Then
    assertFalse(result);
  }

  @Test
  @DisplayName("isNotNull, no null values, should return true")
  public void isNotNullNoneAreNullTest() {
    // When
    final boolean result = Utils.isNotNull("Not Null", "Not Null either");
    // Then
    assertTrue(result);
  }
```

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/internal/UtilsTest.java
```diff
@@ -110,4 +112,36 @@ public void interpolateStringWithParametersTest() {
     // Then
     assertEquals("This is a \"template string\" and the following is code ${NOT_REPLACED}: '1' === '1'; /* END */", result);
   }
+
+  @Test
+  @DisplayName("isNotNull, null, should return false")
+  public void isNotNull() {
+    // When
+    final boolean result1 = Utils.isNotNullOrEmpty((Map)null);
```
manusa: These are not tests for the implemented method `isNotNull`

You can keep these in a sepearete method `isNotNullOrEmpty` (with changed DisplayName too) to add more coverage, but please do add the nullcheck tests for the currently being implemented method `isNotNull`

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/internal/UtilsTest.java
```diff
@@ -110,4 +112,36 @@ public void interpolateStringWithParametersTest() {
...
+    assertFalse(result1);
+    assertFalse(result2);
+    assertFalse(result3);
+  }
+
+  @Test
+  @DisplayName("isNotNull, some null values, should return false")
+  public void isNotNullSomeAreNullTest() {
+    // When
+    final boolean result = Utils.isNotNullOrEmpty(new String[] {"notNullObj", null, null});
```
manusa: Same as the previous comment. We are checking `isNotNull`

rohanKanojia:
retest this please

rohanKanojia:
[merge]

PR #2040, opened 2020-03-04
Fix #1247: URL parameters are not escaped.

### LINKED ISSUE #1247: URL parameters are not escaped in many methods
In some cases this makes it impossible to pass certain arguments to APIs. In other cases, if a user of this client is handling untrusted user input, it can be a security issue since end users would be able to inject arbitrary unintended parameters into URLs.

If you look at the exec function in https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java the code for putting command arguments in the URL is currently written as:
```java
    public ExecWatch exec(String... command) {
        String[] actualCommands = command.length >= 1 ? command : EMPTY_COMMAND;

        sb.append("exec?command=");

        boolean first = true;
        for (String cmd : actualCommands) {
            if (first) {
                first = false;
            } else {
                sb.append("&command=");
            }
            sb.append(cmd);
        }
```

So if you call it with:
```java
exec("/bin/sh", "-c", "echo test 2>&1");
```
the command errors out because `&` gets interpreted by the api servers as starting another query parameter and `2>` alone is invalid shell syntax.

Exec isn't the only place this is an issue, it happens in a lot of other places too. Is there a reason to construct  URLs manually with a StringBuilder instead of using okhttp's HttpUrl.Builder class? http://square.github.io/okhttp/3.x/okhttp/okhttp3/HttpUrl.Builder.html

dev-gaur:
#1247

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java
@@ -66,6 +66,7 @@ import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
 import okhttp3.ResponseBody;
+import okhttp3.HttpUrl;
 
 import static io.fabric8.kubernetes.client.utils.OptionalDependencyWrapper.wrapRunWithOptionalDependency;
 
@@ -140,6 +141,7 @@ public class PodOperationsImpl extends HasMetadataOperation<Pod, PodList, Doneab
   protected String getLogParameters() {
     StringBuilder sb = new StringBuilder();
     sb.append("log?pretty=").append(withPrettyOutput);
+
     if (containerId != null && !containerId.isEmpty()) {
       sb.append("&container=").append(containerId);
     }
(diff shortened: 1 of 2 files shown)


rohanKanojia:
ok to test

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java
```diff
@@ -138,29 +139,40 @@ public PodOperationContext getContext() {
     return (PodOperationContext) context;
   }
   protected String getLogParameters() {
+    HttpUrl.Builder httpUrlBuilder = new HttpUrl.Builder();
     StringBuilder sb = new StringBuilder();
+    httpUrlBuilder.addPathSegment("log");
+    httpUrlBuilder.addQueryParameter("pretty", String.valueOf(withPrettyOutput));
     sb.append("log?pretty=").append(withPrettyOutput);
```
rohanKanojia: Why are you doing this?

oscerd: Yes a bit weird

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationsImpl.java
```diff
@@ -138,29 +139,40 @@ public PodOperationContext getContext() {
     return (PodOperationContext) context;
   }
   protected String getLogParameters() {
+    HttpUrl.Builder httpUrlBuilder = new HttpUrl.Builder();
     StringBuilder sb = new StringBuilder();
```
rohanKanojia: Is this still WIP? Why are you keeping `StringBuilder` too? I don't see it getting used anywhere.

dev-gaur: forgot to remove it .sorry

dev-gaur: I've rolled back to using stringbuilder here only.. since httpurl isn't an ideal replacement for stringbuilder in this method.

dev-gaur:
Opened a new PR #2042. I mistakenly created this PR from fabric8io/fix1247 to fabric8io/master rather than from my own fork.

PR #2042, opened 2020-03-04
Fix #1247: URL parameters are not escaped.

### LINKED ISSUE #1247: URL parameters are not escaped in many methods

dev-gaur:
#1247

### DIFF
--- a/kubernetes-itests/src/test/java/io/fabric8/kubernetes/PodIT.java
+++ b/kubernetes-itests/src/test/java/io/fabric8/kubernetes/PodIT.java
@@ -182,6 +182,18 @@ public class PodIT {
   }
 
   @Test
+  public void readFileEscapedParams() throws IOException {
+    // Wait for resources to get ready
+    ReadyEntity<Pod> podReady = new ReadyEntity<>(Pod.class, client, pod1.getMetadata().getName(), currentNamespace);
+    await().atMost(30, TimeUnit.SECONDS).until(podReady);
+    ExecWatch watch = client.pods().inNamespace(currentNamespace).withName(pod1.getMetadata().getName()).writingOutput(System.out).exec("sh", "-c", "echo 'H$ll* (W&RLD}' > /msg");
+    try (InputStream is = client.pods().inNamespace(currentNamespace).withName(pod1.getMetadata().getName()).file("/msg").read())  {
+      String result = new BufferedReader(new InputStreamReader(is)).lines().collect(Collectors.joining("\n"));
+      assertEquals("H$ll* (W&RLD}", result);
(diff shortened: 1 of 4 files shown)


rohanKanojia:
ok to test

kubernetes-itests/src/test/java/io/fabric8/kubernetes/PodIT.java
```diff
@@ -169,6 +169,39 @@ public void onClose(int i, String s) {
...
+          logger.info("Shell barfed");
+          execLatch.countDown();
+        }
+
+        @Override
+        public void onClose(int i, String s) {
+          logger.info("Shell closed");
+          execLatch.countDown();
+        }
+      }).exec("echo 'H$ll* (W&RLD}'"); // These arguments in the exec command should be escaped.
```
manusa: I think you need to separate the arguments. 
Also the `}` in `(W&RLD}` looks like a typo

dev-gaur: No I added that bracket just to demonstrated that all of these characters are being escaped properly.

manusa: I understand, but then the assertion was inconsistent (changed now).

rohanKanojia:
[merge]

PR #2069, opened 2020-03-16
Fix #2046: OpenshiftClient getVersion returns null for Openshift ContainerPlatform v4

### LINKED ISSUE #2046: OpenshiftClient  getVersion returns null for Openshift ContainerPlatform v4
While working on an issue on the IntelliJ openshift connector (https://github.com/redhat-developer/intellij-openshift-connector/issues/119), I noticed that the client.getversion() does not return proper version info for V4

while on v3, the endpoint is hardcoded to "version/openshift" and return a result, the same endpoint for v4 does return a 404, and then the versionInfo is null.

rohanKanojia:
Fix #2046 

Right now I'm just converting Openshift v4's `ClusterVersion` object to `VersionInfo` which doesn't seem very elegant to me also :confused: . Shall I introduce another method in DSL like this:

```
ClusterVersionList clusterVersion()
```
But then it would be user's responsibility to fetch all the necessary parts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Fix #2057: Fix jar and osgi bundle generation for extensions
 * Fix #2075: KubernetesDeserializer registration for CustomResources
 * Fix #2078: watchLog for Deployment and StatefulSet
+* Fix #2046: OpenshiftClient getVersion returns null for Openshift ContainerPlatform v4
 
 #### Improvements
 * Fix #2019: Added CustomResourceCrudTest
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 Fix #2057: Fix jar and osgi bundle generation for extensions
 Fix #2075: KubernetesDeserializer registration for CustomResources 
 Fix #2078: watchLog for Deployment and StatefulSet 
+* Fix #2057: Fix jar and osgi bundle generation for extensions
```
manusa: This line has been duplicated (see line 5)

manusa:
[merge]

PR #2072, opened 2020-03-18
Fix #2034: Quantity.equals mishandles fractions

### LINKED ISSUE #2034: Quantity.equals mishandles fractions
#1921 introduced https://github.com/fabric8io/kubernetes-client/blob/53f70355d6b140ae4d25f0fe8704aee064fda5d7/kubernetes-model/kubernetes-model/src/main/java/io/fabric8/kubernetes/api/model/Quantity.java#L194-L196 which truncates `BigDecimal` to `BigInteger`. While this makes sense for, say, memory or disk, it does not make sense for CPU measurements where fractions of a unit are meaningful. Thus

```java
assertNotEquals(new Quantity("1"), new Quantity("2"));
```

passes as expected, yet

```java
assertNotEquals(new Quantity("100m"), new Quantity("200m"));
```

fails:

```
Values should be different. Actual: Quantity(amount=200, format=m, additionalProperties={})
```

Simply dropping the cast to `BigInteger` seems to fix this:

```diff
diff --git kubernetes-model/kubernetes-model/src/main/java/io/fabric8/kubernetes/api/model/Quantity.java kubernetes-model/kubernetes-model/src/main/java/io/fabric8/kubernetes/api/model/Quantity.java
index 9438f4aec..417e3e92e 100644
--- kubernetes-model/kubernetes-model/src/main/java/io/fabric8/kubernetes/api/model/Quantity.java
+++ kubernetes-model/kubernetes-model/src/main/java/io/fabric8/kubernetes/api/model/Quantity.java
@@ -192,8 +192,7 @@ public class Quantity  implements Serializable {
 
     Quantity quantity = (Quantity) o;
     return getAmountInBytes(this)
-      .toBigInteger()
-      .equals(getAmountInBytes(quantity).toBigInteger());
+      .compareTo(getAmountInBytes(quantity)) == 0;
   }
 
   @Override
```

While it is dangerous to check exact equality when dealing with floating-point numbers, in this case we are dealing with arbitrary-precision numbers https://github.com/fabric8io/kubernetes-client/blob/53f70355d6b140ae4d25f0fe8704aee064fda5d7/kubernetes-model/kubernetes-model/src/main/java/io/fabric8/kubernetes/api/model/Quantity.java#L127 which should not have rounding errors.

While we are here, I would recommend

```java
@Override
public String toString() {
    StringBuilder b = new StringBuilder();
    if (getAmount() != null) {
        b.append(getAmount());
    }
    if (getFormat() != null) {
        b.append(getFormat());
    }
    return b.toString();
}
```

(I would offer a pull request but I seem to be unable to consistently `mvn compile` this module—fails with weird source generation errors after several minutes of computation—and the IDEs I have tried choke on it.)

rohanKanojia:
Fix #2034

### DIFF
--- a/kubernetes-model/kubernetes-model/src/main/java/io/fabric8/kubernetes/api/model/Quantity.java
+++ b/kubernetes-model/kubernetes-model/src/main/java/io/fabric8/kubernetes/api/model/Quantity.java
@@ -50,6 +50,7 @@ import java.io.Serializable;
 @Buildable(editableEnabled = false, validationEnabled = false, generateBuilderPackage=true, builderPackage = "io.fabric8.kubernetes.api.builder", inline = @Inline(type = Doneable.class, prefix = "Doneable", value = "done"))
 public class Quantity  implements Serializable {
 
+  private static final String AT_LEAST_ONE_DIGIT_REGEX = ".*\\d+.*";
   private String amount;
   private String format;
   private Map<String, Object> additionalProperties = new HashMap<String, Object>();
@@ -118,10 +119,9 @@ public class Quantity  implements Serializable {
     Quantity amountFormatPair = parse(value);
     String formatStr = amountFormatPair.getFormat();
     // Handle Decimal exponent case
-    if ((formatStr.startsWith("e") || formatStr.startsWith("E")) &&
-        formatStr.length() > 1) {
+    if ((formatStr.matches(AT_LEAST_ONE_DIGIT_REGEX)) && formatStr.length() > 1) {
       int exponent = Integer.parseInt(formatStr.substring(1));
-      return new BigDecimal("10").pow(exponent).multiply(new BigDecimal(amountFormatPair.getAmount()));
+      return new BigDecimal("10").pow(exponent, MathContext.DECIMAL64).multiply(new BigDecimal(amountFormatPair.getAmount()));
     }
 
     BigDecimal digit = new BigDecimal(amountFormatPair.getAmount());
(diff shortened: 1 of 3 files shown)


rohanKanojia:
@jglick: Could you please take a look too? I've created this PR as per the changes proposed by you.

manusa:
Please, use this PR to add complete code coverage to [Quantity#getAmountInBytes](https://sonarcloud.io/code?id=kubernetes_client&selected=kubernetes_client%3Akubernetes-model%2Fkubernetes-model%2Fsrc%2Fmain%2Fjava%2Fio%2Ffabric8%2Fkubernetes%2Fapi%2Fmodel%2FQuantity.java&line=102).

Please, also add enough test cases (including corner-cases) for Quantity#equals.

Tests should be implemented declaratively and in a way that document the behavior of this class.

Too many bugs have been open for unexpected behavior of this class, let's anticipate them.

kubernetes-model/kubernetes-model/src/main/java/io/fabric8/kubernetes/api/model/Quantity.java
```diff
@@ -50,6 +50,7 @@
 @Buildable(editableEnabled = false, validationEnabled = false, generateBuilderPackage=true, builderPackage = "io.fabric8.kubernetes.api.builder", inline = @Inline(type = Doneable.class, prefix = "Doneable", value = "done"))
 public class Quantity  implements Serializable {
 
+  public static final String AT_LEAST_ONE_DIGIT_REGEX = ".*\\d+.*";
```
manusa: This one should be private if possible
```suggestion
  private static final String AT_LEAST_ONE_DIGIT_REGEX = ".*\\d+.*";
```

rohanKanojia:
[merge]

PR #2099, opened 2020-03-31
Fix #1586: Replace of Job should add extra selector property to the resource

### LINKED ISSUE #1586: Replace of Job should add extra selector property to the resource
Hi,

I'm trying to use the `createOrReplace` method available on `client.inNamespace(namespace).batch().jobs()`.

I load a job definition from a file and I expect it to be created or replaced in case it already exists. The assumption is that kubernetes will re-run the job in case the replacement is different from the one already deployed.

The problem is that the API of kubernetes answers with 422 and the following type of errors:
```
io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: PUT at: https://cluster-xxxxxx.hcp.westeurope.azmk8s.io/apis/batch/v1/namespaces/dev/jobs/myjob. Message: Job.batch "myjob" is invalid: [spec.selector: Required value, spec.template.metadata.labels: Invalid value....
```

So I added a selector on the job, but then I get the following error:
```
io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: PUT at: https://cluster-xxxxxx.hcp.westeurope.azmk8s.io/apis/batch/v1/namespaces/dev/jobs/optigrid. Message: Job.batch "myjob" is invalid: [spec.template.metadata.labels[controller-uid]: Required value: must be '36910c27-8e83-11e9-bd3c-72b51ec2d813', spec.template.metadata.labels[job-name]: Required value: must be 'myjob', spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/name":"app", "app.kubernetes.io/instance":"app1"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: `selector` not auto-generated, spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/instance":"app1", "app.kubernetes.io/name":"app"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable, ....
```

I feel like `replace` should be reading from the existing job and retrieve the needed values for the selector. A bit like how `kubectl apply` is able to update a resource by merging it with the upstream one.

Without this, I don't see how `createOrReplace` can be used successfuly, but maybe I am mistaken about what `createOrReplace` is there for?

rohanKanojia:
Fix #1586

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/JobOperationsImpl.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/JobOperationsImpl.java
@@ -223,4 +223,24 @@ public class JobOperationsImpl extends HasMetadataOperation<Job, JobList, Doneab
     }
     return null;
   }
+
+  @Override
+  public Job replace(Job job) {
+    if (job == null) {
+      job = getItem();
+    }
+    // Fetch item from server and patch Selector and PodTemplate
+    // metadata in case not present already in order to avoid 422
+    Job jobFromServer = fromServer().get();
+    if (job.getSpec().getSelector() == null) {
+      job.getSpec().setSelector(jobFromServer.getSpec().getSelector());
+    }
+    if (job.getSpec().getTemplate().getMetadata() != null) {
+      job.getSpec().getTemplate().getMetadata().setLabels(jobFromServer.getSpec().getTemplate().getMetadata().getLabels());
+    } else {
+      job.getSpec().getTemplate().setMetadata(jobFromServer.getSpec().getTemplate().getMetadata());
+    }
+
+    return super.replace(job);
+  }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### 4.9-SNAPSHOT
 #### Bugs
+* Fix #1586: Replace of Job should add extra selector property to the resource
 * Fix #2034: Quantity.equals mishandles fractions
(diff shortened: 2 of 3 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/JobOperationsImpl.java
```diff
@@ -223,4 +223,24 @@ public LogWatch watchLog(OutputStream out) {
...
+      job = getItem();
+    }
+    // Fetch item from server and patch Selector and PodTemplate
+    // metadata in case not present already in order to avoid 422
+    Job jobFromServer = fromServer().get();
+    if (job.getSpec().getSelector() == null) {
+      job.getSpec().setSelector(jobFromServer.getSpec().getSelector());
+    }
+    if (job.getSpec().getTemplate().getMetadata() != null) {
+      job.getSpec().getTemplate().getMetadata().setLabels(jobFromServer.getSpec().getTemplate().getMetadata().getLabels());
```
manusa: Is it possible that the replacing job (new) will contain additional labels?
If so, shouldn't we purAll labels from server and keep the current. 

What if replacing job is removing some labels, is this the expected behavior?

rohanKanojia: `Job` resource's labels are different from `PodTemplate`'s labels. I think these are assigned from kube-apiserver after initial creation.

manusa:
[merge]

PR #2100, opened 2020-04-01
Fix #2088: Moved Ingress to networking.k8s.io/v1beta1

### LINKED ISSUE #2088: Does version 4.9.0 supports to CRUD Ingress which 'apiVersion: networking.k8s.io/v1beta1'
Example yaml:
```
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
....
```

rohanKanojia:
Fix #2088 

- Added model and DLS support for networking.k8s.io/v1beta1 Ingress
- Modified kubernetes client handler generator logic to generate handlers with respect to apiVersions
      in order to avoid collisions


rohanKanojia:
Umm, looks like `extensions/v1beta1` is still used by Openshift. But I'm not sure anyone uses `Ingress` in Openshift when we already have `Routes`

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/URLFromServiceUtil.java
```diff
@@ -17,15 +17,16 @@
...
+import io.fabric8.kubernetes.api.model.networking.v1beta1.IngressSpec;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
```
dev-gaur, line 29: can you take care of the wild card import?

kubernetes-model/cmd/generate/generate.go
```diff
@@ -311,6 +312,7 @@ func main() {
     {"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1", "", "io.fabric8.kubernetes.api.model.apiextensions", "kubernetes_apiextensions_"},
     {"k8s.io/apimachinery/pkg/apis/meta/v1", "", "io.fabric8.kubernetes.api.model", "kubernetes_apimachinery_"},
     {"k8s.io/api/networking/v1", "networking.k8s.io", "io.fabric8.kubernetes.api.model.networking", "kubernetes_networking_"},
```
dev-gaur, line 314: Do we have test coverage for Ingress under networking.k8s.io/v1 as well? 
I'm not sure if we need it or not but just asking.

rohanKanojia: I think there is `IngressTest` already

kubernetes-client/src/main/resources/resource-handler-services.vm
```diff
@@ -14,5 +14,10 @@
  * limitations under the License.
  *#
 #foreach ($m in ${model.values()})
-io.fabric8.kubernetes.client.handlers.${m.name}Handler
+#foreach ($annotation in $m.annotations)
+#if ($annotation.getClassRef().getName().equals("PackageSuffix"))
+#set ($packageSuffix = $annotation.getParameters().get("value"))
+#end
```
manusa, line 20: nit: If possible add some indentation

rohanKanojia:
[merge]

PR #2103, opened 2020-04-02
Fix #1987: Added example of Task and TaskRun in tekton with updated model

### LINKED ISSUE #1987: How can I use "withPipelineResourceBinding()" to define taskRun resources?
**Issue description**
I recently use tekton-client extension to make taskRun with Java.

I tried to define new input resources, and I found withPipelineResourceBinding() function in ".withNewSpec().withNewInputs().addNewResource()".

There are "withName(), withNewResourceSpec()" functions in the above funtion, so that I thought this is to define resources.

However, when I use the function, a runtime error occurs.

> Message addmission webhook "webhook.tekton.dev" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "pipelineResourceBinding".

How could I use this function properly?

Or, if that function is not for resource definition, what is the right function?

Thanks for replying.

rohanKanojia:
Fix #1987 
+ Added TaskRunCreate and TaskCreate example
+ Updated Tekton model to v0.11.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,11 +12,13 @@
 * Fix #2046: OpenshiftClient getVersion returns null for Openshift ContainerPlatform v4
 
 #### Improvements
+* Fix #1987: Added an example for Task and TaskRun with updated model
 * Fix #2019: Added CustomResourceCrudTest
 * Fix #2054: JobExample doesn't work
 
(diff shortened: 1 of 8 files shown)


manusa:
[merge]

manusa:
[merge]

Issue #2108, opened 2020-04-03
Split Kubernetes Model into smaller jars corresponding to each apiGroup


rohanKanojia:
Right now our model has lots of classes and if we add Sundrio helper classes too they become a lot. @iocanel checked out number of classes generated and these were his findings:
```
It contains:
- 38 Resource definitions for Kuberntes 
- 20 Resource definitions for Openshift

These objects include properties of around 578 different types.
We do generate 3 classes for each to satisfy the builder needs (Fluent, FluentImpl & Builder).

Which takes us close to 2312 classes.

To support nesting in builder we also use nested classes (should be 2 nested classes per `buildable` property). In total there should be around 5000 of those.

The remaining 2000 classes, could possibly be considered part of optional features, and these are the only classes we could possibly get rid of (by restricting the model DSL).
```
But we don't always need all the classes in most of the cases. This seems to be creating problems in [Quarkus Kubernetes Client Extension](https://github.com/quarkusio/quarkus/tree/master/extensions/kubernetes-client) where the build is taking too much time on JDK11. 

As per discussions in Quarkus channel, we should probably split kubernetes model into several modules each specific to an `apiVersion` in Kubernetes. So in the end `kubernetes-model` would consist of several small modules like this:
```
  <modules>
       <module>kubernetes-model-openshift</module>
       <module>kubernetes-model-admission-v1beta1</module>
       <module>kubernetes-model-apps-v1</module>
       <module>kubernetes-model-batch-v1</module>
       <module>kubernetes-model-networking-v1</module>
       <module>kubernetes-model-storage-v1</module>
       <module>kubernetes-model-storage-v1beta1</module>
       <module>kubernetes-model-apimachinery-meta-v1</module>
       <module>kubernetes-model-networking-v1</module>
       <module>kubernetes-model-autoscaling-v1</module>
       
       ...
  </modules>
```

PR #2110, opened 2020-04-04
Fix #1996: Added module-info.java files for different modules in project

### LINKED ISSUE #1996: Module based library ( support for Java 9+ )
I'm using Java 11 and trying to create a lightweight JRE . Is there a new version for kubernetes-client which uses modules  ? It will make uptake very easy in java 9 + .

rohanKanojia:
Fix #1996


manusa:
## Problem with `ResourceSelector`

There is a package name collision in sundr.io modules `sundr.codegen`and `builder.annotations` as transitive package `io.sundr.builder`  is available/can be read from both.

This is a problem as the conflicting annotation (`io.sundr.codegen.annotations.ResourceSelector`) can't be read as it's not exported in the `transform.annotations` package so it needs a require from `sundr.codegen` (or at least this is my guess). As stated in the previous paragraph, if we require both modules we get the conflict.

In order to fix this, I changed the sundr.io library as in this branch: https://github.com/manusa/sundrio/tree/kc/1996-java-modules

And also did these additional changes to FKC: https://github.com/manusa/sundrio/tree/kc/1996-java-modules

Now I can get past the `IncompleteAnnotationException` problem.

But we reach a new issue as `kubernetes.properties` file is missing (it's not being generated).
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project kubernetes-client: Fatal error compiling: java.lang.RuntimeException: java.io.FileNotFoundException: kubernetes.properties -> [Help 1]
```

manusa:
## Stopper

When fixing `knative-model` I realized that unless we do some major refactor on sundr.io it will be impossible to create modules for the [split model](https://github.com/fabric8io/kubernetes-client/issues/2108) as they all generate builders in the same package hence all modules will export a same package (something forbidden in java modules).

So I'm halting the work on this.

fransguelinckx:
Hi, I was wondering if there is any news about further steps on this PR?

PR #2136, opened 2020-04-15
Modify the dependency management mode of generator to mod

Lionelpang:
delete the dep and use the go mod to manage the dependeces


Lionelpang:
> a bit of context would be of help.
> 
> * Why?
> * Advantages?

When using the generator to update the model, I think it is too cumbersome to use dep to manage the dependency construction of golang, so I will change the secondary dependency management to the mod mode of golang, and then generate the generator without gopath configuration. Submit this modification. I hope to be able to merge this modification to simplify the way generator is built.

rohanKanojia:
ok to test

rohanKanojia:
I think this is a good change since go modules are now preferred way of dependency management in the go world.

Thanks a lot for this.

kubernetes-model/Makefile
```diff
@@ -22,7 +22,7 @@ build: gobuild
 	mvn clean install
 
 gobuild:
-	dep ensure -v
-	CGO_ENABLED=0 GO15VENDOREXPERIMENT=1 go build -a ./cmd/generate/generate.go
+	# dep ensure -v
```
rohanKanojia, line 25: Could you please remove this comment?

Lionelpang: Comments have been submitted

rohanKanojia:
@Lionelpang : Could you please squash your commits so that we can proceed with merging your PR?

Lionelpang:
> @Lionelpang : Could you please squash your commits so that we can proceed with merging your PR?

just add the changlog and remove the commen code。need to Launch a new PR？

rohanKanojia:
@Lionelpang : No no, You don't need to lauch a new PR. You just need to squash your commits into one.

rohanKanojia:
anyway, let me do just squash and merge instead. I want to rebase #2137 with this.

PR #2137, opened 2020-04-15
Fix #2108: Split Kubernetes Model into smaller Jars

### LINKED ISSUE #2108: Split Kubernetes Model into smaller jars corresponding to each apiGroup

rohanKanojia:
Fix #2108 

Divides Kubernetes Model into the following jars:
```
  <modules>
    <module>kubernetes-model-core</module>
    <module>kubernetes-model-rbac</module>
    <module>kubernetes-model-admissionregistration</module>
    <module>kubernetes-model-apiextensions</module>
    <module>kubernetes-model-apps</module>
    <module>kubernetes-model-autoscaling</module>
    <module>kubernetes-model-batch</module>
    <module>kubernetes-model-certificates</module>
    <module>kubernetes-model-coordination</module>
    <module>kubernetes-model-discovery</module>
    <module>kubernetes-model-events</module>
    <module>kubernetes-model-extensions</module>
    <module>kubernetes-model-networking</module>
    <module>kubernetes-model-metrics</module>
    <module>kubernetes-model-policy</module>
    <module>kubernetes-model-scheduling</module>
    <module>kubernetes-model-settings</module>
    <module>kubernetes-model-storageclass</module>
    <module>openshift-model</module>
  </modules>

```


gastaldi:
Can we still have `kubernetes-model` as a POM just depending on these modules just to avoid breaking existing consumers of this library?

geoand:
This is awesome!

gastaldi:
> This is awesome!

Indeed it is, great job @rohanKanojia, as usual :wink:

kubernetes-model/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/KubernetesList.java
```diff
@@ -43,7 +44,7 @@
 })
 @JsonDeserialize(using = JsonDeserializer.None.class)
 @Buildable(editableEnabled = false, validationEnabled = false, generateBuilderPackage=true, builderPackage = "io.fabric8.kubernetes.api.builder", inline = @Inline(type = Doneable.class, prefix = "Doneable", value = "done"))
-public class KubernetesList extends BaseKubernetesList implements KubernetesResource {
+public class KubernetesList extends BaseKubernetesList implements Serializable {
```
manusa, line 47: Why was this changed?

kubernetes-model/kubernetes-model-core/src/test/java/io/fabric8/kubernetes/api/model/ListOptionsTest.java
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.fabric8.kubernetes.api.model;
+
+public class ListOptionsTest {
```
manusa, line 18: Either remove the class or add some tests to it

kubernetes-model/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/KubernetesList.java
rohanKanojia: looks like a mistake. I would revert it to it's previous state

kubernetes-model/kubernetes-model-core/src/test/java/io/fabric8/kubernetes/api/model/StatusTest.java
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.fabric8.kubernetes.api.model;
+
+public class StatusTest {
```
manusa, line 18: Either remove the class or add some tests to it

kubernetes-model/kubernetes-model-coordination/src/main/java/io/fabric8/kubernetes/api/model/coordination/v1/LeaseSpec.java
```diff
@@ -49,7 +49,7 @@
 @JsonDeserialize(using = com.fasterxml.jackson.databind.JsonDeserializer.None.class)
 @ToString
 @EqualsAndHashCode
-@Buildable(editableEnabled = false, validationEnabled = false, generateBuilderPackage = true, lazyCollectionInitEnabled = false, builderPackage = "io.fabric8.kubernetes.api.builder", inline = @Inline(type = Doneable.class, prefix = "Doneable", value = "done"))
+@Buildable(editableEnabled = false, validationEnabled = false, generateBuilderPackage = false, lazyCollectionInitEnabled = false, builderPackage = "io.fabric8.kubernetes.api.builder", inline = @Inline(type = Doneable.class, prefix = "Doneable", value = "done"))
```
manusa, line 52: Why is this changed to false? Class generated with Annotator has it true

kubernetes-model/openshift-model/src/main/java/io/fabric8/openshift/api/model/Template.java
```diff
@@ -61,7 +61,7 @@
 @JsonDeserialize(using = JsonDeserializer.None.class)
 @EqualsAndHashCode
 @ToString
-@Buildable(editableEnabled = false, validationEnabled = false, generateBuilderPackage=true, builderPackage = "io.fabric8.kubernetes.api.builder", inline = @Inline(type = Doneable.class, prefix = "Doneable", value = "done"))
+@Buildable(editableEnabled = false, validationEnabled = false, generateBuilderPackage=false, builderPackage = "io.fabric8.kubernetes.api.builder", inline = @Inline(type = Doneable.class, prefix = "Doneable", value = "done"))
```
manusa, line 64: Why is this changed to false? Class generated with Annotator has it true

kubernetes-model/kubernetes-model-coordination/src/main/java/io/fabric8/kubernetes/api/model/coordination/v1/LeaseSpec.java
rohanKanojia: Actually, I think this flag is for generating sundrio helper classes again. In `kubernetes-model-core`, we use KubernetesCoreTypeAnnotator which sets `generateBuilderPackage=true`, rest of the models are simply depending on `kubernetes-model-core` so we don't need to do it in this case. 

Earlier everything was under one module that's why earlier it's `generateBuilderPackage` was set to `true`

kubernetes-model/kubernetes-model-batch/pom.xml
```diff
@@ -0,0 +1,235 @@
...
+          </execution>
+        </executions>
+        <dependencies>
+          <dependency>
+            <groupId>io.fabric8</groupId>
+            <artifactId>kubernetes-model-annotator</artifactId>
+            <version>${project.version}</version>
+          </dependency>
+        </dependencies>
+      </plugin>
```
manusa, line 107: Make this pom consistent with the others (remove profile + leave just minimal configuration entry for the plugin definition)

kubernetes-model/kubernetes-model-certificates/pom.xml
```diff
@@ -0,0 +1,235 @@
...
+          </execution>
+        </executions>
+        <dependencies>
+          <dependency>
+            <groupId>io.fabric8</groupId>
+            <artifactId>kubernetes-model-annotator</artifactId>
+            <version>${project.version}</version>
+          </dependency>
+        </dependencies>
+      </plugin>
```
manusa, line 107: Make this pom consistent with the others (remove profile + leave just minimal configuration entry for the plugin definition)

kubernetes-model/kubernetes-model-coordination/src/main/java/io/fabric8/kubernetes/api/model/coordination/v1/LeaseSpec.java
manusa: These builders may actually be used by client users. If builders were being generated before, they should be still be generated. So `KubernetesTypeAnnotator` should set generateBuilderPackage=true too in order not to break compatibility.

Or is there any reason why this flag was set to false in `KubernetesTypeAnnotator`?

rohanKanojia: By builder helper classes, I didn't mean builder itself. I meant these
Screenshot from 2020-04-17 12-18-53

If you look at extensions' model they are using `kubernetes-model-core` and they have `generateBuilderPackage=false`. That way these classes won't be generated again and would be consumed from `kubernetes-model-core`
Screenshot from 2020-04-17 12-19-04

manusa: :+1:  nice explanation.
This may also help when dealing with the java 9 modularization.

kubernetes-model/openshift-model/src/main/java/io/fabric8/openshift/api/model/Template.java
manusa: See https://github.com/fabric8io/kubernetes-client/pull/2137#discussion_r410024370

iocanel:
Let's get this in and then we can start thinking of ways we can apply the same principal to the client modules too.

We could possibly have lots of thin client jars that could be combined into the existing `kubernetes-client` and `openshift-client`. 

As we already have the notion of extensions baked into the client, it should be relatively easy to do so.

Again great job @rohanKanojia !

rohanKanojia:
[merge]

PR #2142, opened 2020-04-17
Javac release compiler flag

Fabian-K:
The --release compiler flag is required when building the project with a jdk > 8 to produce the right byte code for jdk8 (see https://github.com/eclipse/jetty.project/issues/3244#issuecomment-495322586).

Otherwise, there are issues e.g. with the PortForwarderWebsocket.


rohanKanojia:
ok to test

Fabian-K:
Missed that the flag is only available starting with jdk9 🙈. This should now work!

rohanKanojia:
[merge]

PR #1600, opened 2019-06-22
Fix #1565: CRD's Enums are prefixed with "Raw: " keyword

rohanKanojia:
Fixes #1565

PR #2151, opened 2020-04-20
Fix json support for openApiV3Schema default, enum, example fields

### LINKED ISSUE #2144: JSONSchemaProps does not support properties of type boolean with default value
Here is a valid CRD spec:

```yaml
apiVersion: apiextensions.k8s.io/v1beta1
metadata:
  name: widgets.test.com
spec:
  group: test.com
  names:
    kind: Widget
    listKind: WidgetList
    plural: widgets
    singular: widget
  scope: Namespaced
  version: v1
  versions:
    - name: v1
      served: true
      storage: true
  preserveUnknownFields: false
  subresources:
    status: {}
  validation:
    openAPIV3Schema:
      type: object
      properties:
        spec:
          type: object
          properties:
            edges:
              type: integer
            hollow:
              type: boolean
              default: false
            dimensions:
              type: object
              properties:
                x:
                  type: integer
                y:
                  type: integer
                zoom:
                  type: boolean
                  default: false
            edgeTemplates:
              type: array
              items:
                type: object
                properties:
                  spiky:
                    type: boolean
                    default: true
```

This is accepted by kubernetes server version 1.15.10. However, this is unparseable via fabric8 CustomResourceDefinition.class:

```java
String yaml = "insert above";
Serialization.unmarshal(yaml, CustomResourceDefinition.class);
```

This fails with the following MismatchedInputException:

```
Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot construct instance of `io.fabric8.kubernetes.api.model.apiextensions.JSON` (although at least one Creator exists): no boolean/Boolean-argument constructor/factory method to deserialize from boolean value (false)
 at [Source: (BufferedInputStream); line: 32, column: 24] (through reference chain: io.fabric8.kubernetes.api.model.apiextensions.CustomResourceDefinition["spec"]->io.fabric8.kubernetes.api.model.apiextensions.CustomResourceDefinitionSpec["validation"]->io.fabric8.kubernetes.api.model.apiextensions.CustomResourceValidation["openAPIV3Schema"]->io.fabric8.kubernetes.api.model.apiextensions.JSONSchemaProps["properties"]->java.util.LinkedHashMap["spec"]->io.fabric8.kubernetes.api.model.apiextensions.JSONSchemaProps["properties"]->java.util.LinkedHashMap["hollow"]->io.fabric8.kubernetes.api.model.apiextensions.JSONSchemaProps["default"])
	at com.fasterxml.jackson.databind.exc.MismatchedInputException.from(MismatchedInputException.java:63)
	at com.fasterxml.jackson.databind.DeserializationContext.reportInputMismatch(DeserializationContext.java:1343)
	at com.fasterxml.jackson.databind.DeserializationContext.handleMissingInstantiator(DeserializationContext.java:1032)
	at com.fasterxml.jackson.databind.deser.ValueInstantiator.createFromBoolean(ValueInstantiator.java:280)
	at com.fasterxml.jackson.databind.deser.std.StdValueInstantiator.createFromBoolean(StdValueInstantiator.java:393)
	at com.fasterxml.jackson.databind.deser.BeanDeserializerBase.deserializeFromBoolean(BeanDeserializerBase.java:1426)
```

If you remove the `default: false` from `hollow` and `zoom`, it works. It seems like the `default: true` in `spiky` _does_ work. Note: the true/false values don't matter. It appears that the parsing works within nested arrays of objects, but not nested objects. Not sure why that is.

bbeaudreault:
Fixes #2144 

In #1600 @rohanKanojia added a hack to enum schema generation which specifically rewrote the spec for the enum to be an array of strings. This fixed the enum case, but I recently ran into a very similar issue when trying to use the `default` field. See #2144 for details.

Upon further investigation into the core kubernetes apiextension models, they utilize a `JSON` struct with a `raw` field for these properties. fabric8's generated schema creates a JSON.java file, with a `getRaw()` field. The difference here is that in kubernetes they have [custom Marshal and Unmarshal](https://github.com/nikhita/kubernetes/blob/6ba1523a8e36c3121a9e08bfac7b4f342bb8ccb3/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1/marshal.go#L121-L134) code for this JSON type. The marshal/unmarshal code ensures that the `raw` field gets handled correctly when serializing and deserializing the type. 

The files generated by the jsonschema2pojo schema do not seem to have the ability to inject a custom de/serialization like they do in the source go code. However, this generic JSON field sounds a lot like jackson's JsonNode to me -- a generic field that handles any type.

My change updates @rohanKanojia's hack to rewrite references of `io.fabric8.kubernetes.api.model.apiextensions.JSON` to instead use `com.fasterxml.jackson.databind.JsonNode`. This solves the enum problem just as well as his did, but also handles `default` (which does need to be typed) and `example`.

Note: I tested this by pulling the change into a downstream artifact i'm working on internally and it worked for my case. I added a test here as rohan did in #1600, but do not have the ability to run these tests. So will need to see what happens with the build.


bbeaudreault:
So unfortunately the version of kubernetes/openshift running on circle CI does not support `default`. I think I will need to remove my test, or not sure if there are other options here.

rohanKanojia:
ok to test

bbeaudreault:
Please advise about how to handle the failing tests (due to openshift compatibility). 

I'm happy to remove my test case or add in a different form. One thing I could try doing it deserialize into CustomResourceDefinition, then re-serializing into yaml and expect it to match (or something along those lines). That would at least prove that we aren't injecting the `Raw:` prefix and are serializing back in the correct format.

bbeaudreault:
I've decided to add a test to SerializationTest, replacing the test in RawCustomResourceIT. The new test at least ensures that we are appropriately deserializing and re-serializing, and spot checks some of the expected fields.

I'd be happy to add back the RawCustomResourceIT test if possible to get around the openshift version issue. I removed it in a single commit that I can just revert if need be.

Let me know what you think.

bbeaudreault:
I'm not sure why the OPENSHIFT_3.11.0 test failed, while 3.10.0 succeeded. Looks like it failed in a totally unrelated test (RoleBindingIT), but I can't re-run the suite. Please re-run or ignore.

rohanKanojia:
@bbeaudreault : Our tests are a bit flaky and fail every now and then. I've restarted them for now. Thanks a lot for your contribution :+1:

manusa:
[merge]

manusa:
GitHub was down this the last past hours, let's retry

manusa:
[merge]

bbeaudreault:
Ah I just noticed there are now merge conflicts. Will take a look

bbeaudreault:
Actually it looks like this PR was already merged by the automation. Not sure if I should just close this PR? Let me know.

PR #2152, opened 2020-04-20
handle possible /status subpath due to using status subresource on crd

bbeaudreault:
When a CRD has the following:

```
subresources:
  status: {}
```

It causes the exposure of a `/status` endpoint on the CRD API. For example: `/apis/hubspot.com/v1/namespaces/test-ns/test-crd/status`.

The KubernetesAttributesExtractor was not properly handling this, which was causing calls to update the status of crds in the test server to not find the right resource to update.


bbeaudreault:
I have https://github.com/fabric8io/kubernetes-client/pull/2151 open which already modifies the changelog. Once that's merged I will rebase this and update the changelog here as well.

bbeaudreault:
ok to test

rohanKanojia:
ok to test

bbeaudreault:
For reference, https://v1-15.docs.kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#status-subresource

kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/KubernetesAttributesExtractor.java
```diff
@@ -53,9 +53,10 @@
   private static final String KIND_GROUP = "/(?<kind>[^/?]+)";
   private static final String NAME_GROUP = "(/(?<name>[^/?]+))?";
   private static final String NAMESPACE_GROUP = "(/namespaces/(?<namespace>[^/]+))?";
+  private static final String STATUS_MAYBE_GROUP = "(/status)?";
```
manusa, line 56: Checking the provided link (https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#subresources)

I see there are 2 subresources `/status` and `/scale`.

Maybe we could change this to:
```
private static final String SUBRESOURCE_MAYBE_GROUP = "(/((status)|(scale)))?";
```
or even (although maybe this will probably break things):
```
private static final String SUBRESOURCE_MAYBE_GROUP = "(/(?<subresource>[^/]+))?";
```

WDYT??

bbeaudreault: Good call. I should have thought to include scale while I was at it. I think perhaps we should hardcode the 2 options. If they add more options in the future we could make it more flexible. But for now it doesn't seem worth the extra risk in terms of matching things we shouldn't.

bbeaudreault: Good call. I should have thought to include scale while I was at it. I think perhaps we should hardcode the 2 options. If they add more options in the future we could make it more flexible. But for now it doesn't seem worth the extra risk in terms of matching things we shouldn't.

I've updated appropriately.

bbeaudreault:
No problem. Would you like me to rebase this and add a changelog line? I'm not sure how strictly you include every PR there.

manusa:
> No problem. Would you like me to rebase this and add a changelog line? I'm not sure how strictly you include every PR there.

Yes, please do. We usually include a line for almost everything.

bbeaudreault:
I've updated to include a changelog entry

manusa:
[merge]

PR #2155, opened 2020-04-21
    Fix #2115: Add Support for keeping both v1alpha1 and v1beta1 apis in tekton extensions

### LINKED ISSUE #2115: [tekton] keep v1alpha1 api
Hi,

with the update 0.11.0, tekton added API version v1beta1. The old API version v1alpha1 is still supported.

The pull request https://github.com/fabric8io/kubernetes-client/pull/2020 removed the model definitions for v1alpha1 and added v1beta1. This forces projects using the client to migrate to the new API version (even if tekton still allows the creation of v1alpha1). Is it possible to keep both and e.g. mark v1alpha1 as deprecated? From the model point of view, this should be quite easy. I'm not sure about the TektonClient API though...

Best regards,
Fabian

rohanKanojia:
Fixes #2115
Added dsl support for `client.v1alpha1().*` and `client.v1beta1().*` for respective apigroups. Kudos to @Fabian-K  for initial work on this

### DIFF
--- a/extensions/tekton/model/src/main/java/io/fabric8/tekton/TektonResourceMappingProvider.java
+++ b/extensions/tekton/model/src/main/java/io/fabric8/tekton/TektonResourceMappingProvider.java
@@ -17,15 +17,6 @@ package io.fabric8.tekton;
 
 import io.fabric8.kubernetes.api.KubernetesResourceMappingProvider;
 import io.fabric8.kubernetes.api.model.KubernetesResource;
-import io.fabric8.tekton.pipeline.v1alpha1.Condition;
-import io.fabric8.tekton.pipeline.v1beta1.ClusterTask;
-import io.fabric8.tekton.pipeline.v1beta1.Pipeline;
-import io.fabric8.tekton.pipeline.v1beta1.PipelineRun;
-import io.fabric8.tekton.pipeline.v1beta1.SidecarState;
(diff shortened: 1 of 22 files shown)


extensions/tekton/model/src/main/java/io/fabric8/tekton/TektonResourceMappingProvider.java
```diff
@@ -38,15 +38,19 @@ public TektonResourceMappingProvider() {
         mappings.put("tekton.dev/v1alpha1#PipelineResource", PipelineResource.class);
         mappings.put("tekton.dev/v1alpha1#Condition", Condition.class);
         mappings.put("tekton.dev/v1beta1#Pipeline", Pipeline.class);
```
manusa, line 40: nit: In order to make it clearer, I thin it would be better to remove global imports and add the canonical path to all classes here:
```suggestion
        mappings.put("tekton.dev/v1beta1#Pipeline", io.fabric8.tekton.pipeline.v1beta1.Pipeline.class);
```

oscerd: +1

manusa:
#2115

manusa:
[merge]

PR #2157, opened 2020-04-22
Knative Generator Rewrite

Fabian-K:
[knative]
- update generator to use shared logic
- replace knative-annotator with shared model-annotator
- remove the generated *Schema class

[tekton]
- update to knative package change

[common]
- add editorconfig for *.go (although this is not applied to all go files yet)

As always, I´m looking forward to feedback!

Thanks,
Fabian


rohanKanojia:
ok to test

Fabian-K:
Hey, thanks for already looking into this! I´d suggest pausing this for now, I´d rebase it on top of https://github.com/fabric8io/kubernetes-client/pull/2160

extensions/knative/generator/cmd/generate/generate.go
```diff
@@ -16,169 +16,91 @@
...
-	}
 
-	typeMap := map[reflect.Type]reflect.Type{
-		reflect.TypeOf(time.Time{}): reflect.TypeOf(""),
-		reflect.TypeOf(struct{}{}):  reflect.TypeOf(""),
+	// the CRD List types for which the model should be generated
+	// no other types need to be defined as they are auto discovered
+	crdLists := []reflect.Type{
+
+		// serving v1
```
rohanKanojia, line 37: Umm, why have you removed all resources from `duck v1beta1`. Are they not part of Knative model?

Fabian-K: 2 things:
- all types that are only used internally are auto-discovered by the generator. Only CRDs need to be passed in as a starting point.
- AFAIK everything related to duck is not a CRD exposed by knative. Instead that is somehow used internally for duck typing. I might be wrong on this, do you have other information?

rohanKanojia: Umm, I don't have much idea about Knative. I think it's okay if duck classes were not exposed earlier.

rohanKanojia:
[merge]

PR #2177, opened 2020-04-29
KubernetesDeserialier lazy loads mappings via current thread (#2109)

### LINKED ISSUE #2109: JsonMappingException "No resource type found for: tekton.dev/v1alpha1#PipelineList" in intellij plugins
Hi guys,

When retrieving the list of pipelines I hit an exception when PipelineList is being deserialized in jackson:

```
client.pipelines().inNamespace(namespace).list().getItems()
```

results in: 
```
com.fasterxml.jackson.databind.JsonMappingException: No resource type found for:tekton.dev/v1alpha1#PipelineList"

io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:64) 
 io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:53)
 io.fabric8.kubernetes.client.utils.Serialization.unmarshal(Serialization.java:248)
 io.fabric8.kubernetes.client.utils.Serialization.unmarshal(Serialization.java:199)
 io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:474)
 io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:430)
 io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:412)
 io.fabric8.kubernetes.client.dsl.base.BaseOperation.listRequestHelper(BaseOperation.java:151)
 io.fabric8.kubernetes.client.dsl.base.BaseOperation.list(BaseOperation.java:621)
 io.fabric8.kubernetes.client.dsl.base.BaseOperation.list(BaseOperation.java:70)
```

For the background:
I'm writing an IntelliJ plugin to display, operate, etc. kubernetes resources in Intellij at https://github.com/redhat-developer/intellij-kubernetes

I'm trying to separate kubernetes- and tekton-pieces into 2 different plugins:
kubernetes base plugin: https://github.com/redhat-developer/intellij-kubernetes/pull/26
tekton extension plugin: https://github.com/redhat-developer/intellij-tekton/pull/36

The kubernetes (base) plugin holds all kubernetes & jackson jars:
* "io.fabric8:kubernetes-client:4.9.0",
* "io.fabric8:kubernetes-model:4.9.0",
* "io.fabric8:kubernetes-model-common:4.9.0",
* "io.fabric8:openshift-client:4.9.0"
* "com.fasterxml.jackson.core:jackson-annotations"
* "com.fasterxml.jackson.core:jackson-core"
* "com.fasterxml.jackson.corejackson-databind"
* "com.fasterxml.jackson.module:jackson-module-jaxb-annotations"

The tekton (extension) plugin holds the tekton jars:
* "io.fabric8:tekton-client:4.9.0"
* "io.fabric8:tekton-model:4.9.0"

IntelliJ plugins each have their own classloader. The kubernetes (base) plugin doesn't "see" the classes in the tekton (extension) plugin. The tekton plugin on the other hand depends on the kubernetes plugin and can thus access all it's classes

In the tekton (extension) plugin I manually register the custom type for the tekton pipeline since I found out that ServiceLoader wouldn't find the ```TektonResourceMappingProvider``` (which would do this automatically in a non-segregated classpath:
```
KubernetesDeserializer.registerCustomKind("tekton.dev/v1alpha1", "Pipeline", Pipeline.class)
```

I'm obviously missing bits for this to work in intellij plugins (which have different classloaders for each plugin). 

I'd highly appreciate some help here. Thanks in advance!

adietish:
fixes #2109

### DIFF
--- a/kubernetes-model/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/internal/KubernetesDeserializer.java
+++ b/kubernetes-model/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/internal/KubernetesDeserializer.java
@@ -21,16 +21,19 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.ServiceLoader;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 import com.fasterxml.jackson.core.JsonParser;
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.DeserializationContext;
 import com.fasterxml.jackson.databind.JsonDeserializer;
 import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.node.ArrayNode;
-import com.fasterxml.jackson.databind.node.ObjectNode;
 
 import io.fabric8.kubernetes.api.KubernetesResourceMappingProvider;
 import io.fabric8.kubernetes.api.model.HasMetadata;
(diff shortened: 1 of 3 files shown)


rohanKanojia:
ok to test

adietish:
@rohanKanojia I just added tests, sonar cloud still complains about 0% test coverage. Is test coverage reporting correct numbers?

rohanKanojia:
@adietish : Hi, @manusa might have some idea about it since he had set up sonar on kubernetes-client.

manusa:
I **think** your test suite is not running, you're using JUnit4 annotations

adietish:
@manusa you were right! I corrected the annotations and tests now seem to run. Nevertheless I still get 0% coverage in sonar. Any idea?

kubernetes-model/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/internal/KubernetesDeserializer.java
```diff
@@ -152,42 +111,133 @@ public static void registerCustomKind(String kind, Class<? extends KubernetesRes
...
+                    return null;
+                }
+                return (Class<? extends KubernetesResource>) clazz;
+            } catch (Throwable t) {
+                return null;
+            }
+        }
+
+        private Predicate<KubernetesResourceMappingProvider> distinctByClassName(
+                Function<KubernetesResourceMappingProvider, Class<? extends KubernetesResourceMappingProvider>> mapperProvider) {
```
dev-gaur, line 237: could you take care of the indentation here?

manusa:
> @manusa you were right! I corrected the annotations and tests now seem to run (had errors when switching to unit5 annotations, fixed them and now there are no errors any more). Nevertheless I still get 0% coverage in sonar. Any idea?

Currently we are not aggregating JaCoCo reports for tests in different modules (neither specifying additional sources for integration tests module), so coverage report is only includes those tests that exist in the same module.

That being said, for your current implementation it shouldn't matter, but I just noticed (sorry for not noticing before) you implemented the test in the `kubernetes-client` module but the tested class is in the `kubernetes-model-core. Is there any reason for this? Coverage for the tested class won't be reported unless the test suite is in the same module.

Thx for contributing. I'll try to do a proper review later, but after a quick read, something that strikes me is why we switched from a ConcurrentHashMap to a synchronized Map?

kubernetes-model/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/internal/KubernetesDeserializer.java
adietish: @dev-gaur this identation is on purpose, the whole line is too long to fit in 120 characters. Spliting parts isn't readble since it's a single parameter with lengthy generdics declaration. I therefore opted to keep the param all together but pushed it onto it's own line that that's indentated by 4. Makes sense?

adietish:
@manusa 
> That being said, for your current implementation it shouldn't matter, but I just noticed (sorry for not noticing before) you implemented the test in the `kubernetes-client` module but the tested class is in the `kubernetes-model-core. Is there any reason for this? Coverage for the tested class won't be reported unless the test suite is in the same module.

ok, I'll correct that, didn't notice.

> 
> something that strikes me is why we switched from a ConcurrentHashMap to a synchronized Map?

I think - and please correct me if wrong - that having atomic retrievals and insertions isn't good enough. Many logical operations are not just retrieving, they are retrieving, transforming and updating the map and shouldn't be interfered with while they're at work.

manusa:
> I think - and please correct me if wrong - that having atomic retrievals and insertions isn't good enough. Many logical operations are not just retrieving, they are retrieving, transforming and updating the map and shouldn't be interfered with while they're at work.

From a functional perspective, I think registration (write/put) will usually happen only at the beginning of the application execution. Retrieval (get) instead will happen through the complete life-cycle of the application and probably invoked simultaneously from multiple threads. Making the methods synchronized will drain the performance.

I'm quite unsure that there is a use case where a deserialization operation will occur before a "kind" registration. If such situation is possible, or we've switched to some sort of lazy registration, maybe we should reconsider IMHO. There are not so many classes to register (low-impact) but there are a lot of deserialization operations.

adietish:
> From a functional perspective, I think registration (write/put) will usually happen only at the beginning of the application execution. 

that's still true, the lazy loading is just postponing type registration to the first retrieval or registration of a type (where it was loaded in the static initializer before)

> Retrieval (get) instead will happen through the complete life-cycle of the application and probably invoked simultaneously from multiple threads. Making the methods synchronized will drain the performance.

I think that retrieval should not be affected performance wise. The only case where performance is affected is where get & put now are altogether synchronized where they were not in the old code. I think that those are justified functionally (we shouldn't allow retrieval while an update could happen). I could use a read-write lock though so that retrievals could still happen in parallel.
Makes sense?

manusa:
[Synchronized Methods](https://docs.oracle.com/javase/tutorial/essential/concurrency/syncmeth.html) _"...it is not possible for two invocations of synchronized methods on the same object to interleave. When one thread is executing a synchronized method for an object, all other threads that invoke synchronized methods for the same object block (suspend execution) until the first thread is done with the object."_

My interpretation of this has always been that no two threads can invoke the same `synchronized` method simultaneously.

Given the fact that now we have: `public synchronized Class<? extends KubernetesResource> getForKey(String key) {` and that this method gets called upon any `public KubernetesResource deserialize(JsonParser jp, DeserializationContext ctxt) throws IOException {` operation (potentially for every OkHttp request) why wouldn't performance be affected for synchronization in  a scenario with multiple simultaneous requests to the cluster?

(I'm probably missing something, I've only read quickly over the code and this is what caught my attention)

adietish:
> [Synchronized Methods](https://docs.oracle.com/javase/tutorial/essential/concurrency/syncmeth.html) _"...it is not possible for two invocations of synchronized methods on the same object to interleave. When 
> Given the fact that now we have: `public synchronized Class<? extends KubernetesResource> getForKey(String key) {` and that this method gets called upon any `public KubernetesResource deserialize(JsonParser jp, DeserializationContext ctxt) throws IOException {` operation (potentially for every OkHttp request) why wouldn't performance be affected for synchronization in a scenario with multiple simultaneous requests to the cluster?

Grocked your concern, this might be limiting in terms of perf I ack. The bottleneck might occur bcs of the synchronization I introduced that protects the mapping from being read & initialized simultaneously. Let me think this through and come up with a suggestion.

manusa:
> Makes sense?

It does. I think that bringing back the concurrent Map, removing synchronized keyword and adding some warning in JavaDoc for registration methods could do the trick without changing much your current implementation.

adietish:
@manusa works for you in the current impl?

adietish:
@manusa thx for allowing me to contribute 👍

rohanKanojia:
[merge]

Issue #2147, opened 2020-04-20
PropagationPolicy is ignored


Fabian-K:
Hi,

when setting a propagation policy for the delete operation this is ignored

```
client.resource(item).withPropagationPolicy("Foreground").delete()
```

Generated request:
```
--> DELETE https://127.0.0.1:32774/apis/sources.knative.dev/v1alpha1/namespaces/site/githubsources/github-source
Content-Type: application/json; charset=utf-8
Content-Length: 67

{"apiVersion":"v1","kind":"DeleteOptions","orphanDependents":false}
--> END DELETE (67-byte body)
```

PS: propagation policy not being propagated. Nice 🤪!

rohanKanojia:
Umm, Need to check. Does it work with you use item with DSL? For example, `client.pods().inNamespace("foo").withName("some").withPropagationPolicy("Foreground").delete()`

I can see propagation policy being appended in code, but need to check whether control comes here when using generic resource API:
https://github.com/fabric8io/kubernetes-client/blob/16bd97fb010fad230ab7851d8e2b9286a739f38a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationSupport.java#L220-L227

Fabian-K:
Currently tracing this... so far I found this where it is lost

https://github.com/fabric8io/kubernetes-client/blob/f70bb953d4b257a774b08a79ca6a3b94e206e9cf/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableImpl.java#L162

rohanKanojia:
Yeah, it then calls resource-specific handler which is a generated class

rohanKanojia:
I think you would need to modify this

https://github.com/fabric8io/kubernetes-client/blob/f70bb953d4b257a774b08a79ca6a3b94e206e9cf/kubernetes-client/src/main/resources/resource-handler.vm#L100-L102

Fabian-K:
Have it working locally. I´ll open a PR

Fabian-K:
😏 applying this to the entire project is no fun, especially knative. IMO it makes more sense to first apply  https://github.com/fabric8io/kubernetes-client/pull/2134 to knative as well and then continue with this

rohanKanojia:
#2178  should be able to fix this issue

rohanKanojia:
#2178 is merged. I think we can close this

PR #2178, opened 2020-04-29
Fix #2107: Set PropagationPolicy to Background by default

### LINKED ISSUE #2107: Confusing message from kubernetes-client: object is being deleted: secrets "XXXX" already exists
During studying this https://github.com/eclipse/che/issues/16512 that might be related with upgrade of  kubernetes-client  from 4.1.0 to 4.9.0  https://github.com/eclipse/che-parent/pull/159 https://github.com/eclipse/che/pull/16345 we noticed very strange error.
```
aused by: io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: POST at: https://172.30.0.1/api/v1/namespaces/skabashn/secrets. Message: object is being deleted: secrets "workspaceyy5nesnxw954tsbz-sshprivatekeys" already exists. Received status: Status(apiVersion=v1, code=409, details=StatusDetails(causes=[], group=null, kind=secrets, name=workspaceyy5nesnxw954tsbz-sshprivatekeys, retryAfterSeconds=null, uid=null, additionalProperties={}), kind=Status, message=object is being deleted: secrets "workspaceyy5nesnxw954tsbz-sshprivatekeys" already exists, metadata=ListMeta(_continue=null, remainingItemCount=null, resourceVersion=null, selfLink=null, additionalProperties={}), reason=AlreadyExists, status=Failure, additionalProperties={}).
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.requestFailure(OperationSupport.java:568)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.assertResponseCode(OperationSupport.java:507)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:471)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:430)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleCreate(OperationSupport.java:251)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.handleCreate(BaseOperation.java:815)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.create(BaseOperation.java:333)
	at org.eclipse.che.workspace.infrastructure.kubernetes.namespace.KubernetesSecrets.create(KubernetesSecrets.java:49)
	... 10 common frames omitted
```
Any ideas?

rohanKanojia:
Fix #2107

This should also take care of #2147

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/PropagationPolicyConfigurable.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/PropagationPolicyConfigurable.java
@@ -15,7 +15,23 @@
  */
 package io.fabric8.kubernetes.client;
 
+import io.fabric8.kubernetes.api.model.DeletionPropagation;
+
 public interface PropagationPolicyConfigurable<T>
 {
-    T withPropagationPolicy(String propagationPolicy);
+  /**
+   * Whether and how garbage collection will be performed.
+   * Either this field or OrphanDependents may be set, but not both.
+   *
+   * <p>The default policy is decided by the existing finalizer set in the metadata.finalizers and the resource-specific default policy.
+   *
+   * <p>Acceptable values are:
+   * <br>'Orphan' - orphan the dependents;</li>
+   * <br>'Background' - allow the garbage collector to delete the dependents in the background;</li>
+   * <br>'Foreground' - a cascading policy that deletes all dependents in the foreground.</li>
+   *
+   * @param propagationPolicy propagation policy in form of string {@link DeletionPropagation}
+   * @return resource
+   */
+    T withPropagationPolicy(DeletionPropagation propagationPolicy);
 }
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationSupport.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationSupport.java
@@ -21,11 +21,11 @@ import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.fabric8.kubernetes.api.model.DeleteOptions;
+import io.fabric8.kubernetes.api.model.DeletionPropagation;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.Status;
 import io.fabric8.kubernetes.api.model.StatusBuilder;
 import io.fabric8.kubernetes.api.model.autoscaling.v1.Scale;
-import io.fabric8.kubernetes.api.model.metrics.v1beta1.PodMetrics;
 import io.fabric8.kubernetes.api.model.extensions.DeploymentRollback;
 import io.fabric8.kubernetes.client.Config;
 import io.fabric8.kubernetes.client.KubernetesClientException;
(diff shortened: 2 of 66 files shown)


rohanKanojia:
retest this please

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/PropagationPolicyConfigurable.java
```diff
@@ -17,5 +17,17 @@
...
+   * Either this field or OrphanDependents may be set, but not both.
+   * The default policy is decided by the existing finalizer set in the metadata.finalizers and the resource-specific default policy.
+   * Acceptable values are:
+   *    'Orphan' - orphan the dependents;
+   *    'Background' - allow the garbage collector to delete the dependents in the background;
+   *    'Foreground' - a cascading policy that deletes all dependents in the foreground.
+   *
+   * @param propagationPolicy propagation policy in form of string
+   * @return resource
+   */
```
manusa, line 31: This JavaDoc will be scrambled when processed (no line breaks), try any of the proposed alternatives.
```suggestion
  /**
   * Whether and how garbage collection will be performed.
   * Either this field or OrphanDependents may be set, but not both.
   *
   * <p>The default policy is decided by the existing finalizer set in the metadata.finalizers and the resource-specific default policy.
   * 
   * <p>Acceptable values are:
   * <ul>
   * <li>'Orphan' - orphan the dependents;</li>
   * <li>'Background' - allow the garbage collector to delete the dependents in the background;</li>
   * <li>'Foreground' - a cascading policy that deletes all dependents in the foreground.</li>
   * </ul>
   *
   * @param propagationPolicy propagation policy in form of string
   * @return resource
   */
```
Or:
```suggestion
  /**
   * Whether and how garbage collection will be performed.
   * Either this field or OrphanDependents may be set, but not both.
   *
   * <p>The default policy is decided by the existing finalizer set in the metadata.finalizers and the resource-specific default policy.
   *
   * <p>Acceptable values are:
   * <br>Orphan' - orphan the dependents;</li>
   * <br>Background' - allow the garbage collector to delete the dependents in the background;</li>
   * <br>Foreground' - a cascading policy that deletes all dependents in the foreground.</li>
   *
   * @param propagationPolicy propagation policy in form of string
   * @return resource
   */
```

manusa:
If propagation policy is a known set of values, why not use an enum instead?

Only advantage of current implementation (free String) is that in the future k8s adds other valid values. But in that case, maybe some warning should be issued if the propagation policy is not within the accepted set, or at least create some publicly accessible constants. 

We want to prevent users who type `withPropagationPolicy("Bakcground")` becoming crazy

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationSupport.java
```diff
@@ -221,7 +224,10 @@ protected void handleDelete(URL requestUrl, long gracePeriodSeconds, String prop
      * Either the propagation policy or the orphan dependent (deprecated) property must be set, but not both.
      */
     if (propagationPolicy != null) {
-      deleteOptions.setPropagationPolicy(propagationPolicy);
+      if (Arrays.stream(DeletionPropagation.values()).noneMatch(deletionPropagation -> deletionPropagation.name().equals(propagationPolicy))) {
+        logger.warn("Unknown DeletionPropagation policy used for deletion: {}", propagationPolicy);
+      }
```
manusa, line 229: This check shouldn't be necessary any more as we are using an enum

manusa:
[merge]

PR #2186, opened 2020-05-04
Fix #2165: list() should also be able to handle ListOptions

### LINKED ISSUE #2165: Dose any method to set resourceVersion in list options?
Dose any method to set resourceVersion in list options that we can query result from apiserver just from apiserver cache?

In our case, we dost not use the k8s recommend style(list + watch cache the resource in client informer cache), we use list resource from apiserver period, that not with resourceVersion "0", so that apiserver will send request to etcd, that way give big pressure to apiserver, the apiserver's memory increase very fast. So a method to set resourceVersion in list options is very important.

rohanKanojia:
Fix #2165

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Listable.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Listable.java
@@ -15,7 +15,28 @@
  */
 package io.fabric8.kubernetes.client.dsl;
 
+import io.fabric8.kubernetes.api.model.ListOptions;
+
 public interface Listable<T> {
   T list();
+
+  /**
+   * List resources from APIServer.
+   * @deprecated : Please use {@link #list(ListOptions)} instead
+   *
+   *
+   * @param limitVal number of resources to list
+   * @param continueVal an offset to pick listing from
+   * @return resource list
+   */
+  @Deprecated
   T list(Integer limitVal, String continueVal);
+
+  /**
+   * List resource from Kubernetes API server.
+   *
+   * @param listOptions ListOptions is the query options to a standard REST list call.
+   * @return list of resource type
+   */
+  T list(ListOptions listOptions);
 }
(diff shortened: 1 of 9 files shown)


rohanKanojia:
ok to test

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Listable.java
```diff
@@ -15,7 +15,10 @@
  */
 package io.fabric8.kubernetes.client.dsl;
 
+import io.fabric8.kubernetes.api.model.ListOptions;
+
 public interface Listable<T> {
   T list();
   T list(Integer limitVal, String continueVal);
```
manusa, line 22: Since the new method already covers the functionality of this one, it maybe good to add a `@Deprecated` annotation (plus some JavaDoc indicating to use the new method) so we can remove it in the next major release.

openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/ProjectRequestsOperationImpl.java
```diff
@@ -119,19 +121,23 @@ public Status list() {
...
+  public Status list(ListOptions listOptions) {
     try {
       URL requestUrl = getNamespacedUrl();
-      if(limitVal != null) {
-        requestUrl = new URL(URLUtils.join(requestUrl.toString(), "?limit=" + limitVal.toString()));
-      } else if(continueVal != null) {
-        requestUrl = new URL(URLUtils.join(requestUrl.toString(), "?limit=" + limitVal.toString() + "&continue=" + continueVal));
+      if(listOptions.getLimit() != null) {
+        requestUrl = new URL(URLUtils.join(requestUrl.toString(), "?limit=" + listOptions.getLimit().toString()));
+      } else if(listOptions.getContinue() != null) {
```
manusa, line 133: This doesn't look good as getLimit will return null.
Why not use the same Url builder you used in BaseOperation

rohanKanojia:
[merge]

Issue #2199, opened 2020-05-06
KubernetesClient#customResources takes CustomResourceDefinition, when CustomResourceDefinitionContext would suffice


bbeaudreault:
As you can see in [KubernetesClient.java#L112](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/KubernetesClient.java#L112), in order to query for custom resources you need to provide a full `CustomResourceDefinition` object. However, in [CustomResourceOperationsImpl.java#L54-L71](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/CustomResourceOperationsImpl.java#L54-L71), you really only use the `crd.spec.group`, `crd.spec.version`, `crd.spec.names.kind`, `crd.spec.scope`, `crd.spec.names.plural`. These are all available in [CustomResourceDefinitionContext.java](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/CustomResourceDefinitionContext.java).

`CustomResourceDefinition`s may have started out simple, but have grown complex with openapi spec validations, defaulting, etc. From an API perspective, it's confusing to expect a full CRD to be passed in here. If one doesn't have access to the full spec, one might feel obligated to fetch it from kubernetes before calling the functions. Slimming the requirements down to just CustomResourceDefinitionContext would reduce the ambiguity in the interface and be simpler to build anyway.

Thoughts? Some other functions in KubernetesClient already take a CustomResourceDefinitionContext, for example the raw custom resource operations function.

rohanKanojia:
@bbeaudreault : Nice catch, I think it's a good idea to replace full `CustomResourceDefinition` with CRDContext :+1:

PR #2208, opened 2020-05-08
Enable querying for customResources using only CustomResourceDefinitionContext instead of full CustomResourceDefinition

bbeaudreault:
Add a customResources method which takes a CustomResourceDefinitionContext instead of CustomResourceDefinition. Use that in CustomResourceOperationsImpl. Add a 'kind' field to CustomResourceDefinitionContext, defaulting to getKind if not specified

Fixes https://github.com/fabric8io/kubernetes-client/issues/2199

~This is a WIP (work in progress) because I am unable to run the tests in my environment. I haven't added any new tests yet, either. I'll take a look at that, though I think this should largely be covered by existing.~ I've taken a look at tests and updated existing tests.

This approach breaks the KubernetesClient interface due to the exposure of a new method for implementers to implement. I imagine most users are simply using one of the existing implementations, so I've left the old method around for backwards compatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 #### Bugs
 
 #### Improvements
+* Fix #2199: KubernetesClient#customResources now accepts CustomResourceDefinitionContext
 
 #### Dependency Upgrade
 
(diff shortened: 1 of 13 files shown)


rohanKanojia:
ok to test

bbeaudreault:
This is ready for review/merge. I've updated the CHEATSHEET and CHANGELOG. SonarCloud is failing on code coverage and duplication, but I believe it's hitting me largely on code that already existed.

rohanKanojia:
[merge]

Issue #1075, opened 2018-05-14
watch should handle etcd old version exception


chenchun:
I am running spark on kubernetes. This is the full issue description https://issues.apache.org/jira/browse/SPARK-24266

I think the exception ```too old resource version: 21648111 (21653211)``` should be better handled in kubernetes-client instead of simply throw it to the caller because resource version is cached by kubernetes-client, not by the caller. https://github.com/fabric8io/kubernetes-client/blob/5b1a57b64c7dcc7ebeba3a7024e8615c91afaedb/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java#L259-L266

chenchun:
cc @foxish

yujiantao:
We run to the same issue, any progress on this?

chenchun:
@yujiantao For a simple fix, you can try comment out these lines https://github.com/fabric8io/kubernetes-client/blob/v4.0.5/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java#L141-L143
We've using it for a long time, everything is fine.

SergSlipushenko:
We also running in the same issue with Spark, would be great to see the fix eventually

SergSlipushenko:
It is still relevant for people using k8s-as-a-service on Azure. We applied the workaround mentioned by @chenchun and it works fine so far...

stijndehaes:
> @yujiantao For a simple fix, you can try comment out these lines https://github.com/fabric8io/kubernetes-client/blob/v4.0.5/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java#L141-L143
> We've using it for a long time, everything is fine.

@chenchun Is this something we maybe could put into the client? That for some watches you don't care about version problems.

SergSlipushenko:
It is still relevant and requires a rebuild of Spark from the sources just for using hacked kubernetes-client :disappointed:

rohanKanojia:
@SergSlipushenko : Actually we added it in past but we had to remove it, see #1800

chenchun:
@stijndehaes I took a look at #1800, is it better to add a bool flag of whether or not do re-watching automatically when receive a version change? So that we won't break the contract of sending HTTP_GONE if resource version is old and also makes people easier when they don't care about the problem.

rohanKanojia:
[...]

PR #2213, opened 2020-05-11
Add support for Namespaced SharedInformers

rohanKanojia:
Seems like passing namespace with `new OperationContext().withNamespace("ns1")` was broken. We were losing all the necessary fields populated by default in `OperationContext` due to a new instance being passed by the user. This
PR fixes it by merging passed `operationContext` with client's existing `operationContext`

Related to #1075


rohanKanojia:
[merge]

PR #2223, opened 2020-05-13
Fix #2195: Annotation processors and build time dependencies transitive

### LINKED ISSUE #2195: Annotation processors and build time dependencies transitive
## Description
After model refactor, some of the project dependencies used at build time (annotation processors, etc.) are being transitively propagated to projects depending on the client. Need to fix pom to prevent this.

io.sundr:builder-annotations
org.jsonschema2pojo:jsonschema2pojo-core

rohanKanojia:
Fix #2195


rohanKanojia:
[merge]

Issue #2212, opened 2020-05-09
kubernetes-client could not work with java 8u252


wangyang0918:
When upgrading the jdk from 8u242 to 8u252, we find that the kubernetes-client could not work. It always throws the following exception.

Even when i bump the kubernetes-client version to 4.9.1, it still could not solve the problem. I know it might be a issue of okhttp or other dependencies of kubernetes-client. However, i could not find a clear clue for that.

```
io.fabric8.kubernetes.client.KubernetesClientException: Operation: [get]  for kind: [Service]  with name: [flink-native-k8s-session-1-rest]  in namespace: [default]  failed.
	at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:64)
	at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:72)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.getMandatory(BaseOperation.java:225)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.get(BaseOperation.java:168)
	at org.apache.flink.kubernetes.kubeclient.Fabric8FlinkKubeClient.getRestService(Fabric8FlinkKubeClient.java:201)
	at org.apache.flink.kubernetes.cli.KubernetesSessionCli.run(KubernetesSessionCli.java:104)
	at org.apache.flink.kubernetes.cli.KubernetesSessionCli.lambda$main$0(KubernetesSessionCli.java:185)
	at org.apache.flink.runtime.security.contexts.NoOpSecurityContext.runSecured(NoOpSecurityContext.java:30)
	at org.apache.flink.kubernetes.cli.KubernetesSessionCli.main(KubernetesSessionCli.java:185)
Caused by: java.net.SocketException: Broken pipe (Write failed)
	at java.net.SocketOutputStream.socketWrite0(Native Method)
	at java.net.SocketOutputStream.socketWrite(SocketOutputStream.java:111)
	at java.net.SocketOutputStream.write(SocketOutputStream.java:155)
	at sun.security.ssl.OutputRecord.writeBuffer(OutputRecord.java:431)
	at sun.security.ssl.OutputRecord.write(OutputRecord.java:417)
	at sun.security.ssl.SSLSocketImpl.writeRecordInternal(SSLSocketImpl.java:894)
	at sun.security.ssl.SSLSocketImpl.writeRecord(SSLSocketImpl.java:865)
[...]

PR #2227, opened 2020-05-14
fix: JDK8 always uses http/1.1 protocol (Prevent OkHttp from wrongly enabling http/2)

manusa:
Fix: #2212


rohanKanojia:
[merge]

PR #2249, opened 2020-05-26
Adds basic support for Kubernetes Pod Eviction API

bbeaudreault:
Evictions allow for safer deletions, respecting PodDisruptionBudget. This PR adds support to the dsl, and adds Eviction to the generated schema. `Evictions` take a `DeleteOptions`, but don't support any of the fields yet except `dryRun` in 1.18. I decided not to add client support for `dryRun` in this implementation because it would involve a bunch more changes to the various generics and interfaces. DryRun support should be a separate contribution if desired, since it applies to more than just Eviction.

For reference, see:
- [Eviction API Docs](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/#the-eviction-api)
- [Eviction type](https://github.com/kubernetes/kubernetes/blob/abe6321296123aaba8e83978f7d17951ab1b64fd/pkg/apis/policy/types.go#L113-L123)
- [Eviction API handling](https://github.com/kubernetes/kubernetes/blob/abe6321296123aaba8e83978f7d17951ab1b64fd/pkg/registry/core/pod/storage/eviction.go#L104)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 #### Improvements
 * Fix #2199: KubernetesClient#customResources now accepts CustomResourceDefinitionContext
+* Adds basic support for Pod Eviction API
 
 #### Dependency Upgrade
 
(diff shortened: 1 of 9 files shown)


rohanKanojia:
ok to test

bbeaudreault:
@rohanKanojia this is all set to review, all tests are passing. As usual, SonarCloud code coverage failed, but there are tests in both PodTest and PodIT.

rohanKanojia:
[merge]

bbeaudreault:
thanks!

PR #2253, opened 2020-05-28
fix: remove dependency jsonschema2pojo-core in kubernetes-model

jorsol:
The dependency jsonschema2pojo-core is not needed at runtime, and it's only used at build time to generate source code by the maven plugin.

This fix the transitive dependency that should not be needed, this transitive dependency pulls others artifacts that should not be used like the Rule failing in the build of Quarkus:
```
[WARNING] Rule 2: org.apache.maven.plugins.enforcer.BannedDependencies failed with message:
Found Banned Dependency: commons-logging:commons-logging:jar:1.1.1
Found Banned Dependency: javax.validation:validation-api:jar:1.0.0.GA
```

A release shortly after this fix is merged should allow a more clean update in Quarkus of the latest version of kubernetes-client.

### DIFF
--- a/kubernetes-model/pom.xml
+++ b/kubernetes-model/pom.xml
@@ -71,12 +71,14 @@
       <artifactId>transform-annotations</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.jsonschema2pojo</groupId>
-      <artifactId>jsonschema2pojo-core</artifactId>
-    </dependency>
-    <dependency>
       <groupId>com.fasterxml.jackson.module</groupId>
       <artifactId>jackson-module-jaxb-annotations</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+      <version>3.10</version>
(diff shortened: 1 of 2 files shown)


rohanKanojia:
ok to test

manusa:
Relates to: https://github.com/fabric8io/kubernetes-client/pull/2223

Sundrio dependencies must be removed too.

rohanKanojia:
retest this please

jorsol:
> Sundrio dependencies must be removed too.

Yes, it looks like sundrio is used only at compile time, but it can't be removed completly since its used on source code, so the option is to mark it as `provided` scope, WDYT?

jorsol:
@manusa It looks like Sundrio is already with the `provided` scope in kuberntes-model, so it won't show up as transitive dependency.

But still I have worked on a refactor and cleaned some declarations and double check that dependencies that should not be transitive are marked as provided, the only issue is that the refactor end up being a bit large (mostly for the renaming), so feel free to use this PR, or close it and use the PR #2254

manusa:
Closed in favor of #2254

PR #2270, opened 2020-06-08
Fixes for Raw Custom Resource API testing

bbeaudreault:
I've been using the raw API recently and noticed two bugs:

1. There was a trailing slash added to the end of POST requests. This was causing the KubernetesAttributeExtractor to fail to extract metadata from the request path. The simple fix here would be to allow a trailing slash in KubernetesAttributeExtractor path regex, but I believe the correct fix is to not add the trailing slash at all. No paths in the kube api docs have trailing slashes. I refactored RawCustomResourceOperationsImpl a bit to fix this issue, and added some tests.
2. Users of the RawCustomResourceOperationsImpl do not need to register a custom kind with the KubernetesSerializer. As such, the KubernetesAttributeExtractor fails to extract metadata from these resources. The attributes in the CrudDispatcher map then do not match across requests. I've added a RawHasMetadata resource to kubernetes-server-mock and use that as a fallback if the toKubernetesResource call fails.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### 4.10-SNAPSHOT
 #### Bugs
+* Fix Raw CustomResource API path generation to not having trailing slash
+* Fix KubernetesAttributesExctractor to extract metadata from unregistered custom resources, such when using Raw CustomResource API 
 
 #### Improvements
 * Fix #2233: client.service().getUrl(..) should be able to fetch URL for ClusterIP based services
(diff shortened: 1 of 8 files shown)


rohanKanojia:
ok to test

bbeaudreault:
`Sonar Scanner` and `test` are both failing on different tests, and both tests succeed locally. Can you kick off a retest?

manusa:
I find everything OK with the PR.

The only thing that disturbs me a little is the name for the [RawHasMetadata](https://github.com/fabric8io/kubernetes-client/pull/2270/files#diff-24d228ae80ae76c5130e7ac7f38680ed) class, as it implies that instances of this type will be returned when interacting with the RawCustomResource API while this is only partially true. This is the type that will be returned as a fallback when the deserializer can't find a more suitable type and the content to deserialize matches the HasMetadata structure which can happen in other cases too.

rohanKanojia:
retest this please

rohanKanojia:
[merge]

PR #2276, opened 2020-06-09
E2E Github Action workflow for minikube

rohanKanojia:
Move Integration tests from CircleCI to Github Actions.

We have been running integration tests on Github Actions in Eclipse JKube and they seem very stable when compared to CircleCi. Let's move them here to see if it's actually CI which is the reason for flaky tests or whether it's actually our test code.


.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,102 @@
...
+            'google-chrome.*'                       \
+            'heroku.*'                              \
+            'libmagic.*'                            \
+            'libmono.*'                             \
+            'mono-.*'                               \
+            'mysql.*'                               \
+            'php.*'                                 \
+            'ruby.*'
+          sudo apt-get autoremove
+          df -h
```
manusa, line 85: Hopefully this part won't be needed. 
This is only needed in JKube because we run parallel tests with different base images and OpenShift won't start Pods if the 80% free space threshold is surpassed.
This step adds an overhead of ~4 minutes in average to the workflow. If we get Disk pressure in this project too, we should try to use lighter images to perform our tests or other approaches (this should be our last resort).

kubernetes-itests/src/test/java/io/fabric8/kubernetes/PodIT.java
```diff
@@ -84,10 +82,11 @@
 
   private String currentNamespace;
 
-  private static final Logger logger = LoggerFactory.getLogger(PodIT.class);
+  private static Logger logger = Logger.getLogger(PodIT.class.getName());
```
manusa, line 85:
```suggestion
  private static final Logger logger = Logger.getLogger(PodIT.class.getName());
```

manusa: If you changed this because of the Sonar warning, it was complaining about the order of the keywords.

kubernetes-itests/src/test/java/io/fabric8/kubernetes/CustomResourceDefinitionIT.java
```diff
@@ -25,15 +25,18 @@
...
 import org.junit.runner.RunWith;
+import org.junit.runners.MethodSorters;
 
 import java.util.Collections;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
 @RunWith(ArquillianConditionalRunner.class)
 @RequiresKubernetes
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
```
manusa, line 39: Is this still needed?

rohanKanojia: ah, no. I forgot to remove it.

manusa:
Once merged we need to remove required steps from settings:
image

rohanKanojia:
[merge]

rohanKanojia:
[merge]

PR #2278, opened 2020-06-10
fix: Code clean up + testing

manusa:
- Added type parameters for KubernetesList in KubernetesClient
- Test verifying waitUntilCondition **always** retrieves resource from server (since 4.10.2)


manusa:
[merge]

PR #2281, opened 2020-06-11
Fix #2131: Kubernetes Client fails to deserialize CustomResourceDefinition with OpenAPIV3Schema using JSONSchemaPropsOr* fields

rohanKanojia:
Fix #2131 

Add our own JSONSchemaPropsOrBool, JSONSchemaPropsOrArray, JSONSchemaPropsOrStringArray
to provide customized serialization/deserialization

Issue #2131, opened 2020-04-14
client.customResourceDefinitions().list(); returns error


gaurav-dhoot:
Hi,

I tried to retrieve the Custom Resource Definitions that I have in my minishift. I have gone through your CRDExample and have followed the steps. However, I end up getting an error. 

This is the code that I used:
```
		Config clientConfig = new ConfigBuilder().withUsername("admin").withPassword("ad").withMasterUrl("https://192.168.99.108:8443/").build();
		OpenShiftClient client=new DefaultOpenShiftClient(clientConfig);
		CustomResourceDefinition crd=new CustomResourceDefinition();
		client.customResourceDefinitions().list();
```

This is the error that I get:
```
Exception in thread "main" io.fabric8.kubernetes.client.KubernetesClientException: An error has occurred.
	at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:64)
	at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:53)
	at io.fabric8.kubernetes.client.utils.Serialization.unmarshal(Serialization.java:248)
	at io.fabric8.kubernetes.client.utils.Serialization.unmarshal(Serialization.java:199)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:474)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:430)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:412)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.listRequestHelper(BaseOperation.java:151)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.list(BaseOperation.java:621)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.list(BaseOperation.java:70)
	at restAPI.CRs.Topic.test.main(test.java:37)
Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot construct instance of `io.fabric8.kubernetes.api.model.apiextensions.JSONSchemaPropsOrBool` (although at least one Creator exists): no boolean/Boolean-argument constructor/factory method to deserialize from boolean value (true)
[...]

PR #2288, opened 2020-06-15
Adds configurable serializers and deserializers to json schema codegen

bbeaudreault:
So, I hit a similar issue to https://github.com/fabric8io/kubernetes-client/issues/2131 over the weekend, but unfortunately I did not see https://github.com/fabric8io/kubernetes-client/pull/2281 when I wrote this implementation. I know you already merged 2281, but I wanted to offer up my implementation as an alternative.

Rohan's approach is totally reasonable, and I considered taking it. The reason I didn't is I felt we could get a cleaner implementation by plugging into jsonschema2pojo. We have had to maintain a "hack" (see removed hack in diff) of rewriting the kube-schema to replace JSON with JsonNode. This new approach solves that hack more cleanly, as well as issue 2131.

The benefit of relying on codegen rather than replacing the generated code with manual code  is we will see build failures if anything changes in the kubernetes implementation that is not compatible with the serializers/deserializers.

Anyway, I'm happy to throw this code away if you prefer, but I figured I'd submit it since I wrote it around the same time Rohan wrote his.

cc @rohanKanojia


bbeaudreault:
Unfortunately the `JSON` handling needs to use `writeRawValue`, which is not compatible with the YAMLGenerator. I can't think of a way around that so I guess I'll punt on this for now. I do think we could similarly remove the hack I mentioned by doing what Rohan did, but for JSON as well. We'd need to modify the class to keep track of a JsonNode, which isn't pretty. But probably better than a regex to update the schema.

manusa:
I like the generator approach because this is really a problem that should be tackled at generation time (Same for classes such as [IntOrString](https://javadoc.io/doc/io.fabric8/kubernetes-model-core/latest/io/fabric8/kubernetes/api/model/IntOrString.html), etc.). IMHO The main problem with apiextensions is that these types (`JSON*Or*`) are somewhat bizarre and don't even work in some K8s version (at least I couldn't).

Regarding `JSON`, type is already being mapped to `JsonNode` in the [generator](https://github.com/fabric8io/kubernetes-client/blob/2d1a9d47d42e6aa37ba3cb0e70ede9429c93e392/kubernetes-model-generator/kubernetes-model-apiextensions/cmd/generate/generate.go#L115). Generator should really have some Map/function (if it doesn't already) to globally take care of these mappings (I agree with the strings replace being a bad idea), before the schema is even generated.

rohanKanojia:
I think adding a hashmap to take care of these mappings should not be that hard. let me try it out.

bbeaudreault:
Ok I'll reopen this PR and let rohan handle JSON in his PR. I'll just handle the 3 `JSON*Or*` classes I've tackled for now, but we can do a followup PR to handle things like IntOrString. Does that sound ok?

bbeaudreault:
@rohanKanojia if you merge yours, since it's already approved, i'll rebase mine on top and remove the JSON handling

rohanKanojia:
retest this please

bbeaudreault:
This PR is ready -- the only remaining failure is the usual flaky test DefaultSharedIndexInformerTest

rohanKanojia:
retest this please

rohanKanojia:
ah, I would fix this flaky test. This started failing randomly when I removed Thread.sleep in tests. Need to revisit this and fix :-(

rohanKanojia:
Could you please squash your commits into one?

bbeaudreault:
@rohanKanojia all set

rohanKanojia:
add to whitelist

rohanKanojia:
I would appreciate if you could address Sonar code smells, most of them can be easily addressed.

bbeaudreault:
@rohanKanojia this one is also all set. The test failure seems like a flaky test, since 3.11 succeeds. Can you re-run and merge if looks good?

bbeaudreault:
@rohanKanojia any chance we can get this merged?

rohanKanojia:
[merge]

Issue #2285, opened 2020-06-13
Raw Custom Resources API createOrReplace does not handle create failures gracefully


bbeaudreault:
I'm hoping to get to this this week but wanted to make an issue to keep track of:

the RawCustomResourceOperationsImpl#createOrReplace does a create, catches the exception, and then does a replace. The idea here I'm guessing is that create would fail if the resource already exists, but create can fail for other reasons as well. Any exception that's not related to the resource already existing should be propagated

PR #2289, opened 2020-06-15
Update createOrReplace to work more like BaseOperation

bbeaudreault:
This fixes https://github.com/fabric8io/kubernetes-client/issues/2285

Rather than try to actually handle exception propagation for create-then-replace, I decided to mirror the BaseOperation implementation which does a get then branch accordingly. I added a couple tests.

### DIFF
--- a/kubernetes-client/src/test/java/io/fabric8/kubernetes/client/dsl/internal/RawCustomResourceOperationsImplTest.java
+++ b/kubernetes-client/src/test/java/io/fabric8/kubernetes/client/dsl/internal/RawCustomResourceOperationsImplTest.java
@@ -15,37 +15,41 @@
  */
 package io.fabric8.kubernetes.client.dsl.internal;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+import static org.mockito.Mockito.*;
+
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mockito;
+
 import io.fabric8.kubernetes.api.model.ListOptionsBuilder;
 import io.fabric8.kubernetes.client.Config;
 import io.fabric8.kubernetes.client.ConfigBuilder;
+import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.dsl.base.CustomResourceDefinitionContext;
 import io.fabric8.kubernetes.client.utils.Utils;
 import okhttp3.Call;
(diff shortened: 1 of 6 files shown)


rohanKanojia:
Ok to test

manusa:
I see the approach of get and then do create / update accordingly if the object exists or not.

I only see one flaw to this approach which is if the object is created (by another thread, process, service, etc.) in the time between get returns null and the current thread/process tries to create the object.

Is it possible to use the try-to-create-or-edit approach by detecting if the POST request failed due to the resource already being there? (I don't know if k8s returns a 409 status or something that will allow us to detect such a case)

Something like:
1. Try to create
2.A.1. Create works
2.A.2. Method returns  
2.B.1 Create fails due to conflict
2.B.2 Try with edit
2.B......
2.C.1 Create fails due to some other error
2.C.2 Throw exception

bbeaudreault:
I agree, that was my original thought as well, but decided to mirror BaseOperation for uniformity. But I will update to use that approach. We should probably update BaseOperation as well, but I don't have time to do that in this PR right now.

manusa:
> I agree, that was my original thought as well...

Thanks, I created an issue (#2292) to keep track of the change.

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/dsl/internal/RawCustomResourceOperationsImplTest.java
```diff
@@ -75,15 +76,42 @@ void testCreateOrReplaceUrl() throws IOException {
...
+    assertEquals("POST", captor.getAllValues().get(1).method());
+    assertEquals("/apis/test.fabric8.io/v1alpha1/hellos/myresource", captor.getAllValues().get(2).url().encodedPath());
+    assertEquals("PUT", captor.getAllValues().get(2).method());
+    assertEquals("/apis/test.fabric8.io/v1alpha1/namespaces/myns/hellos", captor.getAllValues().get(3).url().encodedPath());
+    assertEquals("POST", captor.getAllValues().get(3).method());
+  }
+
+  private Response mockResponse(int code) {
+    Response mockResponse = mock(Response.class);
+    when(mockResponse.isSuccessful()).thenReturn(code < 400);
```
rohanKanojia, line 111: nit, maybe use `HttpUrlConnection` for status codes:
```suggestion
    when(mockResponse.isSuccessful()).thenReturn(code < HttpURLConnection.HTTP_BAD_REQUEST);
```

bbeaudreault: Instead I updated to just build a Response object directly so we can re-use the internal isSuccessful method which does this comparison for us. I updated the other codes to use HttpURLConnection though.

bbeaudreault:
This has been updated as requested

bbeaudreault:
@rohanKanojia this is ready. can you re-run the 3.11 test?

the remaining 2 code smell seem more trouble than they are worth

rohanKanojia:
[merge]

Issue #1512, opened 2019-05-05
Change context/cluster programmatically


kchrusciel:
Is there any way to change context/cluster using kubernetes client?

rohanKanojia:
what kind of context? Are you talking about [config](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/JobExample.java#L41) object?

kchrusciel:
I talking about `kubectl config set-context CONTEXT_NAME`

wemu:
sorry for digging this out - but I ran into the same question. I am looking for the analog feature from kubectl:

```
kubectl config current-context
kubeint

kubectl config use-context kubeint
Switched to context "kubeint".
```

I only found a way to determine the current context:

```
import io.fabric8.kubernetes.api.model.Config;

Config config = KubeConfigUtils.parseConfig(new File(kubeConfig));
List<NamedContext> contexts = config.getContexts();
System.out.println("Current Context: " + config.getCurrentContext());
```

but no way to change the context or pass it into the KubernetesClient.
Is this not yet supported or is it hidden somewhere else?

rohanKanojia:
@wemu : I think all code related to `Config` is inside https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java 

Could you please check if we are supporting setting context there? If not, I think we should try to add this feature.

wemu:
Thanks for the pointer! Indeed the Config class in the client package has that functionality. 

I was using the ConfigBuilder but there is a more suitable method on the Config directly:

```
Config config = new ConfigBuilder().build(); // no context?
Config cfgWithContext = Config.autoConfigure("kubeint");

KubernetesClient client = new DefaultKubernetesClient(cfgWithContext);
```

I thought it should be on the client but kubectl behaves the same way (set the config, then execute commands). So that actually looks ok.
It is a bit inconvenient to read the available contexts via the model.Config api and then use it via the client.Config api - but so far it looks ok for what I plan to do.

Thanks!

pedro93:
+1 on this feature. 

My particular use-case is to check in the client API what context the client is currently in and if it's one of a selected few run additional code (e.g: port-forward a few services).

LeiYangGH:
i also met this issue. need a way to specify the context. following code not work(running but still pointing to default context)
```
[...]

PR #2290, opened 2020-06-15
expose current (named) and all (named) contexts

### LINKED ISSUE #2215: io.fabric8.kubernetes.client.Config should expose all and the current context defined in kubeconfig
Ex. If you need to list all clusters that are present in your kubeconfig there's no other way than to (re-) read and (re-) parse ```~/.kube/config```. ```io.fabric8.kubernetes.client.Config``` should therefore expose this info that is present in ```io.fabric8.kubernetes.api.model.Config```.
There are also more values in the kubeconfig that a 3rd party programmer might want/need (ex. current cluster, etc.).

adietish:
This PR changes `io.fabric8.kubernetes.client.Configuration` so that it stores all contexts that exist in the kubeconfig, does the same with the current context and exposes them via getters.

fixes #2215

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
@@ -16,26 +16,7 @@
 
 package io.fabric8.kubernetes.client;
 
-import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.annotation.JsonProperty;
-import io.fabric8.kubernetes.api.model.AuthInfo;
-import io.fabric8.kubernetes.api.model.Cluster;
-import io.fabric8.kubernetes.api.model.ConfigBuilder;
-import io.fabric8.kubernetes.api.model.Context;
-import io.fabric8.kubernetes.api.model.ExecConfig;
-import io.fabric8.kubernetes.api.model.ExecEnvVar;
-import io.fabric8.kubernetes.client.internal.CertUtils;
-import io.fabric8.kubernetes.client.internal.KubeConfigUtils;
-import io.fabric8.kubernetes.client.internal.SSLUtils;
-import io.fabric8.kubernetes.client.utils.IOHelpers;
-import io.fabric8.kubernetes.client.utils.Serialization;
-import io.fabric8.kubernetes.client.utils.Utils;
-import io.sundr.builder.annotations.Buildable;
-import okhttp3.TlsVersion;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import static okhttp3.TlsVersion.TLS_1_2;
 
 import java.io.BufferedReader;
 import java.io.File;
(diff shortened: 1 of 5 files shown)


rohanKanojia:
Ok to test

adietish:
"test" task failed with an error for JaCoco, not an error in the tests (which ran fine afais):
```
[ERROR] Failed to execute goal org.jacoco:jacoco-maven-plugin:0.8.5:report (default-report) on project kubernetes-karaf-itests: An error has occurred in JaCoCo report generation.: Error while creating report: null: EOFException -> [Help 1]
```

rohanKanojia:
retest this please

rohanKanojia:
@adietish: There is one bug related to NullPtrException being reported by Sonar. Do you think we can fix this?

rohanKanojia:
I'm not sure why but I'm seeing two commits. I think maybe the rebase wasn't done properly.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
```diff
@@ -597,6 +599,20 @@ private static boolean loadFromKubeconfig(Config config, String context, String
 
     return false;
   }
+
+  private static Context setCurrentContext(String context, Config config, io.fabric8.kubernetes.api.model.Config kubeConfig) {
```
rohanKanojia, line 603: Do you think we should maybe move this method to `KubeConfigUtils`

rohanKanojia: I haven't look deeply but would we be able to resolve https://github.com/fabric8io/kubernetes-client/issues/1512 if we expose this as public?

adietish: @rohanKanojia so the idea in #1512 is as far as I understand it is to be able to set the current context to the Config, right? The "problem" with that is that this would  imply an update to plenty of settings like the current cluster, namespace etc. Could be achieved by calling Config#autoConfigure behind the scene. I'd rather fix this in its very own issue.

adietish:
extracted a few code blocks to their own methods in `io.fabric8.kubernetes.client.Config` and added a test.

adietish:
@rohanKanojia ready to be merged if you're ok with it.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
```diff
@@ -460,41 +477,52 @@ public static Config fromKubeconfig(String kubeconfigContents) throws IOExceptio
...
   private static boolean tryKubeConfig(Config config, String context) {
     LOGGER.debug("Trying to configure client from Kubernetes config...");
-    if (Utils.getSystemPropertyOrEnvVar(KUBERNETES_AUTH_TRYKUBECONFIG_SYSTEM_PROPERTY, true)) {
-      String fileName = Utils.getSystemPropertyOrEnvVar(KUBERNETES_KUBECONFIG_FILE, new File(getHomeDir(), ".kube" + File.separator + "config").toString());
+    if (!Utils.getSystemPropertyOrEnvVar(KUBERNETES_AUTH_TRYKUBECONFIG_SYSTEM_PROPERTY, true)) {
+      return false;
+    }
+    File kubeConfigFile = new File(getKubeconfigFilename());
+    if (!kubeConfigFile.isFile()) {
+      LOGGER.debug("Did not find Kubernetes config at: ["+kubeConfigFile.getPath()+"]. Ignoring.");
```
rohanKanojia, line 491: nit, Could you please use format specifiers?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
```diff
@@ -1128,4 +1170,21 @@ public void setOauthTokenProvider(OAuthTokenProvider oauthTokenProvider) {
   public void setCustomHeaders(Map<String, String> customHeaders) {
     this.customHeaders = customHeaders;
   }
+
+  public List<NamedContext> getContexts() {
```
rohanKanojia, line 1174: I would appreciate it if you could add some Javadoc to publicly available methods introduced in this PR.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
adietish: > nit, Could you please use format specifiers?

you mean the slf4j facility, right?

adietish:
@rohanKanojia updated, please review.

rohanKanojia:
[merge]

PR #2291, opened 2020-06-16
Added a manualTypeMapping parameter for providing manual type mappings

rohanKanojia:
This PR modifies `kubernetes-model-generator` to accommodate manual type mapping. This was discussed in https://github.com/fabric8io/kubernetes-client/pull/2288 . With this in place we should be able to provide a HashMap in which we can configure which go structs should be mapped to some already existing java types


bbeaudreault:
@rohanKanojia if you wouldn't mind merging this one too. i'm trying to prune some custom commits in my company's fork now that we handled these cases upstream

rohanKanojia:
[merge]

PR #2310, opened 2020-06-24
move HasMetadataComparator to exported package

### LINKED ISSUE #2309: openshift-model has unresolvable osgi-import
the `io.fabric8:openshift-model:4.10.2` imports package `io.fabric8.kubernetes.internal` in its MANIFEST.MF but this package is private to `kubernetes-model-core`. 

Having a quick look on the sources the culprit seems to this import: https://github.com/fabric8io/kubernetes-client/blob/2ed58668ab6a0f83e038844d96efbbf4eea92a9c/kubernetes-model-generator/openshift-model/src/main/java/io/fabric8/openshift/api/model/Template.java#L30

joshiste:
The openshift-model needs to make use of the HasMetadataComparator and therefore
it should reside in an exported package

fixes #2309


rohanKanojia:
ok to test

rohanKanojia:
[merge]

rohanKanojia:
[merge]

PR #2374, opened 2020-07-28
Fix #2316: Cannot load resource from stream without apiVersion

### LINKED ISSUE #2316: Cannot load resource from stream without apiVersion
I am rendering a template and loading the resulting resource directly as inputstream. When loading the resource directly, in version `4.10.1` it was previously possible to simply specify `kind: Deployment` and have the client figure out the apiVersion. Since `4.10.2` this leads to a Jackson deserialization error. I am not sure whether this implied change of behavior was intended, but here's a simple reproducer for the problem that will throw the exception on `4.10.2` but work on `4.10.1`.

I understand that just specifying the `apiVersion` as `apps/v1` works in this case, but i am not sure if this change was intended.

```java

    @Test
    public void testTemplate() {
        final DefaultKubernetesClient client = new DefaultKubernetesClient();

        final String deployment = "kind: Deployment\n" +
                "metadata:\n" +
                "  name: \"test\"\n" +
                "  labels:\n" +
                "    app: \"test\"\n" +
                "spec:\n" +
                "  selector:\n" +
                "    matchLabels:\n" +
                "      app: \"test\"\n" +
                "  replicas: \"1\"\n" +
                "  template:\n" +
                "    metadata:\n" +
                "      labels:\n" +
                "        app: \"test\"\n" +
                "    spec:\n" +
                "      containers:\n" +
                "        - name: \"test\"\n" +
                "          image: \"busybox:latest\"\n" +
                "          command: [\"/bin/sh\", \"-c\"]\n" +
                "          args:\n" +
                "            - |\n" +
                "             sleep 60";

        final var deploymentObject = client.load(new ByteArrayInputStream(deployment.getBytes()));
    }
```

rohanKanojia:
Fix #2316 
In case of duplicate Kubernetes resources found, when no apiVersion
is provided return first item found instead of null

## Type of change

## Checklist

### DIFF
--- a/kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/internal/KubernetesDeserializer.java
+++ b/kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/internal/KubernetesDeserializer.java
@@ -126,6 +126,9 @@ public class KubernetesDeserializer extends JsonDeserializer<KubernetesResource>
     static class Mapping {
 
         private static final String KEY_SEPARATOR = "#";
+
+        // n.b. Packages sorted in order of precedence, deserialization of resources with no
+        // specific version will default to first available Class in one of these packages:
         private static final String[] PACKAGES = {
                 "io.fabric8.kubernetes.api.model.",
                 "io.fabric8.kubernetes.api.model.admission",
@@ -145,7 +148,6 @@ public class KubernetesDeserializer extends JsonDeserializer<KubernetesResource>
                 "io.fabric8.kubernetes.api.model.coordination.",
                 "io.fabric8.kubernetes.api.model.coordination.v1.",
                 "io.fabric8.kubernetes.api.model.discovery.",
-                "io.fabric8.kubernetes.api.model.extensions.",
                 "io.fabric8.kubernetes.api.model.events.",
                 "io.fabric8.kubernetes.api.model.networking.",
                 "io.fabric8.kubernetes.api.model.networking.v1beta1.",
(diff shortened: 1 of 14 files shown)


kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/internal/KubernetesDeserializer.java
```diff
@@ -248,16 +248,20 @@ private String getClassName(String key) {
...
+              for (Class<? extends KubernetesResource> result : possibleResults) {
                 String defaultKeyFromClass = getKeyFromClass(result);
                 if (key.equals(defaultKeyFromClass)) {
-                    return result;
+                  return result;
                 }
+              }
+              // Not able to compare with apiVersions, perhaps it was not provided;
+              // return first item found
+              return possibleResults.get(0);
```
manusa, line 261: Are we sure we want to return the first item found.
This means that `PACKAGES` must be sorted by order of preference (for packages providing same Types/Kinds). If so maybe we should document this somewhere.

rohanKanojia: I am also not sure whether returning first item found is good or not. But I thought if user wanted some specific apiVersion related resource. He should be providing apiVersion inside object

manusa: What I mean is that probably the PACKAGE array definition should have an n.b comment e.g.:
`// n.b. Packages sorted in order of precedence, deserialization of resources with no specific version will default to first available Class in one of these packages:`

manusa: Also unit tests for conflicting classes would help enforce list of entries doesn't get messed up (and can probably avoid the comment)

manusa:
[merge]

PR #2446, opened 2020-08-28
Quickstarts fails on OCP 4.6

rnetuka:
From downstream: https://github.com/jboss-fuse/kubernetes-client/pull/12

On newer Kubernetes, when WebSockets are not available, the server returns 503 Service Unavailable status code (formerly it returned 200 OK). Watch service needs to consider this response code too when trying to fall back to HTTP watch.

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
@@ -808,8 +808,8 @@ public class BaseOperation<T extends HasMetadata, L extends KubernetesResourceLi
     } catch (MalformedURLException e) {
       throw KubernetesClientException.launderThrowable(forOperationType("watch"), e);
     } catch (KubernetesClientException ke) {
-
-      if (ke.getCode() != 200) {
+      List<Integer> furtherProcessedCodes = Arrays.asList(200, 503);
+      if (! furtherProcessedCodes.contains(ke.getCode())) {
         if (watch != null) {
           //release the watch
           watch.close();
@@ -824,7 +824,7 @@ public class BaseOperation<T extends HasMetadata, L extends KubernetesResourceLi
         watch.close();
       }
 
-      // If the HTTP return code is 200, we retry the watch again using a persistent hanging
+      // If the HTTP return code is 200 or 503, we retry the watch again using a persistent hanging
       // HTTP GET. This is meant to handle cases like kubectl local proxy which does not support
       // websockets. Issue: https://github.com/kubernetes/kubernetes/issues/25126
       try {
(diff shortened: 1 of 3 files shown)


oscerd (approved):

rohanKanojia:
There is one test failing:
```
[ERROR] Failures: 
[ERROR]   WatchTest.testHttpErrorReconnect:193
[INFO] 
[ERROR] Tests run: 593, Failures: 1, Errors: 0, Skipped: 4
```

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
```diff
@@ -808,8 +808,8 @@ public Watch watch(ListOptions options, final Watcher<T> watcher) {
     } catch (MalformedURLException e) {
       throw KubernetesClientException.launderThrowable(forOperationType("watch"), e);
     } catch (KubernetesClientException ke) {
-
-      if (ke.getCode() != 200) {
+      List<Integer> furtherProcessedCodes = Arrays.asList(200, 503);
```
manusa, line 811: Maybe we can move this to a static concurrent set

rnetuka: Why static concurrent set? This is just more readable version of `ke.getCode() != 200 && ke.getCode() != 503`. It doesn't have any deeper purpose.

oscerd: Maybe a concurrent set is a bit overkilling in this case.

rnetuka:
@rohanKanojia Strange, it works for me. Could you please post a stack trace why does it fails?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
manusa: It just doesn't seem optimal to create a new instance of a List each time the method gets called especially if the list is never changed.
It doesn't make sense either to make it concurrent since the list is never written to (besides when it's initialized).

Anyway, it's fine as it is. :heavy_check_mark:

manusa (approved):

rohanKanojia:
@rnetuka : I checked by pulling your PR locally but the test was still failing for me. I think the reason it's failing is due to it assuming that 503 would refuse connection:
https://github.com/fabric8io/kubernetes-client/blob/a007adfd50e6603167aae5eac481215a4159f900/kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/WatchTest.java#L173

I think modifying server mocking to some other status code should fix it:
```diff
-    server.expect().withPath(path).andReturn(503, new StatusBuilder().withCode(503).build()).times(6);
+    server.expect().withPath(path).andReturn(HttpURLConnection.HTTP_NOT_FOUND, new StatusBuilder().withCode(HttpURLConnection.HTTP_NOT_FOUND).build()).times(6);

```

I tried pushing to your branch but it seems like you have disabled updates to your branch from maintainers. Could you please update your PR so that we can proceed with merge?

rnetuka:
@rohanKanojia Thank you for the review. I've updated the PR with your suggested changes.

rohanKanojia (approved):

rohanKanojia:
[merge]

rohanKanojia:
[merge]

manusa:
[merge]

PR #2468, opened 2020-09-07
Changes the resource kind in the ProjectRequestHandler from "Project" to "ProjectRequest"

### LINKED ISSUE #2442: ProjectRequestHandler is registered as handler for Project
After updating from the client version 4.10.2 to 4.11.0 I'm experiencing the following bug:

I'm using the client to create all resources of a KubernetesList in the following way:
`client.resourceList(resources).createOrReplace()`

The list also contains one resource of the kind "Project" and apiGroup "project.openshift.io/v1". When it's the project's turn to be created, I get the following error with new version of the client while it worked with the old one: `class io.fabric8.openshift.api.model.Project cannot be cast to class io.fabric8.openshift.api.model.ProjectRequest`.

I think the culprit lies somewhere in the resuscitation of the ProjectRequestHandler and I'd like to verify that. In [this line ](https://github.com/fabric8io/kubernetes-client/blob/master/openshift-client/src/main/java/io/fabric8/openshift/client/handlers/ProjectRequestHandler.java#L42) I guess `return Project.class.getSimpleName();` might need to be replaced with `return ProjectRequest.class.getSimpleName();`.

With the way the ProjectRequestHandler is written now, a call to `Handlers.get("Project", "project.openshift.io/v1")` will return the ProjectRequestHandler instead of the ProjectHandler and after that the erroneous cast happens. That's exactly what happens in the "acceptVisitors" method of NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl after the wrong handler is returned.

Obirah:
Fix #2442 

## Description
Changes the resource kind in the ProjectRequestHandler from "Project" to "ProjectRequest" so that the client does not wrongfully try to cast a Project to ProjectRequest.

## Type of change

## Checklist


rohanKanojia:
Thanks for your PR. Could you please add a test of loading a ProjectRequest and verifying that it's successful?

Obirah:
> Thanks for your PR. Could you please add a test of loading a ProjectRequest and verifying that it's successful?

I'll try, but unfortunately my IntelliJ has some problems to properly import and index the project. If I'm able to fix that, of course I'll code the test.

Obirah:
@rohanKanojia, I added it to the `HandlersTest` in the same fashion the other resources are tested. Is that sufficient or do we need a separate test that does even more?

rohanKanojia:
Umm, I think this should work. Are you able to resolve your issue while loading a generic resoucelist with this change?
```
client.resourceList(resources).createOrReplace();
```

Obirah:
> Umm, I think this should work. Are you able to resolve your issue while loading a generic resoucelist with this change?
> 
> ```
> client.resourceList(resources).createOrReplace();
> ```

Yes, I just created a small, temporary test on my machine that creates a KubernetesList containing a Project directly on one of my OpenShift clusters and it worked.

rohanKanojia (approved):

manusa (approved):

rohanKanojia:
[merge]

PR #2471, opened 2020-09-09
correct spelling on curd (crud) mode

### LINKED ISSUE #1106: what is 'curdMode' in mock/KubernetesServer

matthyx:
Rename 'curdMode' to 'crudMode' and add a comment what it does.
This is the same PR as #1107 but generalized to all other mocks.
Fixes #1106.

### DIFF
--- a/extensions/knative/mock/src/main/java/io/fabric8/knative/mock/KnativeServer.java
+++ b/extensions/knative/mock/src/main/java/io/fabric8/knative/mock/KnativeServer.java
@@ -33,7 +33,10 @@ public class KnativeServer extends ExternalResource {
   private KnativeClient client;
 
   private boolean https;
-  private boolean curdMode;
+  // In this mode the mock web server will store, read, update and delete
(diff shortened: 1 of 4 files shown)


matthyx:
Curd is a kind of cheese (https://en.wikipedia.org/wiki/Curd)...

oscerd (approved):

rohanKanojia:
Here in India, curd is used for yogurt

manusa (approved):

rohanKanojia (approved):

rohanKanojia:
[merge]

Issue #2458, opened 2020-09-02
creating new instance of the DefaultKubernetesClient is hanging 


umutkocasarac:
Creating new instance of the DefaultKubernetesClient is hanging and never responding. I could see the following messages  on the logs

```
16:36:38.471 [main] DEBUG io.fabric8.kubernetes.client.Config - Trying to configure client from Kubernetes config...
16:36:38.471 [main] DEBUG io.fabric8.kubernetes.client.Config - Found for Kubernetes config at: [/Users/----/.kube/config].
```

The reason for that is kubernetes client using the config file and trying to get token from the AWS with the aws cli. However there is an issue with the following code

```java
Process p = pb.start();
      if (p.waitFor() != 0) {
        LOGGER.warn(IOHelpers.readFully(p.getErrorStream()));
      }
      ExecCredential ec = Serialization.unmarshal(p.getInputStream(), ExecCredential.class);
```

ProcessBuilder should read the output of the command before waiting the result of the execution. It seems like an easy fix, I could provide a pull request if you want

rohanKanojia:
@umutkocasarac: Please go ahead and create a PR to fix this.

rohanKanojia:
Are you facing this issue after upgrade to v4.11.0 or are you using client for the first time?

rohanKanojia:
I think you should be facing this issue on older versions too since we haven't changed this code in recent releases.

umutkocasarac:
I am getting this error on both v4.10.3 and v4.11.0 versions

rohanKanojia:
pheww, It's not a regression :-) . Actually, since we released 4.11.0 people have reported a few regressions. I was a bit worried if this is another regression.

umutkocasarac:
hey @rohanKanojia I am getting permission error while pushing my changes. Do I need to do anything before pushing my changes to branch?

kubernetes-client % git push origin iss_2458
ERROR: Permission to fabric8io/kubernetes-client.git denied to umutkocasarac.
fatal: Could not read from remote repository.

rohanKanojia:
oh, you need to fork the repository and then push to your own fork.

1. Fork Fabric8 Github repository to your account
2. Add your fork remote `git remote add my-fork https://github.com/umutkocasarac/kubernetes-client.git`
3. Push to your fork `git push my-fork iss_2458`
4. Make PR against fabric8 master from Github UI :rocket:

manusa:
[...]

PR #2487, opened 2020-09-15
Read process output stream before waiting the exit code of the process

umutkocasarac:
## Type of change

## Checklist


umutkocasarac:
I accidentally deleted the previous branch while resolving conflict :( so I pushed my changes again

manusa:
Fixes #2458

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java
```diff
@@ -634,15 +634,23 @@ protected static ExecCredential getExecCredentialFromExecConfig(ExecConfig exec,
...
+        while ((line = in.readLine()) != null) {
+          sb.append(line);
+        }
+      }
       if (p.waitFor() != 0) {
-        LOGGER.warn(IOHelpers.readFully(p.getErrorStream()));
+        LOGGER.warn(sb.toString());
       }
-      ExecCredential ec = Serialization.unmarshal(p.getInputStream(), ExecCredential.class);
+      ExecCredential ec = Serialization.unmarshal(sb.toString(), ExecCredential.class);
```
manusa, line 653: In this excerpt you are no longer considering the error stream.

Would something like this work with your solution?
```suggestion
      final String errorOutput = IOHelpers.readFully(p.getErrorStream());
      final String standardOutput = IOHelpers.readFully(p.getInputStream());
      if (p.waitFor() != 0) {
        LOGGER.warn(errorOutput);
      }
      ExecCredential ec = Serialization.unmarshal(standardOutput, ExecCredential.class);
```

umutkocasarac: I am redirecting the error stream to outputstream with the following code
pb.redirectErrorStream(true);
and if the command exit code is other than 0 then reading the outputstream as error log, so we don't need to read ErrorStream anymore

manusa: Oh, sorry I missed that line.
However, I still don't see why we removed `IOHelpers#readFully` method in favor of the read stream boilerplate code.

umutkocasarac: @manusa because we have redirected all output to outputstream of the process and that will be read with the following code.

try (BufferedReader in = new BufferedReader(new InputStreamReader(p.getInputStream()))) {
        String line;
        while ((line = in.readLine()) != null) {
          sb.append(line);
        }
      }

and in case of failure **sb** variable will have the error messages
In case of error sb variable will have the error output

manusa: Yes, but can't seem to understand why that stream can't be read using
```
final String output = IOHelpers.readFully(p.getInputStream());
```
which in essence encapsulate that expanded code (or should).

umutkocasarac: I could read the output with that command however that helper method was not closing the anything. If you want I could change the 
`try (BufferedReader in = new BufferedReader(new InputStreamReader(p.getInputStream()))) {
        String line;
        while ((line = in.readLine()) != null) {
          sb.append(line);
        }
      }`

with `final String output = IOHelpers.readFully(p.getInputStream());`

manusa: This should take care of closing the stream and avoid extra boilerplate code:
```java
String output;
try (InputStream is = p.getInputStream()) {
  output = IOHelpers.readFully(is);
}
```

My main concern is that I wouldn't like to keep same read stream logic lying around several parts of the codebase. It would be hard then to maintain buffer sizes or other stream reading configurations.
IOHelpers also requires a refactor. Besides, in case we want to delegate this to some library (e.g. commons-io), we'll only need to change this in a single place.

umutkocasarac: It is updated with IOHelpers

manusa: Great, thx!

manusa (approved):
LGTM, thx!

rohanKanojia (approved):

oscerd (approved):

PR #2601, opened 2020-11-12
Allow to pick the CustomResourceDefinitionVersion with the highest priority (#2599)

### LINKED ISSUE #2599: Should be able to sort CustomResourceDefinitionVersion(s) according to their priority
When listing custom resources one has by default to display them in the version with the highest priority. One thus has to be able to sort available versions (CustomResourceDefinitionVersion) by their priority.

The kubernetes docs state the following at https://v1-17.docs.kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/:
> Version priority
>
>    Regardless of the order in which versions are defined in a CustomResourceDefinition, the version with the highest priority is used by kubectl as the default version to access objects. The priority is determined by parsing the name field to determine the version number, the stability (GA, Beta, or Alpha), and the sequence within that stability level.
>
>    The algorithm used for sorting the versions is designed to sort versions in the same way that the Kubernetes project sorts Kubernetes versions. Versions start with a v followed by a number, an optional beta or alpha designation, and optional additional numeric versioning information. Broadly, a version string might look like v2 or v2beta1. Versions are sorted using the following algorithm:
>
>* Entries that follow Kubernetes version patterns are sorted before those that do not.
>* For entries that follow Kubernetes version patterns, the numeric portions of the version string is sorted largest to smallest.
>* If the strings beta or alpha follow the first numeric portion, they sorted in that order, after the equivalent string without the beta or alpha suffix (which is presumed to be the GA version).
>* If another number follows the beta, or alpha, those numbers are also sorted from largest to smallest.
>* Strings that don’t fit the above format are sorted alphabetically and the numeric portions are not treated specially. Notice that in the example below, foo1 is sorted above foo10. This is different from the sorting of the numeric portion of entries that do follow the Kubernetes version patterns.

adietish:
fixes #2599 

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/CustomResourceDefinitionContext.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/CustomResourceDefinitionContext.java
@@ -18,10 +18,15 @@ package io.fabric8.kubernetes.client.dsl.base;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.Namespaced;
 import io.fabric8.kubernetes.api.model.apiextensions.v1beta1.CustomResourceDefinition;
+import io.fabric8.kubernetes.api.model.apiextensions.v1beta1.CustomResourceDefinitionVersion;
 import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.utils.ApiVersionUtil;
 import io.fabric8.kubernetes.client.utils.Pluralize;
 import io.fabric8.kubernetes.client.utils.Utils;
+import io.fabric8.kubernetes.api.model.apiextensions.v1beta1.CustomResourceDefinitionSpec;
+import io.fabric8.kubernetes.client.utils.KubernetesVersionPriority;
+
+import java.util.stream.Collectors;
 
 public class CustomResourceDefinitionContext {
   private String name;
@@ -81,12 +86,23 @@ public class CustomResourceDefinitionContext {
   public static CustomResourceDefinitionContext fromCrd(CustomResourceDefinition crd) {
     return new CustomResourceDefinitionContext.Builder()
       .withGroup(crd.getSpec().getGroup())
-      .withVersion(crd.getSpec().getVersion())
+      .withVersion(getVersion(crd.getSpec()))
       .withScope(crd.getSpec().getScope())
       .withName(crd.getMetadata().getName())
       .withPlural(crd.getSpec().getNames().getPlural())
       .withKind(crd.getSpec().getNames().getKind())
       .build();
+  }
+
+  private static String getVersion(CustomResourceDefinitionSpec spec) {
+    String version = KubernetesVersionPriority.highestPriority(
+            spec.getVersions().stream()
+                    .map(CustomResourceDefinitionVersion::getName)
+                    .collect(Collectors.toList()));
+    if (version == null) {
+      version = spec.getVersion();
+    }
+    return version;
   }
 
   public static class Builder {
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesVersionFactory.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesVersionFactory.java
@@ -1,0 +1,248 @@
+/**
+ * Copyright (C) 2015 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.fabric8.kubernetes.client.utils;
+
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * A factory that is creating a version instance for a given version string.
+ * Returns {@link KubernetesVersion} instances for kubernetes versions (starts with 'v')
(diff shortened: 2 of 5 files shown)


oscerd (approved):

adietish:
I don't agree with sonar complaining about 2 assertions in my tests (the 2 code smells reported above): it wants me to use #contains instead of #equals which afaics don't have the same meaning.

rohanKanojia:
You can simply ignore Sonar warnings if they don't make sense in your opinion :-)

rohanKanojia:
In my opinion, it would be nice if we have some method like this:
```java
public static String highestPriorityVersion(CustomResourceDefinition crd);
public static String highestPriorityVersion(List<CustomResourceDefinitionVersion> versions);

```

adietish:
> In my opinion, it would be nice if we have some method like this:
> 
> ```java
> public static String highestPriorityVersion(CustomResourceDefinition crd);
> public static String highestPriorityVersion(List<CustomResourceDefinitionVersion> versions);
> ```

I had typed parameters and return values using CustomResourceDefinitionVersion in my first take on this. I got off it bcs there are different classes for each version: `io.fabric8.kubernetes.api.model.apiextensions.v1` and `io.fabric8.kubernetes.api.model.apiextensions.v1beta1`. Using these classes would force me to provide overloads for each of them. Using simple Strings thus looks better to me.

adietish:
I'll update my PR to also include a change in `CustomResourceDefinitionContext#fromCrd` which picks the version indicated in `CustomResourceDefinitionSpec#getVersion` where it should pick the version with the highest priority within `CustomResourceDefinitionSpec#getVersions`

adietish:
I just discovered the changes to `CustomResourceDefinitionContext` made in 9662c38. I rather doubtful that the following would work correctly in `fromCustomResourceType(Class<? extends HasMetadata>)`:
```
String version = ApiVersionUtil.apiVersion(instance, "v1");
```
What happens if there's no "v1" but a "v1alpha1" and "v1beta1"?

rohanKanojia:
Not sure if I understand...Where is this being used? Are you talking about this place?

https://github.com/fabric8io/kubernetes-client/blob/5bb774dc16c17d09b1fc1d446d6b254d729306d9/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationContext.java#L88

Are you referring to `CustomResourceDefinitionContext` using `v1beta1` CustomResourceDefinition in `fromCrd` method?

adietish:
> Not sure if I understand completely...Where is this being used? Are you talking about this place?
> 
> https://github.com/fabric8io/kubernetes-client/blob/5bb774dc16c17d09b1fc1d446d6b254d729306d9/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationContext.java#L88
> 
> Are you referring to `CustomResourceDefinitionContext` using `v1beta1` CustomResourceDefinition in `fromCrd` method?

Sorry, my bad, was mislead by sub optimal UI in IntelliJ. The correct commit is 258ffb8. What I was referring to was :
https://github.com/fabric8io/kubernetes-client/commit/258ffb8#diff-0a2faee0e48ea089c119f1ce3b464f87493d36fddba3275eb21b96231988ea47R66
`CustomResourceDefinitionContext`
```
  public static CustomResourceDefinitionContext fromCustomResourceType(Class<? extends HasMetadata> customResource) {
    HasMetadata instance;
    try {
      instance = customResource.newInstance();

      String kind = Utils.isNullOrEmpty(instance.getKind()) ? customResource.getSimpleName() : instance.getKind();
      String name = kind.toLowerCase();
      String plural = Pluralize.toPlural(name);
      String group = ApiVersionUtil.apiGroup(instance, null);
      String version = ApiVersionUtil.apiVersion(instance, "v1");   <---------------------------------------------------
      String scope = instance instanceof Namespaced ? "Namespaced" : "Cluster";

      return new CustomResourceDefinitionContext.Builder()
        .withGroup(group)
        .withVersion(version)
        .withScope(scope)
        .withName(name)
        .withPlural(plural)
        .withKind(kind)
        .build();
    } catch (InstantiationException | IllegalAccessException e) {
    }
  }
``` 
I'm doubtful that assuming "v1" as version would be the right choice when judging from the spec. How would kube behave if there's no such version exist? Shouldn't we request the version in the custom resource instead?

rohanKanojia:
oh, I think this is related to the new CustomResource DSL improvements we recently introduced. We now have a new method in DSL with which you can use any CustomResource with the client like this:
```java
MixedOperation<Foo, KubernetesResourceList<Foo>, Resource<Foo>> fooClient = k8sClient.customResource(Foo.class);
```
Here client just tries to build an opinionated `CustomResourceDefinitionContext` from the provided type. At this point, we don't have access to CustomResource; I'm not sure we can extract the version from CustomResource.

If the user wants to override this behavior, he can simply use DSL method with explicit `CustomResourceDefinitionContext`:
```java
CustomResourceDefinitionContext fooContext = new CustomResourceDefinitionContext.Builder()
      .withName("foo.example.com")
      .withGroup("example.com")
      .withVersion("v1alpha1")
      .withPlural("foos")
      .withScope("Namespaced")
      .build();
MixedOperation<Foo, KubernetesResourceList<Foo>, Resource<Foo>> fooClient = k8sClient.customResource(fooContext, Foo.class, FooList.class);
```

adietish:
ok, my bad, misread the code, oversaw that the instance is created within the method from a type. "v1" is the default that it falls back to. I still believe that it's nothing more than a possible value, there's no guarantee or heightened plausibility for it. Nevertheless, I can't suggest any better solution for now.

manusa:
Hi @adietish, tests are not being run in the CI env. Would you mind switching to JUnit5?
(`org.junit.jupiter.api.Test` instead of `org.junit.Test`)

We'll probably release a 5.0.0-alpha-3 version soon and it would be nice to have this in.

adietish:
@manusa changed the import, rebased and pushed again.

adietish:
@manusa weird, looks as if test now dont run. K8S and OpenShift checks stay yellow, wont turn green as they did before. Any idea?

rohanKanojia:
I've restarted the workflow for now. I hope it'll pass

rohanKanojia:
Could you please try pushing again? Strange, even after restarting it seems stuck

manusa (approved):

rohanKanojia (approved):

manusa:
[merge]

manusa:
[merge]

manusa:
[merge]

PR #2604, opened 2020-11-15
make build near reproducible (jandex index is not)

hboutemy:
see https://maven.apache.org/guides/mini/guide-reproducible-builds.html
from my test, jandex index is not reproducible but everything else is: this improvement will have to be done later

on next release published to Central Repository, I'll independently rebuild and check against reference build, to publish result to Reproducible Central https://maven.apache.org/guides/mini/guide-reproducible-builds.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 ## CHANGELOG
 
+### 5.1-SNAPSHOT
+
+#### Bugs
+
+#### Improvements
+
+#### Dependency Upgrade
+
+#### New Features
+
 ### 5.0.1 (2021-01-27)
 
 #### Bugs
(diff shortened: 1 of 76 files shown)


oscerd (approved):

manusa:
Sorry, I broke this PR when updating the branch. See #2760.

hboutemy:
Jandex Maven Plugin PR opened https://github.com/wildfly/jandex-maven-plugin/pull/26

other issues found when rebuilding: https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/io/fabric8/kubernetes-client/kubernetes-client-project-5.1.0.diffoscope
will need to investigate and fix

Issue #2614, opened 2020-11-20
Improve Watch.close so that error can always be reported


metacosm:
@manusa's idea:
> Change the `onClose` signature to have a dedicated exception parameter i.e. `WatcherException`. This exception should only have a constructor with the message and cause. So on the calling methods, we should always propagate the cause.

manusa:
Relates to: #2273

PR #2616, opened 2020-11-20
refactor: Watcher.onClose has dedicated WatcherException as parameter.

manusa:
## Description

refactor: `Watcher#onClose` has dedicated `WatcherException` as parameter.

 - Refactored common parts of Watcher implementations into `AbstractWatchManager`

#### Watcher interface refactor

I've refactored the `Watcher` interface and the methods calling it.

The interface now provides a dedicated default argument-less `onClose()` method. This method should always be called upon Watcher completion. Implementations should override this method in case operations (e.g. cleanup) should be performed once the Watcher is closed.

The previous `onClose(WatcherException cause)` is kept and only invoked whenever an exception occurs during the Watcher life-cycle (once the watch connection is open and established).

To be clear, Exceptions that happen during Watch initiation should be captured with code like (this has not changed):
```java
try (
  Watch watch = kubernetesClient.pods().watch(new WatcherImpl())
) {
   // ...
} catch (Exception exception) {
  // act upon exceptions that may happen synchronously when opening the watch
}
```

Maybe further improvements can be performed or [method signatures](https://github.com/fabric8io/kubernetes-client/blob/10d55aa5a022b176bc06667abac809a579b31bfd/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Watchable.java#L29) changed so that they throw checked exceptions.

Fixes: #2614

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
@@ -63,6 +63,14 @@ public abstract class AbstractWatchManager<T> implements Watch {
     watcher.onClose(cause);
   }
 
+  final void closeEvent() {
+    if (forceClosed.getAndSet(true)) {
+      logger.debug("Ignoring duplicate firing of onClose event");
+      return;
+    }
+    watcher.onClose();
+  }
+
   static void closeWebSocket(WebSocket webSocket) {
     if (webSocket != null) {
       logger.debug("Closing websocket {}", webSocket);
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
@@ -1,0 +1,78 @@
+/**
+ * Copyright (C) 2015 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.fabric8.kubernetes.client.dsl.internal;
+
+import io.fabric8.kubernetes.api.model.ListOptions;
+import io.fabric8.kubernetes.client.Watch;
+import io.fabric8.kubernetes.client.Watcher;
+import io.fabric8.kubernetes.client.WatcherException;
+import okhttp3.OkHttpClient;
+import okhttp3.WebSocket;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+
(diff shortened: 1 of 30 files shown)


metacosm (changes requested):
This doesn't address the fact that `WatchConnectionManager.close` calls `closeEvent(null)`, which is why you get empty errors.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
```diff
@@ -0,0 +1,78 @@
...
+    this.listOptions = listOptions;
+    this.reconnectLimit = reconnectLimit;
+    this.reconnectInterval = reconnectInterval;
+    this.maxIntervalExponent = maxIntervalExponent;
+    this.clonedClient = clonedClient;
+    this.resourceVersion = new AtomicReference<>(listOptions.getResourceVersion());
+    this.forceClosed = new AtomicBoolean();
+  }
+
+  final void closeEvent(WatcherException cause) {
```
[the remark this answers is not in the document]
manusa: It's a helper method only used by the *Manager implementations. I'm still working on that, but this method is not supposed to be called from the outside neither improved.
As I said, I'm trying to move as much as possible from the vaious *Managers to this class, but it's a WIP.

manusa:
> This doesn't address the fact that `WatchConnectionManager.close` calls `closeEvent(null)`, which is why you get empty errors.

The problem is that many watchers may want to get informed of when the connection was closed.
This close can happen due to the natural lifecycle of the Watcher (it was either closed from the server or from the client because the watch ended), or due to an exception.


What we can do to improve this is provide 2 methods instead of 1: `onCloseExcpetion` and `onClose`. The former is called when the connection is closed due to errors (broken pipe, connection, etc.), the latter when the connection has closed due to natural Watcher life-cycle reasons.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
```diff
@@ -331,7 +331,7 @@ public void waitUntilReady() {
   @Override
   public void close() {
     logger.debug("Force closing the watch {}", this);
-    closeEvent(null);
+    closeEvent();
```
metacosm, line 334: I don't think that this addresses the issue because very often what triggers the call to `close` here is an exception, which is still not propagated in that case.

manusa: I really don't see how an Exception will trigger this close since this method will most probably be called by try-with-resources Closable.

In the event of an Exception triggering this close, the client would also be aware of the Exception via the catch block.

Do you have any code which shows this "because very often what triggers the call to close here is an exception" so I can understand better?

metacosm (approved):
LGTM

manusa:
Hi @jonathanvila, maybe some of these changes are of interest to you.

manusa:
[TestWatcher.zip](https://github.com/fabric8io/kubernetes-client/files/5583721/TestWatcher.zip)

Previous script can be used to test the new implementation.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
```diff
@@ -107,7 +85,7 @@ public WatchConnectionManager(final OkHttpClient client, final BaseOperation<T,
     this(client, baseOperation, listOptions, watcher, reconnectInterval, reconnectLimit, websocketTimeout, 5);
   }
 
-  private final void runWatch() {
+  private void runWatch() {
     logger.debug("Connecting websocket ... {}", this);
```
metacosm, line 89: Output target API URL instead of rather useless manager object reference
```suggestion
    logger.debug("Connecting websocket to {}...", requestURL);
```

manusa:
[merge]

Issue #2668, opened 2020-12-16
BuildConfig watch performs request on wrong endpoint


manusa:
## Description

The code introduced in #2490, more specifically the following lines

https://github.com/fabric8io/kubernetes-client/blob/579d9728635bb1346b3c4e97eff1cb517e0af249/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/OpenShiftOperation.java#L110-L116

cause the watch connection to be performed to the following (invalid) endpoint:
```
/apis/build.openshift.io/v1/namespaces/test/builds/$resourceName?watch=true
```
instead of:
```
/apis/build.openshift.io/v1/namespaces/test/builds?fieldSelector=metadata.name%3D$resourceName&watch=true
```

This is originated in the following code:
https://github.com/fabric8io/kubernetes-client/blob/579d9728635bb1346b3c4e97eff1cb517e0af249/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java#L126-L133
Since  `isApiGroup()` call returns `true` for OpenShift resources. In case of standard model types, apiVersion field is null, so method returns `false`.

The #2490 PR reverts #2373, however, those lines weren't part of that initial PR.

Relates to: https://github.com/openshift/origin/issues/25759

manusa:
Removing the `updateApiVersion()` and its invocation in the constructor fixed the issue in JKube (Using the code provided inhttps://github.com/eclipse/jkube/pull/524 and updating the dependency to `5.0-SNAPSHOT` with the fixed client installed locally).

So this seems to be the fix, once we find out why those lines were added in the first place and we make sure we don't break anything else by removing them.

rohanKanojia:
I did a git bisect between v4.10.3 and v4.12.0 and it's pointing to this commit: a9caeb1a8be898df13f9ca03cf90398cac0876a2

Here are git bisect logs:
```
kubernetes-client : $ git bisect log
git bisect start
# bad: [9dc84ecafba374b23324cff60bee56c53737315a] [RELEASE] Updated project version to v4.12.0
git bisect bad 9dc84ecafba374b23324cff60bee56c53737315a
# good: [9d05eb1fbb6134518c9b2991b3b372b13bd501d0] Updated project version to v4.10.3
git bisect good 9d05eb1fbb6134518c9b2991b3b372b13bd501d0
[...]

PR #2671, opened 2020-12-16
Fixes around exceptional cases in watchers

bbeaudreault:
## Description
These changes have been running for us internally for 6 months or so, but I never got a chance to upstream them. We'd like to bring our fork more in-line with mainstream, so here is a group of changes related to watcher reliability:

- _If the target of a watch is deleted, it will never satisfy the watch condition_
We saw this case where you are watching a resource hoping it will hit a particular condition, if the resource gets deleted while you are waiting the watcher will simply go on forever until the timeout. This change makes it fail much faster in the case where it will never succeed.

- _Path segment based filters have spotty support in newer kube api versions. Field selector seems to work everywhere_
It looks like this was added [3.5 years ago](https://github.com/fabric8io/kubernetes-client/pull/816) because field selectors weren't working. The patch to make CRDs support field selectors was then added [a few months later](https://github.com/kubernetes/kubernetes/pull/53345). If you take a look at the API spec (even for the really old v1.11 [here](https://v1-16.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#read-pod-v1-core)), the read API url we are using here does not support the `watch` query param. If you want to watch an object, you need to use either the Watch API (with `/api/v1/watch` path) or the List API, which _does_ support the `watch` query param. We noticed that watches for specific resources were not working correctly for us, instead they were spamming our API servers because they were actually short polling (basically just a get that returns instantly). Note that I linked to the Pods API, but the same seems to be true for CustomResources (or as the original PR called them, API Groups). You can test this out pretty easily:

```
kubectl proxy
# then in another terminal:

# this returns instantly, with no watch:
curl http://127.0.0.1:8001/apis/zk.hubspot.com/v1/namespaces/zk/zookeeperclusters/zk-test-qa\?watch=1

# this properly watches only the resource in question:
 curl http://127.0.0.1:8001/apis/zk.hubspot.com/v1/namespaces/zk/zookeeperclusters/\?watch=1\&fieldSelector=metadata.name=zk-test-qa

# this similarly works:
curl http://127.0.0.1:8001/apis/zk.hubspot.com/v1/watch/namespaces/zk/zookeeperclusters/\?watch=1\&fieldSelector=metadata.name=zk-test-qa

# this watches ALL resources of the type, as you'd expect:
curl http://127.0.0.1:8001/apis/zk.hubspot.com/v1/namespaces/zk/zookeeperclusters/\?watch=1
```

This is obviously a CRD we have internally, you can create your own CRD of any type to test it. It also works the same on standard APIs. I chose to simply remove this apiGroup handling here so that we always use the fieldSelector filtering. It's possible that another change we could make is to move over to the `/watch` API, but that would be more complicated because we'd need to use a different URL than what we have available from `getNamespacedUrl()`. I could hook that up if you want, but it mean somewhat wider scope changes. I'd prefer to handle that separately/later if possible.

- _Handle unsuccessful watch statuses better -- we should either be retrying, or passing an error event back to the user. Throwing an exception will result in a transparent failure that no user can handle, because it happens in a thread_
We noticed this issue where our watchers were failing silently with no real feedback. This is because the client was throwing a failure exception in the callback which does not get executed in the main thread. So the exception really goes nowhere. I refactored  callback so that when `!isSuccessful()` we do the same status handling that we do in the case of a successful response, we just pass in the unsuccessful Status object. This way the watcher will receive an appropriate event and the caller can be notified that the watch failed.

- _long polls return a response once data is available. we should only backoff in exceptional cases, otherwise immediately connect so we can receive followup data in a more timely manner_
Another situation we saw is we some watch events would not make their way to us in a timely manner. This was because the WatchHTTPManager was always backing off after processing a response body. The way long polling works is, when data is ready, the connection immediately returns a response with the data in the body. At that point, the long poll needs to be reinitialized. This is not an error state, the long poll should _immediately_ be reinitialized. So I updated the onResponse callback and scheduleReconnect to only backoff in the case of errors.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### 5.0-SNAPSHOT
 
 #### Bugs
+* Fix: Reliability improvements to watchers
 * Fix #2592: ConcurrentModificationException in CRUD KubernetesMockServer
 * Fix #2519: Generated schemas contains a valid meta-schema URI reference (`http://json-schema.org/draft-05/schema#`)
 * Fix #2631: Handle null values when getting current context on OIDC interceptors
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/WaitForConditionWatcher.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/WaitForConditionWatcher.java
@@ -48,6 +48,8 @@ public class WaitForConditionWatcher<T extends HasMetadata> implements Watcher<T
       case DELETED:
         if (condition.test(null)) {
           future.complete(null);
+        } else {
+          future.completeExceptionally(new WatcherException("Unexpected deletion of watched resource, will never satisfy condition"));
         }
         break;
       case ERROR:
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchConnectionManager.java
@@ -98,25 +98,16 @@ public class WatchConnectionManager<T extends HasMetadata, L extends KubernetesR
     String fieldQueryString = baseOperation.getFieldQueryParam();
     String name = baseOperation.getName();
 
-    // for API groups we can use the name in the path rather than a fieldSelector
-    // which is more likely to work well for API Groups
     if (name != null && name.length() > 0) {
-      if (baseOperation.isApiGroup()) {
-        httpUrlBuilder.addPathSegment(name);
-      } else {
-        if (fieldQueryString.length() > 0) {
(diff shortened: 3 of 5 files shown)


manusa:
Timing is impecable on this one. We identified #2668 yesterday.

I'm still wary about why these lines were added in the first place:
```java
 if (baseOperation.isApiGroup()) { 
   httpUrlBuilder.addPathSegment(name); 
 } 
```

Do you think it has something to do with K8s initial CRD implementations/spec?

manusa:
Relates to #2616 #2614

manusa (approved):
LGTM

manusa:
I confirm this PR also fixed the issue reported in #2668

rohanKanojia (approved):

manusa:
[merge]

Issue #2118, opened 2020-04-08
local development setup


Fabian-K:
Hi,

I find it rather difficult and time-consuming to compile the project locally. For the latest commit on master (be6b4722a8381b8c6117a14430f081c096424776) this is even failing.

running `mvn clean` completes successfully. Afterwards, I execute `mvn install` which fails after quite some time with

````
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Fabric8 :: Kubernetes :: Project 4.9-SNAPSHOT:
[INFO] 
[INFO] Fabric8 :: Kubernetes :: Project ................... SUCCESS [  2.694 s]
[INFO] Fabric8 :: Kubernetes Model Parent ................. SUCCESS [  0.313 s]
[INFO] Fabric8 :: Kubernetes Model :: Common .............. SUCCESS [  1.211 s]
[INFO] Fabric8 :: Kubernetes Model Annotator .............. SUCCESS [  1.028 s]
[INFO] Fabric8 :: Kubernetes Model ........................ SUCCESS [03:12 min]
[INFO] Fabric8 :: Kubernetes :: Java Client ............... SUCCESS [01:32 min]
[INFO] Fabric8 :: Kubernetes :: Server Mock ............... SUCCESS [ 16.610 s]
[INFO] Fabric8 :: Openshift :: Java Client ................ SUCCESS [ 11.143 s]
[INFO] Fabric8 :: Kubernetes :: Extensions ................ SUCCESS [  0.029 s]
[INFO] Fabric8 :: Knative :: Extension .................... SUCCESS [  0.027 s]
[INFO] Fabric8 :: Knative :: Model :: Annotator ........... SUCCESS [  0.751 s]
[INFO] Fabric8 :: Knative :: Model ........................ SUCCESS [ 47.059 s]
[INFO] Fabric8 :: Knative :: Client ....................... SUCCESS [  3.867 s]
[INFO] Fabric8 :: Knative :: Server Mock .................. SUCCESS [  0.310 s]
[INFO] Fabric8 :: Knative :: Examples ..................... SUCCESS [  0.133 s]
[INFO] Fabric8 :: Knative :: Tests ........................ SUCCESS [  5.537 s]
[INFO] Fabric8 :: Tekton :: Extension ..................... SUCCESS [  0.032 s]
[INFO] Fabric8 :: Tekton :: Model :: Annotator ............ SUCCESS [  0.657 s]
[...]

PR #2676, opened 2020-12-17
Allow specifying PropagationPolicy when using deleteExisting

bbeaudreault:
## Description
As the title says, this is something the API supports but kubernetes-client's generics were just not set up properly to allow crafting such a request. This updates the generics to support that use-case. Added an IT and unit test for the new functionality.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/DeleteAndCreateHelper.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/DeleteAndCreateHelper.java
@@ -23,15 +23,25 @@ import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.ResourceHandler;
 import okhttp3.OkHttpClient;
 
+import java.net.HttpURLConnection;
+import java.util.Objects;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 import java.util.function.UnaryOperator;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 public class DeleteAndCreateHelper<T extends HasMetadata> {
+  private static final Logger LOG = LoggerFactory.getLogger(DeleteAndCreateHelper.class);
+
   private final UnaryOperator<T> createTask;
+  private final Function<T, Boolean> awaitDeleteTask;
   private final Function<T, Boolean> deleteTask;
 
-  public DeleteAndCreateHelper(UnaryOperator<T> createTask, Function<T, Boolean> deleteTask) {
+  public DeleteAndCreateHelper(UnaryOperator<T> createTask, Function<T, Boolean> deleteTask, Function<T, Boolean> awaitDeleteTask) {
     this.createTask = createTask;
+    this.awaitDeleteTask = awaitDeleteTask;
     this.deleteTask = deleteTask;
   }
 
@@ -40,15 +50,40 @@ public class DeleteAndCreateHelper<T extends HasMetadata> {
     if (Boolean.FALSE.equals(deleted)) {
       throw new KubernetesClientException("Failed to delete existing item:" + item.getMetadata().getName());
     }
-    return createTask.apply(item);
+    try {
+      return createTask.apply(item);
+    } catch (KubernetesClientException e) {
+      // depending on the grace period, the object might not actually be deleted by the time we try to create
+      // if that's the case, give it some time.
+      if (e.getCode() == HttpURLConnection.HTTP_CONFLICT) {
+        if (Boolean.FALSE.equals(awaitDeleteTask.apply(item))) {
+          throw new KubernetesClientException("Timed out waiting for item to be deleted before recreating: " + item.getMetadata().getName(), e);
+        }
+        return createTask.apply(item);
+      }
+      throw e;
+    }
   }
 
   public static HasMetadata deleteAndCreateItem(OkHttpClient client, Config config, HasMetadata meta, ResourceHandler<HasMetadata, HasMetadataVisitiableBuilder> h, String namespaceToUse, DeletionPropagation propagationPolicy) {
     DeleteAndCreateHelper<HasMetadata> deleteAndCreateHelper = new DeleteAndCreateHelper<>(
       m -> h.create(client, config, namespaceToUse, m),
-      m -> h.delete(client, config, namespaceToUse, propagationPolicy, m)
+      m -> h.delete(client, config, namespaceToUse, propagationPolicy, m),
+      waitUntilDeletedOrInterrupted(client, config, h, namespaceToUse)
     );
 
     return deleteAndCreateHelper.deleteAndCreate(meta);
+  }
+
+  private static <T extends HasMetadata> Function<T, Boolean> waitUntilDeletedOrInterrupted(OkHttpClient client, Config config, ResourceHandler<HasMetadata, HasMetadataVisitiableBuilder> h, String namespaceToUse) {
+    return m -> {
+      try {
+        return h.waitUntilCondition(client, config, namespaceToUse, m, Objects::isNull, 30 , TimeUnit.SECONDS) == null;
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        LOG.warn("interrupted waiting for item to be deleted, assuming not deleted");
+        return false;
+      }
+    };
   }
 }
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/DeleteAndCreateHelper.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/DeleteAndCreateHelper.java
@@ -34,7 +34,8 @@ import org.slf4j.LoggerFactory;
 
 public class DeleteAndCreateHelper<T extends HasMetadata> {
   private static final Logger LOG = LoggerFactory.getLogger(DeleteAndCreateHelper.class);
-
+  private static final int MAX_WAIT_SECONDS = 30;
+  
   private final UnaryOperator<T> createTask;
   private final Function<T, Boolean> awaitDeleteTask;
   private final Function<T, Boolean> deleteTask;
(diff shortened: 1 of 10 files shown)


bbeaudreault:
I have a very hard time running kubernetes-client tests locally or in my company's build environment. Will mark this PR ready for review once the checks above have passed.

rohanKanojia:
Which tests are you referring to? Mock Server tests in `kubernetes-tests` module or regression tests in `kubernetes-itests` module?

bbeaudreault:
I can basically never get the project to fully build in IntelliJ, usually due to weird configuration issues with all of the generated models or some other compile issue in one of the various extensions. One major issue is that fabric8's generated classes all go to `target/generated-sources` directly, rather than to a subdirectory within there. I think they should probably be sent to `target/generated-sources/schemagen` or something like that. The problem with going directly to generated-sources is that intellij assumes `io` is the container src dir, not part of the package. This messes up imports. So I have a hard time running any even unit tests within the IDE. That slows things down a lot. Do you have any tips on getting kubernetes-client to build smoothly in intellij?

I _can_ get it to work using `mvn clean verify` on the command line, but it takes forever and eventually fails in the karaf-itests module. In the past I've tried limiting to just particular modules using `-pl`, but there is a lot of interdependency so it's like whack-a-mole trying to figure out the write set of modules to include.

I can't run the regression tests in `kubernetes-itests`, because I don't have a test environment to run them against. My machine is very much set up by my company's configuration scripts to be tightly coupled to our kubernetes environments, and we really dont want all of these actions executing one of our live clusters. I think I'd probably have to do this all in some sort of container setup specifically for fabric8 to ensure it's totally isolated, and I'm not sure how performant that would be.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/DeleteAndCreateHelper.java
```diff
@@ -40,15 +50,40 @@ public T deleteAndCreate(T item) {
     if (Boolean.FALSE.equals(deleted)) {
       throw new KubernetesClientException("Failed to delete existing item:" + item.getMetadata().getName());
     }
-    return createTask.apply(item);
+    try {
+      return createTask.apply(item);
+    } catch (KubernetesClientException e) {
```
bbeaudreault, line 55: I don't think this is related to my original change to support PropagationPolicy, but my new integration test caught this issue. The default grace period is 30s, so unless you explicitly specify 0 seconds (which we don't here) it's possible that the DELETE call returns a 202 instead of 200, meaning it hasn't actually been deleted yet.

BaseOperation/OperationSupport/etc assume that a non-error response means the item was deleted. This is sort of true, but not really since it may have actually been queued for asynchronous deletion (per the 202 response). I looked into whether I could pipe the response Status through as the return value of the various `delete()` functions, but that proved to be a significant change (and probably a breaking one, since right now we expect to return a simple Boolean).

The simpler fix here seems to be to add a waiting period as I did here. Alternatively we could continually retry createTask for a period of time, but this might be less impactful on API servers?

bbeaudreault:
I think this is ready for review. The E2E tests all passed, and the unit tests all pass locally. The builds both failed due to some sort of timeout downloading dependencies. Any chance you can kick off another build?

rohanKanojia:
You're not the first one pointing this out :-( (#2118). This project has grown a lot over time and became quite bulky. I don't know how other manage but I usually load project as maven project in my IntelliJ after `mvn clean install -DskipTests`(which takes 6-7 minutes). If I have to run/debug some test in IntelliJ, I usually comment out `kubernetes-model-generator` in root pom.xml and then reload. With this I'm able to run/debug stuff. 

For integration tests, I have to add `kubernetes-itests` module to my IDE and run it on some local kubernetes cluster(`minikube`/`minishift` etc)

bbeaudreault:
Looks like builds are still having issues. Not sure if there's something bigger going on with builds currently

bbeaudreault:
@manusa any chance I can get this PR and https://github.com/fabric8io/kubernetes-client/pull/2678 merged and included in the next 5.0.0 release? At that point I'd be happy to reset our internal fork to be based on 5.0.0 and help test there. We make heavy use of CustomResources, so interested to see how the 5.0.0 changes affect that.

manusa:
> Do you have any tips on getting kubernetes-client to build smoothly in intellij?

I hear you. It's really painful especially when changing to older branches where module structure is different from the current one.

The following setup works more or less well most of the times.
image

However still need to run manual compilation from time to time or rebuild the project. Again, lots of problems when switching to branches with deprecated modules and vice-versa. Need to manually tune those from time to time.

So any other tip to improve the experience is welcome ;)

bbeaudreault:
Thanks, I'll give that a shot. Any other comments on this PR? It's ready to be merged I think. The build is failing for unrelated reasons, but I can't kick it off myself. It succeeds locally.

metacosm:
@bbeaudreault there's a conflict

bbeaudreault:
Whoops, sorry! I'll take care of that shortly

bbeaudreault:
Is it possible to get access to retry these PR checks? They seem flaky and I hate to keep bugging you guys. 

The K8S v1.18.6 failed this time, but it has been succeeding consistently before. It failed in a test unrelated to my change.
The Java 11 and 8 builds are in some sort of hung state -- i can't access their logs. I was tailing the java 11 logs before and everything looked healthy. Last I saw, it had made it into the karaf tests. Not sure what is causing it to hang.

manusa:
> Is it possible to get access to retry these PR checks?

I'm not sure you can through GH interface

Today it's specially bad, Maven Central is also not very reliable today :(

bbeaudreault:
Sounds good. I see someone kicked off a new build, and the E2E all passed 🎉 . Unfortunately the build java tasks are still stuck from the previous run -- they didn't restart.

manusa:
All of the build tasks are getting stuck. I think it might have to do with the tests and the new retry behavior.
(we need to add proper timeout configs to our CI workflows BTW)

rohanKanojia:
While testing this change, I found out that `deletingExisting` API is actually broken since v4.13.0. If we try creating an object which doesn't exist in server like this:
```java
Pod pod = client.pods().load(CreateOrReplaceResource.class.getResourceAsStream("/test-pod.yml")).get();
client.resource(pod).inNamespace("default").deletingExisting().createOrReplace();
```
This would fail with this exception:
```
Exception in thread "main" io.fabric8.kubernetes.client.KubernetesClientException: Failed to delete existing item:example-pod
	at io.fabric8.kubernetes.client.utils.DeleteAndCreateHelper.deleteAndCreate(DeleteAndCreateHelper.java:51)
	at io.fabric8.kubernetes.client.utils.DeleteAndCreateHelper.deleteAndCreateItem(DeleteAndCreateHelper.java:75)
```

The same code works without any problems on v4.12.0

I think we should add a check whether resource exists before deleting it here:
https://github.com/fabric8io/kubernetes-client/blob/7918764df29d34d99547d999c8d03c37f615ca38/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/DeleteAndCreateHelper.java#L38-L42

It's not really related to this PR since I had done this refactor. Maybe we can tackle this as a part of this PR or in separate issue.

bbeaudreault:
@rohanKanojia I'm happy to add this here. Would you be ok with my just catching and handling the KubernetesClientException, rather than adding a new API call?

rohanKanojia:
I checked and looks like Kubernetes Client only returns `false` on deletion when an item doesn't exist.
https://github.com/fabric8io/kubernetes-client/blob/7918764df29d34d99547d999c8d03c37f615ca38/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java#L699-L703
I think we should not throw the exception when `deleteTask.apply(item)` returns `false`. We should simply proceed to creation.

bbeaudreault:
Col, I'm making that change now

bbeaudreault:
FYI that change was of course easy enough to make, but I realized that my local maven actually _wasn't_ running all of the tests. It was failing on karaf, which I thought was the last one, but actually was missing kubernetes-tests. I excluded kubernetes-tests and noticed that the build is actually hanging locally as well! It appears to be hanging in one of my deleteExisting tests. I'm trying to debug it but it's slow going because I can't run it in intellij, so can't easily debug. Still working through it. There appears to be a 400 client error being thrown in the test which causes one of the `getLastRequest()` calls to hang forever.

bbeaudreault:
I've pushed the change we discussed, along with a little sanity checking. I also pushed the fix for the hung tests. Sorry for the runaround on that one... the lack of a local test env really got me. I was able to hack around it a bit, hence how i discovered the actual issue. Note that I can't use your suggested "Delegate IDE to maven" setting because there are no 5.0.0-SNAPSHOTs uploaded to maven central. I can't do a mvn install first because of my company's required local maven settings.

bbeaudreault:
Looks like everything passed!

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/DeleteAndCreateHelper.java
```diff
@@ -23,32 +23,75 @@
...
-      throw new KubernetesClientException("Failed to delete existing item:" + item.getMetadata().getName());
+    if (!deleted) {
+      LOG.debug("did not delete because item did not exist, continuing to create {}", item.getMetadata().getName());
+    }
+
+    try {
+      return createTask.apply(item);
+    } catch (KubernetesClientException e) {
+      // depending on the grace period, the object might not actually be deleted by the time we try to create
+      // if that's the case, give it some time.
```
metacosm, line 58: Is this comment still accurate?

bbeaudreault: Yes, still accurate. All of the stuff I mentioned about DELETE returning a 200 or 202 still applies. So we still need to wait for the deletion to occur before creating, otherwise we'll get a conflict.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/DeleteAndCreateHelper.java
```diff
@@ -23,32 +23,75 @@
...
+      waitUntilDeletedOrInterrupted(client, config, h, namespaceToUse)
     );
 
     return deleteAndCreateHelper.deleteAndCreate(meta);
   }
+
+  private static <T extends HasMetadata> Function<T, Boolean> waitUntilDeletedOrInterrupted(OkHttpClient client, Config config, ResourceHandler<HasMetadata, HasMetadataVisitiableBuilder> h, String namespaceToUse) {
+    return m -> {
+      try {
+        return h.waitUntilCondition(client, config, namespaceToUse, m, Objects::isNull, 30 , TimeUnit.SECONDS) == null;
```
manusa, line 89: It would make sense to at least declare 30 as a static constant.

manusa (approved):
LGTM (besides the constant comment). Thx!

bbeaudreault:
@manusa all set, thanks!

rohanKanojia (approved):

bbeaudreault:
Anything else you guys want to see here? Everything but the code coverage check is passing. I did add tests here, which seem sufficient? Let me know

manusa:
Thanks a lot @bbeaudreault 
I'm going to start the release process of 5.0.0-beta1 as a first step to the 5.0.0 final release. Hopefully your team can provide some feedback on the new API and changes :pray: :wink:

Issue #266, opened 2015-12-16
Cannot handle non-base64 encoded cert/key data


JoshuaAndrew:
io.fabric8.kubernetes.client.internal.CertUtils.getInputStreamFromDataOrFile(String data, String file)
and  io.fabric8.kubernetes.client.internal.CertUtils.decodePem(InputStream keyInputStream)
getInputStreamFromDataOrFile  do not need  "-----BEGIN"  but  decodePem need it

anyone come across this issue!

jimmidyson:
Haven't seen this. Can you give a bit more detail on how to reproduce?

davsclaus:
do you mind changing the title to be english

JoshuaAndrew:
ConfigBuilder builder = new ConfigBuilder();
	builder.withMasterUrl(server);
	builder.withApiVersion(ReplicationController.ApiVersion.V_1.toString());
	builder.withConnectionTimeout(3600000);
	builder.withWatchReconnectInterval(2);
	builder.withWatchReconnectLimit(10);
	builder.withRequestTimeout(36000);
	builder.withUsername(username);
		
	//builder.withCaCertData(caCertData);
        //builder.withClientCertData(clientCertData);
	//builder.withClientKeyData(clientKeyData);
		
	builder.withCaCertFile("E:\\isoftstone\\workspacedeis\\svc-container\\svc-container-provider\\src\\main\\resources\\ca.crt");
	builder.withClientCertFile("E:\\isoftstone\\workspacedeis\\svc-container\\svc-container-provider\\src\\main\\resources\\client-develop.crt");
	builder.withClientKeyFile("E:\\isoftstone\\workspacedeis\\svc-container\\svc-container-provider\\src\\main\\resources\\client-develop.key");
		
	Config kubeConfig = builder.build();
	KubernetesClient kubeClient = new DefaultKubernetesClient(kubeConfig);

withCaCertFile,withClientCertFile,withClientKeyFile  is ok
but withCaCertData withClientCertData  withClientKeyData   is error

JoshuaAndrew:
Today I fixed this issue
io.fabric8.kubernetes.client.internal.CertUtils.getInputStreamFromDataOrFile(String data, String file)
change this ByteArrayInputStream(ByteString.decodeBase64(data).toByteArray())     to ByteArrayInputStream(data.getBytes())

jimmidyson:
The use case was using kube config files which include the cert data base 64 encoded.

[...]

PR #2685, opened 2020-12-21
Add an Integration test for Typed CustomResource support

rohanKanojia:
Added an Integration test for new improved typed `customResources()` API whichtests standard operations like create, createOrReplace, get, list, watch, update, updateStatus, delete

We had an integration test for custom resources but that was using the raw API. This test should cover most of the use cases with respect to typed api

## Type of change

## Checklist


oscerd (approved):

kubernetes-itests/src/test/java/io/fabric8/crd/PetSpec.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+package io.fabric8.crd;
+
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import io.fabric8.kubernetes.api.model.KubernetesResource;
+
+@JsonDeserialize(
+  using = JsonDeserializer.None.class
+)
```
manusa, line 24: Not necessary

kubernetes-itests/src/test/java/io/fabric8/crd/PetSpec.java
```diff
@@ -0,0 +1,42 @@
...
+public class PetSpec implements KubernetesResource {
```
manusa, line 25: Don't implement KubernetesResource (unless there is a strong reason to)

kubernetes-itests/src/test/java/io/fabric8/crd/PetStatus.java
```diff
@@ -0,0 +1,41 @@
...
+package io.fabric8.crd;
+
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import io.fabric8.kubernetes.api.model.KubernetesResource;
+
+@JsonDeserialize(
+  using = JsonDeserializer.None.class
+)
+public class PetStatus implements KubernetesResource {
```
manusa, line 25: Same as with spec, remove deserializer config and don't implement KubernetesResource

kubernetes-itests/src/test/java/io/fabric8/kubernetes/TypedCustomResourceIT.java
```diff
@@ -0,0 +1,239 @@
...
+  private String currentNamespace;
+
+  private MixedOperation<Pet, KubernetesResourceList<Pet>, Resource<Pet>> petClient;
+
+  @ClassRule
+  public static final AssumingK8sVersionAtLeast assumingK8sVersion = new AssumingK8sVersionAtLeast("1", "16");
+
+  @BeforeClass
+  public static void init() {
+    ClusterEntity.apply(TypedCustomResourceIT.class.getResourceAsStream("/pet-crd.yml"));
```
manusa, line 69: Why don't we declare the CRD using Java instead?

I really think we should push our users to think in Java instead of YAML.
It works against us to mix unnecessary YAMLs, this sends the wrong message. I've seen this in many of the open issues when users post examples of infrastructure defined in YAML resources and then pushed to k8s using our client. "Why are they using YAML in the first place?".
We should really be working our way with our own types and try to send this message: https://www.youtube.com/watch?v=QFn_H_LmMfQ

This point should be true for test suites too unless we're either testing serialization/deserialization, or we're actually testing the given k8s resource and we want to prepare it using YAML first (but then that should use other means of deployment aka kubectl).

For this case we're testing CRs, so defining the CRD using our Java API should be perfectly fine and IMO preferred.

kubernetes-itests/src/test/java/io/fabric8/kubernetes/TypedCustomResourceIT.java
```diff
@@ -0,0 +1,239 @@
...
+      assertEquals(currentStatus, pet.getStatus().getCurrentStatus());
+    }
+  }
+
+  @AfterClass
+  public static void cleanup() {
+    ClusterEntity.remove(TypedCustomResourceIT.class.getResourceAsStream("/pet-crd.yml"));
+  }
+
+  private Pet getNewPet(String name, String type, String currentStatus) {
```
manusa, line 224: nit: You are not getting a pet from a store, you are creating a new Pet from scratch `createNewPet`.

kubernetes-itests/src/test/java/io/fabric8/crd/PetSpec.java
rohanKanojia: ohk, We should update our tests[0] and examples[1] too to remove this

[0] https://github.com/fabric8io/kubernetes-client/tree/master/kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/crd
[1] https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/crds/DummySpec.java#L24

manusa: I'm dealing with the examples. Most of them are completely outdated.

kubernetes-itests/src/test/java/io/fabric8/kubernetes/TypedCustomResourceIT.java
```diff
@@ -59,14 +62,53 @@
...
+    .endVersion()
+    .withScope("Namespaced")
+    .withNewNames()
+    .withPlural("pets")
+    .withSingular("pet")
+    .withKind("Pet")
+    .withShortNames("pt")
+    .endNames()
+    .endSpec()
+    .build();
```
manusa, line 102: Most of the boilerplate can be skipped, see https://github.com/fabric8io/kubernetes-client/blob/22e2e6fe942b9a740812ba2d45968a31e0e045da/kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/CustomResourceV1Example.java as reference

rohanKanojia: Umm, strange. Why is this method part of `CustomResourceDefinitionContext`? It doesn't seem to be related to `CustomResourceDefinitionContext`

manusa: It's not, this was discussed in scope #266.
IIRC we all agreed these methods had to live somewhere for now.

manusa (approved):

kubernetes-itests/src/test/java/io/fabric8/kubernetes/TypedCustomResourceIT.java
rohanKanojia: Maybe you linked the wrong issue... When I looked at your example, it seemed a bit weird. I thought maybe `CustomResource` is the right place for that method

manusa: Wrong issue indeed. I don't see `CustomResource` a good fit either for that, I think it might even be worse since it will add even more confusion to the CustomResource vs. CustomResourceDefinition situation.
The helper is used to create a CustomResourceDefinition from a CustomResource pojo. It might make sense to make it part of CustomResourceDefinition class, but of course that's a generated model, and there are (as of now) 2 of them.
IMHO it makes sense to have these in an utils class. I'm positive `CustomResourceDefinitionContext` isn't a good place either.

Issue #2596, opened 2020-11-10
Extension models should have buildable references to core kubernetes-model types.


iocanel:
Methods like: 
- withNewMetadata
- withNewContainer
- addNewContainer

and so on are no longer generated, due to the absence of `BuildableReference` to types like: 
- ObjectMeta
- Container
- etc

While this is relatively easy to work around, having those method is one of the little things that add to the overall experience.

The root cause of this issue seems to be the removal of the specific annotators.

manusa:
Caused by: #2157 

https://github.com/fabric8io/kubernetes-client/blob/bec9ac92d02234b4b6bbcd23f9072285854793cd/kubernetes-model-generator/kubernetes-model-annotator/src/main/java/io/fabric8/kubernetes/annotator/KubernetesTypeAnnotator.java

vs.

https://github.com/fabric8io/kubernetes-client/blob/bec9ac92d02234b4b6bbcd23f9072285854793cd/model-annotator/src/main/java/io/fabric8/kubernetes/ModelAnnotator.java

Probably merging the ModelAnnotator with the others (or making it extend form KubernetesCoreTypeAnnotator), or at least move it to the same package would make sense too.

PR #2734, opened 2021-01-20
fix(#2596) Add buildable ref for container, port and volume

iocanel:
This is a backport from master. Since, we've already broken the model jar into pieces, it does make sense to add it on 4.x too.


oscerd (approved):

manusa (approved):
LGTM, thx

rohanKanojia (approved):

PR #2739, opened 2021-01-20
Automatically instantiate spec and status fields

### LINKED ISSUE #2736: Add more crd annotations to cover subresources and additional printer columns.
We currently have a nice set of annotations for defining core crd configuration.

External tooling could benefit form having access to additional annotations for specifying subresources and additional printer columns.
Even though these annotations, won't be used directly from the client (yet). It does make sense to propose them for downstream projects to use.

Dekorate does provide those and we could directly port them here: https://github.com/dekorateio/dekorate/tree/master/annotations/crd-annotations/src/main/java/io/dekorate/crd/annotation

The annotations of interest are:

- Status
- PrinterColumn
- SpecReplicas
- StatusReplicas
- LabelSelector

### LINKED ISSUE #2737: @Version annotation should include a flags for specifying if version is served or storage.
This can be useful for downstream projects that want to use the annotation for crd generation.

### LINKED ISSUE #2744: Automatically instantiate Spec and Status fields
With 5.0.0, `CustomResource` are now parameterized with status and spec types allowing very compact, boilerplate-free definition of custom resource implementations. However, the status and spec fields are not currently automatically created which leads to awkward code to avoid NPEs.
The client should therefore attempt to instantiate the associated field where possible, provided the availability of a no-arg constructor for these types.

metacosm:
## Description
Makes typical use cases automatic, provide `initSpec` and  `initStatus` method on `CustomResource` to let user provide their own custom implementation if the default doesn't work for their use case. Fixes #2744.
 
## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/CustomResource.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/CustomResource.java
@@ -20,6 +20,8 @@ import java.lang.reflect.Modifier;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.Locale;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
@@ -254,7 +256,7 @@ public abstract class CustomResource<S, T> implements HasMetadata {
   
   private final static String TYPE_NAME = CustomResource.class.getTypeName();
   private final static String VOID_TYPE_NAME = Void.class.getTypeName();
-  private Instantiator[] instantiators = new Instantiator[2];
+  private final static Map<String, Instantiator> instantiators = new ConcurrentHashMap<>();
   
   /**
    * Encapsulates an instantiation means. Needed to provide no-op when needed.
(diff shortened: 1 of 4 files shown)


manusa:
I'm failing to understand how or when are either of the methods `initSpec` or `initStatus` going to be invoked more than once per instance.

In other words, what advantages does keeping/caching the instantiator instances have?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/CustomResource.java
```diff
@@ -229,4 +253,97 @@ public T getStatus() {
...
+              "Cannot instantiate interface/abstract type " + className);
+          }
+          
+          // record the instantiator associated with the identified type
+          instantiator = () -> {
+            final Constructor<?> constructor;
+            try {
+              // get the no-arg (declared needed if implicit) constructor
+              constructor = clazz.getDeclaredConstructor();
+              constructor.setAccessible(true); // and make it accessible
```
manusa, line 321: _Thought:_
This line worries me.
Can't think  of a case where the no-arg constructor will not be accessible. 
And in case it's not, we should probably desist in providing the empty instance since the user might have added some functionality to the alternate constructor. Invoking the hidden no-arg constructor will probably defeat the user's expectations and security assumptions.

Anyway, still not sure if this is a big deal. Everything else LGTM.

manusa (approved):

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/CustomResource.java
metacosm: I had to add this to make the most common case (a class that doesn't provide any constructor at all, i.e. one that relies on the implicit one) work so it's pretty much required. That surprised me as well.

manusa: :flushed: that's really strange, not sure how Jackson deals with this then.

rohanKanojia (approved):

Issue #2732, opened 2021-01-19
Question on fips compliance


SubhraB:
We are using the k8s client version 4.13.0. 

I was looking at this class. `io.fabric8.kubernetes.client.internal.CertUtils`

Specifically the code below, which I see multiple instances of.
` KeyStore trustStore = KeyStore.getInstance("JKS");`

The above line of code becomes invalid when we run our application on fips mode, since the truststore format becomes BCFKS. We expect something like this.
`KeyStore trustStore = KeyStore.getInstance(KeyStore.getDefaultType());`

Is there a way to override that to make us fips compliance. FYI, we are not allowed to make insecure https calls, we have to use a truststore in BCFKS format

DimoDonchev:
I am having the same issuе here. Really hope this gets checked out.

gulyaev13:
@SubhraB , @DimoDonchev , I've created PR with a fix. Hope that it will be merged and solve our problem.
https://github.com/fabric8io/kubernetes-client/pull/2788

SubhraB:
I got around it by not using the default k8s client. Buts its a little pain.

manusa:
#2788 was merged and will be released as part of [5.4.0](https://github.com/fabric8io/kubernetes-client/milestone/23)

PR #2788, opened 2021-02-05
Support FIPS mode in kubernetes-client with BouncyCastleFipsProvider

gulyaev13:
Support FIPS mode in kubernetes-client with BouncyCastleFipsProvider сonnected externally.
Solving the issue described in https://github.com/fabric8io/kubernetes-client/issues/2732

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/CertUtils.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/CertUtils.java
@@ -90,7 +90,7 @@ public class CertUtils {
   }
 
   public static KeyStore createTrustStore(InputStream pemInputStream, String trustStoreFile, char[] trustStorePassphrase) throws IOException, CertificateException, KeyStoreException, NoSuchAlgorithmException {
-    KeyStore trustStore = KeyStore.getInstance("JKS");
+    KeyStore trustStore = KeyStore.getInstance(KeyStore.getDefaultType());
 
     if (Utils.isNotNullOrEmpty(trustStoreFile)) {
       trustStore.load(new FileInputStream(trustStoreFile), trustStorePassphrase);
@@ -112,7 +112,7 @@ public class CertUtils {
       Collection<? extends Certificate> certificates = certFactory.generateCertificates(certInputStream);
       PrivateKey privateKey = loadKey(keyInputStream, clientKeyAlgo);
 
-      KeyStore keyStore = KeyStore.getInstance("JKS");
+      KeyStore keyStore = KeyStore.getInstance(KeyStore.getDefaultType());
       if (Utils.isNotNullOrEmpty(keyStoreFile)){
         keyStore.load(new FileInputStream(keyStoreFile), keyStorePassphrase);
       } else {
@@ -145,7 +145,7 @@ public class CertUtils {
         @Override
         public PrivateKey call() {
           try {
-            if (Security.getProvider("BC") == null) {
+            if (Security.getProvider("BC") == null && Security.getProvider("BCFIPS") == null) {
               Security.addProvider(new org.bouncycastle.jce.provider.BouncyCastleProvider());
             }
             PEMKeyPair keys = (PEMKeyPair) new PEMParser(new InputStreamReader(keyInputStream)).readObject();


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/CertUtils.java
```diff
@@ -145,7 +145,7 @@ private static PrivateKey handleECKey(InputStream keyInputStream) throws IOExcep
         @Override
         public PrivateKey call() {
           try {
-            if (Security.getProvider("BC") == null) {
+            if (Security.getProvider("BC") == null && Security.getProvider("BCFIPS") == null) {
               Security.addProvider(new org.bouncycastle.jce.provider.BouncyCastleProvider());
```
metacosm, line 149: Shouldn't you be adding the `BouncyCastleFipsProvider` here if it's not found, instead of the regular BC provider?

gulyaev13: According to [Bouncy Castle FIPS Java API User Guide](https://downloads.bouncycastle.org/fips-java/BC-FJA-UserGuide-1.0.2.pdf) security provider can be added in **java.security** file.

I chose to support only the external configuration due to the following reasons:
- Additional flag for FIPS mode isn't required
- No need to add dependency with BC FIPS and update versions
- Adding FIPS provider at the state of initialization of Kubernetes client is too late. So much code can be executed in an application using kubernetes-client. Initialization on startup of JVM is preferable.

manusa: As I understand, this statement prevents the the BouncyCastleProvider to be added if you configured your JVM to use the BCFIPS provider, right?

manusa: On the other hand, these changes should be harmless in case no BC configuration is provided, right?

oscerd (approved):

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/CertUtils.java
metacosm: Hmm, still not convinced… Doesn't this mean that someone might be requesting the FIPS provider and actually getting the regular one?

gulyaev13: @manusa , yes! Changes should be harmless in case no BC configuration is provided.
There are 4 cases:
1. None of the BouncyCastle security providers configured in JVM -> BC will be added in runtime
2. BC configured in JVM -> BC will not be added in runtime
3. BCFIPS configured in JVM -> BC will not be added in runtime
4. Another FIPS provider configured in JVM (is it exist?) -> BC will be added in runtime

@metacosm , looks like you are asking about case №4, aren't you?
I see 2 solutions: 
1. The additional flag can be added to prevent adding BC provider in case if someone would like to use an external provider.
2. We can say that kubernetes-client works only with the family of BouncyCastle security providers (BC/BCFIPS).

metacosm: What I don't understand is why we add the BC provider when the user asks for `BCFIPS` and not the `BouncyCastleFipsProvider`? How will that address the issue?

gulyaev13: @metacosm , because there is no option in kubernetes-client configuration for add `BouncyCastleFipsProvider(BCFIPS)` instead of `BouncyCastleProvider(BC)`.

But if the user will configure JVM for use `BouncyCastleFipsProvider(BCFIPS)` kubernetes-client will not add `BouncyCastleProvider(BC)`. See case №3 in my previous message.

gulyaev13:
@oscerd , hello!
If my PR will be approved what should I do to backport changes to kubernetes-client 4.11.*?
I've planned to add these changes in Jenkins plugin with Kubernetes API and looks like they are still using 4.11.* releases. Updating the Jenkins plugin to the 5.1.0 release seems a big project for this small patch.
https://github.com/jenkinsci/kubernetes-client-api-plugin/blob/master/pom.xml#L16

oscerd:
I don't think 4.11 is an active branch, only for critical CVE.

gulyaev13:
@oscerd , understood, thanks!
I will try to propose an update of kubernetes-client version in Jenkins plugins to 5 and more.

gulyaev13:
@manusa , hello!
Are there some blockers for merge?

manusa:
Hi @gulyaev13,
Everything is fine with the PR. We're waiting on a security team to double check everything is OK (unfortunately it's taking longer than expected :sweat:, sorry).

DimoDonchev:
@manusa , hello!
Is there any update on the double checking from the security team?

manusa:
Hi @DimoDonchev, sorry for the delay. We'll try to include this in our upcoming release.

oleg-nenashev:
Hi @manusa . Is there a ETA for the 2.4.0 release? We use this library in Jenkins Kubernetes Plugin and in a few developer tools. It would be great to hqve a release. Please let me know if any help needed

manusa:
> Hi @manusa . Is there a ETA for the 2.4.0 release? We use this library in Jenkins Kubernetes Plugin and in a few developer tools. It would be great to hqve a release. Please let me know if any help needed

I guess you mean 5.4.0. Release is imminent, we are long overdue our initial ETA. Not sure if we'll be able to deliver by the end of the week, but our intention is to release ASAP.

oleg-nenashev:
Yeah, thanks for the update and sorry for the typo. Thanks for working on
this and for maintaining the library

FuyaoLi2017:
@gulyaev13 @manusa Hello Team, I am using Apache Flink 1.13.0 and Flink references the 4.9.2 version of the Fabric8 kubernetes client. I am writing an kubernetes operator for Flink and 5.4.0 is not compatible with Flink since the APIs changes a lot between these two versions because I need to call some Flink internal methods which is actually using 4.9.2 Fabric library.

However, I still want to make Fabric8 version 4.9.2 FIPS compliant, is it possible to do to monkey patch for such changes?

manusa:
I don't know how heavily Apache Flink relies on the client, but it would make sense to update their dependency.
4.9.2 was released almost a year ago, so it might be worthier to spend the time upgrading Flink to use the latest release of the Fabric8 client instead (as I said, depending on how complex this turn out to be).

PR #2882, opened 2021-03-08
Fix #2746: SharedInformerFactory should use a key based on OperationContext rather than Type in Map storing registered informers

### LINKED ISSUE #2746: Informer creation behavior for custom resource with CRDContext in SharedInformerFactory
The factory has a  method [sharedIndexInformerForCustomResource](https://github.com/fabric8io/kubernetes-client/blob/2b9158528257a2de030ca67c776c5caf37e61d40/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/SharedInformerFactory.java#L132) to create an informer with a CRDContext for CRs. It's is using [sharedInformerFor](https://github.com/fabric8io/kubernetes-client/blob/2b9158528257a2de030ca67c776c5caf37e61d40/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/SharedInformerFactory.java#L152) to register the informer inside the factory object. This causes a problem when registering new informers using several different CRDContexts, but using same set of classes (extending CustomResource and CustomResourceList). Case will result in creation of several informers, but registering only the one which is has been created the latest (since factory's internal map [uses only classname as key](https://github.com/fabric8io/kubernetes-client/blob/2b9158528257a2de030ca67c776c5caf37e61d40/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/SharedInformerFactory.java#L158)).
Example:

```java
        SharedInformerFactory factory = defaultKubernetesClient.informers();

        CustomResourceDefinitionContext firstCrdContext = getFirstCrdContext();
        CustomResourceDefinitionContext secondCrdContext = getSecondCrdContext();

        SharedIndexInformer<MyCustomResourceClass> lostIndexInformer = factory.sharedIndexInformerForCustomResource(
                firstCrdContext,
                MyCustomResourceClass.class,
                MyCustomResourceClassList.class,
                0);

        SharedIndexInformer<MyCustomResourceClass> savedIndexInformer = factory.sharedIndexInformerForCustomResource(
                secondCrdContext,
                MyCustomResourceClass.class,
                MyCustomResourceClassList.class,
                0);

        factory.startAllRegisteredInformers();
```

As I have said `lostIndexInformer` will be rewritten by the time we reach to `startAllRegisteredInformers`. I don't think this is the right behavior, since idea of passing the different CRDContexts to factory methods means that several informers should be registered and that they should operate on different CRs which has been described in CRDContext.

rohanKanojia:
Fix #2746 

    SharedInformerFactory has `informers` and `startedInformers` Map fields
    which use `Type` as key. This limits us to use only one
    SharedIndexInformer corresponding to a particular type. We should make a
    key based on OperationContext's elements like apigroup, apiversion,
    namespace, name.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@
 * Fix #2857: Fix the log of an unexpected error from an Informer's EventHandler
 * Fix #2853: Cannot change the type of the Service from ClusterIP to ExternalName with PATCH
 * Fix #2783: OpenIDConnectionUtils#persistKubeConfigWithUpdatedToken persists access token instead of refresh token
-* Fix #2871: Change longFileMode to LONGFILE_POSIX for creating tar in PodUpload, improve exception handling in PodUpload.
+* Fix #2871: Change longFileMode to LONGFILE\_POSIX for creating tar in PodUpload, improve exception handling in PodUpload.
+* Fix #2746: SharedInformerFactory should use key formed from OperationContext 
 
 #### Improvements
 * Fix #2781: RawCustomResourceOperationsImpl#delete now returns a boolean value for deletion status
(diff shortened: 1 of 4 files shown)


oscerd (approved):

jorsol (commented):
I'm not sure if having `OperationContext` as key in HashMap would be safe or correct, the class doesn't implement hashCode() and equals(), and even if they are implemented (I could add them), fields like OkHttpClient are probably not good candidates for hashCode() and equals().

manusa:
As discussed during today's stand-up call, the key should be the `self` path for the resource in the Kubernetes REST API.

This will provide consistency with how the resource is identified in k8s, and will also fix the issue at hand.

In case this is not enough, or causes problems, we could always consider using `metadata.uid` as the key.

manusa (approved):
LGTM, thx!

PR #2899, opened 2021-03-12
Change badges for the list of client modules

jorsol:
This changes the badges for maven central from shields.io to maven-badges.herokuapp.com, the issue with shields.io is that they aren't updated correctly (or the cache fails to refresh) when there is a new release showing old versions. Also, they are now in a table instead of a list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,27 +4,25 @@ This client provides access to the full [Kubernetes](http://kubernetes.io/) &
 
 [![Build](https://github.com/fabric8io/kubernetes-client/workflows/Build/badge.svg)](https://github.com/fabric8io/kubernetes-client/actions?query=workflow%3ABuild)
 [![Sonar Scanner](https://github.com/fabric8io/kubernetes-client/workflows/Sonar%20Scanner/badge.svg)](https://github.com/fabric8io/kubernetes-client/actions?query=workflow%3A%22Sonar+Scanner%22)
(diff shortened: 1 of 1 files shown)


oscerd (approved):

manusa:
Thx, this was something very much needed.

The list of modules probably needs updating too (I for one see that volumesnapshot extension is missing)

jorsol:
> Thx, this was something very much needed.
> 
> The list of modules probably needs updating too (I for one see that volumesnapshot extension is missing)

Is the `kubernetes-model-core` needed here?

manusa:
> Is the `kubernetes-model-core` needed here?

:sweat_smile: No idea, as a consumer user you probably have a better idea of which modules make sense to expose here. I'm aware some users are just interested in model types, others in extensions, etc. Personally I think it doesn't belong to the set. @rohanKanojia ?

jorsol:
> > Is the `kubernetes-model-core` needed here?
> 
> sweat_smile No idea, as a consumer user you probably have a better idea of which modules make sense to expose here. I'm aware some users are just interested in model types, others in extensions, etc. Personally I think it doesn't belong to the set. @rohanKanojia ?

I only use `kubernetes-client` 😅 , so I also think the model doesn't belong to the set.

I added `volumesnapshot-client` and removed `kubernetes-model-core`.

rohanKanojia:
We have kubernetes-model related badged in kubernetes-model-generator README

manusa (approved):
LGTM, thx!

PR #2951, opened 2021-03-24
using the custom resource class if possible

shawkins:
## Description
The CustomResource constructor has side effects such as setting the version and kind from annotations.  There should be similar CustomResourceList logic.  The closest non-breaking change is to have an overridable method to provide the custom resource class.  If you are open to something like this, I can round out this pr with tests, javadocs, etc.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/CustomResourceList.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/CustomResourceList.java
@@ -23,6 +23,7 @@ import io.fabric8.kubernetes.api.model.KubernetesResourceList;
 import io.fabric8.kubernetes.api.model.ListMeta;
 import lombok.ToString;
 
+import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -43,6 +44,19 @@ public class CustomResourceList<T extends HasMetadata> implements KubernetesReso
 
   @JsonProperty("metadata")
   private ListMeta metadata;
+
+  public CustomResourceList() {
+    try {
+      Class<T> customResourceClass = getCustomResourceClass();
+      if (customResourceClass != null) {
+        HasMetadata instance = customResourceClass.getDeclaredConstructor().newInstance();
+        this.apiVersion = instance.getApiVersion();
+        this.kind = instance.getKind() + "List";
+      }
+    } catch (InstantiationException | IllegalAccessException | IllegalArgumentException | InvocationTargetException
+            | NoSuchMethodException | SecurityException e) {
+    }
+  }
 
   public String getApiVersion() {
     return apiVersion;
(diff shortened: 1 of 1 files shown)


oscerd (approved):

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/CustomResourceList.java
```diff
@@ -44,6 +45,19 @@
   @JsonProperty("metadata")
   private ListMeta metadata;
 
+  public CustomResourceList() {
+    try {
+      Class<T> customResourceClass = getCustomResourceClass();
+      if (customResourceClass != null) {
+        HasMetadata instance = customResourceClass.getDeclaredConstructor().newInstance();
```
metacosm, line 52: You don't need to instantiate the class here. You can just call `HasMetadata.getApiVersion(customResourceClass)` and `HasMetadata.getKind(customResourceClass)` instead.

metacosm (changes requested):

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/CustomResourceList.java
shawkins: @metacosm I was just trying to be consistent with how it's down in CustomResource - there it is calling the instance method, which defaults to the static method.  Is there a chance that someone can override getKind?

metacosm: It's indeed possible… not likely, but possible. Good point.

metacosm: The thing is I'd like to avoid instantiating the CR if possible. Did you run into specific issues with the current implementation? If yes, how did you address them? Would using the first item work? Or do you need that information before the list is "filled"?

shawkins: > Did you run into specific issues with the current implementation? 

Not really, the empty api and version was noticed on a list returned by some of mock logic - it's not something that's needed by our implementation code.

> Would using the first item work?

My assumption was there are instances where the list may be empty.

metacosm: Empty list is indeed the issue. Then again, if there isn't an actual issue, I'm a little bit reluctant to add reflection-based instantiation because I worry of side-effects (in particular, for native compilation).

shawkins: I was hoping it might just be a quick thing to make CustomResourceList more consistent with CustomResource.  However it does require the developer to know to provide this override to supply the class and it's not something that we actually require at this point - so if there's reasons to not pursue this, I'm fine with that.

metacosm:
I think we have several things we need to clean up around resources management. If you're not hitting a specific issue that this PR would fix, I'd rather defer until we can find a "better/cleaner" solution. 

I will close this PR for now, if that's OK with you, @shawkins. Feel free to open an issue if you ever run into problems with this.

shawkins:
> I will close this PR for now, if that's OK with you, @shawkins

I'm fine with that.

PR #2973, opened 2021-04-08
Save and re-use random

cutiechi:
## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/Utils.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/Utils.java
@@ -55,6 +55,7 @@ public class Utils {
   public static final String OS_NAME = "os.name";
   public static final String PATH_WINDOWS = "Path";
   public static final String PATH_UNIX = "PATH";
+  private static final Random random = new Random();
 
   private Utils() {
   }
@@ -237,7 +238,6 @@ public class Utils {
   }
 
   public static String randomString(int length) {
-    Random random = new Random();
     StringBuilder sb = new StringBuilder();
     for (int i = 0; i < length; i++) {
       int index = random.nextInt(ALL_CHARS.length());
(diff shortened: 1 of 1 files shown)


oscerd (changes requested):
I'd prefer SecureRandom if i have to store the instanxe

cutiechi:
@oscerd Ok, let me modify it

cutiechi:
@oscerd Please help me review again

oscerd (approved):

metacosm:
What's the motivation for this change if I may ask?

oscerd:
I guess it is just for decreasing the objects allocation. Nothing critical or important.

cutiechi:
@metacosm Creating a new Random object each time a random value is needed is inefficient and may produce numbers which are not random depending on the JDK. For better efficiency and randomness, create a single Random, then store, and reuse it.

The Random() constructor tries to set the seed with a distinct value every time. However there is no guarantee that the seed will be random or even uniformly distributed. Some JDK will use the current time as seed, which makes the generated numbers not random at all.

metacosm:
I don't feel like we need a cryptographically strong random generator which might have an impact on performance but also on native compilation (though, I'm not quite sure about that last point) so I think that staying with `Random` would be good enough, and possibly better, than using `SecureRandom`. If we worry about distribution due to seeding, we could provide our own seed as well when we initialize the object.

Regarding minimizing instantiations, I don't think that the impact is all that great either (i.e. I'm pretty sure there are lots of other things we could be optimizing first) 😄

oscerd:
I think this was much more as starting point for contributing.

oscerd:
Btw, it's up to the team, maybe get back to Random or close this PR directly.

cutiechi:
@metacosm @oscerd Please help me review again, thx

rohanKanojia (approved):

oscerd:
About SecureRandom vs Random, sometimes it is marked as bad practice by sonar. Nothing important, but something to take a look at.

metacosm (approved):

oscerd (approved):

cutiechi:
@rohanKanojia @oscerd @metacosm The sonar quality gate failed, what should I do？

metacosm:
@cutiechi nothing as the error is expected

cutiechi:
@metacosm Thank you very much

cutiechi:
@metacosm Thank you very much

PR #2995, opened 2021-04-14
handling the indexer delete with DeletedFinalStateUnknown

shawkins:
## Description
Possible fix for #2994 - updating the indexer state for deletes generated by resync
Fix #2994 

## Type of change
 - [x] Bug fix (non-breaking change which fixes an issue)

## Checklist
 - [x] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [x] I Added [CHANGELOG](../CHANGELOG.md) entry regarding this change
 - [x] I have implemented unit tests to cover my changes

PR #2993, opened 2021-04-14
cleaning up the reflectorwatcher logging

shawkins:
## Description
Fix for #2991 - event received logged at too high of a level.

## Type of change
 - [x] Bug fix (non-breaking change which fixes an issue)

## Checklist
 - [x] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [x] I Added [CHANGELOG](../CHANGELOG.md) entry regarding this change

Issue #2812, opened 2021-02-15
SharedIndexInformer EventHandler sees double updates at resync interval 


shawkins:
Creating an informer with an Eventhandler similar to this:

```
informerFactory = client.informers();
informer = informerFactory.sharedIndexInformerFor(MyCustomResource.class, MyCustomResourceList.class, 30000);
informer.addEventHandler(new ResourceEventHandler() {
   public void onUpdate(ManagedKafkaAgent oldObj, ManagedKafkaAgent newObj) {
     //some action             
   }
   ...
});
informerFactory.startAllRegisteredInformers();

```

Results in the onUpdate method being called twice for every resource on the resync interval.  Is that expected?

manusa:
We should compare what happens in the analogous operation for client-go.

rohanKanojia:
I tried this on client-go and it doesn't look like they also get double updates. So a code like this:
```go
func (c *PodLoggingController) podAdd(obj interface{}) {
        pod := obj.(*v1.Pod)
        klog.Infof("POD CREATED: %s/%s", pod.Namespace, pod.Name)
}

func (c *PodLoggingController) podUpdate(old, new interface{}) {
        oldPod := old.(*v1.Pod)
        newPod := new.(*v1.Pod)
        klog.Infof(
                "POD UPDATED. %s/%s %s",
                oldPod.Namespace, oldPod.Name, newPod.Status.Phase,
        )
}

func (c *PodLoggingController) podDelete(obj interface{}) {
        pod := obj.(*v1.Pod)
        klog.Infof("POD DELETED: %s/%s", pod.Namespace, pod.Name)
}

// NewPodLoggingController creates a PodLoggingController
func NewPodLoggingController(informerFactory informers.SharedInformerFactory) *PodLoggingController {
        podInformer := informerFactory.Core().V1().Pods()

        c := &PodLoggingController{
                informerFactory: informerFactory,
                podInformer:     podInformer,
        }
        podInformer.Informer().AddEventHandler(
                // Your custom resource event handlers.
                cache.ResourceEventHandlerFuncs{
                        // Called on creation
                        AddFunc: c.podAdd,
                        // Called on resource update and every resyncPeriod on existing resources.
                        UpdateFunc: c.podUpdate,
                        // Called on resource deletion.
                        DeleteFunc: c.podDelete,
                },
        )
        return c
}
```
This seems to produce the following output:
```
informer : $  KUBECONFIG=~/.kube/config go run informer.go
[...]

PR #3008, opened 2021-04-17
refactoring informer logic for type safety and to refine the role of stores

shawkins:
## Description
While looking at changes for #2993, #2995, and another bug that has not been logged here.  It appears there are some clarifications that could be made to improve the readability and the safety of the informer code.  The central change is that Store should be more type specific and have a narrower interface - I realize this differs from the go implementation and this does change the public interface of SharedIndexInformer.getIndexer because the Store will no longer have replace, resync, or isPopulated.  If it's a problem to make such a change it can of course be refined (previous code, all no-ops, or throw not implemented exceptions).  In general it seems problematic that SharedIndexInformer.getIndexer returns as broad of an interface as it does - for example in DefaultSharedIndexInformer.addIndexers you can't add an indexer once running, but if you call SharedIndexInformer.getIndexer.addIndexers you can.

If you'd prefer another interface such as ResyncableStore vs direct usage of the DeltaFIFO in ResyncRunnable, Reflector, and ReflectorWatcher - that can be done as well, it just didn't seem needed given the single implementation of DeltaFIFO.  I do see several implementations of Store types in go.

This also moves the responsibility for the lastResourceVersion to the DeltaFIFO - having it updatable outside the DeltaFIFO lock is dangerous.  With the code pre #2812 it appears to allow for timing issues in threads updating the DeltaFIFO/lastResourceVersion:

1. resyncExecutor thread performs scheduled relist - sees state s1, sets the last resourceVersion to s1
2. watch thread via ReflectorWatcher - sees an update to state s2, calls DeltaFIFO.update and sets last resourceVersion to s2
3. resyncExecutor thread continues - calls DeltaFIFO.replace with state s1, adds a syncronization event to state s1 after the update event

The event generated from 2 will be an update showing s1 as the old state and s2 as the new
The event generated from 3 will be an update showing s2 as the old state and s1 as the new 

A reconnect won't fix that as the last resourceVersion is s2.

In theory though on the next relist there will be a synchronization and event to s2 (or whatever the latest state is). However with the pre #2812 code any uncaught exception from relist will inhibit the job from running again.

Granted this should no longer happen after #2812 because the scheduled relist has been removed, but it is still something that should be guarded against.  

@rohanKanojia or others let me know if you are open to these types of changes, I can refine this PR as needed.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/ResyncRunnable.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/ResyncRunnable.java
@@ -15,7 +15,8 @@
  */
 package io.fabric8.kubernetes.client.informers;
 
-import io.fabric8.kubernetes.client.informers.cache.Store;
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.fabric8.kubernetes.client.informers.cache.DeltaFIFO;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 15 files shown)


shawkins:
I'll rebase this after 5.3.1 - it appears that quite a few changes will make it into that release.  And some of this is already being addressed by things like #3031

manusa:
> I'll rebase this after 5.3.1

Hi  @shawkins, 5.3.1 was just released

shawkins:
@manusa thank you.  I think the way forward will be to split this into several prs.  I've started with #3048 - that incorporates some the changes from here.

After that, can we look at #3029

Then I will include the changes that move the last resource version responsibility onto the store.

Finally I'll open a pr that refines the Store interface.

shawkins:
I'll now replace this pr with one that refines the store interface.  This will be for after the deltafifo removal #3061 and ideally after a cleanup of the threading concerns for informers: https://github.com/shawkins/kubernetes-client/pull/2

Issue #2010, opened 2020-02-18
Refactor SharedInformers


manusa:
## Description

Shared Informers module was implemented simply by porting the Go-client implementation.

Although everything works fine, there's still room for improvement.
- Create more testable code (clear separation of concerns, and interface segregation)
- Add unit tests to verify implementation
- Check concurrency implementations and thread safety

Relates to:
- #1961
- #1894 - #1951 
- #1384 - #1728
- #1075 
- #2489

shawkins:
After the Informable change I think this can be resolved.  The only thing I don't like about the remaining impl is how the resync logic functions - but it matches the go client so it's probably not worth tweaking further.

shawkins:
Marking as resolved based upon the last comment.

manusa:
@shawkins Thank you very much for your tremendous effort making this happen.

I'm sure that our Shared Informer implementation and all the extra-features around it are some of the main differentiating traits that make the Fabric8 client special and maybe better as compared to its competition.

:raised_hands:

PR #3018, opened 2021-04-20
allowing the auto-reconnect logic to work after an onClose exception

### LINKED ISSUE #2992: SharedInformer does not survive to an API server restart
SharedInformer created to watch `ImageStreams` or `BuildConfig` does not survive a k8s API server restart.

Please note, that this probably related to a bug in k8s api, as I was abled to reproduce the behaviour using the  `oc` command.

As user using `oc`, if I restart the api server, while watching `imagestreams`, I got the following error:
```
STATUS                REASON            MESSAGE
Failure               InternalError     an error on the server ("unable to decode an event from the watch stream: stream error: stream ID 3; INTERNAL_ERROR") has prevented the request from succeeding
```
The same operation using `oc get secrets -w` does not fail.

In the kubernetes-client, this materialized by an `EOFException` caught in the `io.fabric8.kubernetes.client.dsl.internal.WatcherWebSocketListener` which does not restart the WebSocket, but instead discards it from the manager.

```
2021-04-14 12:07:02 WARNING io.fabric8.kubernetes.client.dsl.internal.WatcherWebSocketListener onFailure Exec Failure
java.io.EOFException
	at okio.RealBufferedSource.require(RealBufferedSource.java:61)
	at okio.RealBufferedSource.readByte(RealBufferedSource.java:74)
	at okhttp3.internal.ws.WebSocketReader.readHeader(WebSocketReader.java:117)
	at okhttp3.internal.ws.WebSocketReader.processNextFrame(WebSocketReader.java:101)
	at okhttp3.internal.ws.RealWebSocket.loopReader(RealWebSocket.java:274)
	at okhttp3.internal.ws.RealWebSocket$2.onResponse(RealWebSocket.java:214)
	at okhttp3.RealCall$AsyncCall.execute(RealCall.java:203)
	at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:32)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

This is silent as a user point of view.

As a possible fix, we can considered having an `else` statement here: https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatcherWebSocketListener.java#L105

As, for an existing and started websocket, it is still possible to get a null response, which may have the signification that the websocket has been started previously, but became unaivalable.

edit:
Discussing with apiserver team, it seems that it is also impacting core objects, not only openshift specific. In my test, I was deleting the openshift apiserver part only. But, the same error is raised then in case of any other objects.

shawkins:
## Description
As seen in #2992 if there is a failure with the shared informer watch, and then an exception thrown from the ReflectorWatcher.onClose(WatcherException) method, the watch is effectively terminated.

Related to this, there isn't a great way from the available apis to determine when a index informer has failed.
Fix #2992 

## Type of change

## Checklist


manusa:
Hi @shawkins, does this one require backporting to 5.3.1 too? (like #2995)

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Watcher.java
```diff
@@ -19,6 +19,14 @@
 
 public interface Watcher<T> {
 
+  /**
+   * If the Watcher can reconnect itself after an error
+   * @return
+   */
+  default boolean reconnecting() {
```
manusa, line 26: Is the reconnect-after-error behavior working for all Watcher managers?
I see in the issue that we're tackling this in the SharedInformer implementations, but we're probably missing this feature for the rest.
It might be safer to move the method out of the interface to the ReflectorWatcher implementation (in case we're unsure if the reconnect feature works everywhere).

oscerd (approved):

shawkins:
> Hi @shawkins, does this one require backporting to 5.3.1 too? (like #2995)

Yes, we'd like this in 5.3.1 as well.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Watcher.java
shawkins: > Is the reconnect-after-error behavior working for all Watcher managers?

I only see isForcedClosed being referenced by WatcherWebSocketListener and WatchHTTPManager wrt reconnect behavior.  This change would have the same affect in both - they will still schedule reconnects after an exception.  But I do see an issue in that if it reaches the point where cannotReconnect returns false, it will hard loop - I'll make an update for that.

> It might be safer to move the method out of the interface to the ReflectorWatcher implementation (in case we're unsure if the reconnect feature works everywhere).

Do you mean doing an unwrap of the toggle and an instance of check for ReflectorWatcher in the AbstractWatchManager, rather than a reconnecting method on the Watcher interface?

rohanKanojia (approved):

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Watcher.java
manusa: OK, sorry, I just saw this was being handled in the AbstractWatchManager, so it will be taken into account for any Manager implementation (unless it's not using the AbstractWatchManager ---> hopefully not).

manusa: Just as a note, now there won't be a reconnection limit for those Watchers.

shawkins: > Just as a note, now there won't be a reconnection limit for those Watchers.

@manusa @rohanKanojia that is my understanding - an informer watch should always try to stay alive.  If it's possible for them to go down without user action, can we consider #3034 - so that it's easy to determine if the index state may be getting stale.

manusa: +1 #3034 makes a lot of sense in the new context after this PR

manusa:
Relates to: #2010

PR #3021, opened 2021-04-20
Fix #3015: Log InterruptedException in debug when it is a nominal terminating behavior

### LINKED ISSUE #3015: DefaultController#processLoop interrupt should not display stackTrace
When stopping an informer, the DefaultController#processLoop displays the interruption stacktrace.
This is annoying as it makes think that this is an unexpected behaviour.

It should be replaced by:
```
log.warn("DefaultController#processLoop got interrupted {}", t.getMessage());
```

hibnico:
## Description

Here is a review of the catch of the InterruptedException and their logs. I have done more than looking to the core code of the client. These suggestions are up for debate, and thus can be reverted in this PR.

So, every log about an nominal InterruptedException is now in debug, with the message of the exception but not the stack trace.

Special cases on some catch that I found:
* `BaseOperation#createOrReplace`: there used to be a `printStackTrace`. I tried to improve it by a log and propagate the status of the interrupted thread. But considering the case here, maybe there is more to do.
* `CreateOrReplaceHelper#createOrReplaceItem`: same considerations
* `CHEATSHEET.md`: I have found these weird piece of catch without the `try` keyword. I took the liberty to juste remove it. Javac will raise an error about the InterruptedException to handle, I see no need to do it explicitely here in the cheat sheet.
* the code exemples: to make things uniform, I made the logs at the warn level

A side note on `PortForwardEquivalent` example. I don't see how the `countDownLatch` is usefull there, `countDown` is never called on it.

Fix #3015 
## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/cache/Controller.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/cache/Controller.java
@@ -159,7 +159,7 @@ public class Controller<T extends HasMetadata, L extends KubernetesResourceList<
       try {
         this.queue.pop(this.processFunc);
       } catch (InterruptedException t) {
-        log.warn("DefaultController#processLoop got interrupted {}", t.getMessage(), t);
+        log.debug("DefaultController#processLoop got interrupted: {}", t.getMessage());
         Thread.currentThread().interrupt();
         return;
       } catch (Exception e) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 * Fix #2788: Support FIPS mode in kubernetes-client with BouncyCastleFipsProvider
 * Fix #2910: Move crd-generator tests from kubernetes-itests to kubernetes-tests
 * Fix #3005: Make it possible to select which CRD version is generated / improve output
+* Fix #3015: Thread interruption in a nominal case (like closing the client) are now logged in debug
 
 #### Dependency Upgrade
 * Fix #2979: Update Kubernetes Model to v1.21.0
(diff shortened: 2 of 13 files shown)


rohanKanojia (approved):

oscerd (approved):

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/cache/Controller.java
```diff
@@ -157,7 +157,7 @@ private void processLoop() throws Exception {
       try {
         this.queue.pop(this.processFunc);
       } catch (InterruptedException t) {
-        log.warn("DefaultController#processLoop got interrupted {}", t.getMessage(), t);
+        log.debug("DefaultController#processLoop got interrupted: {}", t.getMessage());
```
akram, line 160: Just a side note: In some tests that I ran, the `t.getMessage()` just returned `null` which is not very informative.

akram (approved):

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/cache/Controller.java
hibnico: From experience I never saw something usefull as a message of an `InterruptedException`. I kept the message in the log to continue as much as possible the current behavior. And maybe some jvm implementations provide additionnal info ? I don't mind removing the message if required.

akram: Let's keep it, better that nothing.

manusa (approved):
LGTM, thx!

PR #3051, opened 2021-04-27
fix: Pre-register String and its dependencies before processing

iocanel:
This PR adds a quick and dirty fix for the conflict between the two processors.
A proper fix needs to be implemented in sundrio:
- Isolate repositories
- Pre-populate core types (e.g. String, primitives etc).
- Prevent SO when dealing with cyclic definitions


metacosm:
I have to admit I don't understand this change… and I'm not sure we should go for a quick and dirty fix. 😢

iocanel:
> I have to admit I don't understand this change… and I'm not sure we should go for a quick and dirty fix.

So, the CRD generaator registers a String version that is very deep, in comparison with the shallow version used by sundro builder processor, the end result is that the builder processor falls into an infinite loop. This quick and dirty registered a simpler version of `String`, thill the issue is fixed in the builder processor.

Since, the root cause has been dealt with, let's kill this pr now.

PR #3058, opened 2021-04-29
fix: Potential NullPointerException on custom resource delete

christophd:
## Description
Fixes potential NPE when delete CR response is empty.

Delete operation response can be empty in https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/RawCustomResourceOperationsImpl.java#L1030

This lead to NPE when trying to read status object from response in https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/RawCustomResourceOperationsImpl.java#L1034

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/RawCustomResourceOperationsImpl.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/RawCustomResourceOperationsImpl.java
@@ -1038,7 +1038,7 @@ public class RawCustomResourceOperationsImpl extends OperationSupport implements
 
     // In most cases Status object is sent on deletion, but when deprecated DeleteOptions.orphanDependents
     // is used; object which is being deleted is sent
-    if (response.get("kind").toString().equals("Status")) {
+    if (!response.isEmpty() && response.get("kind").toString().equals("Status")) {
       return response.get("status").toString().equals("Success");
     }
     return true;
--- a/kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/CustomResourceTest.java
+++ b/kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/CustomResourceTest.java
@@ -282,6 +282,23 @@ class CustomResourceTest {
   }
 
   @Test
+  void testDeleteWithEmptyResponse() throws InterruptedException, IOException {
+    // Given
+    server.expect().delete().withPath("/apis/test.fabric8.io/v1alpha1/namespaces/ns1/hellos/example-hello").andReturn(HttpURLConnection.HTTP_OK, "").once();
+
+    // When
+    boolean result = client.customResource(customResourceDefinitionContext)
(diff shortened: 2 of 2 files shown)


oscerd (approved):

christophd:
I see unit tests failing in checks that are not related to the changes made in the PR: ```DefaultSharedIndexInformerIT.testPodSharedIndexInformerGetsSingleUpdates:85 expected:<[0]L> but was:<[3]L>```

rohanKanojia:
@christophd: Sorry for that, We have a flaky test that we need to fix. I'll restart the workflow for now.

rohanKanojia:
Hello, when you say empty response. Do you mean you're seeing cases where kubernetes apiserver returns empty response on DELETE calls?

manusa (approved):
LGTM, thx!

rohanKanojia (approved):

manusa:
> The Javadoc for `Deletable` contain:
> 
> > It is not guaranteed that the returned list will contain all values marked for deletion - see Issue #3058
> 
> It's not clear if this issue was fixed (in which case the above text should be removed from the Javadoc), or what the list can/will contain if this is not the case... Can you please elaborate? Thank you.


OK, so this Javadoc is a little bit unclear and probably outdated.

Its verbosity is mostly because there was a change in the method return type (from `boolean` to the `List<StatusDetails>`) some time ago.
Users were relying on the boolean response to assert that a resource was actually deleted.

However, there is no guarantee ([from the Kubernetes API server](https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-deletion)) that a resource will be deleted if a DELETE request is performed against such resource.

In order to make sure that a resource is deleted, you need to check that it's marked for deletion (has a `deletionTimestamp`) and that all the finalizers have been cleared.
Eventually, the resource will be deleted.

To ensure that a resource is deleted you'd need a further check such as:
```java
client.pods().withName($name).waitUntilCondition(Objects::isNull, 60, TimeUnit.SECONDS);
```

We can probably remove the reference to this PR and add a few extra tips to the Javadoc.

/cc @shawkins

shawkins:
> OK, so this Javadoc is a little bit unclear and probably outdated.

I don't think if was clarified later under what conditions exactly no response was returned from the server. I think this is what kubectl would do if it hit an empty response: https://github.com/kubernetes/kubectl/blob/cee2a595e4c6c501fd730bedf0f95c24a5395207/pkg/cmd/delete/delete.go#L425

> client.pods().withName($name).waitUntilCondition(Objects::isNull, 60, TimeUnit.SECONDS);

It would probably be perferable to use:

client.pods().withName($name).withTimeout(60, TimeUnit.SECONDS).delete();

That will wait only for what was actually deleted - someone recreating it won't affect the call.

manusa:
> client.pods().withName($name).withTimeout(60, TimeUnit.SECONDS).delete();

Forgot about this one.
Was that also working for multiple resources?

Issue #3067, opened 2021-04-30
Patch calls should use the base resource version to create a diff


shawkins:
On calls to edit or patch that pass in a resource the underlying patch logic is fetching the latest resource to generate the diff - this opens up the possibility that you are creating a diff against a different starting point than what you used (although that's much more unlikely with edit).  see https://github.com/fabric8io/kubernetes-client/blob/b54d41f854d4a34ebf82cef8310f4226f702f834/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/HasMetadataOperation.java#L114

The logic should instead check if the resource specifies a resource version and use that as the base version for the diff.  Only if there is no resource version specified should latest be assumed - or alternatively the client could require a resource version.

manusa:
Relates to #3078

shawkins:
Based upon discussion on the pr, it does not seem desirable to directly expose a patch(base, item) method.  Instead a patch(patchcontext, item) will be exposed - for consistency with the new patch method and flexibility.

shawkins:
This was resolved by #3070

PR #3070, opened 2021-04-30
adding a patch item operation with a context and other patch / patch derived operation changes

shawkins:
## Description
This is to address #3067 the current patch operation with only the modified item is dangerous when there are concurrent modifications.

This requires a two argument patch call.

I looked at adding the resourceVersion to a GET call, but the kubernetes support is "no older than" - https://kubernetes.io/docs/reference/using-api/api-concepts/#the-resourceversion-parameter so you are not guaranteed to get back the object of that version.  Based upon that, this is the simpler and backwards compatible change.

I'm not sure about the the appropriate way to clone the item in the patch call - or even if that is needed.  I can't see any guidance on whether it's expected that the operation may modify the item passed into it.

I also removed the retry logic on patching - it seems you will get a 422 error if the patch can't be applied rather than a conflict.

@rohanKanojia let me know if this is on the right track.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/HasMetadataOperation.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/HasMetadataOperation.java
@@ -49,7 +49,15 @@ public class HasMetadataOperation<T extends HasMetadata, L extends KubernetesRes
   @Override
   public T edit(UnaryOperator<T> function) {
     T item = getMandatory();
-    return patch(Serialization.clone(item), function.apply(item), false);
+    T clone = Serialization.clone(item);
+    return patch(null, clone, function.apply(item), false);
+  }
+  
+  @Override
+  public T editStatus(UnaryOperator<T> function) {
+    T item = getMandatory();
+    T clone = Serialization.clone(item);
+    return patch(null, clone, function.apply(item), true);
   }
 
   @Override
@@ -57,7 +65,7 @@ public class HasMetadataOperation<T extends HasMetadata, L extends KubernetesRes
     T item = getMandatory();
     T clone = Serialization.clone(item);
     consumer.accept(item);
-    return patch(clone, item, false);
+    return patch(null, clone, item, false);
   }
   
   protected VisitableBuilder<T, ?> createVisitableBuilder(T item) {
@@ -68,7 +76,7 @@ public class HasMetadataOperation<T extends HasMetadata, L extends KubernetesRes
   public T edit(Visitor... visitors) {
     T item = getMandatory();
     T clone = Serialization.clone(item);
-    return patch(clone, createVisitableBuilder(item).accept(visitors).build(), false);
+    return patch(null, clone, createVisitableBuilder(item).accept(visitors).build(), false);
   }
   
   /**
@@ -83,9 +91,21 @@ public class HasMetadataOperation<T extends HasMetadata, L extends KubernetesRes
     }
     throw new KubernetesClientException("Name not specified. But operation requires name.");
   }
-
+  
   @Override
   public T replace(T item) {
+    return replace(item, false);
+  }
+  
+  @Override
+  public T replaceStatus(T item) {
+    return replace(item, true);
+  }
+
+  /**
+   * base replace operation, which is effectively a forced update with retries
+   */
+  protected T replace(T item, boolean status) {
     String fixedResourceVersion = getResourceVersion();
     Exception caught = null;
     int maxTries = 10;
@@ -109,7 +129,7 @@ public class HasMetadataOperation<T extends HasMetadata, L extends KubernetesRes
         final UnaryOperator<T> visitor = resource -> {
           try {
             resource.getMetadata().setResourceVersion(resourceVersion);
-            return handleReplace(resource, false);
+            return handleUpdate(resource, status);
           } catch (Exception e) {
             throw KubernetesClientException.launderThrowable(forOperationType("replace"), e);
           }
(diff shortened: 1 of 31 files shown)


metacosm:
The way we clone in the Java operator SDK: https://github.com/java-operator-sdk/java-operator-sdk/blob/master/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/CustomResourceCache.java#L64-L71 (serialize -> deserialize)

shawkins:
@metacosm thanks, I've added that clone method as a utility method on the Serialization class.

@rohanKanojia I see there are a lot of internal calls to the single argument path function - given that these have concurrency issues, is that something that needs addressed as well?

shawkins:
@iocanel I see a relatively recent change to use edit(visitors) via a patch call.  I've updated this pr to base that upon a protected method to create the builder when possible - is there currently any way to find/construct the builder for a given type programatically so that this method is not needed?

@iocanel and others - what is the expectation for cloning via constructing a builder with a reference object?  for example - new PodBuilder(pod) - from what I can see in the code this is a shallow copy.  So for example in calls to edit where an item is associated with the context, that item may be edited.  Is this expected?  I see in some places this is referred to as a clone: https://github.com/fabric8io/kubernetes-client/blob/74cc63df9b6333d083ee24a6ff3455eaad0a6da8/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/apps/v1/DeploymentRollingUpdater.java#L41

@rohanKanojia another concern is that some of the calls to patch have overriding logic, such as to toggle the cascading flag - that logic will be applied in the case of a call to patch with an object, but it will not be applied when calling the new methods with a patch string.  Should there be common enforcement of that logic?

shawkins:
> @iocanel and others - what is the expectation for cloning via constructing a builder with a reference object? for example - new PodBuilder(pod) - from what I can see in the code this is a shallow copy. 

Actually this seems to only be an issue for Resource derived from CustomResource (I just assumed it was the same for other classes like Pod). I'll spawn this off as a separate issue - #3076

shawkins:
The other approach instead of introducing the two argument method would be to expose the withItem method on the Resource interface, so that you could make the patch call:

client.configMaps().inNamespace(currentNamespace).withName(name).withItem(base).patch(modified);

rohanKanojia:
I don't remember why this `cascading` is used in combination with `patch`. Since cascading is deprecated, I don't think we should use it in our newer methods

kubernetes-itests/src/test/java/io/fabric8/kubernetes/PatchIT.java
```diff
@@ -48,12 +51,12 @@
...
   @BeforeClass
   public static void init() {
-    ClusterEntity.apply(CronJobIT.class.getResourceAsStream("/patch-it.yml"));
+    ClusterEntity.apply(PatchIT.class.getResourceAsStream("/patch-it.yml"));
   }
 
   @Before
   public void initNamespace() {
-    this.currentNamespace = session.getNamespace();
+    this.currentNamespace = ClusterEntity.getArquillianNamespace();
```
rohanKanojia, line 59: Why is this change required?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/Serialization.java
```diff
@@ -312,4 +312,19 @@ private static String readSpecFileFromInputStream(InputStream inputStream) {
     }
     return JSON_MAPPER.readerFor(KubernetesResource.class).readValue(jsonString);
   }
+  
+  /**
+   * Create a copy of the resource via serialization
+   * @param <T>
+   * @param resource
+   * @return
+   */
```
rohanKanojia, line 321: Please complete javadoc

kubernetes-itests/src/test/java/io/fabric8/kubernetes/PatchIT.java
shawkins: It may not be.  When I was running the test through my ide I was seeing setup and initNamespace using different namespaces - this made sure they were using the same one.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/Serialization.java
shawkins: It's updated now.

kubernetes-itests/src/test/java/io/fabric8/kubernetes/PatchIT.java
rohanKanojia: I remember I had added `ClusterEntity.getArquillianNamespace()` to find out active namespace used in tests from static context(where session is not available). I think `session.getNamespace()` should work for your use case.

shawkins: Perhaps because of killing the tests prematurely in the ide or something else, they can return different namespaces - so the setup happens in one namespace, while the session is in another.  Then referencing the session namespace here causes the test to fail.  I'll remove this change.

rohanKanojia:
Some tests are failing. Are they related to this PR?

```
 Error:  Failures: 
Error:    SecurityContextConstraintsTest.testEdit:117 expected: <1> but was: <0>
Error:  Errors: 
Error:    OpenshiftRoleBindingTest.testPatchWithOnlySubjects:160 » NullPointer
Error:    OpenshiftRoleBindingTest.testPatchWithUserNamesAndGroupsAndNoSubjects:183 » NullPointer
Error:    OpenshiftRoleBindingTest.testPatchWithUserNamesAndGroupsAndOverwriteSubjects:204 » NullPointer
[INFO] 
Error:  Tests run: 744, Failures: 1, Errors: 3, Skipped: 4
[INFO] 
[INFO] ------------------------------------------------------------------------
```

shawkins:
> Some tests are failing. Are they related to this PR?

Yes they are.  There's an assumption down in OperationSupport.checkName that a HasMetadata object will always return a non-null metadata.  However these and another test were not written this way.  I've now relaxed that assumption, rather than continuing to refine the tests.

rohanKanojia (approved):

metacosm (commented):
LGTM apart from minor wording in exception message.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/HasMetadataOperation.java
```diff
@@ -42,18 +49,63 @@ public HasMetadataOperation(OperationContext ctx) {
...
+   * Get the current item from the server, consulting the metadata for the name if needed
+   */
+  protected Gettable<T> fromServer(ObjectMeta metadata) {
+    if (getName() != null) {
+      return fromServer();
+    }
+    if (metadata != null) {
+      return withName(metadata.getName()).fromServer();
+    }
+    throw new KubernetesClientException("Name not specified. But operation requires name.");
```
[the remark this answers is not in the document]
shawkins: I took this wording from OperationSupport - do you want all of the similar messages updated?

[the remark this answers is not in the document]
metacosm: Nevermind then. I found the wording strange but if it's already used elsewhere, let's keep it as-is.

metacosm (approved):

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/HasMetadataOperation.java
```diff
@@ -42,18 +49,63 @@ public HasMetadataOperation(OperationContext ctx) {
...
+  public T edit(Visitor... visitors) {
+    T item = getMandatory();
+    T clone = Serialization.clone(item);
+    return patch(null, clone, createVisitableBuilder(item).accept(visitors).build(), false);
+  }
+  
+  /**
+   * Get the current item from the server, consulting the metadata for the name if needed
+   */
+  protected Gettable<T> fromServer(ObjectMeta metadata) {
```
manusa, line 85: why protected and not private?

shawkins: It can be whatever you want.  BaseOperation.fromServer is public, but I was unsure if this should be as well, so I went with protected.

manusa: I really don't see how a children class would need to reimplement this, anyway, let's leave it and get this merged :rocket:

Issue #3072, opened 2021-04-30
Simplify/consolidate thread utilization by informers


shawkins:
With the current implementation each informer directly consumes a thread for:

1. Controller - resync executor thread pool
2. DefaultSharedIndexInformer - controller thread
3. SharedProcessor - executor, which is 1 thread per listener.

This becomes an issue when running lots of informers.

We should be able to consolidate or remove several of these.   #3061 for example allows the controller thread to complete.

Additionally each watch has a websocket reader thread and a scheduler thread.

manusa:
Relates to: #2010

manusa:
Should be fixed addressed in #3108

PR #3108, opened 2021-05-12
overhauling all threading concerns related to informers

shawkins:
## Description
Fix for #3072 and #3143, which includes:

- fully removing the controller thread and making run blocking - this could be viewed as an important behavioral change.  This requires not using the forkJoin pool in the informer factory because the async run calls will be blocking.  However this allows anyone directly calling run to know immediately if the informer has been established without using a listener.  An alternative is to add a separate runBlocking method and go back to passing in the executorservice into the DefaultSharedIndexInformer.  Note the go implementation has a blocking run method - https://github.com/kubernetes/client-go/blob/f6ce18ae578c8cca64d14ab9687824d9e1305a67/tools/cache/controller.go#L127 https://github.com/kubernetes/client-go/blob/f6ce18ae578c8cca64d14ab9687824d9e1305a67/tools/cache/shared_informer.go#L157

- made all thread pools daemon - by using a utility method
to create the daemon thread factory

- inverting the paradigm for processorlistener execution

- merging the controller class with defaultsharedindexinformer, since controller only had a minimal amount of responsibilities

- since resync is in-memory only there is now only 1 shared scheduler for all informers, which is also reused by abstractwatchmanager

- event handlers are processed in a thread off of main informer threadpool - but retain serial execution.

- to remove the code smell of the run exception handling, #3143 was addressed by moving concerns about the informer listerners to the factory.

## Type of change

## Checklist


oscerd (approved):

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/cache/SharedProcessor.java
```diff
@@ -148,10 +108,30 @@ public boolean shouldResync() {
   public void stop() {
     lock.writeLock().lock();
     try {
+      toClose.forEach(c->c.run());
```
rohanKanojia, line 111:
```suggestion
      toClose.forEach(Runnable::run);
```

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/informers/SharedSchedulerTest.java
```diff
@@ -0,0 +1,52 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class SharedSchedulerTest {
+  
+  @Test
+  void testAutoShutdown() throws InterruptedException {
+    SharedScheduler scheduler = new SharedScheduler(50);
+    assertFalse(scheduler.hasExecutor());
+    CountDownLatch latch = new CountDownLatch(2);
+    ScheduledFuture<?> future = scheduler.scheduleWithFixedDelay(()->latch.countDown(), 50, 50, TimeUnit.MILLISECONDS);
```
rohanKanojia, line 35:
```suggestion
    ScheduledFuture<?> future = scheduler.scheduleWithFixedDelay(latch::countDown, 50, 50, TimeUnit.MILLISECONDS);
```

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/informers/SharedSchedulerTest.java
```diff
@@ -0,0 +1,52 @@
...
+    assertTrue(scheduler.hasExecutor());
+    latch.await();
+    assertFalse(future.isDone());
+    future.cancel(true);
+    for (int i = 0; i < 10; i++) {
+      if (!scheduler.hasExecutor()) {
+        break;
+      }
+      Thread.sleep(100);
```
rohanKanojia, line 44: We are already using awaitility in kubernetes-tests and kubernets-itests. Maybe you can avoid Thread.sleep by adding awaitility in kubernetes-client module

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/SharedScheduler.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Maintains a single thread daemon scheduler.
+ * It is not intended for long-running tasks
+ */
+public class SharedScheduler {
+  
+  public static long DEFAULT_TTL_MILLIS = TimeUnit.SECONDS.toMillis(10);
```
rohanKanojia, line 31: Do you think 10 milliseconds is too short for TTL or is it fine?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/ResourceEventHandler.java
```diff
@@ -54,4 +54,12 @@
...
    */
   void onDelete(T obj, boolean deletedFinalStateUnknown);
+  
+  /**
+   * If the event handler should be called in a separate thread
+   * @return
+   */
+  default boolean isCalledAsync() {
+    return true;
+  }
```
rohanKanojia, line 64: Shall we add some example/test for the case when the user overrides this?

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/informers/SharedSchedulerTest.java
shawkins: Yes, I had seen that, but since there wasn't yet a dependency I went ahead with the sleeps.  Since you are okay with adding the dependency, I'll switch.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/ResourceEventHandler.java
shawkins: This is just to prevent the proliferation of handler threads - I'm on the fence as to whether this is a good idea.  See the pr description - if you (or anyone else) would rather limit the threads dedicated to events in a simpler way, such as just 1 thread for all handlers and/or reusing the executor passed into the informerfactory for this purpose, I'd be glad to get rid of this method.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/SharedScheduler.java
shawkins: That actually evaluates to 10 seconds worth of millis.

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/informers/SharedSchedulerTest.java
shawkins: updated now

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/ResourceEventHandler.java
shawkins: @rohanKanojia @manusa are you comfortable with an approach that would use a single thread for all handlers instead of an isCalledAsync method?

shawkins: took a different approach of reusing the overall executor informer.

shawkins:
If the approaches here seem good, then logic in dsl.internal would be good to review for consolidating threading concerns.

manusa:
Relates to: #2010

shawkins:
I've expanded the scope to include the the AbstractWatchManager thread as well - it will now use shared resources, rather than a dedicated thread.  The only dedicated threads now related to informers will be held by okhttp.

All of the resync related changes are potentially moot based upon #3159

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/Utils.java
```diff
@@ -57,6 +61,9 @@
   public static final String PATH_WINDOWS = "Path";
   public static final String PATH_UNIX = "PATH";
   private static final Random random = new Random();
+  
+  private static final ExecutorService SHARED_POOL = Executors.newCachedThreadPool();
```
manusa, line 65: Thread creation in Cached Thread Pools are probably lazy by default (threads are only created when needed and then reused. However, I'm not sure if eagerly initializing this variable might have other consequences when users are not using any of the features that consume this pool.

shawkins: > threads are only created when needed and then reused

That is correct.

> However, I'm not sure if eagerly initializing this variable might have other consequences when users are not using any of the features that consume this pool.

It will not, but it can of course be instantiated on demand if you want.

manusa: > It will not, but it can of course be instantiated on demand if you want.

Not sure if it's worth the effort

This was more a disclaimer comment ;P

manusa (approved):
LGTM, thx!

PR #3141, opened 2021-05-18
fix #3122 normalizing the service create exception to be a conflict

### LINKED ISSUE #3122: Repeated createOrReplace with a service fails
If you do:

```
    service = client.services().inNamespace(session.getNamespace()).createOrReplace(service);
    service = client.services().inNamespace(session.getNamespace()).createOrReplace(service);
```

The second call will fail with:

```
io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: POST at: https://192.168.39.135:8443/api/v1/namespaces/itest-08651507/services. Message: Service "createorreplaceit-svc" is invalid: spec.clusterIPs: Invalid value: []string{"10.101.116.8"}: failed to allocated ip:10.101.116.8 with error:provided IP is already allocated. Received status: Status(apiVersion=v1, code=422, details=StatusDetails(causes=[StatusCause(field=spec.clusterIPs, message=Invalid value: []string{"10.101.116.8"}: failed to allocated ip:10.101.116.8 with error:provided IP is already allocated, reason=FieldValueInvalid, additionalProperties={})], group=null, kind=Service, name=createorreplaceit-svc, retryAfterSeconds=null, uid=null, additionalProperties={}), kind=Status, message=Service "createorreplaceit-svc" is invalid: spec.clusterIPs: Invalid value: []string{"10.101.116.8"}: failed to allocated ip:10.101.116.8 with error:provided IP is already allocated, metadata=ListMeta(_continue=null, remainingItemCount=null, resourceVersion=null, selfLink=null, additionalProperties={}), reason=Invalid, status=Failure, additionalProperties={}).
```

shawkins:
## Description
fixing #3122 - normalizing the 422 create exception to a 409 conflict

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/core/v1/ServiceOperationsImpl.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/core/v1/ServiceOperationsImpl.java
@@ -16,9 +16,6 @@
 package io.fabric8.kubernetes.client.dsl.internal.core.v1;
 
 import io.fabric8.kubernetes.api.builder.Visitor;
-import io.fabric8.kubernetes.api.model.Service;
-import io.fabric8.kubernetes.api.model.ServiceBuilder;
-import io.fabric8.kubernetes.api.model.ServiceList;
 import io.fabric8.kubernetes.api.model.*;
 import io.fabric8.kubernetes.client.*;
 import io.fabric8.kubernetes.client.Config;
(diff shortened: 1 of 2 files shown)


oscerd (approved):

manusa:
This is a little hacky because the new Exception is based on the assumption of what happens in the createOrReplace method. If that implementation is changed in the future, then this piece of code will become completely outdated and odd.

Anyway it does the job.

manusa:
I'm thinking of a more generic approach that would involve having a list of status for which createOrReplace should perform the replacement.
This default list of status (which includes only 409) could then be overridden per-resource operations (such as ServiceOperationsImpl) to include additional status like 422.

WDYT? maybe we can implement something like this in the future.

manusa (approved):

shawkins:
> This is a little hacky because the new Exception is based on the assumption of what happens in the createOrReplace method.

Yes and no.  It's the createOrReplace logic that is specifically checking for the 409.  However anyone calling create and expecting to see a 409 would have a similar issue.  Alternatively if we want the methods to return the codes as directly as possible, the I agree with you create should still throw the 422.

> WDYT? maybe we can implement something like this in the future.

I see this a workaround to a kubernetes bug.  It looks like this was captured a long time ago, and nothing was done - https://github.com/kubernetes/kubernetes/issues/54425

It just comes down to the scope of the workaround for their bug - is it only for createOrReplace, or all creates?

shawkins:
On the changelog I wasn't sure if this needs to be called out as a breaking change of some kind.

manusa:
> Alternatively if we want the methods to return the codes as directly as possible, the I agree with you create should still throw the 422.

For me this is what makes it "hacky". Since we are forcing a different status code just to trigger a behavior of a (internal) consumer method.

> It just comes down to the scope of the workaround for their bug - is it only for createOrReplace, or all creates?

Agreed, maybe it's too much for just this given scenario (I'm unsure this (retrying based on different status codes) would apply to other scenarios)

> On the changelog I wasn't sure if this needs to be called out as a breaking change of some kind.

Maybe some users have already tackled the issue and are expecting a 422 status which now will never come, so at least a note makes sense.

Maybe including the original status code in the exception message might also be interesting.

shawkins:
After thinking about @rohanKanojia email a little more, I'm going to close this pr - it's not worth the behavior change.  Our calling logic will take responsibility for setting the clusterIp to null in a repeated create scenario.

PR #3165, opened 2021-05-24
fix #3159: simplifying the resync logic

### LINKED ISSUE #3159: Simplify the informer resync logic
In reviewing the informer resync logic it seems a little complicated and it's also imprecise.  

For example try starting an informer with a resync of 1 second and adding a handler with the default resync and another with a value of 1.1 seconds.  You'll notice that the latter handler is called much less often than expected.  That's because the logic is driven by the shortest interval, rather than by a least common denominator.  Unless you only add handlers with resyncs shorter or longer than the lowest interval by an even multiple you won't actual get the behavior you expect.

Also if the informer is already started, then you cannot successfully add a handler with a smaller interval than what is already in use.

As an alternative I propose the following behavior - let there be only one resync interval, which is the one defined when the informer is created.  Using addEventHandlerWithResyncPeriod should be deprecated and the meaning changed to allow a 0 value to indicate resync is disabled for a given handler.  The replacement method should be addEventHandlerWithoutResync.

With that change almost all of the resync code will go away and have a much precise behavior.

cc @rohanKanojia @manusa

shawkins:
## Description
This treats the informer resync period as the only meaningful resync interval.  It is however a breaking change and does diverge from the go client.
 
## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@
 #### New Features
 * Fix #3133: Add DSL Support for `authorization.openshift.io/v1` resources in OpenShiftClient
 
+#### _**Note**_: Breaking changes in the API
+##### DSL Changes:
+- #3159 The resync period that an informer is created with is the only relevant period for that informer.  Handlers that are added will use the informer resync period unless resync is disabled when the handler is added.
+
 ### 5.4.0 (2021-05-19)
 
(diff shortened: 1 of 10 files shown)


shawkins:
After better understanding the expectations from the go code, this is too much of a breaking change.

Issue #3091, opened 2021-05-08
Refine the Informer Store interface


shawkins:
There are several options to improve the Store interface.

The most aggressive, which would deviate from the go client, would be to remove all of the mutative / internal methods to a separate interface - add, update, delete, replace, resync, isPopulated, hasSynced would be part of a  - which would then no longer be directly exposed via the Indexer interface.  The rationale is that these either have no-op behavior or should not be called directly.

A lesser refactoring would simply align the interface with the go client, such that newer methods added in fabric8 are moved to a separate interface.

Once #3061 is committed all of the cache methods can be type-safe - no need to reference the object type.

#3090 proposes removing the isPopulated method altogether.

shawkins:
@rohanKanojia  a full proposal for refactoring after #3108 looks like https://github.com/shawkins/kubernetes-client/pull/3

rohanKanojia:
@shawkins : Thanks, let me review it later today :+1:

shawkins:
Closed via #3167

PR #3167, opened 2021-05-24
fix #3090 #3091 refining the store interface and hasSynced

### LINKED ISSUE #3013: IndexInformer cant restart after call function stop
@Override
  public void run() {
    if (started) {
      return;
    }

    started = true;

    this.processor.run();
    controllerThread.start();
  }

  @Override
  public void stop() {
    if (!started) {
      return;
    }

    stopped = true;
    controller.stop();
    controllerThread.interrupt();

    processor.stop();
  }

After call function stop；var "started" is still true;  so when call run() again, this just return after check "started" is true.

is this correct? or is a bug?

### LINKED ISSUE #3090: Informer.hasSynced and Store.isPopulated
A change in #1961 was to set hasSynced to false when re-listing.  This differs from the go client behavior - where once true, it won't flip back to false see https://github.com/kubernetes/client-go/blob/f6ce18ae578c8cca64d14ab9687824d9e1305a67/tools/cache/delta_fifo.go and the doc https://github.com/kubernetes/client-go/blob/f6ce18ae578c8cca64d14ab9687824d9e1305a67/tools/cache/shared_informer.go#L160

Can we return hasSynced to the original meaning and introduce more specific methods for additional checks - such as isWatching?

A change along these lines would allow us to eliminate the Store.isPopulated method

cc @rohanKanojia

shawkins:
## Description
Refining the Store interface by splitting internal and mutative methods to SyncableStore

Removing isPopulated and directly wiring the cache to be used
this adds an isRunning method to the SharedInformer.  this clarifies the
cache methods to be more like a map and the underlying map is now
concurrent to remove read locks

This could be seen as a breaking change since Store is somewhat of a public interface - however it seems much clearer and safer to have a narrower interface.

Closes #3013

## Type of change

## Checklist


oscerd (approved):

manusa:
Relates to: #2010

rohanKanojia (approved):

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/cache/Cache.java
```diff
@@ -46,15 +47,15 @@
   private Map<String, Function<T, List<String>>> indexers = new HashMap<>();
 
   // items stores object instances
-  private Map<String, T> items = new HashMap<>();
+  private volatile ConcurrentHashMap<String, T> items = new ConcurrentHashMap<>();
```
manusa, line 50: Is it really needed for the ConcurrentHashMap to be declared as volatile?

manusa: Oh, I see it's being replaced later on

manusa (approved):
LGTM, thx!

PR #3172, opened 2021-05-26
fix: Use File.toURI() to create the generated crd uri.

iocanel:
This is meant to fix the issue with the malformed URI on windows.

The fix has not been verified manually, but usually this short of issues are handled better when you delegate to java itself.


rohanKanojia (approved):

manusa (approved):
LGTM, thx!

PR #3181, opened 2021-05-27
fix: handling of JsonProperty when generating crds.

iocanel:
This fixes an issue with when generating schema for items annotated with the `JsonProerty` annotation.

### DIFF
--- a/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
+++ b/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
@@ -23,15 +23,14 @@ import io.fabric8.kubernetes.api.model.IntOrString;
 import io.fabric8.kubernetes.api.model.Quantity;
 import io.sundr.builder.internal.functions.TypeAs;
 import io.sundr.codegen.functions.ClassTo;
+import io.sundr.model.AnnotationRef;
 import io.sundr.model.ClassRef;
 import io.sundr.model.Method;
 import io.sundr.model.PrimitiveRefBuilder;
 import io.sundr.model.Property;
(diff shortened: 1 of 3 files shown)


oscerd (approved):

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
```diff
@@ -138,6 +139,7 @@ protected T internalFrom(TypeDef definition, String... ignore) {
       final Property[] updated = {property};
 
       if (property.isStatic() || ignores.contains(name)) {
+        System.out.println("\tIgnoring!");
```
manusa, line 142: I think the LOGGER is now configured to print to System in case of APT.

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
```diff
@@ -166,10 +168,8 @@ protected T internalFrom(TypeDef definition, String... ignore) {
           // if we found an annotated accessor, override the property's name if needed
           .ifPresent(a -> {
             final String fromAnnotation = (String) a.getParameters().get("value");
-            if(!updated[0].getName().equals(fromAnnotation)) {
-              updated[0] = new Property(property.getAnnotations(), property.getTypeRef(),
-                fromAnnotation, property.getComments(),
-                property.getModifiers(), property.getAttributes());
+            if(!Strings.isNullOrEmpty(fromAnnotation) && !updated[0].getName().equals(fromAnnotation)) {
```
metacosm, line 171: So the issue was that there was a JSONProperty annotation without a name? Could we add a test case for that, please?

manusa (approved):

PR #3156, opened 2021-05-21
fix fabric8io#3144: making the crud mock logic more crd aware

shawkins:
## Description
#3144 this adds broader crd awareness to the mock server - primarily for determining if status is a subresource.  If there is no crd, it's assumed that status is not a subresource.

There's also a minor change to not send events from patches that result in no changes to the resource.

## Type of change

 - [x] Bug fix (non-breaking change which fixes an issue)
 - [ ] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [x] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [x] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [x] I Added [CHANGELOG](../CHANGELOG.md) entry regarding this change
 - [x] I have implemented unit tests to cover my changes
 - [ ] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/master/doc/CHEATSHEET.md) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [ ] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

Issue #3023, opened 2021-04-20
custom resources: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update


lburgazzoli:
I'm trying to use an "unstructured" client using custom resource as workaround and I have something like this:

```java
CustomResourceDefinitionContext ctx = ... ;
Map<String, Object> unstructured = ...;

return kubernetesClient
    .customResource(ctx)
    .inNamespace(namespace)
    .createOrReplace(unstructured);
```

Sometimes, I get an exception like:

```
Message: kamelets.camel.apache.org "aws-sqs-source-v1alpha1" is invalid: metadata.resourceVersion:
 Invalid value: 0x0: must be specified for an update. Received status: Status(apiVersion=v1, code=422,
 details=StatusDetails(causes=[StatusCause(field=metadata.resourceVersion, message=Invalid value: 0x0:
 must be specified for an update, reason=FieldValueInvalid, additionalProperties={})], 
group=camel.apache.org, kind=kamelets, name=aws-sqs-source-v1alpha1, retryAfterSeconds=null, uid=null, 
additionalProperties={}), kind=Status, message=kamelets.camel.apache.org "aws-sqs-source-v1alpha1" is 
invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update, 
metadata=ListMeta(_continue=null, remainingItemCount=null, resourceVersion=null, selfLink=null, 
additionalProperties={}), reason=Invalid, status=Failure, additionalProperties={})
```

manusa:
Extra logic added in #2501 and contained in [CreateOrReplaceHelper](https://github.com/fabric8io/kubernetes-client/blob/db7d4eba49a5dac3ce916f782048aaecc2e40cbd/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/CreateOrReplaceHelper.java) is not applicable for Untyped/Raw custom resources.

matthyx:
Should be add it for Raw custom resources? What's the plan to them, because as I see it, it's unusable for any replacement.

rohanKanojia:
We don't have it in our sprint plan right now. Do you think you have some time to fix this? Just need to do what we're doing in CreateOrReplaceHelper..

matthyx:
Thanks @rohanKanojia I will check once we merge the other PR...

Issue #2929, opened 2021-03-18
Provide a dynamic client


lburgazzoli:
The official Kubernetes provides two types of clients 
- **typed** which relies on k8s.io/api structs and offer a type safe, programming model
- **untyped** which uses `unstructured.Unstructured` to represent any king of Kubernetes resource

The Fabric8 Kubernetes Client, is similar to the **typed** one as it requires the Java model representing the Kubernetes resource to be known but it forces to be tightly coupled with versions and the type used and for some use cases, having unstructured capabilities could be very useful. 

We can then introduce an `Unstructured` class, like:

```java
/**
 * Represent a generic Kubernetes resource and provides helper methods
 * to access common attributes such as namespace, spec, status, etc
 */
class Unstructured implements Map<String, Object> {
}
```

Which can be leverage for something like:

```java
var gvr = new GroupVersionResourceBuilder()
    .withGroup("apps")
    .withVersion("v1")
    .withResource("deployments")
    .build()

var res = new UnstructuredBuilder()
    .withField("apiVersion", "apps/v1")
    .withField("kind", "Deployment")
    .withSection("metadata")
        .withField("name", "my-deployment")
    .build()

// update or create     
Unstructured result = client.resource(gvr).inNamespace("ns").createOrUpdate(res);

// watch
client.resource(gvr).inNamespace("ns").watch(u -> {
    System.out.printf("a=%s, k=%s, n=%s\n", u.getApiVersion(), u.getKind(), u.getName());
})
```

rohanKanojia:
Is your use case similar to this[0]

[0] https://github.com/fabric8io/kubernetes-client/blob/master/doc/CHEATSHEET.md#customresource-typeless-api

lburgazzoli:
Yes similar to that one but not limited to `CustomResources`

rohanKanojia:
@lburgazzoli : I was trying out using `ConfigMap` with this raw API as per our discussion in chat but unfortunately there is a bug in Raw API which doesn't allow it to work with code API(where apiGroup is ""). So this code:
```java
        try (KubernetesClient kubernetesClient = new DefaultKubernetesClient()) {
            CustomResourceDefinitionContext context = new CustomResourceDefinitionContext.Builder()
                    .withPlural("configmaps")
                    .withVersion("v1")
                    .withGroup("")
                    .withScope("Namespaced")
                    .build();

            Map<String, Object> output = kubernetesClient.customResource(context).inNamespace("default").list();
[...]

PR #3185, opened 2021-05-27
feat: Introducing GenericKubernetesResource

manusa:
### Introduction of the GenericKubernetesResource

This class holds the minimum structure for any valid resource served by the Kubernetes API. i.e. A HasMetadata implementation with an `ObjectMeta metadata` field and a Map for any additional property/field.

The main purpose of this class is to be able to Serialize/Deserialize any resource served by Kubernetes regardless of a specific type previously registered for that resource.

The Serializer now looks for any registered type for a given resource and in case none is found, the resource is deserialized as a GenericKubernetesResource. The immediate benefit is that deserialization will never fail.

This resource can be later on used in a specific [`OperationSupport`](https://github.com/fabric8io/kubernetes-client/blob/71c1a8c09035f09d65c16d7fcface9e509518a52/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationSupport.java) implementation that will provide **consistent** operation support for any resource.

### RawCustomResourceOperationsImpl uses GenericKubernetesResourceOperationsImpl

This brings consistency to the client, but no specific behavioral changes should be noticed. Some changes might be noticed in case of the MockServer and white box testing, since some of the requests have changed, but the result should be no different.

The only major difference is that a `.get()` for a non-existent resource previously threw an Exception and now returns null (just like the rest of OperationSupport implementations for any other resource).

Redundant methods have been marked as deprecated. We should consider removing them definitely after the next few releases.

### Next steps

Improve `KubernetesClient#load` to support operations for GenericKubernetesResource. This means requesting the applicable CustomResourceDefinition from the cluster in order to infer the plural and scope of the managed resource.

## Type of change

## Checklist


shawkins:
This change will be needed to rework/remove the raw websocket watch logic - it seems broken in the current state because it doesn't fully understand the events (http gone, updating the resource version, lists, etc.).

shawkins:
Should we hold off on #3156 until after this work?  If so, I will split off the additional logic that addresses #3194 as a separate pr.

manusa:
> Should we hold off on #3156 until after this work? If so, I will split off the additional logic that addresses #3194 as a separate pr.

My intention is to at least finish the initial work on this by today (I already spent 2 days working on it and can't spend more time).

This PR introduces the Generic Resource that can be further used in many areas of the client but those should be tackled separately.

shawkins:
> This PR introduces the Generic Resource that can be further used in many areas of the client but those should be tackled separately.

Okay, I'll hold off on adopting this for now in the mock.  In the meantime I've rebased #3156 to master, when you get a chance please review.

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/internal/KubernetesDeserializer.java
```diff
@@ -82,7 +82,7 @@ private static KubernetesResource fromObjectNode(JsonParser jp, JsonNode node) t
         if (key != null) {
```
shawkins, line 82: I tried these changes with removing FallbackHasMetadata.  Some of the raw logic was written with tests that do not specify the kind in a create.  I'm not sure if that's allowed, but we could easily account for that here by using the generic logic when either the key or the resourceType is null.

shawkins (changes requested):

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/internal/KubernetesDeserializer.java
manusa: I cannot think of a case (besides JKube's Fragments) where you can define a resource and not specify the kind and the apiVersion.

We could modify to always try to deserialize to GenericKubernetesResource and let Jackson fail. However I'm not sure of the implications or backwards compatibility side-effects of this change:

```java
    private static KubernetesResource fromObjectNode(JsonParser jp, JsonNode node) throws IOException {
        String key = getKey(node);
        if (key != null) {
            Class<? extends KubernetesResource> resourceType = mapping.getForKey(key);
            if (resourceType != null && KubernetesResource.class.isAssignableFrom(resourceType)){
                return jp.getCodec().treeToValue(node, resourceType);
            }
        }
        return jp.getCodec().treeToValue(node, GenericKubernetesResource.class);
    }
```

shawkins: After trying to modify test-rawcustomresource.yml, I see kind is definitely required.  So CustomResourceCrudTest and a unit test that was added in the mock logic need to be updated to include kind to incorporate your changes.  With that understanding what you had to start with is good.

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/GenericKubernetesResource.java
```diff
@@ -0,0 +1,70 @@
...
+public class GenericKubernetesResource implements HasMetadata {
+
+  @JsonProperty("apiVersion")
+  private String apiVersion;
+  @JsonProperty("kind")
+  private String kind;
+  @JsonProperty("metadata")
+  private ObjectMeta metadata;
+  @JsonIgnore
+  private transient Map<String, Object> additionalProperties = new HashMap<>();
```
shawkins, line 54: Can we use a LinkedHashMap here instead?  For the mock server logic I'll be able to manipulate GenericKubernetesResource objects directly and that will keep the output json looking very similar to the input.

manusa: I think we can set the default value to LinkedHashMap.
However, I'm not sure if we can guarantee the same for deserialization purposes.

shawkins: I tested locally using 
private transient LinkedHashMap<String, Object> additionalProperties = new LinkedHashMap<>(); 
and everything worked as expected.

shawkins (approved):
LGTM

manusa:
Hi @lburgazzoli, since (I think) you are an active user of the RawCustomResources, it would be very interesting to count with your review and check in case something is broken for you with the changes suggested in this PR. :pray:

lburgazzoli:
@manusa I'll try to build the PR and use it in my project beginning of next week, my knowledge of the k8s client internals is too limited to give any valuable review. If there is an example that show how to use the new resource, I may take a look at it today to see if I spot something missing.

manusa:
> If there is an example that show how to use the new resource, I may take a look at it today to see if I spot something missing.

Basically there is no way to explicitly use the new Resource (as of now). The partial purpose of this PR RawCustomResourceOperationsImpl instead of re-implementing everything from scratch (and leading to issues like some of the ones you reported (#3023, and indirectly #2929)) it will use the regular implementations by delegating the logic to the new `GenericKubernetesResourceOperationsImpl`.

So if you're still using the "unstructured" RawCustomResourceOperationsImpl ([`kubernetesClient.customResource(CustomResourceDefinitionContext)`](https://github.com/fabric8io/kubernetes-client/blob/c0a0107e2420fad3d2ad9c76028066401364536e/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/KubernetesClient.java#L206)), the new functionality will kick in. Any troubles you may have had previously with this DSL should remain fixed, and functionality should be complete.

This is also the base to any enhancement or feature we want to perform regarding #2929.

lburgazzoli:
ok, so I'll try to compile it and see if it breaks something

rohanKanojia (approved):

manusa:
> @manusa I've updated my operator to use the code from this PR and nothing seems to be broken, so LGTM

Great, thx for trying it out. Merging :rocket:

PR #3212, opened 2021-06-03
fix #3179: creating a common base for the default kube and os clients

### LINKED ISSUE #3179: Functional differences between DefaultOpenShift and DefaultKubernetes clients
What are the expected functional differences between these clients?

We're seeing that at least custom resource handling differs because the defaultkubernetes client specifies a default delete propagation of background, while the openshift client does not so it default to orphan.

cc @rohanKanojia @manusa

shawkins:
## Description
Creating a common base for the default kube and os clients.  This seems to require a generic form of the DefaultKubernetesClient, which I've named here BaseKubernetesClient.  Any of the problematic methods dealing with NamespacedXClient are left for the Default classes to implement.  Otherwise everything else can be shared from the base.

## Type of change

## Checklist


manusa (approved):
LGTM, thx!

manusa:
What bug are we fixing with this?

rohanKanojia:
I think it's https://github.com/fabric8io/kubernetes-client/issues/3179 maybe

manusa:
> I think it's #3179 maybe

Let me elaborate

That's not a bug, the issue is marked as a bug-fix on top of a chore/refactor.
image

I want to know if this PR is actually fixing something that wasn't working OK besides the refactor of the default clients.

shawkins:
> That's not a bug, the issue is marked as a bug-fix on top of a chore/refactor.

From our projects perspective #3179 is a bug - prior to using operator sdk 1.8.2 custom resource deletes were cascading.  After the upgrade they weren't because of the difference in the implementation in the openshift client, which we were forced to use instead.

If it was intentional, or at least now expected, to have the openshift client default to orphaned deletes, then this should be noted as a breaking change as well.

metacosm (approved):

PR #3221, opened 2021-06-05
fix #3098: making the operation logic more consistent

### LINKED ISSUE #3098: Review if patch(context, string) should handle the rolling case
https://github.com/fabric8io/kubernetes-client/blob/74cc63df9b6333d083ee24a6ff3455eaad0a6da8/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/apps/v1/RollableScalableResourceOperation.java#L171 the old patch and other operations consider the rolling case separately.  This should be reviewed if it is needed for the new patch(context, string) method and/or if the rolling case could be handled differently given the additional base logic that's now available.

This is related to #3070

shawkins:
## Description
After reviewing #3098 I can see that we should only modify and use alternate rolling logic when the full object is expected by a patch - which is only for a JSON patch currently.

Fixes #3098 

## Type of change

## Checklist


oscerd (approved):

rohanKanojia (approved):

manusa (approved):
LGTM, thx!

manusa:
Needs rebase :disappointed:

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/HasMetadataOperation.java
```diff
@@ -196,7 +216,13 @@ protected T patch(PatchContext context, T base, T item, boolean status) {
         if (item.getMetadata() == null) {
           item.setMetadata(new ObjectMeta());
         }  
-        item.getMetadata().setResourceVersion(base.getMetadata().getResourceVersion());  
+        item.getMetadata().setResourceVersion(base.getMetadata().getResourceVersion());
+      }
+      final T current = base;
+      try {
+        item = modifyItemForReplaceOrPatch(() -> current, item);
```
rohanKanojia, line 223: Hello, why are we modifying item for replace or patch only in case of `base == null`? Shouldn't this be called always?

rohanKanojia: We're facing an error during patching Service in Eclipse JKube https://github.com/eclipse/jkube/issues/756 . In that case, a base is a valid object; `modifyItemForReplaceOrPatch` overridden method in ServiceOperationsImpl doesn't get invoked due to which we get 422 from API Server

shawkins: Only when the base is unknown and this produces a json patch do you need to modify what the user passed in as it may not match what is current - that corresponds to the old logic in ServiceOperationImpl,patch(Service item)

If the base is non-null then this operation is a edit - meaning that the base is already current and you are producing a patch based upon your changes to the current.  You don't need to modify in that case.

shawkins: @rohanKanojia what dsl method is being called specifically in the linked issue?  Is it patch(Service item)?

rohanKanojia: It's this in JKube's [PatchService](https://github.com/eclipse/jkube/blob/master/jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/PatchService.java#L173) :
```java
client.services().inNamespace("default").withName(svc.getMetadata().getName())
                    .edit(s -> new ServiceBuilder(svc).build());
```

shawkins: That's calling edit, but not using the value at all.  While not prohibited, that's not the expected usage pattern.  It could instead be:

```
client.services().inNamespace("default").withName(svc.getMetadata().getName()).patch(new ServiceBuilder(svc).build());
```

If you want to guard against your current usage style, yes we could move the modification code.

rohanKanojia: You're right. I'll update code in JKube to use patch instead. Thanks a lot for help!

PR #3230, opened 2021-06-09
Fix #2937: Add `SharedInformerFactory#getExistingSharedIndexInformers()` to return list of registered informers

### LINKED ISSUE #2937: Incorrect SharedInformerFactory#getExistingSharedIndexInformer
The caching key of informers has changed in the client 5.2. And as far as I can see, the lookup function is not coherent with the create functions. The function `getExistingSharedIndexInformer` is sensitive to the name I chose for the Java class of my custom resource.

For instance, if I have a custom resource under the API `com.acme/myapp:v1`, and that I have created the Java class `MyAppCustomResource` annotated to be a proper fabric8 `CustomResource`, in the following code `existingInformer` will be `null`, but I would expect that `existingInformer` is the same instance as `createdInformer`.

    SharedIndexInformer createdInformer = sharedInformerFactory.sharedIndexInformerForCustomResource(MyAppCustomResource.class, MyAppCustomResourceList.class)
    SharedIndexInformer existingInformer = sharedInformerFactory.getExistingSharedIndexInformer(MyAppCustomResource.class)

### LINKED ISSUE #3235: getExistingSharedIndexInformer may Incorrect
https://github.com/fabric8io/kubernetes-client/blob/b0dbeef2f6795934ea0a0c57b2d4ed3748aa0a18/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/SharedInformerFactory.java#L282

```java
 private static <T> boolean isKeyOfType(String key, Class<T> apiTypeClass) {
        String plural = HasMetadata.getPlural(apiTypeClass);
        return key.contains(plural);
    }
```

this may incorrect： return key.contains(plural); 
for example： virtualservice and service

rohanKanojia:
## Description
Fix #2937
Fix #3235

+ Add a new method `getExistingSharedIndexInformers()` to
  SharedInformerFactory which would return a list of entries containing
  registered SharedIndexInformer and their respective OperationContext.

  This is added to overcome limitation of old `getExistingSharedIndexInformer(Class<T> apiTypeClass)`
  in case there are two informers registered with same type. Based on @hibnico 's suggestion https://github.com/fabric8io/kubernetes-client/issues/2937#issuecomment-815836081

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/SharedInformerFactory.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/SharedInformerFactory.java
@@ -47,7 +47,10 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+
 import okhttp3.OkHttpClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -330,9 +333,11 @@ public class SharedInformerFactory extends BaseOperation {
     }
 
     if (!informerExecutor.isShutdown()) {
+      List<Map.Entry<SharedIndexInformer, Future<Boolean>>> startInformerTasks = new ArrayList<>();
       for (Map.Entry<OperationContext, SharedIndexInformer> entry : informers) {
-        startInformer(entry.getValue());
+        startInformerTasks.add(new AbstractMap.SimpleEntry<>(entry.getValue(), startInformerAsync(entry.getValue())));
       }
+      waitForAllInformersToStart(startInformerTasks);
     }
   }
 
@@ -362,11 +367,26 @@ public class SharedInformerFactory extends BaseOperation {
     this.eventListeners.add(event);
   }
 
-  private synchronized void startInformer(SharedIndexInformer informer) {
-    try {
+  private synchronized Future<Boolean> startInformerAsync(SharedIndexInformer informer) {
+    return informerExecutor.submit(() -> {
       informer.run();
-    } catch (RuntimeException e) {
-      this.eventListeners.forEach(listener -> listener.onException(informer, e));
+      return true;
+    });
+  }
+
+  private void waitForAllInformersToStart(List<Map.Entry<SharedIndexInformer, Future<Boolean>>> startInformerTasks) {
+    for (Map.Entry<SharedIndexInformer, Future<Boolean>> startedInformerTask : startInformerTasks) {
+      waitUntilInformerStartedAndHandleFailure(startedInformerTask);
+    }
+  }
+
+  private void waitUntilInformerStartedAndHandleFailure(Map.Entry<SharedIndexInformer, Future<Boolean>> startedInformerTask) {
+    try {
+      startedInformerTask.getValue().get();
+    } catch (InterruptedException interruptedException) {
+      Thread.currentThread().interrupt();
+    } catch (ExecutionException e) {
+      this.eventListeners.forEach(listener -> listener.onException(startedInformerTask.getKey(), e));
       log.warn("Informer start did not complete", e);
     }
   }
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/SharedInformerFactory.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/SharedInformerFactory.java
@@ -46,14 +46,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 
 import okhttp3.OkHttpClient;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * SharedInformerFactory class constructs and caches informers for api types.
@@ -62,7 +60,6 @@ import org.slf4j.LoggerFactory;
  * which is ported from offical go client https://github.com/kubernetes/client-go/blob/master/informers/factory.go
  */
 public class SharedInformerFactory extends BaseOperation {
-  private static final Logger log = LoggerFactory.getLogger(SharedInformerFactory.class);
   private final List<Map.Entry<OperationContext, SharedIndexInformer>> informers = new ArrayList<>();
 
   private final ExecutorService informerExecutor;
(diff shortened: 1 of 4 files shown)


oscerd (approved):

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/SharedInformerFactory.java
```diff
@@ -52,6 +56,7 @@
 public class SharedInformerFactory extends BaseOperation {
   private static final Logger log = LoggerFactory.getLogger(SharedInformerFactory.class);
   private final Map<String, SharedIndexInformer> informers = new HashMap<>();
```
shawkins, line 58: Can the Map of informers be removed?  A Map doesn't seem like the right construct since you can overwrite existing entries with additional creates and the lookup is a linear scan.  Since the SharedInformer now exposes getApiTypeClass the getExistingSharedIndexInformer should just look at that.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/SharedInformerFactory.java
```diff
@@ -325,12 +334,7 @@ public synchronized void stopAllRegisteredInformers(boolean shutDownThreadPool)
     if (informers.isEmpty()) {
       return;
     }
-    informers.forEach(
-      (informerType, informer) -> {
-        if (startedInformers.remove(informerType) != null) {
-          informer.stop();
-        }
-      });
+    informers.forEach(e -> stopInformer(e.getValue()));
```
shawkins, line 337: It would be simplest to just call stop on every informer.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/SharedInformerFactory.java
```diff
@@ -51,9 +55,9 @@
  */
 public class SharedInformerFactory extends BaseOperation {
   private static final Logger log = LoggerFactory.getLogger(SharedInformerFactory.class);
-  private final Map<String, SharedIndexInformer> informers = new HashMap<>();
+  private final List<Map.Entry<OperationContext, SharedIndexInformer>> informers = new ArrayList<>();
 
-  private final Map<String, Future> startedInformers = new HashMap<>();
+  private final List<Map.Entry<OperationContext, SharedIndexInformer>> startedInformers = new ArrayList<>();
```
shawkins, line 60: I think you'll need to leave the start logic as async.  The reason to track statedInformers is to cancel the futures - I think stop should acutally stop everything, rather than only those that have been started async.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/SharedInformerFactory.java
```diff
@@ -340,39 +337,16 @@ public void addSharedInformerEventListener(SharedInformerEventListener event) {
...
-
-  static String getInformerKey(OperationContext operationContext) {
-    StringBuilder keyBuilder = new StringBuilder();
-    if (operationContext.getApiGroupName() == null) {
-      keyBuilder.append(operationContext.getApiGroupVersion());
-    } else {
-      keyBuilder.append(operationContext.getApiGroupName()).append("/").append(operationContext.getApiGroupVersion());
+  private synchronized void startInformer(SharedIndexInformer informer) {
+    try {
+      informer.run();
```
shawkins, line 342: With other informer changes, run is a sync call.  It probably makes sense for the factory to start async though.  However if you think sync is better then that would remove the need for the listeners.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/SharedInformerFactory.java
```diff
@@ -296,16 +308,11 @@ public synchronized void startAllRegisteredInformers() {
...
-                } catch (RuntimeException e) {
-                  this.eventListeners.forEach(listener -> listener.onException(informer, e));
-                  log.warn("Informer start did not complete", e);
-                }
-              })));
+      List<Map.Entry<SharedIndexInformer, Future<Boolean>>> startInformerTasks = new ArrayList<>();
+      for (Map.Entry<OperationContext, SharedIndexInformer> entry : informers) {
+        startInformerTasks.add(new AbstractMap.SimpleEntry<>(entry.getValue(), startInformerAsync(entry.getValue())));
+      }
+      waitForAllInformersToStart(startInformerTasks);
```
shawkins, line 315: waiting doesn't match previous behavior, the other java client, nor the go client.  Perhaps returning a future would be the best compromise.  It would be easy to create a CompleteableFuture of all the starts.

shawkins (commented):
This could be done in a separate issue, but in looking at the other implementations I see that the drift here was adding support for the OperationContext - at which point things here weren't just class based and since there's no good comparison methods provided for the context it doesn't seem possible to treat it as fully shared.

That change probably should have been made as a "non" shared informer and left the shared factory logic just based upon class type.  

For consistency with the other dsl it would be straight-forward to not directly expose the OperationContext for this purpose, as you can base the logic off of WatchListDeletable (an existing interface that is both Watchable and Listable).  That is you would do something like:

```
SharedIndexInformer<Deployment> informer = kubernetesClient.apps().deployments().inAnyNamespace().withLabels(...).inform( resourceEventHandler, resync); // probably sync call to run
```

Does an approach like that make considering #3233 as well?

shawkins:
@rohanKanojia @manusa see #3239 for a dsl inform method.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/SharedInformerFactory.java
```diff
@@ -277,36 +276,37 @@ public Watch watch(ListOptions params, String namespace, OperationContext contex
...
+  }
+
+  /**
+   * Gets a list of entries of ({@link OperationContext}, {@link SharedIndexInformer}) registered
+   * by the user.
+   *
+   * @return a list of entries of {@link OperationContext} and {@link SharedIndexInformer}
+   */
+  public List<Map.Entry<OperationContext, SharedIndexInformer>> getExistingSharedIndexInformers() {
+    return this.informers;
```
manusa, line 294: Encapsulate list in `Collections.unmodifiableList(this.informers)`

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/SharedInformerFactory.java
```diff
@@ -277,36 +276,37 @@ public Watch watch(ListOptions params, String namespace, OperationContext contex
...
 
   /**
    * Starts all registered informers in an asynchronous fashion.
    * <br>use {@link #addSharedInformerEventListener(SharedInformerEventListener)} to receive startup errors. 
    */
-  public synchronized void startAllRegisteredInformers() {
-    if (informers.isEmpty()) {
-      return;
-    }
+  public synchronized List<Map.Entry<SharedIndexInformer, CompletableFuture<Boolean>>> startAllRegisteredInformers() {
```
shawkins, line 301: To encapsulate things could this just be ```Future<Void> startAllRegisteredInformers()```

With 

```return CompletableFuture.allOf(startInformerTasks.values().toArray(new CompletableFuture[] {}));```

File: kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/DefaultSharedIndexInformerTest.java [hunk 8]
```diff
@@ -384,7 +386,12 @@ public void onDelete(Pod oldObj, boolean deletedFinalStateUnknown) { }
 
     // When
     factory.addSharedInformerEventListener(exception -> failureCallbackReceived.countDown());
-    factory.startAllRegisteredInformers();
+    Future<Void> informerStartFutures = factory.startAllRegisteredInformers();
+    try {
+      informerStartFutures.get();
+    } catch (Exception e) {
+      e.printStackTrace();
```
shawkins, line 393: Could this be a timed get and no catch?

shawkins (approved):
Added a couple of nits, but otherwise seems good.  I've opened a WIP pr showing how we could encapsulate the usage of OperationContext for informers - which may mean we could deprecate or revert their direct usage for this factory.

rohanKanojia:
With this change our `addSharedInformerEventListener` doesn't work anymore. Now user can directly handle exceptions on calling `future.get()` returned from `startAllRegisteredInformers()`.

Shall I deprecate `addSharedInformerEventListener` ?

shawkins:
> With this change our addSharedInformerEventListener doesn't work anymore. Now user can directly handle exceptions on calling future.get() returned from startAllRegisteredInformers().

It should still work, correct?  You're just saying it may not be needed.  

> Shall I deprecate addSharedInformerEventListener ?

The issue would be making sense of the exception - if there are multiple start exceptions, do you know which informers have failed to start?  If on the other hand we just want users to know whether all started and that they'll use the dsl or other mechanism to start if they care about which may fail, then I'm fine with deprecating.

rohanKanojia:
Sorry, You're right. I think I did a mistake while testing yesterday. It's working. Now it's just that failure can be found out by `startedFuture.get()` too. 

You're right `addSharedInformerEventListener` would still make sense in case there are multiple informer failures. Let's keep it.

manusa (approved):
LGTM, thx!

manusa:
Relates to: #2010

PR #3233, opened 2021-06-09
Fix #3228: Add support for Dynamic informers for custom resources in KubernetesClient

rohanKanojia:
## Description
Fix #3228 
Introduce a new method in SharedInformerFactory which will only accept
CustomResourceDefinitionContext, types would automatically be assumed
to be GenericKubernetesResource and GenericKubernetesResourceList.

Till now we were deciding whether a resource is namespaced or not
depending upon whether it implements Namespaced interface or not. But
having dynamic informers required change to consider `scope` field in
CustomResourceDefinitionContext as well, hence I introduced a new class
SharedInformerOperationsImpl which overrides `isResourceNamespaced()` so
that it can be configured from calling methods.

## Type of change

 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [X] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [ ] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [X] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [X] I Added [CHANGELOG](../CHANGELOG.md) entry regarding this change
 - [X] I have implemented unit tests to cover my changes
 - [X] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/master/doc/CHEATSHEET.md) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [X] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

PR #3239, opened 2021-06-12
Adding a dsl inform method

shawkins:
## Description
A possible implementation for the informer mentioned in https://github.com/fabric8io/kubernetes-client/pull/3230#pullrequestreview-682321798

This implementation is just a simple wrapper over a Watch - no additional threads are consumed.

It is not a SharedIndexInformer because of the simplification that it is already started, which means that Indexers cannot be added.  Since I haven't needed to use Indexers I'm probably biasing this toward simplicity.

For generic functionality I've added a second commit to expand the existing RawCustomResourceOperationsImpl - it's a little cumbersome.  Narrowing to a more specific interface than the SharedInformer (run isn't needed, etc.) would help a little. The pro is that this doesn't require yet exposing the GenericKubernetesResource for #3233 - but as mentioned above it's not quite functionally equivalent since the factory supports indexing and additional sharing.

If this seems sensible, then the SharedInformerFactory methods dealing with an OperationContext could be deprecated.  

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/KubernetesClient.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/KubernetesClient.java
@@ -161,12 +161,12 @@ public interface KubernetesClient extends Client {
   <T extends HasMetadata, L extends KubernetesResourceList<T>> MixedOperation<T, L, Resource<T>> customResources(CustomResourceDefinitionContext crdContext, Class<T> resourceType, Class<L> listClass);
   
   /**
-   * Semi-Typed API for managing CustomResources.
+   * Semi-Typed API for managing {@link GenericKubernetesResource}s which can represent any resource.
    *
-   * @param crdContext CustomResourceDefinitionContext describes the core fields used to search for CustomResources
-   * @return returns a MixedOperation object with which you can do basic CustomResource operations
+   * @param crdContext CustomResourceDefinitionContext describes the core fields
+   * @return returns a MixedOperation object with which you can do basic operations
    */
-  default MixedOperation<GenericKubernetesResource, GenericKubernetesResourceList, Resource<GenericKubernetesResource>> genericCustomResources(
+  default MixedOperation<GenericKubernetesResource, GenericKubernetesResourceList, Resource<GenericKubernetesResource>> genericKubernetesResources(
       CustomResourceDefinitionContext crdContext) {
     return customResources(crdContext, GenericKubernetesResource.class, GenericKubernetesResourceList.class);
   }
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/KubernetesClient.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/KubernetesClient.java
@@ -27,6 +27,8 @@ import io.fabric8.kubernetes.api.model.ComponentStatus;
 import io.fabric8.kubernetes.api.model.ComponentStatusList;
 import io.fabric8.kubernetes.api.model.Endpoints;
 import io.fabric8.kubernetes.api.model.EndpointsList;
+import io.fabric8.kubernetes.api.model.GenericKubernetesResource;
+import io.fabric8.kubernetes.api.model.GenericKubernetesResourceList;
 import io.fabric8.kubernetes.api.model.LimitRange;
 import io.fabric8.kubernetes.api.model.LimitRangeList;
 import io.fabric8.kubernetes.api.model.Namespace;
@@ -157,6 +159,17 @@ public interface KubernetesClient extends Client {
    */
   @Deprecated
   <T extends HasMetadata, L extends KubernetesResourceList<T>> MixedOperation<T, L, Resource<T>> customResources(CustomResourceDefinitionContext crdContext, Class<T> resourceType, Class<L> listClass);
+  
+  /**
+   * Semi-Typed API for managing CustomResources.
+   *
+   * @param crdContext CustomResourceDefinitionContext describes the core fields used to search for CustomResources
+   * @return returns a MixedOperation object with which you can do basic CustomResource operations
+   */
+  default MixedOperation<GenericKubernetesResource, GenericKubernetesResourceList, Resource<GenericKubernetesResource>> genericCustomResources(
+      CustomResourceDefinitionContext crdContext) {
+    return customResources(crdContext, GenericKubernetesResource.class, GenericKubernetesResourceList.class);
+  }
 
   /**
    * Discovery API entrypoint for APIGroup discovery.k8s.io
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Informable.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Informable.java
@@ -1,20 +1,67 @@
+/**
+ * Copyright (C) 2015 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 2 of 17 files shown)


shawkins:
It would also probably make sense if SharedInformerFactory were not a BaseOperation - that opens it up to a lot of methods, and it means it both has a and is a BaseOperation.

shawkins:
@rohanKanojia @manusa this does not yet make any changes to the SharedInformerFactory - we'd want Rohan's pending changes to go in then add appropriate deprecations.

The proposed solution looks like:

```
SharedIndexInformer<Pod> informer = client.pods().withLabel("my-label").inform(handler);

// some time later

informer.stop();
```

It tightens up how handlers are added to ensure that it's not possible to miss an event because the add events are sent while locked.

The proposed generic solution is:

```
SharedIndexInformer<GenericKubernetesResource> informer =
        client.genericCustomResources(context).withLabel("my-label").inform(handler);
```

This requires a new client method genericCustomResources - but it should have the same logic that Rohan added to check for an existing handler otherwise deserialized values will have actual types and not be generic - which also means that you can never register the given type with the deserializer while the informer is running.

The rationale to use this is for situations where sharing the informer is not a primary concern, or you track the informer yourself.  

Scenarios that watch for all of given resource / class would still be handled by the existing factory logic.

shawkins:
@rohanKanojia now that all of your changes are in, I'll refine this more to propose deprecations on the informer factory.

shawkins:
@rohanKanojia I think this is up-to-date now.  SharedInformerFactory is no longer a BaseOperation, public methods dealing with the OperationContext, withName, and withNamespace are deprecated.  I also removed the validation on whether there is a handler because we don't do that check for the other custom resource / raw logic - instead there is a check in the ReflectorWatcher to convert the WatchEvent value to the expected type.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Informable.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.function.Function;
+
+public interface Informable<T> {
+  
+  /**
+   * The indexers to add to {@link SharedInformer}s created by subsequent inform calls;
+   * @param indexers to customize the indexing
+   * @return the current {@link Informable}
+   */
+  public Informable<T> withIndexers(Map<String, Function<T, List<String>>> indexers);
```
rohanKanojia, line 35:
```suggestion
  Informable<T> withIndexers(Map<String, Function<T, List<String>>> indexers);
```

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Informable.java
```diff
@@ -0,0 +1,67 @@
...
+   * <p>This call will be blocking for the initial list and watch.
+   * <p>You are expected to call stop to terminate the underlying Watch.
+   * <p>Additional handlers can be added, but processing of the events will be in the websocket thread, 
+   * so consider non-blocking handleroperations for more than one handler.
+   * 
+   * @param handler to notify
+   * @param resync the resync period or 0 for no resync
+   * @return a running {@link SharedIndexInformer}
+   */
+  public SharedIndexInformer<T> inform(ResourceEventHandler<T> handler, long resync);
```
rohanKanojia, line 65:
```suggestion
  SharedIndexInformer<T> inform(ResourceEventHandler<T> handler, long resync);
```

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/SharedInformerFactory.java
```diff
@@ -84,21 +104,22 @@ public SharedInformerFactory(OkHttpClient okHttpClient, Config configuration) {
...
   /**
    * Configure Namespace for {@link SharedInformerFactory}
    *
    * @param namespace namespace to configure
    * @return {@link SharedInformerFactory} with namespace configured
+   * @deprecated use {@link Informable} instead
    */
-  @Override
+  @Deprecated
   public SharedInformerFactory inNamespace(String namespace) {
```
rohanKanojia, line 121: Umm, why are we deprecating this? I think using namespaced informers with SharedInformerFactory can be a very common use case. Even client-go's informer factory provides Namespaced informers via [`NewFilteredSharedInformerFactory`/`NewSharedInformerFactoryWithOptions`](https://github.com/kubernetes/client-go/blob/master/informers/factory.go#L109)

kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/InformTest.java
```diff
@@ -0,0 +1,231 @@
...
+class InformTest {
+
+  private static final Long EVENT_WAIT_PERIOD_MS = 10L;
+
+  KubernetesMockServer server;
+  private KubernetesClient client;
+
+  @BeforeEach
+  void setUp() {
+    client = server.createClient().inNamespace("test");
```
rohanKanojia, line 53: I'm not sure if this is necessary since by default namespace is set to `test` in mockwebserver's configuration

shawkins: This just came from the watch test.  I can remove it.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/SharedInformerFactory.java
shawkins: It behaves differently than the withName and inNamespace dsl methods - it modifies the current context rather than creating a new one.  I also find this behavior confusing wrt an informer factory - if you can set this post construction, then some informers can end watching all namesapces, while others would be in a specific namespace.  Also a inAnyNamespace is not implemented.  At least to me it seems like these methods are problematic.

rohanKanojia: I agree mutating context is not good. We should think about providing similar alternatives to the above-mentioned client-go methods in the future.

shawkins: Direct alternatives would be setName and setNamespace - but I still wonder if it may not be better to offer either:

Informable.informers() - where the factory is off of the current context.

or 

KubernetesClient.informers(InformerOptions) - with options similar to the go client.

Neither of those require additional methods on the factory.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/KubernetesClient.java
```diff
@@ -157,6 +159,17 @@
...
   @Deprecated
   <T extends HasMetadata, L extends KubernetesResourceList<T>> MixedOperation<T, L, Resource<T>> customResources(CustomResourceDefinitionContext crdContext, Class<T> resourceType, Class<L> listClass);
+  
+  /**
+   * Semi-Typed API for managing CustomResources.
+   *
+   * @param crdContext CustomResourceDefinitionContext describes the core fields used to search for CustomResources
+   * @return returns a MixedOperation object with which you can do basic CustomResource operations
+   */
+  default MixedOperation<GenericKubernetesResource, GenericKubernetesResourceList, Resource<GenericKubernetesResource>> genericCustomResources(
```
rohanKanojia, line 169: I think this method can also be used for regular Kubernetes resources. It seems to address (https://github.com/fabric8io/kubernetes-client/issues/2929) . Do you think `genericKubernetesResources()` or `dynamic()` would be better?

shawkins: > Do you think genericKubernetesResources() or dynamic() would be better?

I wasn't sure what to name it.  Definitely if its usage with regular resources isn't discouraged, then the names you are suggesting make sense.  Generic is consistent with the resource class names, but is more overloaded as a term.  I'm fine either way - @manusa WDYT?

shawkins: I'll update to genericKubernetesResources

shawkins:
@manusa @rohanKanojia I believe this implements the changes we've discussed.  The generic method name has been changes and Informable will work off of Resource as well.

I think the only behavior that people may not like is that the initial list operation will result in add events that are processed by the thread calling inform - this may not be desirable for some usage with a lot of resources.  The only thing addressing that is the javadocs suggesting that you use a non-blocking resource handler.  The alternative is to allow the users control over when run is called.

manusa:
Relates to: #2010

manusa (approved):
LGTM, thx!

rohanKanojia (approved):

PR #3274, opened 2021-06-25
fix #3271: refining the wait logic

### LINKED ISSUE #3271: waitUntilCondition resource too old exception
I tried to do a waitUntilCondition call, that eventually errorred out with a resource too old exception.  This seems to come from WatcherException isShouldRetry - it explicitly is false when there's an http gone exception.

It should either be the other way around, or to retry no matter what the exception is.

shawkins:
## Description
To address #3271 the only time to retry is on http gone - the watch framework will already handle all of the other retries.
delete does not need special handling in the watcher

## Type of change

## Checklist


oscerd (approved):

shawkins:
Like most things this turned out to be bigger than I thought at first glance...  It makes sense to me to switch the implementation to be based upon an informer instead - as that already has the necessary retry logic and it's simple to express.

With this change it is unnecessary to specifically configure an exponential backoff for this operation - it should use the defaults for watches / informers.  I did not remove those fields from all of the relevant contexts as that greatly expands the scope of these changes.

This change also reuses the existing KubernetesClientTimeoutException instead of using an IllegalArgumentException for timed out wait calls - I'll note that as a breaking change.

This also simplifies the logic in NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl - it was somewhat idiosyncratic and should be using the fork join pool for the potentially long running waitUntilCondition tasks.  At some point the resource handlers and underlying logic should exposes the underlying futures as that would prevent this needless usage of threads.

I commented out ServiceTest.testWaitUntilReady as it doesn't appear that a service is a readiable resource.

This addresses having duplicate metadata.name fields in the watch url, but the real problem is #3275

manusa (approved):
LGTM, thx!

shawkins:
As mentioned on the issue, the subsequent commit removes the declared interruptedexception to simplify internal code and for consistency with the rest of the api.  @manusa if that doesn't seem reasonable I can pull that out.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.java
```diff
@@ -85,116 +83,64 @@
...
-      if (latch.getCount() == 0) {
-        return result;
-      } else {
-        throw new KubernetesClientTimeoutException(itemsWithConditionNotMatched, amount, timeUnit);
-      }
-    } finally {
-      executor.shutdown();
-    }
+  public List<HasMetadata> waitUntilReady(final long amount, final TimeUnit timeUnit) {
+    return waitUntilCondition(resource -> Objects.nonNull(resource) && getReadiness().isReady(resource), amount, timeUnit);
```
manusa, line 87: The statement
```java
getReadiness().isReady(resource)
```
is not equivalent to
```java
handlerOf(resource).waitUntilReady(...)
```

However the former doesn't make much sense since the effective call of waitUntilRady should be that of the applicable *OperationsImpl (e.g. ServiceHandler->ServiceOperationsImpl), especially considering the ServiceTest. Are we sure the Endpoints are being queried after these changes?

manusa: https://github.com/fabric8io/mockwebserver/issues/62

rohanKanojia (approved):

PR #3277, opened 2021-06-27
fix #3275: filterable should not modify the current context

### LINKED ISSUE #3275: Filterable modifies the current context
All of the Filterable methods modify then return the current context.  Should they be creating new instances?

cc @rohanKanojia @manusa

shawkins:
## Description
This addresses #3275 - the Filterable with methods modify the context objects directly such that the current value and any other contexts previously created are modified.

This also introduces a builder for near backwards compatibility:

```
    FilterNested<FilterWatchListDeletable<Pod, PodList>> filter = pods().inNamespace(namespace).withNewFilter();
    if (selector.getMatchLabels() != null) {
      filter.withLabels(selector.getMatchLabels());
    }
...
    return filter.endFilter().list();
```
Rather than:
```
    FilterWatchListDeletable<Pod, PodList> podLister = pods().inNamespace(namespace);
    if (selector.getMatchLabels() != null) {
      podLister.withLabels(selector.getMatchLabels());
    }
...
    return podLister.list();
```
If that level of backwards compatibility is not needed, the builder construct can be removed or left unexposed and the user can instead do:
```
    FilterWatchListDeletable<Pod, PodList> podLister = pods().inNamespace(namespace);
    if (selector.getMatchLabels() != null) {
      podLister = podLister.withLabels(selector.getMatchLabels());
    }
...
    return podLister.list();
```

@manusa @rohanKanojia let me know what seems to be the right approach here and I'll make the necessary changes and add a changelog entry.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/FilterBuilderImpl.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/FilterBuilderImpl.java
@@ -1,0 +1,203 @@
+/**
+ * Copyright (C) 2015 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.fabric8.kubernetes.client.dsl.base;
+
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.fabric8.kubernetes.api.model.KubernetesResourceList;
+import io.fabric8.kubernetes.api.model.LabelSelector;
+import io.fabric8.kubernetes.api.model.LabelSelectorRequirement;
+import io.fabric8.kubernetes.api.model.ObjectReference;
+import io.fabric8.kubernetes.client.dsl.FilterBuilder;
+import io.fabric8.kubernetes.client.dsl.FilterWatchListDeletable;
+import io.fabric8.kubernetes.client.dsl.Resource;
+import io.fabric8.kubernetes.client.utils.Utils;
+
+import java.lang.reflect.Array;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+final class FilterBuilderImpl<T extends HasMetadata, L extends KubernetesResourceList<T>, R extends Resource<T>> implements FilterBuilder<FilterWatchListDeletable<T, L>> {
+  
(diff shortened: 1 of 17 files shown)


shawkins:
https://github.com/fabric8io/kubernetes-client/pull/3277/files#diff-b63603660c0c2488b6e1590fe167246b669504b3e219252a7cca56fc46f13448R665 highlights this issue as well - the only logic was apply the selector labels from the Deployment - and they were picked up still even in later calls.  With these changes now we need the selector lablels on the ReplicaSet as well for the test to pass (since there is no mock for getting all pods).

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/FilterBuilderImpl.java
```diff
@@ -0,0 +1,203 @@
...
+  private static final String INVOLVED_OBJECT_UID = "involvedObject.uid";
+  private static final String INVOLVED_OBJECT_RESOURCE_VERSION = "involvedObject.resourceVersion";
+  private static final String INVOLVED_OBJECT_API_VERSION = "involvedObject.apiVersion";
+  private static final String INVOLVED_OBJECT_FIELD_PATH = "involvedObject.fieldPath";
+  
+  private final BaseOperation<T, L, R> baseOperation;
+  private OperationContext context;
+
+  /**
+   * @param baseOperation
```
rohanKanojia, line 48: nit, either complete this javadoc or remove it.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/FilterBuilderImpl.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+    if (objectReference.getFieldPath() != null) {
+      context.fields.put(INVOLVED_OBJECT_FIELD_PATH, objectReference.getFieldPath());
+    }
+    return this;
+  }
+
+  @Override
+  public FilterWatchListDeletable<T, L> build() {
+    return this.baseOperation.newInstance(context);
```
rohanKanojia, line 201: Umm, In my opinion this can be a bit confusing:
```java
client.apps().deployments().withFilter().withLabel("key1", "value1").build().list();
```
I think maybe if we name this to something like this it might be better:
```java
client.apps().deployments().withNewFilter().withLabel("key1", "value1").endFilter().list();
```

manusa: Yes, this seems also more consistent with the rest of builders generated by sundr.io

shawkins: Right, I should have gone with nested rather than a builder.  I've committed a change to nested.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/FilterBuilderImpl.java
shawkins: No problem, I should probably turn that off in the ide...

manusa (approved):
LGTM, thx!

PR #3031, opened 2021-04-21
Fix #3016: Use scheduleWithFixedDelay for resync task execution

rohanKanojia:
+ When using `resyncExecutor.scheduleAtFixedRate` we're seeing inconsistent
  resync intervals since it doesn't wait for previous task to be
  terminated. Hence using `resyncExecutor.scheduleWithFixedDelay` instead.
+ Remove unused reflectExecutor from Controller and Reflector since we
  don't use scheduled reflector relist since https://github.com/fabric8io/kubernetes-client/pull/2964

## Type of change

 - [X] Bug fix (non-breaking change which fixes an issue)
 - [ ] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [ ] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [X] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [X] I Added [CHANGELOG](../CHANGELOG.md) entry regarding this change
 - [X] I have implemented unit tests to cover my changes
 - [ ] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/master/doc/CHEATSHEET.md) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [X] I tested my code in Kubernetes
 - [X] I tested my code in OpenShift

Issue #3016, opened 2021-04-20
should resync be executed in resyncExecutor.scheduleWithFixedDelay instead of resyncExecutor.scheduleAtFixedRate  ?


akram:
The `resync` task in `kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/cache/Controller.java` is currently submitted into the `resyncExecutor` with a `scheduleAtFixedRate` .

The `scheduleAtFixedRate` of `ServiceExecutor` does not wait for a previous task to be terminated to run the next one. 
If the resync tasks takes too long, that may make the Controller to restart the resync quickly. In recent kubernetes versions the requests can then be throttled which may cascade other issues.

By replacing it with `scheduleWithFixedDelay`, we can be sure too have a reasonnable delay between syncs.

rohanKanojia:
Is this related to  https://github.com/fabric8io/kubernetes-client/issues/2977?

rohanKanojia:
Thanks a lot for investigating this issue. As you pointed out using scheduleWithFixedDelay seems to be producing correct results.

behavior with scheduleAtFixedRate. For a resync period of 30 seconds, I see resync happening sometimes after 30 seconds and sometimes after 60 seconds
```
kubernetes-examples : $ mvn exec:java -Dexec.mainClass="io.fabric8.kubernetes.examples.SharedInformerExample"ample"

[INFO] Scanning for projects...
[INFO] Inspecting build with total of 1 modules...
[INFO] Installing Nexus Staging features:
[INFO]   ... total of 1 executions of maven-deploy-plugin replaced with nexus-staging-maven-plugin
[INFO] 
[INFO] -------------------< io.fabric8:kubernetes-examples >-------------------
[INFO] Building Fabric8 :: Kubernetes :: Client Examples 5.4-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- exec-maven-plugin:3.0.0:java (default-cli) @ kubernetes-examples ---
[io.fabric8.kubernetes.examples.SharedInformerExample.main()] INFO :: - Informer factory initialized.
[io.fabric8.kubernetes.examples.SharedInformerExample.main()] INFO :: - Starting all registered informers
[informer-controller-Pod] INFO io.fabric8.kubernetes.client.informers.cache.Controller - informer#Controller: ready to run resync and reflector runnable
[...]

PR #3286, opened 2021-06-30
fix #3285: adding waiting for list contexts

### LINKED ISSUE #3285: Implement waiting for lists
Related to #3278 it would be good if wait methods were available off of the list context.

I propose:

```List<T> waitUntilListCondition(Predicate<List<T>> condition, long amount, TimeUnit timeUnit);```

In addition it would be good if there were versions of the wait methods that also returned futures so that dependent code, like the NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl did not need to create threads that will simply wait.

### PR DESCRIPTION
**Pull request by shawkins**
## Description
Adds waiting methods for list contexts as per #3285.  Also adds methods to directly return wait futures - to expose these the
resource handler generation was simplified.

## Type of change

## Checklist

### CODE REVIEW

**Comment by shawkins** [comment]
After implementing the logic in NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl to directly use futures exposed by waitalbe methods, I decided that does not seem worth it yet.  While it does same some threads, we are still creating 2 threads per watch.  The cleanup to the resourcehandler generation seems valid, so I'm leaving that in.

The function is now proposed as:
```
  /**
   * Wait for the entire list available at this context to satisfy the given predicate.
   * @param condition the {@link Predicate} to test
   * @param pollInterval the polling interval to use, a value less than 1 indicates that every event should be tested
   * @param timeout after which a {@link KubernetesClientTimeoutException} is thrown
   * @param timeUnit unit of the pollInterval and timeout
   * @return the list of items after the condition is met
   * @throws KubernetesClientTimeoutException if time runs out before the condition is met
   */
  List<T> waitUntilListCondition(Predicate<List<T>> condition, long pollInterval, long timeout, TimeUnit timeUnit);
```

Such that you can test at an interval or test on every event - under the covers it is trivial to filter the interval tests to only when the resourceVersion has changed, but I've left that out for now. 

The usage looks like https://github.com/fabric8io/kubernetes-client/pull/3286/files#diff-e3b6fec210f5b68dea994bde82d20006708a6e34d5e64f7f62211290dde44f0fR216

The only other thought would be if anyone doesn't want this method on WatchListDeletable, something similar could be added directly to SharedInformer.

After a preliminary review by @manusa or @rohanKanojia, I'll add a changelog and more tests.

**Comment by shawkins** [comment]
Separated the template changes as #3301

File: kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/apps/v1/RollingUpdater.java [hunk 1]
```diff
@@ -207,40 +206,27 @@ public T rollUpdate(T oldObj, T newObj) {
...
-    } catch (InterruptedException e) {
-      Thread.currentThread().interrupt();
-      poller.cancel(true);
-      logger.cancel(true);
-      executor.shutdown();
-      LOG.warn("Only {}/{} pod(s) ready for {}: {} in namespace: {}  after waiting for {} seconds so giving up",
-          podCount.get(), requiredPodCount, obj.getKind(), obj.getMetadata().getName(), namespace, TimeUnit.MILLISECONDS.toSeconds(rollingTimeoutMillis));
+        podCount.set(count);
+        return count == requiredPodCount;
+      }, 1000, rollingTimeoutMillis, TimeUnit.MILLISECONDS);
```
rohanKanojia, line 227: why is `pollingInterval` hardcoded to 1000 here?

shawkins: That's what it was previously.

shawkins:
It may be worth mentioning that you could also just make this a method of the Informable interface rather than something that is directly on WatchListDeletable.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/WatchListDeletable.java
```diff
@@ -16,13 +16,30 @@
...
+  /**
+   * Wait for the entire list available at this context to satisfy the given predicate.
+   * @param condition the {@link Predicate} to test
+   * @param pollInterval the polling interval to use, a value less than 1 indicates that every event should be tested
+   * @param timeout after which a {@link KubernetesClientTimeoutException} is thrown
+   * @param timeUnit unit of the pollInterval and timeout
+   * @return the list of items after the condition is met
+   * @throws KubernetesClientTimeoutException if time runs out before the condition is met
+   */
+  List<T> waitUntilListCondition(Predicate<List<T>> condition, long pollInterval, long timeout, TimeUnit timeUnit);
```
manusa, line 43: should we provide a
```java
default List<T> waitUntilListCondition(Predicate<List<T>> condition) {
  return waitUntilListCondition(condition, 0, TimeUnit.SECONDS);
}
```
for a non-polling default method?

shawkins: Okay, I'll add:
```
default List<T> waitUntilListCondition(Predicate<List<T>> condition, long timeout, TimeUnit timeUnit) {
  return waitUntilListCondition(condition, 0, timeout, TimeUnit.SECONDS);
}
```

And move the methods onto Informable.

metacosm: I'm not a big fan of multiplying method overrides when they don't provide an obvious value. This leads to higher cognitive load and potential confusion in my opinion.

manusa: Well, my suggestion is based on usability. For the average user I don't see the added value of being able to specify a polling interval.

Right now, I wouldn't be able to say what's the recommended approach for any given scenario. I suspect that we'll end up getting questions about this.

I think that the reason to add this variable is based on the underlying implementation and the usage or not of informers. So in the end, I guess the question boils down to the question that given informers+cache should provide the exact same behavior, why should we want to use a polling mechanism instead?

shawkins: > So in the end, I guess the subject boils down to the question that given informers+cache should provide the exact same behavior, why should we want to use a polling mechanism instead?

The polling approach was to match the existing logic of the RollingUpdater - it would of course work without polling as well.  The other thoughts as to why a polling option made sense are for cases where a resource is rapidly changing (and therefore testing on every event may not be desired) or where you want the test to have a temporal component - for example, test every 500 ms, but every 5 seconds emit a log or some other more involved action.  We have a scenario where while waiting for a resource to be ready we periodically check for unscheduled pods.

Another view point is that the polling is like a resync, but at a list level.  If that and other handling were added to the ResourceEventHandler - onRefresh(List<T>), onResync(List<T>) - then we don't necessarily need a dsl method as this scenario could be implemented directly on the Informer.

```
Predicate<List<Pod>> predicate = ...
CompletableFuture<Void> future = ...
SharedInformer<Pod> informer = ...pods().inform(null, 1000);
informer.addEventHandler(new ResourceEventHandler<HasMetadata>() {
    public void onRefresh(List<HasMetadata> objs) {
      test(objs);
    }

    public void onResync(List<HasMetadata> objs) {
      test(objs);
    }
   
    public void onAdd(HasMetadata obj) {
      // do something if I want to test every add
      // requires access to the Informer Cache
    }
  
    ...

    void test(List<HasMetadata> objs) {
      // test the predicate, complete the future
    }
});

``` 
A built-in implementation of this ResourceEventHandler could be supplied - that would be somewhat similar to the ReadinessWatcher (although that looks like it's no longer needed).

manusa: So if it does make sense to have both options, IMHO it is clearer to have the 2 methods, one that implies the non-polling behavior and another one where it's actually expected.
For me it's confusing that a "magic" 0 will disable polling and assume you want the async behavior instead.
Anyway, we can discuss this point in today's meeting.

shawkins:
I've made a couple of changes/simplifications.

1. The ResourceEventHandler now has a method onNothing to be notified when there is nothing.  This was the simplest / most consistent change I could think of to put this behavior on ResourceEventHandler.   Alternatives would include adding methods such as onList - and even possibly delegating the behavior to the event handler, but that gets complicated.

2.  Changed the method signature to Informable.informOnCondition(Predicate<List<T>) - and it returns a CompletableFuture.  That provides the flexibility setup multiple of these conditions and then wait across all of them rather than trying to force everything into a single one and a poll/resync for alternative behavior.  For callers on Java 9+ this will look like:

```
    client.pods()
        .inNamespace(namespace)
        .withLabel("test", "WaitUntilReadyIT")
        .informOnCondition(Collection::isEmpty).orTimeout(60, TimeUnit.SECONDS).get();
```

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/apps/v1/RollingUpdater.java
```diff
@@ -222,25 +220,22 @@ private void waitUntilPodsAreReady(final T obj, final String namespace, final in
...
-    };
-
-    ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
-    ScheduledFuture poller = executor.scheduleWithFixedDelay(readyPodsPoller, 0, 1, TimeUnit.SECONDS);
-    ScheduledFuture logger = executor.scheduleWithFixedDelay(() -> LOG.debug("Only {}/{} pod(s) ready for {}: {} in namespace: {} seconds so waiting...",
-        podCount.get(), requiredPodCount, obj.getKind(), obj.getMetadata().getName(), namespace), 0, loggingIntervalMillis, TimeUnit.MILLISECONDS);
+      return count == requiredPodCount;
+    });
+
+    ScheduledFuture<?> logger = Utils.scheduleAtFixedRate(Utils.getCommonExecutorSerive(),
```
manusa, line 226: ` Th `Utils#scheduleAtFixedRate` is really misleading. We should rename it to match the delegated`CachedSingleThreadScheduler#scheduleWithFixedDelay` method name.

manusa (approved):
LGTM, thx!

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/apps/v1/RollingUpdater.java
shawkins: It's a little confusing, but it actually is scheduleAtFixedRate.

The logic is handing the task off to another executor, so unless it's an executor that executes in the current thread, it means that fixed delay is actually a fixed rate.

manusa: OK, makes sense, I missed the `executor.execute(...)` part of the method.

manusa: Related to #3016 -> #3031

metacosm (approved):

PR #3299, opened 2021-07-06
Fix #3240: MicroTime serialises incorrectly; Add custom serializer/deserializer for MicroTime class

### LINKED ISSUE #3240: MicroTime serialises incorrectly
As per my investigation recorded on [this ticket here](https://github.com/fabric8io/kubernetes-client/issues/3178#issuecomment-856621169),  `io.fabric8.kubernetes.api.model.MicroTime` serialises incorrectly. 

When creating an `Event`, fabric8 serialises `MicroTime`  as an `object`  but the k8s server is expecting it as a `string` (in RFC-3339 format). 

Looking into how the generated models are derived (Go struct public members reflected -> JSON schema -> jsonschema2pojo) I can see how this arose, MicroTime is indeed a struct with a field called `Time`, but it serialises to/from a `string` [on the server](https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/micro_time.go#L167).

I have a workaround, which is to subclass MicroTime (thank you so much for not making it `final`!) and override the JSON serialisation logic (I won't embed it inline for the purposes of conciseness)

https://gist.github.com/LiamClarkeNZ/65f0888d2bcb1dfe9807c873ff4f18cb

### Who would fix this?

I'd be very happy to submit a PR once I understand what a good solution looks like from the maintainers' point of view.

### The question is, what does that good solution look like?

The current Go struct -> JSON schema -> pojo seems to work ideally for 99.9% of cases. `v1.MicroTime` looks to be a rather special case. So I can see several options.

1. Modify the Go struct -> JSON schema step to take advantage of Go types exposing `OpenAPISchemaType()` and `OpenAPISchemaFormat()` to ensure that the Go type and the serialised type align.
2. Special case MicroTime - exclude it from generation, and rely on a hand-crafted copy. 

Special casing always feels like a bad choice, but if this struct/serialisation representation mismatch only affects one type, it might be the easiest approach. It also offers the advantage that a hand-crafted model could make it easier in converting a Java date-time representation to a K8s one. 

E.g., in my Gist, the workaround requires a `ZonedDateTime` (as one is needed to serialise to a format with TZ) and formats the string to ensure that K8s can deserialise it, removing some burden on end users.

TL;DR - MicroTime serialisation is currently not working, I'm very happy to submit a PR to fix it, I just want to find out what maintainers think is the best way to do so :)

rohanKanojia:
## Description

Fix #3240 

Add Custom MicroTime with custom Serializer/Deserializers in order to
convert to a timestamp string 

## Type of change

## Checklist
 - [X] I tested my code in OpenShift

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Fix #3295: Fix wrong kind getting registered in KubernetesDeserializer in SharedInformerFactory
 * Fix #3318: Informer relist add/update should not always be sync events
 * Fix #3328: Allow for generic watches of known types
+* Fix #3240: MicroTime serialises incorrectly; add custom serializer/deserializer for MicroTime
 
 #### Improvements
 * Fix #3284: refined how builders are obtained / used by HasMetadataOperation
--- a/kubernetes-itests/src/test/java/io/fabric8/kubernetes/EventsIT.java
+++ b/kubernetes-itests/src/test/java/io/fabric8/kubernetes/EventsIT.java
@@ -18,7 +18,7 @@ package io.fabric8.kubernetes;
 import io.fabric8.kubernetes.api.model.Event;
 import io.fabric8.kubernetes.api.model.EventBuilder;
 import io.fabric8.kubernetes.api.model.EventList;
-import io.fabric8.kubernetes.api.model.ObjectReference;
+import io.fabric8.kubernetes.api.model.MicroTimeBuilder;
 import io.fabric8.kubernetes.api.model.ObjectReferenceBuilder;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import org.arquillian.cube.kubernetes.api.Session;
@@ -27,6 +27,9 @@ import org.arquillian.cube.requirement.ArquillianConditionalRunner;
 import org.jboss.arquillian.test.api.ArquillianResource;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+
+import java.time.OffsetDateTime;
+import java.time.format.DateTimeFormatter;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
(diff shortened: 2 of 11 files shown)


manusa (approved):
As we discussed after today's stand-up, it would be much better if these fields that represent time were typed as OffsetDateTime , or any other type that properly represents timestamps with timezone information.

However, for consistency reasons and since the rest of fields we already have are parsed to timestamp strings, we can leave this for now.

For v6.0.0 it might be a good idea to convert all of those fields into a proper type.

These are some examples (there are many more) of timestamp fields that are currently typed as strings:
- https://github.com/fabric8io/kubernetes-client/blob/13387c759e7144d0c879b0903442a572ad54b030/kubernetes-model-generator/kubernetes-model-core/src/generated/java/io/fabric8/kubernetes/api/model/ObjectMeta.java#L53
- https://github.com/fabric8io/kubernetes-client/blob/13387c759e7144d0c879b0903442a572ad54b030/kubernetes-model-generator/kubernetes-model-storageclass/src/generated/java/io/fabric8/kubernetes/api/model/storage/VolumeError.java#L56
- https://github.com/fabric8io/kubernetes-client/blob/13387c759e7144d0c879b0903442a572ad54b030/kubernetes-model-generator/kubernetes-model-apps/src/generated/java/io/fabric8/kubernetes/api/model/apps/DeploymentCondition.java#L58

rohanKanojia:
I think it would be better if we have a new issue for addressing this.

manusa:
> I think it would be better if we have a new issue for addressing this.

Indeed, a new issue should be open to keep track of this and properly plan its implementation.

LiamClarkeNZ:
> As we discussed after today's stand-up, it would be much better if these fields that represent time were typed as OffsetDateTime , or any other type that properly represents timestamps with timezone information.

@manusa as an end user, that would be nice, it would make it easier to not get it wrong, especially when the K8s documentation isn't overly explicit about accepted formats. :)

PR #3323, opened 2021-07-14
Allow delete with optimistic lock

### LINKED ISSUE #3316: Resource delete with optimistic lock
This is maybe a feature request? Anyway, I'd like to be able to issue a DELETE request to some resource with _delete options_ locking the resource version as [allowed by the Kubernetes API](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#delete-deployment-v1-apps).

I don't see how can I do that using `kubernetes-client`. I've tried this:

```java
Resource<RoleBinding> k8sClientResource; // or NetworkPolicy, Service, Deployment ...
k8sClientResource.lockResourceVersion(resourceVersion).delete();
```

But `delete` isn't usable after `lockResourceVersion`. Is this somehow doable with the current client?

thiagoarrais:
## Description

This PR allows using optimistic locking with deletion operations, like so:

```java
k8sClientResource.lockResourceVersion(resourceVersion).delete();
```

Fixes #3316

## Type of change

## Checklist
 - [x] ~~I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/master/doc/CHEATSHEET.md) accordingly~~ (not applicable)
 - [x] ~~I tested my code in OpenShift~~ (not sure if applicable)

### DIFF
--- a/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/image/ImageSignatureOperationsImpl.java
+++ b/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/image/ImageSignatureOperationsImpl.java
@@ -39,7 +39,7 @@ public class ImageSignatureOperationsImpl extends CreateOnlyResourceOperationsIm
   @Override
   public Boolean delete() {
     try {
-      handleDelete(getResourceUrl(), context.getGracePeriodSeconds(), context.getPropagationPolicy(), null, context.getCascading());
+      handleDelete(getResourceUrl(), context.getGracePeriodSeconds(), context.getPropagationPolicy(), context.getResourceVersion(), context.getCascading());
       return true;
     } catch (IOException | ExecutionException exception) {
       return false;
--- a/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/image/ImageSignatureOperationsImpl.java
+++ b/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/image/ImageSignatureOperationsImpl.java
@@ -39,7 +39,7 @@ public class ImageSignatureOperationsImpl extends CreateOnlyResourceOperationsIm
   @Override
   public Boolean delete() {
     try {
-      handleDelete(getResourceUrl(), context.getGracePeriodSeconds(), context.getPropagationPolicy(), context.getCascading());
+      handleDelete(getResourceUrl(), context.getGracePeriodSeconds(), context.getPropagationPolicy(), null, context.getCascading());
       return true;
     } catch (IOException | ExecutionException exception) {
       return false;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Bugs
 
 #### Improvements
+* Fix #3316: allow locking deletion to resource version
 
 #### Dependency Upgrade
 
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/EditReplacePatchDeletable.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/EditReplacePatchDeletable.java
@@ -18,7 +18,7 @@ package io.fabric8.kubernetes.client.dsl;
 
 import io.fabric8.kubernetes.client.GracePeriodConfigurable;
 
-public interface EditReplacePatchDeletable<T> extends EditReplacePatchable<T>, Deletable,
+public interface EditReplacePatchDeletable<T> extends EditReplacePatchable<T>, Deletable, ReplaceDeletable<T>,
                                                           GracePeriodConfigurable<Deletable>
 
 {
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/EditReplacePatchDeletable.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/EditReplacePatchDeletable.java
@@ -18,7 +18,7 @@ package io.fabric8.kubernetes.client.dsl;
 
 import io.fabric8.kubernetes.client.GracePeriodConfigurable;
 
-public interface EditReplacePatchDeletable<T> extends EditReplacePatchable<T>, Deletable, ReplaceDeletable<T>,
+public interface EditReplacePatchDeletable<T> extends EditReplacePatchable<T>, ReplaceDeletable<T>,
                                                           GracePeriodConfigurable<Deletable>
 
 {
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/ReplaceDeletable.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/ReplaceDeletable.java
@@ -13,14 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package io.fabric8.kubernetes.client.dsl;
 
-import io.fabric8.kubernetes.client.PropagationPolicyConfigurable;
-
-public interface CascadingEditReplacePatchDeletable<T> extends
-  EditReplacePatchDeletable<T>,
-  Cascading<EditReplacePatchDeletable<T>>,
-  PropagationPolicyConfigurable<EditReplacePatchDeletable<T>>,
-  Lockable<ReplaceDeletable<T>> {
+public interface ReplaceDeletable<T> extends Replaceable<T>, Deletable {
 }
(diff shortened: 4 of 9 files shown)


thiagoarrais:
I've fixed the issues raised by the Sonar run. Don't know if it will do a recheck, though...

thiagoarrais:
It seems like there was a Sonar re-run. Do I need to add a new test somewhere else besides the test class? I've added a test method to `CreateOrReplaceResourceTest`. It runs in the Sonar build ([log line 21658](https://github.com/fabric8io/kubernetes-client/commit/9c10d649d29a891c3d8915fd5171c10883bdcb75/checks/3104889071/logs) goes from 16 to 17 tests). But the coverage report for the [relevant line](https://sonarcloud.io/code?id=fabric8io_kubernetes-client&pullRequest=3323&selected=fabric8io_kubernetes-client%3Akubernetes-client%2Fsrc%2Fmain%2Fjava%2Fio%2Ffabric8%2Fkubernetes%2Fclient%2Fdsl%2Fbase%2FOperationSupport.java&line=276) in `OperationSupport` does not light up.

shawkins (commented):
This looks good as a narrow change, but it may be good to consider adding delete(DeleteOptions) - that aligns with the list/watch operations.

If that is ever added it introduces overlap with this logic.

thiagoarrais:
> but it may be good to consider adding delete(DeleteOptions)

Agreed. I guess that is doable on top of this change by injecting the options in `handleDelete` instead of constructing them inside it. There will be some minor level of confusion due to precedence rules when `DeleteOptions` is used in conjunction with `lockResourceVersion`, though.

Should we create an issue for this as backlog for the future?

shawkins:
> There will be some minor level of confusion due to precedence rules when `DeleteOptions` is used in conjunction with `lockResourceVersion`, though.

Would you consider not going the route of lockResourceVersion and just using the DeleteOptions directly?

thiagoarrais:
That's possible. But I feel the `lockResourceVersion` route fits better with the fluent approach taken in the rest of the client.

shawkins:
> But I feel the lockResourceVersion route fits better with the fluent approach taken in the rest of the client.

After thinking about it more and seeing the other existing delete option context properties, it seems fine to move forward with this change.

On lockResourceVersion vs. delete(DeleteOptions) there is another option.

cc @rohanKanojia @manusa This may to broad of a discussion for just here but to support something similar that was added for filtering one could do:

```
public interface DeleteOptionsNested<T> extends DeleteOptionsFluent<DeleteOptionsNested<T>>, Nested<T> {
  
  default T endDeleteOptions() {
    return and();
  }
  
}
...
public interface Deletable {

  ...
  DeleteOptionsNested<Deletable> withDeleteOptions();

}

```
That allows for calls like:
```
client.configMaps().withName("map1").withDeleteOptions().withNewPreconditions("800", null).and().delete()
```
Obviously more verbose than just the lock call, but allows for any delete option based upon inlining the fluent and could clarify the relationship between the context and XXXOptions.

To implement the context needs a reference to DeleteOptions and an extended DeleteFluentImpl to maintain a reference to the operation so that it can be returned with the updated context when the fluent nested ends. 

Any property that is specific to a given operation (gracePeriodSeconds, propagationPolicy, cascading) would just move to a DeleteOptions on the context, rather than having an individual value.  However common things like dryRun and this locked resourceVersion don't map neatly and would still need placeholders in the context and would presumably influence the default values of the DeleteOptions returned by withDeleteOptions.  

To be honest it's not clear right now the relationship between the context properties and options passed into list/watch.  For example if you do:

```
client.configMaps().withField("y", "z").list(new ListOptionsBuilder().withFieldSelector("x=y").build())
```
It will currently send both:
```
GET /api/v1/namespaces/test/configmaps?fieldSelector=x%3Dy&fieldSelector=y%3Dz
```

shawkins (approved):
LGTM

we can move the broader discussion to another place if there is anyone who thinks it looks like a good idea.

manusa:
`delete(DeleteOptions)` + `DeleteOptions.preconditions` + `Preconditions.resourceVersion`


IMHO This is the most K8s REST API compliant way of providing this functionality.

However, in the client we introduced the Optimistic Lock feature through the [Lockable](https://github.com/fabric8io/kubernetes-client/blob/36c49e3d3f65179cf4d448bf2e3cfe8a2780a845/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Lockable.java) interface.

This behavior is Fabric8 _proprietary_ and already overlaps with any imperative behavior that we might provide when performing replace/update operations. e.g. for the replace operation, the natural non-fabric8 way would be for the user to manually set the new object's resource version to the previous known version of the object we want to replace.

So since we already have this Lockable interface, it makes sense for me to reuse it in any context that this lock might apply, in this case delete operations.

The only drawback I see (in general) is the inconsistency with the ListOptions, but some of the DeleteOptions are already part of the DSL (cascading, dryRun, etc.)


> On lockResourceVersion vs. delete(DeleteOptions) there is another option...

I really think we shouldn't complicate more the DSL, especially for duplicating functionality. But as you say, we can discuss this in a separate issue.


So the only drawback I see so far is the lack of support to lock by UID which is provided by DeleteOptions#preconditions#uid

openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/image/ImageSignatureOperationsImpl.java
```diff
@@ -39,7 +39,7 @@ public ImageSignatureOperationsImpl(OperationContext context, String apiGroupNam
   @Override
   public Boolean delete() {
     try {
-      handleDelete(getResourceUrl(), context.getGracePeriodSeconds(), context.getPropagationPolicy(), context.getCascading());
+      handleDelete(getResourceUrl(), context.getGracePeriodSeconds(), context.getPropagationPolicy(), null, context.getCascading());
```
manusa, line 42: Why null instead of `context.getResourceVersion()`?

@rohanKanojia do you know why we have a specific ImageSignatureOperationsImpl#delete override?

shawkins: My confusion here is that I assumed that ImageSignature is just a KubernetesResource because the operation extends from CreateOnlyResourceOperationsImpl, which is not a HasMetadataOperation - however it is, so the resourceVersion is applicable.

We can make the change here to pass the resourceVersion, but I'd vote for either having CreateOnlyResourceOperationsImpl be a HasMetadataOperation, or to remove CreateOnlyResourceOperationsImpl altogether.  

With #3327 as it stands you can't register anything that extends CreateOnlyResourceOperationsImpl as an operation (this is case currently as well, there's no ImageSignatureResourceHandler that is generated).  

To remove it entirely means creating an interface to replace references to NamespacedCreateOnlyResourceOperationsImpl and that if you use resource(ImageSignature) - you could attempt the other write operations, but they'd simply fail.

shawkins:
> I really think we shouldn't complicate more the DSL, especially for duplicating functionality. But as you say, we can discuss this in a separate issue.

Moved to #3349

openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/image/ImageSignatureOperationsImpl.java
manusa: OK, I missed the `CreateOnlyResourceOperationsImpl`.

There are some non-standard Kubernetes->OpenShift resources that don't have full CRUD support. I think we started to add proper OperationsImpl to those in the scope of the latest OpenShift resource type implementations (#3223).

ImageSignatures are deletable, createable, but cannot be replaced.

Regarding #3327, I don't know what the best way to handle this situation would be.

shawkins: > Regarding #3327, I don't know what the best way to handle this situation would be.

Would you be opposed to removing CreateOnlyResourceOperationsImpl?  For standard dsl calls it should be sufficient that the OperationImpl or associated (NamespacedCreateOnlyResourceOperationsImpl) is accessed through the relevant interfaces.

manusa: Implementation-wise I think it's not important.

What we should preserve is the DSL syntax, i.e. there's no `client.imageSignatures().withName("is").replace()` method (neither `edit`, `patch`, etc.).

Having the DSL method return a proper interface sound like a cleaner solution for me, yes.

rohanKanojia (approved):

openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/image/ImageSignatureOperationsImpl.java
thiagoarrais: Not sure if the original sugestion is needed anymore, but changed it anyway. Please let me know if this should be reverted.

manusa (approved):
LGTM, thx!

thiagoarrais:
Merged! Thanks y'all!

🎉

PR #3327, opened 2021-07-17
Removing generated ResourceHandlers

shawkins:
## Description
Before I go too far down this path, I wanted to solicit some feedback.  The initial changes to just refine the usage of CustomResource (the first two commits - which had not yet rationalized the SharedInformerFactory) made me want to see what things would look like without any generated ResourceHandlers.   This pr has that work mostly completed for the kubernetes/openshift client modules - there are more validations and tests that should be added, but the kubernetes-tests run cleanly.

The idea is to replace the generated classes with ResourceHandlerImpl and ResourceOperationsImpl (which is also used by the resources/customResources dsl entry points).  This means that any HasMetadata that is properly annotated can be used as a Resource - whether it's custom, built-in, or from a different client version provided HasMetadata and the annotations are compatible.  Some of the SharedInformerFactory logic could be re-targeted to this as well.

The downsides to this approach:
- it means that instances of the generated resourcehandlers / operations don't exist, so you have to use static Handlers methods to obtain them.
- has reflective association/usage of Builders (based upon similar to the logic for automatically obtaining custom list type)
- requires explicit registration of specialty HasMetadataOperations - which for now was put as a static block in BaseKubernetesClient and DefaultOpenShiftClient.  This allows those operations to be used from the NamespaceVisitFromServer* classes rather than the default ResourceOperationsImpl.

Not really related to these changes it appears that OpenShiftNamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableImpl is not being used.

I'm also thinking it would be good to have more guardrails on KubernetesDeserializer.registerCustomKind given its static nature.

## Type of change

## Checklist


iocanel (commented):
I like where this is going.
This will reduce the size of the client.
If we manage to completely eliminate the use of velocity transformations we will further reduce the number of transitives too.

The tricky part seemed to be the handling of custom impls, which you seem to handle already with the static block.

+1

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/BaseKubernetesClient.java
```diff
@@ -132,6 +127,21 @@
  * It is thread safe.
  */
 public abstract class BaseKubernetesClient<C extends Client> extends BaseClient implements GenericKubernetesClient<C> {
+  
+  static {
+    Handlers.register(Pod.class, PodOperationsImpl::new);
```
metacosm, line 132: Perhaps the `Handlers` class could be generated during the build process so that we don't need to manually register new ones?

shawkins: There are a couple of approaches to be weighed:

- something generated - but that will need the some templating mechanism, which we should otherwise be able to remove
- just a reflective association by convention similar to List and Builder classes - bit it seemed more involved to determine the package name, so I didn't take that approach yet.  There is some validation logic in my next commit thought that could easily determine if an OperationsImpl is not being picked up by the Handler system that would be applicable here.
- more logic around the initialization - this is what I've started to added in the next commit.  If the *ExtensionAdapter class uses references the *Client then that can trigger the static block to register the handlers.  Alternatively that registration logic could be moved onto the *ExtensionAdapter (which is probably a better option than the former).

shawkins:
This next commit rationalizes the SharedInformerFactory logic and updates the ServiceCatalog client to the new approach.  The DefaultSharedIndexInformerTest.testCustomResourceInformerWithDifferentVersions is not working yet because the direct usage of the OperationContext means that I'll need to add logic to override the ResourceDefinitionContext.

shawkins:
I've combined the last two commits and updated things mentioned in the last couple of comments to where this now compiles and runs through the kubernetes-tests

shawkins:
@rohanKanojia @manusa this now should be pretty representative of these changes - it should build without error not withstanding the latest unsuccessful run.  With this only the SharedInformerFactory.sharedIndexInformerFor(Class<T> apiTypeClass, long resyncPeriodInMillis) is not deprecated.

manusa:
I haven't deeply reviewed it, but I really like what's going on.

metacosm:
> I haven't deeply reviewed it, but I really like what's going on.

Same here, though I see several Todos in different spots… Are these expected to be addressed in subsequent PRs or is this PR still a work in progress?

shawkins:
@metacosm @manusa I've resolved the TODOs that seem important for now.  The ones that remain are a reminder about KubernetesDeserializer type registration and a reminder to remove some of the remaining overlap between the SharedInformerFactory and BaseOperation wrt creating informers.  I can change those comments if needed.

shawkins:
Took a pass through to rebase and cleanup the import changes.

- The things to look at for review - deprecated / new methods on KubernetesClient and SharedInformerFactory

There's nothing special about the CustomResource base class after these changes, so methods dealing with CustomResources have been deprecated, and similar methods involving "resources" were added.  This means that nearly every method on sharedinformerfactory is now deprecated

- replaced CustomResourceDefinitionContext with ResourceDefinitionContext in the api where appropriate

Does that seem like a good classname?

- what were generated handlers are now obtained through Handlers methods, 

- There's new reflective access was added for Builders.  Where we look for List/Builder classes was expanded to include the classloader of the type itself.

- We're now passing the type/listType to the base HasMetadataOperation to validate
that the operation is registered

- moved OperationImpl registration onto the ExtensionAdapters, with a static block in BaseKubernetesClient to load the default operations and trigger the load of all the others.

shawkins:
@manusa I'm afraid that keeping this up-to-date will become a full-time job.  If possible let's come up with a plan for inclusion to minimize conflicts.

extensions/service-catalog/client/src/main/java/io/fabric8/servicecatalog/client/DefaultServiceCatalogClient.java
```diff
@@ -38,39 +38,44 @@ public DefaultServiceCatalogClient(OkHttpClient httpClient, Config configuration
     super(httpClient, configuration);
   }
 
-  public NonNamespaceOperation<ClusterServiceBroker, ClusterServiceBrokerList, ClusterServiceBrokerResource> clusterServiceBrokers() {
+  @Override
+public NonNamespaceOperation<ClusterServiceBroker, ClusterServiceBrokerList, ClusterServiceBrokerResource> clusterServiceBrokers() {
     return new ClusterServiceBrokerOperationsImpl(this.getHttpClient(), this.getConfiguration());
   }
```
manusa, line 44: I guess this is related to #3350

extensions/tekton/client/src/main/java/io/fabric8/tekton/client/V1beta1APIGroupClient.java
```diff
@@ -49,26 +45,26 @@ public V1beta1APIGroupClient(OkHttpClient httpClient, final Config config) {
...
   @Override
   public MixedOperation<TaskRun, TaskRunList, Resource<TaskRun>> taskRuns() {
-    return new TaskRunOperationsImpl(this.getHttpClient(), this.getConfiguration());
+    return Handlers.getOperation(TaskRun.class, TaskRunList.class, this.getHttpClient(), this.getConfiguration());
   }
 
   @Override
   public NonNamespaceOperation<ClusterTask, ClusterTaskList, Resource<ClusterTask>> clusterTasks() {
-    return new ClusterTaskOperationsImpl(this.getHttpClient(), this.getConfiguration());
+    return Handlers.getOperation(ClusterTask.class, ClusterTaskList.class, this.getHttpClient(), this.getConfiguration());
```
manusa, line 68: https://github.com/fabric8io/kubernetes-client/pull/3327#pullrequestreview-715741612

I'm confused now. I assumed this was possible until I saw the code in the linked comment. Now I don't get why in those lines we can follow the same procedure.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/KubernetesListOperationsImpl.java
```diff
@@ -126,7 +126,7 @@ public KubernetesList get() {
   }
 
   private <T extends HasMetadata, V extends VisitableBuilder<T, V>> T create(T resource) {
-    ResourceHandler<T, V> handler = Handlers.get(resource.getKind(), resource.getApiVersion());
+    ResourceHandler<T, ?> handler = Handlers.get(resource);
```
manusa, line 129: Why not `ResourceHAndler<T, V> handler`?

If not possible, then the V type parameter declaration should be dropped from the signature

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/core/v1/ServiceOperationsImpl.java
```diff
@@ -65,7 +61,10 @@ public Service waitUntilReady(long amount, TimeUnit timeUnit) {
     // if awaiting existence took very long, let's give at least 10 seconds to awaiting readiness
     long remaining = Math.max(10_000, timeUnit.toNanos(amount) - alreadySpent);
 
-    EndpointsOperationsImpl endpointsOperation = new EndpointsOperationsImpl(context);
+    HasMetadataOperation<Endpoints, EndpointsList, Resource<Endpoints>> endpointsOperation =
+        (HasMetadataOperation<Endpoints, EndpointsList, Resource<Endpoints>>) Handlers
+            .getOperation(Endpoints.class, EndpointsList.class, client, config)
+            .newInstance(context);
```
manusa, line 67: Why the newInstance?
Doesn't `Handlers#getOperation` always return a new instance?

extensions/service-catalog/client/src/main/java/io/fabric8/servicecatalog/client/DefaultServiceCatalogClient.java
shawkins: No, the IDE is just set to automatically add missing Override annotations.

extensions/tekton/client/src/main/java/io/fabric8/tekton/client/V1beta1APIGroupClient.java
shawkins: You could.  ClusterServiceBrokerOperationsImpl will be registered so you can get it through Handlers.  However I was attempting to create the minimal changes at first, which would just be to remove the usage of the generated ones.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/KubernetesListOperationsImpl.java
shawkins: We can use V there now.  Initially Handlers was more loosely typed.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/core/v1/ServiceOperationsImpl.java
shawkins: It does return a new instances, but for the given http client and config.  Here you need an instance at a specific context - for which we already have a factory method for.

manusa: Understood, thx!

extensions/tekton/client/src/main/java/io/fabric8/tekton/client/V1beta1APIGroupClient.java
shawkins: Captured as #3357

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/KubernetesListOperationsImpl.java
shawkins: Updated now.

metacosm:
Initial tests look positive though, surprisingly, for the same app, the native binary size is larger now than it previously was… For comparison, the same code went from 122 MB with 5.4.1 to 138MB with 5.6 to 140MB with 5.7-SNAPSHOT and this PR, which is quite surprising to me. :(

shawkins:
> which is quite surprising to me. :(

It is surprising but indicates that the native compilation is very efficient.  So the takeaway is that the jar sizes should go down, but the native doesn't really change.

rohanKanojia (approved):

metacosm (approved):

manusa (approved):
LGTM, thx!

image :heart_eyes:

PR #3358, opened 2021-07-28
fix #3350: refining CreateOnlyResourceOperationsImpl and related classes

### LINKED ISSUE #3350: Remove or refine CreateOnlyResourceOperationsImpl
The usage of CreateOnlyResourceOperationsImpl creates a parallel set of logic for HasMetadata resources that is not needed to just satisfy the dsl.

After the commit of #3327 this could be done with:

Create 
```
public interface NamespacedInOutCreateable<I,O> extends Namespaceable<InOutCreateable<I, O>>, InOutCreateable<I, O>  {}
```
and 
```
public interface InOutCreateableDeletable<I,O> extends InOutCreateable<I, O>, Deletable, Nameable<InOutCreateableDeletable<I,O>>  {}
```
To replace references to NamespacedCreateOnlyResourceOperationsImpl and ImageStreamXXXOperationsImpl in the dsl api.

Handlers.getCreateOnlyOperation(Class<T> type, OkHttpClient client, Config config) -- just returns the HasMetadataOperation with an inferred list type (so it can do more than just creates). 

Handlers.getNamespacedCreateOnlyOperation(Class<T> type, OkHttpClient client, Config config) -- will create a wrapper class around the operation returned above to implement the inNamespace method.

And a similar method/wrapper for InOutCreateableDeletable.

For all instances where in and out are the same type and there's no special handling in the operation, the classes can be replaced with these invocations - which appears to be all of them currently.  Those operations will also be available for resource(item) calls - but should fail for anything other than a create.  It might also be a good idea to just use Createable<T>, rather than InOutCreatable<T, T> in the api as well.

@rohanKanojia @manusa WDYT?

shawkins:
## Description
Addresses #3350 by consolidating operationimpl logic.  It introduces several new interfaces so that impls can be removed from the dsl.

## Type of change

## Checklist


rohanKanojia:
There seem to be some compilation failures :-(

shawkins:
@rohanKanojia sorry, it's a draft

shawkins:
@rohanKanojia this is in a better state now.  The CreateOnlyResourceOperationsImpl cannot be removed for a couple of reasons:
The openshift resources are not hasmetadata
The openshift resources don't have the necessary annotations for determining the metadata for reuse with the other logic paths.

So this moves everything that is hasmetadata onto the common handlers logic and everything else is still using CreateOnlyResourceOperationsImpl and NamespacedCreateOnlyResourceOperationsImpl - which are no longer used by the core client, they could be moved to openshift or have docs added to not use them with hasmetadata resources.

New interfaces (to review the names / purpose):
NamespacedInOutCreateable - replaces references of NamespacedCreateOnlyResourceOperationsImpl in the api.  This could just be replaced by direct usage of Namespaceable<InOutCreateable<I, O>>.
ImageSignatureOperation - replaces ImageSignatureOperationsImpl in the api.
CreateOrDeleteable - narrows the result of a ImageSignatureOperation.withName call and allows the Resource to be passed directly as the result

ImageStreamImages doesn't need a dedicated interface, and can be defined as Namespaceable<Nameable<? extends Gettable<ImageStreamImage>>>

shawkins:
The test errors are related to missing metadata - that affects both hasmetadata and just kubernetesresource that lack a list class.

Would it be possible to change KubernetesCoreTypeAnnotator to something like:

```
     if (resourceClass != null) {
        ...
        resourceClass.annotate(Version.class).param(ANNOTATION_VALUE, apiVersion);
        resourceClass.annotate(Group.class).param(ANNOTATION_VALUE, apiGroup);
        resourceClass.annotate(PackageSuffix.class).param(ANNOTATION_VALUE, packageSuffix);
        if (resourceListClass != null) {
          resourceListClass.annotate(Version.class).param(ANNOTATION_VALUE, apiVersion);
          resourceListClass.annotate(Group.class).param(ANNOTATION_VALUE, apiGroup);
          resourceListClass.annotate(PackageSuffix.class).param(ANNOTATION_VALUE, packageSuffix);
        }
```

Such that the annotations are still added even if there is no list class?

shawkins:
@rohanKanojia @manusa the next commit adjusts the logic in KubernetesCoreTypeAnnotator to add annotations to types that don't have lists.  It seemed to require adjusting one of the go.sum files.  I also have locally a lot of modifications to the validation-schema.json files - but that seemed to occur with or without the change to KubernetesCoreTypeAnnotator, so they seem unrelated.  This does also end up adding the affected resources to the generated manifest files - is that okay?  

If this seems like an okay change, the other logic can be consolidate more.

shawkins:
https://github.com/fabric8io/kubernetes-client/pull/3358/checks?check_run_id=3195183735#step:4:24351 is occurring because ImageSignature is being generated as Namespaced, but it's actually not.  Where is that controlled?

rohanKanojia:
All cluster scoped resources need to be declared in here in generator script: https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-model-generator/pkg/schemagen/generate.go#L587

shawkins:
@rohanKanojia thanks, that seemed to work

shawkins:
This seems pretty consistent now.  We do need to keep CreateOnlyResourceOperationsImpl, but it is now effectively a dual of HasMetadataOperationsImpl.  

For now the logic for creating CreateOnlyResourceOperationsImpl is only in OpenShiftHandlers and is not caching the ResourceDefinitionContexts.  

In the same way that we have a client.resources(Class<HasMetadata>) method, we could support a client.createOnlyResource(Class<KubernetesResource>) - does that seem useful?

shawkins:
I've update the generation logic again and went ahead and checked in the validation-schema.json files - however the resource iteration order is effectively random, so the default resource is not guarenteed. @manusa is that something that needs addressed?

Other issues:
- This assumes that a kubernetes resource with a namespace field is namespaced - that could be handled differently, such as by adding another list into the go logic.
- This will add annotations to XXXOptions classes because it can't tell them apart from other non-hasmetadata kubernetes resources which could be api roots.
- It it quite a bit of changes to address increasingly edge cases.   I could walk this back to the point where the hasmetadata stuff was consolidated and leave the kubernetes resource stuff alone if that makes more sense.

manusa:
>  I also have locally a lot of modifications to the validation-schema.json files

These files are non-reproducible, for each go execution they change (order of the fields is randomized).
(We ignore those changes when [checking](https://github.com/fabric8io/kubernetes-client/blob/8d380a350231c7d26710b68f02731659418aa999/.github/workflows/gobuild.yml#L38) the model generation)

manusa:
I think the `@PackageSuffix` annotation was used in the Handlers (#2100) and it might not longer be needed.

openshift-client/src/main/java/io/fabric8/openshift/client/dsl/ImageSignatureOperation.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.fabric8.openshift.client.dsl;
+
+import io.fabric8.kubernetes.client.dsl.CreateOrDeleteable;
+import io.fabric8.kubernetes.client.dsl.Nameable;
+import io.fabric8.openshift.api.model.ImageSignature;
+
+public interface ImageSignatureOperation extends Nameable<CreateOrDeleteable<ImageSignature>>, CreateOrDeleteable<ImageSignature> {
```
manusa, line 22: This interface is specific to a single Resource, why can't we use something more generic?

manusa:
>This does also end up adding the affected resources to the generated manifest files - is that okay?

> There were also changes in the objects themselves. @manusa is that something that needs addressed?

Are you referring to the generated objects and the extra annotations or change of scope? If this is the end result of your changes (which I think it is), it's OK. Generated classes should reflect those changes.
Same goes for the manifests.

openshift-client/src/main/java/io/fabric8/openshift/client/dsl/ImageSignatureOperation.java
shawkins: It can, I'll replace it with NameableCreateOrDeleteable.

shawkins:
> I think the @PackageSuffix annotation was used in the Handlers (#2100) and it might not longer be needed.

Yes it is no longer used.  I'll remove it.

manusa (approved):
LGTM, thx!

rohanKanojia (approved):

PR #3365, opened 2021-08-01
fix #3349: adding consistency to the handling of option values

### LINKED ISSUE #3349: Design of methods with XXXOptions
This topic originated on #3316.  list/watch both support ListOptions as an argument.  Delete does not, but does support quite a few context methods.

It is not clear if delete should also support delete(DeleteOptions) for consistency with the list/watch methods.

There is also a consideration of what passing options directly makes sense given the usage of conflicting context values.

For example if you do:

```
client.configMaps().withField("y", "z").list(new ListOptionsBuilder().withFieldSelector("x=y").build())
```

It will currently send both:

```
GET /api/v1/namespaces/test/configmaps?fieldSelector=x%3Dy&fieldSelector=y%3Dz
```

An alternative design to directly passing the options is to inline the fluent construction, similar to what is being done for filtering:

```
public interface DeleteOptionsNested<T> extends DeleteOptionsFluent<DeleteOptionsNested<T>>, Nested<T> {
  
  default T endDeleteOptions() {
    return and();
  }
  
}
...
public interface Deletable {

  ...
  DeleteOptionsNested<Deletable> withDeleteOptions();

}
```

That allows for calls like:

```
client.configMaps().withName("map1").withDeleteOptions().withNewPreconditions("800", null).and().delete()
```

The understanding here is that the constructed deleted options will be the ones passed to the operation.  If you override the default value passed in from the context, then it is explicit what you want to have happen.

The implementation with inline fluents is straight-forward.  The context needs a reference to DeleteOptions and an extended DeleteFluentImpl to maintain a reference to the operation so that it can be returned with the updated context when the fluent nested ends.

Any property that is specific to a given operation (gracePeriodSeconds, propagationPolicy, cascading) would just move to a DeleteOptions on the context, rather than having an individual value. However common things like dryRun and this locked resourceVersion don't map neatly and would still need placeholders in the context and would presumably influence the default values of the DeleteOptions returned by withDeleteOptions.

@rohanKanojia @manusa WDYT: 

- should delete support all options?  
- should we stick with passing options or inline their construction?  
- if we stick with passing in the options, should we clarify what the relationship is between the existing context values and those in the options?

shawkins:
## Description
This is to address #3349.  After trying out the changes accepting a unaryoperator with the options, it seemed simpler to just have the context override whatever the user passes in - which is the expected behavior with forcing the watch property to true on a watch.  With this change that is being done consistently for the labels, fields, and resourceVersion as well, and the javadocs have been updated appropriately.

For consistency this means if you try to do something like
```
.withResourceVersion("x").watch("y", watcher);
```
that y is ignored (since that watch method was already deprecated that shouldn't be an issue).  Alternatively an exception could be thrown when the values conflict - similar to what happens when withName and the item name conflict.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/ListerWatcher.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/ListerWatcher.java
@@ -29,7 +29,7 @@ import io.fabric8.kubernetes.client.Watcher;
 public interface ListerWatcher<T, L> {
   Watch watch(ListOptions params, Watcher<T> watcher);
 
-  L list(ListOptions params);
+  L list();
 
   String getNamespace();
 }
--- a/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/build/BuildConfigOperationsImpl.java
+++ b/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/build/BuildConfigOperationsImpl.java
@@ -32,7 +32,6 @@ import io.fabric8.openshift.api.model.Build;
 import io.fabric8.openshift.api.model.BuildConfig;
 import io.fabric8.openshift.api.model.BuildConfigBuilder;
 import io.fabric8.openshift.api.model.BuildConfigList;
-import io.fabric8.openshift.api.model.BuildList;
 import io.fabric8.openshift.api.model.BuildRequest;
 import io.fabric8.openshift.api.model.WebHookTrigger;
 import io.fabric8.openshift.client.OpenShiftConfig;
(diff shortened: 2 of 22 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/ListerWatcher.java
```diff
@@ -29,7 +29,7 @@
 public interface ListerWatcher<T, L> {
   Watch watch(ListOptions params, Watcher<T> watcher);
 
-  L list(ListOptions params);
+  L list();
```
metacosm, line 32: Isn't this a breaking change?

openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/build/BuildConfigOperationsImpl.java
```diff
@@ -163,39 +162,10 @@ public Void trigger(WebHookTrigger trigger) {
   }
 
   @Override
-  public BuildConfigResource<BuildConfig, Void, Build> withResourceVersion(String resourceVersion) {
+  public BuildConfigOperationsImpl withResourceVersion(String resourceVersion) {
```
metacosm, line 165: Why switch to the implementation instead of the interface, here?

openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/build/BuildConfigOperationsImpl.java
```diff
@@ -163,39 +162,10 @@ public Void trigger(WebHookTrigger trigger) {
...
     return new BuildConfigOperationsImpl(getContext().withResourceVersion(resourceVersion));
   }
 
-  /*
-   * Labels are limited to 63 chars so need to first truncate the build config name (if required), retrieve builds with matching label,
-   * then check the build config name against the builds' build config annotation which have no such length restriction (but
-   * aren't usable for searching). Would be better if referenced build config was available via fields but it currently isn't...
-   */
-  private void deleteBuilds() {
```
metacosm, line 175: Why was this method removed? This seems like a change that's not related to this PR…

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/ListerWatcher.java
shawkins: No this should not be considered a public facing interface.

openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/build/BuildConfigOperationsImpl.java
shawkins: Only because of the other modifications in this class.  It's been unused for a year and a half.

openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/build/BuildConfigOperationsImpl.java
shawkins: I'll be making an update to remove that method altogether.  It is not needed because the base logic that calls newInstance will handle it.

manusa (approved):
LGTM, thx!

shawkins:
A couple of things to call out:

1. Versionable is tied to waitable - but that is no longer the case as we're using an informer, not just a watch to wait.  That is why the constructor calls for the informers are calling .withResourceVersion(null).  I'll add a comment about that in the code.
2. I changed the deprecation message for https://github.com/fabric8io/kubernetes-client/pull/3365/files#diff-aa9f754695e7c28c1fde6238b8a8c2d010cf3c0cad2219d1dec562ef3e511938R49 because it seems more concise - but Versionable is not supported by NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableImpl so it's not generally applicable.  I'll revert that.

rohanKanojia (approved):

oscerd (approved):

PR #3415, opened 2021-08-20
fix: enum values should honor JsonProperty annotations

### LINKED ISSUE #3411: Allow to specify POJO enum values for the corresponding generated CRD enum
When a POJO contains an enum, I would like to serialize it into the corresponding CRD with a specific name instead of the field one.
For example, if an enum is defined this way:

```java
public enum MyEnum {
       Foo,
       Bar,
```

The corresponding CRD will be the following:

```yaml
name:
    enum:
    - Foo
    - Bar
    type: string
```

while I would like to have:

```yaml
name:
    enum:
    - foo
    - bar
    type: string
```

or anyway any other name (i.e. if for any crazy reason I want `Foo` to be `aaa` in the CRD).
I tried the `JsonProperty("foo")` on the enum but with no luck.

### LINKED ISSUE #3431: CRD generator doesn't properly add renamed properties to `required` list
The CRD generator currently doesn't account for renamed properties when checking for required field: the field will always be added with the original name without taking any potential overrides.

metacosm:
Fixes #3411
FIxes #3431 

## Type of change

## Checklist

### DIFF
--- a/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
+++ b/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
@@ -23,25 +23,12 @@ import io.fabric8.kubernetes.api.model.IntOrString;
 import io.fabric8.kubernetes.api.model.Quantity;
 import io.sundr.builder.internal.functions.TypeAs;
 import io.sundr.codegen.functions.ClassTo;
-import io.sundr.model.AnnotationRef;
-import io.sundr.model.ClassRef;
-import io.sundr.model.Method;
-import io.sundr.model.PrimitiveRefBuilder;
-import io.sundr.model.Property;
-import io.sundr.model.TypeDef;
-import io.sundr.model.TypeRef;
+import io.sundr.model.*;
 import io.sundr.utils.Strings;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.util.*;
 
 /**
  * Encapsulates the common logic supporting OpenAPI schema generation for CRD generation.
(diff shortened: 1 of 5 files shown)


oscerd (approved):

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
```diff
@@ -132,85 +120,208 @@ protected T internalFrom(TypeDef definition, String... ignore) {
...
 
-  private void updatePropertyFromAnnotationIfNeeded(Property property, String name, Property[] updated,
-    AnnotationRef a) {
-    if (updated[0] != null) {
-      LOGGER.debug("Property {} has already been renamed to {}", name, updated[0].getName());
-      return;
+  private Map<String, Method> indexPotentialAccessors(TypeDef definition) {
+    final List<Method> methods = definition.getMethods();
+    final Map<String, Method> accessors = new HashMap<>(methods.size());
+    methods.stream()
```
iocanel, line 157: what about inherited properties and accessors?

iocanel (commented):
Overall, looks good. The only thing I missed is what happens with inherited types.

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
metacosm: Aren't the TypeDef that are passed there "complete"? Plus, this is not different from what was done before i.e. the previous implementation was only checking the TypeDef's methods for accessors. This bit only makes it slightly more efficient by avoiding to iterate over all methods for each property.

manusa (approved):

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
metacosm: @iocanel if you think this is an issue, I think we should open a different issue.

rohanKanojia (approved):

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
iocanel: `TypeDef` instances do not include properties and method available in super classes. This is why I recently added `RichTypeDef` that allows accessing to superclass properties and methods.

As long as we are aware of this, we can proceed and cover this in  a future pr.

iocanel (approved):

PR #3468, opened 2021-09-14
add non-running informer to Informable interface

pH14:
## Description

We make heavy use of CRDs and are in the process of migrating from 5.0 to the latest. One thing we've run into in transitioning from the deprecated `client.informer().sharedIndexInformerFor(...)` family of methods to `client.resource(...).inform(...)` is that the new methods / `Informable` interface always return an Informer that is already running, whereas the deprecated methods return Informers that need an explicit `run()`. The primary difference for us is that we pass around the Informer and add a bunch of event handlers / indexers after it's created, which can't be done if the Informer is already running.

This PR adds a `runnableInformer` method to `Informable` that gives the option of returning an `Informer` that is not already running. We're running this patch on our operators internally already and it's been looking good so far 😄 

## Type of change

## Checklist


oscerd (approved):

shawkins:
We debated about whether it made sense to add something like this initially.  The idea was that simplicity and consistency with something like a watch was desirable.

What about a signature like informer(long resync) or runnableInformer(long resync) - and omit the handler, that would be the most consistent with the old factory method.

pH14:
Updated the method to simply be `runnableInformer(long resync)` + added a changelog entry. LMK if anything else is needed, and thank you for the quick response! 🙇

shawkins (commented):
LGTM

>  The primary difference for us is that we pass around the Informer and add a bunch of event handlers / indexers after it's created, which can't be done if the Informer is already running.

You should be able to add handlers after an informer is running.  There is currently no facility to remove them though.

If it makes sense for your usage, adding indexers to a running informer seems like a pretty small set of changes.

rohanKanojia (approved):

shawkins (approved):
Marking as approved this time.

manusa:
Hi, there are merge conflicts. Since the branch belongs to HubSpot and I don't have permissions I cannot resolve them myself. Could you rebase please?

pH14:
Resolved the changelog entry. 

I don't have merge permissions, so if the change looks good could someone merge it in?

PR #3405, opened 2021-08-14
fix #3294 #3303 exposing apigroups and apiresource

shawkins:
## Description
Adds client methods for apigroups and apiresources.  Also consolidates existing logic about openshift adapting.

## Type of change

 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [x] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [ ] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [x] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [x] I Added [CHANGELOG](../CHANGELOG.md) entry regarding this change
 - [ ] I have implemented unit tests to cover my changes
 - [x] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/master/doc/CHEATSHEET.md) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [ ] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

Issue #3184, opened 2021-05-27
OpenshiftExtensionAdapter not working when KubernetesClient is a mock


r00ta:
Hi!

In my application I'm trying to use the `KubernetesClient` in some classes and the `OpenshiftClient` in some others when the underlying platform is Openshift. In my code I have something like 
```java
@Dependent
public class NetworkingServiceBeanProducer {

    private static final Logger LOGGER = LoggerFactory.getLogger(NetworkingServiceBeanProducer.class);

    @ConfigProperty(name = "baaas.k8s.platform", defaultValue = "openshift")
    String platform;

    @Inject
    KubernetesClient kubernetesClient;

    @Produces
    public NetworkingService produceNetworkingServiceBean() {
        LOGGER.info("Platform is {}", platform);
        if ("openshift".equals(platform)) {
            OpenShiftExtensionAdapter adapter = new OpenShiftExtensionAdapter();
            adapter.adapt(kubernetesClient);
            return new OpenshiftNetworkingService(adapter.adapt(kubernetesClient));
        }
        return new KubernetesNetworkingService(kubernetesClient);
    }
}
```

And it works like a charm. But in the tests I have the following mock of the `KubernetesClient`

```java
@ApplicationScoped
public class PlatformClientProducer {

    private final KubernetesServer mockKubernetesServer;

    private final KubernetesClient mockKubernetesClient;

    public PlatformClientProducer() {
        List<CustomResourceDefinitionContext> crds = new ArrayList<>();
        crds.add(PIPELINE_RUN_CONTEXT);
        this.mockKubernetesServer = new KubernetesServer(false, true, crds);
        this.mockKubernetesServer.before();
        this.mockKubernetesClient = mockKubernetesServer.getClient();
    }

    @Produces
    KubernetesServer getMockKubernetesServer() {
        return mockKubernetesServer;
    }

    @Produces
    KubernetesClient getMockKubernetesClient() {
        return mockKubernetesClient;
    }
}
```

And it crashes when I try to run the test suite with the Openshift client

```
Caused by: io.fabric8.openshift.client.OpenShiftNotAvailableException: OpenShift is not available. Root paths at: http://jrota.remote.csb:38987/ do not include oapi.
	at io.fabric8.openshift.client.OpenShiftExtensionAdapter.adapt(OpenShiftExtensionAdapter.java:46)
[...]

Issue #1587, opened 2019-06-14
OpenShiftAdapter uses deprecated APIs to check if cluster is an OpenShift cluster


adambkaplan:
The OpenShift DevEx team has identified two issues with the `OpenShiftAdapter.isOpenShift` [1] method used by kubernetes client:

1. It checks the `/oapi/` path, which has been deprecated in 4.1
2. It starts walking from the root path`/` to see if the apiserver supports OpenShift components. The OpenShift auth team wants to prevent discovery at the root level - this blocks their effort [2].

[1] https://github.com/fabric8io/kubernetes-client/blob/master/openshift-client/src/main/java/io/fabric8/openshift/client/OpenshiftAdapterSupport.java#L48-L76
[2] https://github.com/openshift/origin/pull/23049

gabemontero:
according to @enj `the fix would be to just use `/apis` directly
`

gabemontero:
hey @akram would opening a corresponding jira somewhere be a good idea wrt getting the appropriate visibility / priority on this ?

rohanKanojia:
this doesn't look complicated; Could anyone please raise a PR to fix this? :-)

gabemontero:
ok @waveywaves and I looked at this a bit today as part of the transfer of openshift/jenkins to the devtools team

I gave him a couple of suggestions of how to switch this, but that was after only looking at the code for the first time while with him

Ultimatlely, we need a masterURL/api http rest invocation in BaseOperation that serializes into a class similar RootPath.class

But of course if @rohanKanojia knows better please chime in

do be aware that this problem will be blocking the 4.2 release of openshift, so it will a high priority for our managment

associated bugzilla https://bugzilla.redhat.com/show_bug.cgi?id=1729522

akram:
@gabemontero we are now tracking it also as a blocker in under DEVEXP-390 jira

gabemontero:
We are doing some research here today.

the k8s model in this client does not appear to have an APIGroupList a la https://docs.okd.io/latest/rest_api/apis/v1.APIGroupList.html or an APIGroup a la https://docs.okd.io/latest/rest_api/apis/v1.APIGroup.html

those are the items returned when you hit <master url>/apis

there is a io.fabric8.kubernetes.model.annotation.ApiGroup but that must be related to the common apiGroup field in all objects (or that is my best guess)

[...]

PR #3487, opened 2021-09-22
Exposing apigroups and apiresources

### LINKED ISSUE #3294: how to get apiGroupList use kubernetes-client?
how to get apiGroupList use kubernetes-client?

### LINKED ISSUE #3303: Support for fetching APIResourceList
Similar to APIGroup which is accessible via `/apis` endpoint(see https://github.com/fabric8io/kubernetes-client/issues/3294), there is another NON-Restful endpoint in Kubernetes related to all available resources in a particular API group. It specifies information regarding a resource like the name of a resource and whether it is namespaced and supported operations.

Here is how it's used:
```bash
~ : $ curl localhost:8001/apis/batch/v1
{
  "kind": "APIResourceList",
  "apiVersion": "v1",
  "groupVersion": "batch/v1",
  "resources": [
    {
      "name": "jobs",
      "singularName": "",
      "namespaced": true,
      "kind": "Job",
      "verbs": [
        "create",
        "delete",
        "deletecollection",
        "get",
        "list",
        "patch",
        "update",
        "watch"
      ],
      "categories": [
        "all"
      ],
      "storageVersionHash": "mudhfqk/qZY="
    },
    {
      "name": "jobs/status",
      "singularName": "",
      "namespaced": true,
      "kind": "Job",
      "verbs": [
        "get",
        "patch",
        "update"
      ]
    }
  ]
}
```
I observed is that it's only possible to get a list of APIResource objects in a particular API group. It's not possible to get a specific APIResource object.

manusa:
## Description
Supersedes #3405
Fixes #3294
Fixes #3303 

Fixes issue when compiling in JDK 8.

Sundrio has an issue when traversing the AST path for OkHttpClient (return type in previos Config#adaptClient).
The new commits on top of #3405 remove the adaptClient from the Config classes and places a similar method in the BaseClient class. The Config class is a little more anemic but at least OkHttpClient adaptation has been abstracted :shrug:. 

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationSupport.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationSupport.java
@@ -46,6 +46,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.Collections;
@@ -766,4 +767,27 @@ public class OperationSupport {
     }
     return STRATEGIC_MERGE_JSON_PATCH;
   }
+  
+  public <R1> R1 restCall(Class<R1> result, String... path) {
+    try {
+      URL requestUrl = new URL(config.getMasterUrl());
+      String url = requestUrl.toString();
+      if (path != null && path.length > 0) {
+        url = URLUtils.join(url, URLUtils.pathJoin(path));
+      }
+      Request.Builder req = new Request.Builder().get().url(url);
+      return handleResponse(req, result);
+    } catch (KubernetesClientException e) {
+      if (e.getCode() != HttpURLConnection.HTTP_NOT_FOUND) {
+        throw e;
+      }
+      return null;
+    } catch (InterruptedException ie) {
+      Thread.currentThread().interrupt();
+      throw KubernetesClientException.launderThrowable(ie);
+    } catch (ExecutionException | IOException e) {
+      throw KubernetesClientException.launderThrowable(e);
+    }
+ }
+  
 }
--- a/openshift-client/src/main/java/io/fabric8/openshift/client/OpenshiftAdapterSupport.java
+++ b/openshift-client/src/main/java/io/fabric8/openshift/client/OpenshiftAdapterSupport.java
@@ -26,8 +26,12 @@ import okhttp3.Authenticator;
 import okhttp3.OkHttpClient;
 
 import java.net.URI;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 public class OpenshiftAdapterSupport {
+
+  private static final Map<String, Boolean> API_GROUPS_ENABLED_PER_URL = new ConcurrentHashMap<>();
 
   public Boolean isAdaptable(Client client) {
     OpenShiftConfig config = OpenShiftConfig.wrap(client.getConfiguration());
@@ -48,8 +52,15 @@ public class OpenshiftAdapterSupport {
    * @return         True if oapi is found in the root paths.
    */
   public static boolean isOpenShift(OkHttpClient client, OpenShiftConfig config) {
-    return new BaseClient(adaptOkHttpClient(client, config), config) {}.getApiGroups()
-        .getGroups().stream().anyMatch(g -> g.getName().endsWith("openshift.io"));
+    if (config.isDisableApiGroupCheck()) {
+      return true;
+    }
+    String url = config.getMasterUrl();
+    return API_GROUPS_ENABLED_PER_URL.computeIfAbsent(url,
+        k -> new BaseClient(adaptOkHttpClient(client, config), config) {}.getApiGroups()
+            .getGroups()
+            .stream()
+            .anyMatch(g -> g.getName().endsWith("openshift.io")));
   }
 
     /**
--- a/openshift-client/src/main/java/io/fabric8/openshift/client/OpenshiftAdapterSupport.java
+++ b/openshift-client/src/main/java/io/fabric8/openshift/client/OpenshiftAdapterSupport.java
@@ -35,7 +35,7 @@ public class OpenshiftAdapterSupport {
 
   public Boolean isAdaptable(Client client) {
     OpenShiftConfig config = OpenShiftConfig.wrap(client.getConfiguration());
-    return hasCustomOpenShiftUrl(config) || isOpenShift(client.adapt(OkHttpClient.class), config);
+    return hasCustomOpenShiftUrl(config) || isOpenShiftAPIGroups(client.adapt(OkHttpClient.class), config);
   }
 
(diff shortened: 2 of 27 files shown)


oscerd (approved):

openshift-client/src/main/java/io/fabric8/openshift/client/OpenshiftAdapterSupport.java
```diff
@@ -54,63 +43,11 @@ public Boolean isAdaptable(Client client) {
    * @param config {@link OpenShiftConfig} OpenShift Configuration
    * @return         True if oapi is found in the root paths.
    */
-  static boolean isOpenShift(Client client, OpenShiftConfig config) {
-    return isOpenShiftAPIGroups(client, config);
+  public static boolean isOpenShift(OkHttpClient client, OpenShiftConfig config) {
+    return new BaseClient(client, config) {}.getApiGroups()
```
shawkins, line 47: One of the reasons I put it on http client handling on the config was for this call, so that it could use the BaseClient.  Now that the adapt logic is in the openshiftclient, does that mean we'll have to introduce a different mechanism here to adapt the http client for openshift use?

manusa: Yes, I saw this after all of the changes.
We might need to find some other place for the adapter.

shawkins (approved):
LGTM

manusa:
> LGTM

Still getting some errors in the tests (I assume that the original PR would have those same errors, did you build it with tests locally?)

Errors are due to the `OpenShiftAdapterSupport#isOpenShift` call in the DefaultOpenShiftClient constructor, which fails in those unit tests that don't have a valid connection. I mocked that class for the openshift-client module tests, but there are still some failures in the Karaf module, still need to check those.

manusa:
OK, I think it might be related with the removal of the negation:
`if (config.isDisableApiGroupCheck()) {`

shawkins:
> OK, I think it might be related with the removal of the negation:

The problem should be with https://github.com/fabric8io/kubernetes-client/pull/3487/files#diff-7b405d7bf5ccb2d183a3bdd42f24aad3cf3e7d1bb419323b2a8a4d422bc6bac4R233

If there is no endpoint, then this call will fail.  Previously the exception was swallowed in the OpenshiftAdapterSupport.

I guess karaf logic is expecting to construct the openshift client without a failure.

shawkins:
After looking at this some more, I'm not sure it was ever working as intended.  Between the boolean confusion, exception swallowing, and API_GROUPS_ENABLED_PER_URL.containsKey - rather than a get, it seems to have a very narrow path to work consistently.

I've tried in the next commit to satisfy #3184 as well.  Essentially the behavior is:

if isDisableApiGroupCheck is true, then we'll report that it's openshift.  That somewhat addresses #3184
if isOpenshiftApiGroupsEnabled is true, then we'll assume the use of api group / version in openshift calls.
if isOpenshiftApiGroupsEnabled is false, then we'll effectively rely on isDisableApiGroupCheck via the isOpenShift call.  If it's false, we'll do the check.  If it's true, then we'll assume it's openshift - that's a difference from old logic.

manusa:
Your fix seems to be passing the tests now, thx.

Shall I remove this one https://github.com/fabric8io/kubernetes-client/pull/3487/commits/e59f4337cc25b8def21c5b17860e045edc385e44? (It didn't seem to have any effect either) _(We really need some proper unit tests to document the logic of the configuration)_

I'm a little worried about possible regressions introduced by this PR when working on OpenShift. Shall we skip it from today's release and delay it to 5.9.0?

shawkins:
> I'm a little worried about possible regressions introduced by this PR when working on OpenShift. Shall we skip it from today's release and delay it to 5.9.0?

I can understand that.

On the one hand this wasn't functioning like it should.  After the first time through the DefaultOpenShiftClient constructor, the check https://github.com/fabric8io/kubernetes-client/pull/3487/files#diff-7b405d7bf5ccb2d183a3bdd42f24aad3cf3e7d1bb419323b2a8a4d422bc6bac4L235 would assume "enabled" for any url previously checked.  However since it's not altering the config in that path it's really not doing anything to the config flags at that point.

Only on the first time through this logic was there a possibility of checking the api groups.

If disableApiGroupCheck was false, then openshiftApiGroupsEnabled became false - regardless of what was set on the OpenShiftConfig (which is also confusing because in the OpenShiftConfig openshiftApiGroupsEnabled is treated as the primary property).  

If disableApiGroupCheck was true (which appears to be only for mock logic), then it would perform the group check and use that result as openshiftApiGroupsEnabled - that seems completely contrary to the disable flag.

The proposed changes are trying somewhat to honor the original intent of the flags, and thus introducing the need to set the disable flag in more places.  You could make the case to further change this logic for clarity.  

Since the way we check for openshift is based upon the presence of openshift apigroups, in effect isOpenShift == isOpenshiftApiGroupsEnabled.  Also since OpenShift 3.6 and earlier is no longer supported, at this point I'd say we don't need to preserve isOpenshiftApiGroupsEnabled at all.  The rationale for keeping disableApiGroupCheck is for making the isAdaptable check work in mock scenarios.

I've added a commit that does this.  The concern there is that it's actually behavioral different from before - what I can see/infer if you leave everything at the defaults, we'll only ever use oapi.  With this change we'll use apis.  @rohanKanojia is there anything that is required to use oapi that is an openshift.io group?

So it's probably best to get more eyes and testing on this.

shawkins:
Also my apologizes for applying the "Whenever I run into a problem I can't solve, I always make it bigger" principle here.

manusa:
> So it's probably best to get more eyes and testing on this.

Since there is nothing critical here, I'll schedule this for 5.9.0 (due in < 3 weeks)

> On the one hand this wasn't functioning like it should

I added a breaking change note in the changelog, at least to warn the users. It wasn't working as supposed, but has been like this for a while. The recent changes now enforce the check which might (or might not) break some user's current setup. The warning should help them figure out the issue.

rohanKanojia:
As far as I remember `/oapi` path was used by OpenShift versions < 3.11 and got deprecated in favor of `/apis` . I had made this change in the scope of https://github.com/fabric8io/kubernetes-client/issues/1587 . Not 100% sure but I don't think there is any resource that'll use `/oapi` if it's in `openshift.io` apigroup.

shawkins:
> The warning should help them figure out the issue.

I've updated that to reflect these next changes.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationSupport.java
```diff
@@ -721,4 +722,27 @@ private MediaType getMediaTypeFromPatchContextOrDefault(PatchContext patchContex
     }
     return STRATEGIC_MERGE_JSON_PATCH;
   }
+  
+  public <R1> R1 restCall(Class<R1> result, String... path) {
```
manusa, line 726: TIL, I always run out of letters from the alphabet :)

PR #3496, opened 2021-09-26
fix #3408: escape upload file path

### LINKED ISSUE #3408: upload file throw exception when file name contains special char
when I upload a file with a special file name, it will throw exception.
the file name is `001_special_!@#\$^&(.mp4`

I do some research, in the `PodUpload.uploadFile`:
```
  final String file = context.getFile();
  final String command = String.format("mkdir -p %s && base64 -d - > %s", directory, file);
```
I think the `file` should be escaped for special char with commons-text `StringEscapeUtils::escapeXSI`

ttbadr:
## Description
escape the file path. In case it contains special characters like space.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/pom.xml
+++ b/kubernetes-client/pom.xml
@@ -193,6 +193,11 @@
       <artifactId>generex</artifactId>
       <version>${generex.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-text</artifactId>
+      <version>${commons-text.version}</version>
+    </dependency>
 
     <!-- Compile Only Dependencies -->
     <dependency>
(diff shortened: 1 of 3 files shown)


oscerd (changes requested):
Looks good but avoid * imports

rohanKanojia:
Is it possible to avoid dependency on `commons-text`? Since we already have `commons-compress` dependency as optional, maybe we can make this optional as well...

shawkins (commented):
I don't see a good way to make the dependency optional.  Would it be possible to use quoted strings instead?  That is instead of substituting the string in the command directly, use:

```
"'" + value.replaceAll("'", "'\\\\''") + "'"
```

ttbadr:
Since @shawkins create a new PR to solve this issue more simply, I will close this PR.

PR #3502, opened 2021-09-29
fix #3448 adding a kubernetes specific version

### LINKED ISSUE #3448: Openshift client: should be able to retrieve kubernetes version
For a given OpenShift client, if one wants to retrieve the kubernetes version of the cluster, one has to create a new Kubernetes client:
```
new DefaultKubernetesClient(client.getConfiguration())
```
This is clueless given that the OpenShift client has it's own Kubernetes client already (instance var `delegate`)

shawkins:
## Description
fixes #3448 by adding a getKubernetesVersion to the openshift client.  Also further consolidates the version handling logic.

@manusa this implements what was mentioned in the issue, but it may not be the best approach.  DefaultOpenShiftClient may return 1 of 3 versions.

If you try to retrieve version/openshift and fail (which may be due to some unrelated issue), you will instead try to get the ClusterVersion.  If that doesn't exist, you'll instead use the kubernetes version.

It would seem better to have explicit, getKubernetesVersion and getOpenShiftVersion on the Kubernetes and Openshift clients respectively - with additional handling of the exception when trying out the v3 version/openshift endpoint.

An alternative would be:

VersionInfo getVersions(Version) - where Version is an enum of values such as kubernetes, openshift_v3, openshift_v4, etc. - such that there is no guessing about what the user wants as the version.

## Type of change

## Checklist


kubernetes-tests/src/test/java/io/fabric8/openshift/client/server/mock/OpenShiftVersionInfoTest.java
```diff
@@ -46,7 +46,7 @@ void testClusterVersioningOpenshift3() throws ParseException {
     assertEquals("3", client.getVersion().getMajor());
     assertEquals("11+", client.getVersion().getMinor());
     assertEquals(119, client.getVersion().getBuildDate().getYear());
-    assertEquals(new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'").parse("2019-10-31T21:06:55Z").getTime(), client.getVersion().getBuildDate().getTime());
+    assertEquals(new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssX").parse("2019-10-31T21:06:55Z").getTime(), client.getVersion().getBuildDate().getTime());
```
metacosm, line 49: Curious about this change: has the date format changed?

shawkins: 'Z' expects the literal character Z to appear in the date.  X will parse it as the timezone.  The logic was switched to using jackson for the deserialization, so it's now respecting the timezone.

manusa:
I discussed this with Rohan during our team's Sprint plan.
The current `getVersion` method is really ambiguous. IMO as a user you might expect that the version it returns is the one for the Client itself instead of the cluster's.

I think that a `getKubernetesVersion` + `getOpenShiftVersion` + deprecation of the current method would be the most clear approach given the point we're at right now. WDYT?

shawkins:
> I think that a getKubernetesVersion + getOpenShiftVersion + deprecation of the current method would be the most clear approach given the point we're at right now. WDYT?

I'm fine with that.

I wasn't quickly able to standup an openshift 3.1x instance, so I'm not sure why those e2e tests are failing.  Everything else looks good with base kubernetes and openshift v4.  I'll see if I can get that straightened out.

shawkins:
Looking at things again, it seems like there could be 3 separate methods.  One for the base version, one for openshift v3 (which also returns a versioninfo).  And one for openshift 4 - which I don't think should return a versioninfo, but rather just the version string from the ClusterVersion.  Mapping into versioninfo leaves most fields blank and the metadata timestamp is not a substitute for the build timestamp.  There's also a note on the javadoc that the user can do resources(ClusterVersion.class).list() - if they need more information.

shawkins:
> Is the wip label still applicable? can I merge?

If you are good with the 3 new methods, then WIP can be removed.  I'll update the changelog for this.

The only thing holding up the merge is then the failing openshift 3.11 test - I'm not sure what is going on there.

manusa:
> > We changed the behavior here, did you check the assertion?
> 
> No the behavior should not have changed. See the 3.10 runs without a problem.


There was a small regression for OpenShiftClient#getVersion. Instead of falling back to KubernetesClient#getVersion in OCP 3.11, it was throwing a RuntimeException.

I added a subsequent commit to address this issue, please @shawkins check if this is OK.

Issue #3547, opened 2021-10-26
Create an abstraction layer for okhttp usage


shawkins:
In order to eventually do #2764 and related issues we should isolate the direct usage of okhttp.  My proposal is to introduce a JDK like set of interfaces that we can further align to the built-in ones once fabric8's min version is bumped to 11.

There will be some breakage / deprecations in doing this, including:
- client constructors that take OkHttpClient
- OperationContext withOkHttpClient
- PatchType getMediaType
- ExecListener usage of Response
- Config TlsVersions
- you could not mix versions of client and extension jars that use Adapters to register handlers.

The intent would be to leave okhttp compatibility for some time though.

shawkins:
1, 2, and 5 are complete

I've updated the pr for review to complete 3.

4 is no longer a priority - we are talking about making these changes for 5.11, not 6, such that the initial incarnation will still only offer the okhttp implementation.

6 is nearly ready - https://github.com/shawkins/httpclient - there's just 1 issue left to resolve around expect continue behavior.

I started looking at what is needed for the implementation of a vert.x client.  At first glance that does seem to be more involved.  It would use the vert.x core HttpClient for websockets and the WebClient for everything else, which complicates the factory / initialization logic - it would make sense to just introduce a base builder class.  The conversion of InputStreams to ReadStreams seems to require an additional class https://stackoverflow.com/questions/66158455/how-to-create-a-vert-x-readstream-from-an-inputstream And the handling of send (instead of the body handling being part of the request builder, it's something you do on the request) and the additional indirection of futures / handlers will just take more time in general to wire up.

joschi:
@shawkins After running into a few issues with the new HTTP client abstraction in https://github.com/dropwizard/dropwizard-kubernetes/pull/147, I think it would be useful to customize the `OkHttpClient` (not `OkHttpClient.Builder`) instance.

Specifically the ability to add a wrapper around the actual `OkHttpClient` instance would be great, for example for adding instrumentation to the OkHttp client via https://github.com/raskasa/metrics-okhttp.

shawkins:
@joschi feel free to open a new issue or pr.  The goal moving forward will be to de-emphasize okhttp.

shawkins:
Related to this issue, there's a possible abstraction for creating the clients is at https://github.com/fabric8io/kubernetes-client/pull/3654/commits/2d22886a71419d0fa5c35a20b752ade204973619 - this makes use of the client factory directly.

[...]

PR #3549, opened 2021-10-26
Adding an api layer to isolate Okhttp

shawkins:
## Description
This is to address #3547 

Lower level implementation details can be discussed here.   This will be somewhat long running as there is quite a bit of okhttp usage.

## Type of change

## Checklist


shawkins:
This should be mostly* working up to the kubernetes-tests.  I haven't yet pulled some of the deprecated methods off of DefaultKubernetesClient and related.  Other than that the only remaining direct references to okhttp in non-test code are in the io.fabric8.kubernetes.client.internal.okhttp package.

The HttpClientUtils will need broken apart.  The are some utility methods that need to be free of okhttp, like the creation of basic credentials.  There is also the init logic for okhttp which will need to be encapsulated through a service/factory

OpenIDConnectionUtils - needs to be rewritten in terms of the new api, or if support gets bumped to java 11 could just be directly written for the java 11 client.  Related to this, the TokenRefreshInterceptorTest has been disabled for now.

I've been writing a java 11 implementation on the side.  There are a couple of pain points that may require a few additional tweaks to the api layer.

shawkins:
Everything other than the TokenRefreshInterceptorTest has been prepped for the separation of the okhttp dependency.

There are still other test dependencies to okhttp - but that is allowable, or at the least can be removed later.

@manusa @rohanKanojia @iocanel can you see if what has been done so far makes sense?  The new api is an amalgam of the okhttp client usage and the java 11 apis - it obviously still needs documented more thoroughly.    

I do have an implementation using java 11 - see https://github.com/shawkins/httpclient  The obvious pain point is adding the interceptor pattern - which needs to also cover websocket requests.  The java api completely hides that from the user, so the Interceptor interface in the new api is stripped down as much as possible.  One minor issue is that it's not fully non-blocking as the calls to refresh the tokens will be blocking - but I think that can be addressed if needed.  I have not yet added a similar createHttpClient method - not all config properties will be applicable, and some of the logic will end up moving to a common spot back in HttpClientUtils..

TODOs have been added anywhere that needs additional thought.  Another thing to note is that the adapt logic for the openshift client has been made to piggy-back more on the main configuration.

I have not yet made the client available from the config, I would like to tackle that after the initial work is done.

maxandersen:
I raised this pr in quarkus chat - details at https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/kubernetes.20client.20okhttp.20isolation - but summary is:

With a Java 11 enabled http client we can always use https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/HttpClient.Builder.html#executor(java.util.concurrent.Executor) to wire in a vert.x/quarkus based executor so the same thread/execution model is used. Simplifies configuration and avoids unnecessary thread creation.

It would be nice to have a vert.x http client natively implemented so even in native compiled apps we can strip out even more classes; but that is secondary. Big win not having to deal with okhttp at least :)

So from Quarkus perspective we should be fine here and we can evolve from this point.

shawkins:
We are now talking about targeting this for 5.11 without explicit support for another client implementation.  This means that client constructors using okhttp should be added back, but deprecated, for convenience.  And it's also unnecessary to split out new modules.

@manusa @rohanKanojia after looking over all of the usage patterns around passing the client around, I'm not sure putting it on the Config works well.  The life-cycle becomes less clear in that case.  

Essentially there is little that can be done to cut down on the proliferation of constructors as BaseClient is initialized as a side-effect of construction.  But we can still cut down on the explicit references to HttpClient.  The narrowest change is to replace replace the direct HttpClient argument with HttpClientAware - which does not seem like it would be worth the effort.  

The larger change would be to replace both HttpClient and Config arguments with something like a ClientState interface that extends both ConfigAware and HttpClientAware.  The latter seems more future proof as any additional derived state could easily be added.  This would look like registering handlers with Function<ClientState, HasMetadataOperation<T, L, R>>, and having Client, OperationSupport, and other classes extend ClientState so that the internal methods would be like:

```
public CamelKClient adapt(Client client) {
  return new DefaultCamelKClient(client);
}

public MixedOperation<Pod, PodList, PodResource<Pod>> pods() {
  return new PodOperationsImpl(this);
}
```

Does that seem worthwhile?

shawkins:
@rohanKanojia @manusa the next commit implements the change I was talking about in the previous comment https://github.com/fabric8io/kubernetes-client/pull/3549/commits/c16835405a5fb9cd6d609c626fe77ecdac1233a7 - that introduces a ClientState parameter interface to provide a level of indirection for the HttpClient - and potentially any other derived client state that may need to be passed around in the future. With this initial change I've made the clients, OperationSupport, and OperartionContext all extend the same SimpleClientState class.  With some more thought that relationship could be refined more - "has a" rather than "is a".  The net effect is that 2/3 of the references to HttpClient have been removed.

shawkins:
I've run the test suites as best as possible through the potential jdk client https://github.com/shawkins/httpclient - there are just a couple of issues remaining - one of which is addressed in Java 16.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/HttpClientUtils.java
```diff
@@ -285,41 +105,103 @@ private static boolean isIpAddress(String ipAddress) {
...
+        }
+        if (config.getUserAgent() != null && !config.getUserAgent().isEmpty()) {
+          builder.setHeader("User-Agent", config.getUserAgent());
+        }
       }
-      return chain.proceed(request);
     });
     // Impersonator Interceptor
-    interceptors.add(new ImpersonatorInterceptor(config));
+    interceptors.put("IMPERSONATOR", new ImpersonatorInterceptor(config));
```
metacosm, line 132: Maybe move the name to the interceptor class, similarly to what's done for `TokenRefreshInterceptor`?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/ClientState.java
```diff
@@ -16,8 +16,6 @@
 
 package io.fabric8.kubernetes.client;
 
-@Deprecated
-public interface ClientAware<C extends Client> {
-
-  C getClient();
+public interface ClientState extends HttpClientAware, ConfigAware<Config> {
```
metacosm, line 19: Not convinced by the `ClientState` name, which seems to imply it provides a state rather than the fact that it holds the client and its config… Maybe `ClientStateHolder` (or something similar) would be better?

kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/KubernetesCrudDispatcher.java
```diff
@@ -188,23 +189,23 @@ public MockResponse handlePatch(RecordedRequest request) {
 
         MediaType mergeType;
         if (contentType == null) {
-          mergeType = OperationSupport.JSON_PATCH;
+          mergeType = OkHttpClientImpl.JSON_PATCH;
```
metacosm, line 192: Why are we referencing the Okhttp implementation here?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/okhttp/OkHttpClientBuilderImpl.java
```diff
@@ -0,0 +1,207 @@
...
+import java.net.InetSocketAddress;
+import java.net.Proxy;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import static okhttp3.ConnectionSpec.CLEARTEXT;
+
+class OkHttpClientBuilderImpl implements Builder {
```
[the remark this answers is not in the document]
shawkins: That is correct.  That was taken off the table for now when we decided to keep targeting the 5.x line and not provide the jdk implementation yet.

kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/KubernetesCrudDispatcher.java
shawkins: For now yes.  The mock server logic is still dependent on okhttp, that will introduce at least a test dependency.  This particular reference will be easy to remove/move as it's just a constant.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationSupport.java
```diff
@@ -305,13 +297,11 @@ protected void handleDelete(URL requestUrl, long gracePeriodSeconds, DeletionPro
...
    * @return returns de-serialized version of apiserver response in form of type provided
-   * @throws ExecutionException Execution Exception
    * @throws InterruptedException Interrupted Exception
    * @throws IOException IOException
    */
-  protected <T, I> T handleCreate(I resource, Class<T> outputType) throws ExecutionException, InterruptedException, IOException {
-    RequestBody body = RequestBody.create(JSON, JSON_MAPPER.writeValueAsString(resource));
-    Request.Builder requestBuilder = new Request.Builder().post(body).url(getResourceURLForWriteOperation(getResourceUrl(checkNamespace(resource), null)));
+  protected <T, I> T handleCreate(I resource, Class<T> outputType) throws InterruptedException, IOException {
+    HttpRequest.Builder requestBuilder = httpClient.newHttpRequestBuilder().post(JSON, JSON_MAPPER.writeValueAsString(resource)).url(getResourceURLForWriteOperation(getResourceUrl(checkNamespace(resource), null)));
```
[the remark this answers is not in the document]
shawkins: Updated

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/ExecListener.java
```diff
@@ -15,32 +15,38 @@
...
-
-    /**
-     * Called when the transport or protocol layer of this web socket errors during communication.
-     *
-     * @param t Throwable
-     * @param response Present when the failure is a direct result of the response (e.g., failed
-     * upgrade, non-101 response code, etc.). {@code null} otherwise.
-     */
-    void onFailure(Throwable t, Response response);
+  default void onOpen() {}
```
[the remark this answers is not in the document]
metacosm: OK, was asking because I noticed one of our example operator in the operator SDK was failing to compile with this PR because it's using a listener. It uses the response's message for some logging, probably nothing critical, just don't have experience with this specific feature so was wondering the reason behind the change. Thanks!

metacosm (approved):
LGTM, hopefully we will be able to wire in a vert.x client in a not-so-distant future! 😉

shawkins:
> LGTM, hopefully we will be able to wire in a vert.x client in a not-so-distant future!

Hopefully - I did start down that path, but quickly got bogged down - https://github.com/fabric8io/kubernetes-client/issues/3547#issuecomment-974829390  Maybe there is someone on their team who could pick that up if they want to go beyond the jdk impl.

shawkins:
On the 4 sonarcloud bugs https://sonarcloud.io/project/issues?id=fabric8io_kubernetes-client&open=AXziolikSmmHNwo2Qbf6&pullRequest=3549&resolved=false&types=BUG - 3 are with the usage of the WritableByteChannel.  Since we close the underlying streams, there's no need for an additional close.  I'm just using that logic to write out the ByteBuffer.  We also need a reference to the underlying streams to perform a flush - that is not done through a WritableByteChannel.  It would probably be a good idea to add flushes to the error calls - I can do it with this pr or separately.  

On the other one the isPresent call is there.  It probably just doesn't like multiple calls to previousResponse

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/ClientState.java
shawkins: @rohanKanojia @manusa what name would you like to see for ClientState?

shawkins:
@manusa it probably goes without saying, but feel free to squash this as the individual commits at this point don't convey much.  alternatively I can rebase it if you want.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/ClientState.java
rohanKanojia: Hmm, I'm not good with naming things :roll_eyes: . Does `ClientContext` make sense?

shawkins: @manusa we have:

- ClientState
- ClientStateHolder
- ClientContext

If you support any one of these, let's use it.  Or do you want something else?

rohanKanojia (approved):
Great effort, Thanks a lot! :weight_lifting_man:

metacosm:
> Great effort, Thanks a lot! 🏋️‍♂️

Herculean effort, indeed! 💪🏼

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/ExecListener.java
```diff
@@ -15,32 +15,38 @@
...
 
+  /**
+   * Called when the transport or protocol layer of this web socket errors during communication.
+   *
+   * @param t Throwable
+   * @param failureResponse non-null if the failure is caused by the handshake
+   */
+  default void onFailure(Throwable t, Response failureResponse) {}
```
manusa, line 46: Shouldn't the second parameter (`Response failureResponse`) be removed here?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationSupport.java
```diff
@@ -601,7 +586,7 @@ protected Response retryWithExponentialBackoff(OkHttpClient client, Request requ
       } catch (IOException ie) {
         if (numRetries < requestRetryBackoffLimit) {
           retryInterval = retryIntervalCalculator.getInterval(numRetries);
-          LOG.debug(String.format("HTTP operation on url: %s should be retried after %d millis because of IOException", request.url(), retryInterval), ie);
+          LOG.debug(String.format("HTTP operation on url: %s should be retried after %d millis because of IOException", request.uri(), retryInterval), ie);
```
manusa, line 589: Not introduced by this PR, but these placeholders `%s` should be replaced by `{}`

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/ExecListener.java
manusa: Or replaced by the new HttpResponse, or  response code + message

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/core/v1/PodOperationsImpl.java
```diff
@@ -321,23 +307,32 @@ private Boolean handleEvict(HasMetadata eviction) {
...
+            URL url = getURLWithCommandParams(actualCommands);
+            HttpClient clone = httpClient.newBuilder().readTimeout(0, TimeUnit.MILLISECONDS).build();
+            final ExecWebSocketListener execWebSocketListener = new ExecWebSocketListener(in, out, err, errChannel, inPipe, outPipe, errPipe, errChannelPipe, execListener, bufferSize);
+            CompletableFuture<WebSocket> startedFuture = clone.newWebSocketBuilder()
+                .setHeader("Sec-WebSocket-Protocol", "v4.channel.k8s.io")
+                .uri(url.toURI())
+                .buildAsync(execWebSocketListener);
+            startedFuture.whenComplete((w, t) -> {
+              if (t != null) {
+                execWebSocketListener.onError​(w, t);
```
manusa, line 319: There's a ZeroWidthCharacter between `onError` and `(w, t)`

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/uploadable/PodUploadWebSocketListener.java
```diff
@@ -45,39 +43,28 @@
...
-            return rb.string();
-          } catch (IOException e) {
-            return e.getMessage();
-          }
-        }).orElse("");
-      error.set(String.format("%s - %s%n%s", response.code(), response.message(), responseBody));
-    } else {
-      error.set("PodUploadWebSocketListener failed with no response");
-    }
+  public void onError​(WebSocket webSocket, Throwable t) {
```
manusa, line 66: There's another Zero Width Character between onError and (, maybe it's worth doing a quick find to remove them.

manusa:
```
Targets
    Occurrences of '​' in Project
Found Occurrences in Project  (13 usages found)
    ExecWebSocketListener.java  (2 usages found)
        188 onError​(webSocket, e);
        197 public void onError​(WebSocket webSocket, Throwable t) {
    OkHttpWebSocketImpl.java  (1 usage found)
        70 listener.onError​(new OkHttpWebSocketImpl(webSocket), t);
    PodOperationsImpl.java  (1 usage found)
        319 execWebSocketListener.onError​(w, t);
    PodUpload.java  (1 usage found)
        177 podUploadWebSocketListener.onError​(w, t);
    PodUploadWebSocketListener.java  (1 usage found)
        66 public void onError​(WebSocket webSocket, Throwable t) {
    PodUploadWebSocketListenerTest.java  (1 usage found)
        68 podUploadWebSocketListener.onError​(null, new IOException("here"));
    PortForwarderWebsocket.java  (2 usages found)
        274 public void onError​(WebSocket webSocket, Throwable t) {
        321 listener.onError​(w, t);
    README.md  (1 usage found)
        104 - kondoumh: [​Kubernetes ネイティブなカオスエンジニアリングツール Chaos Mesh を使ってみる](https://blog.kondoumh.com/entry/2020/10/23/123431)
    WatchConnectionManager.java  (1 usage found)
        125 listener.onError​(w, t);
    WatcherWebSocketListener.java  (1 usage found)
        49 public void onError​(WebSocket webSocket, Throwable t) {
    WebSocket.java  (1 usage found)
        35 default void onError​(WebSocket webSocket, Throwable error) {}
```

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/ExecListener.java
shawkins: This is a local ExecListener.Response that contains just the code and body (which should be available on later java versions with the jdk client)

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationSupport.java
shawkins: I can see why this would be done, the slf4j docs should make it clearer about the convention of the last vararg being a throwable.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/uploadable/PodUploadWebSocketListener.java
shawkins: I have no idea how that got in there.  Should be removed now.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/ClientState.java
manusa: Sorry for the delayed reply. I really have no preference about the name for this.

Is this something that users are going to actually implement? I don't see it very likely.

I see that OperationContext extends this interface, so maybe ClientContext might be a reasonable choice.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/ExecListener.java
manusa: :flushed: I missed the line where it was defined. I knew we discussed this internally.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/OperationSupport.java
manusa: OK; I also missed the String.format part :facepalm: 
But, anyway, not urgent but it should be changed to the standard approach.

manusa:
> @manusa it probably goes without saying, but feel free to squash this as the individual commits at this point don't convey much. alternatively I can rebase it if you want.

The commit history for the PR is not adding much value at this point. I do think it's better if we squash.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/http/HttpRequest.java
```diff
@@ -0,0 +1,100 @@
...
+
+    default Builder post(String contentType, String writeValueAsString) {
+      return method("POST", contentType, writeValueAsString);
+    }
+    
+    Builder post(String contentType, byte[] writeValueAsBytes);
+    
+    Builder post(String contentType, InputStream stream, long length);
+
+    default Builder delete(String contentType, String writeValueAsString) {
```
manusa, line 53: Out of curiosity, are we sure we want/need to provide a delete with body method?
Is any of the Kubernetes HTTP endpoints actually expecting requests like this one?

manusa: arrg. OK, delete options go in the body of the request.

manusa: https://github.com/kubernetes/kubernetes/issues/34856

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/ClientState.java
shawkins: > Is this something that users are going to actually implement? I don't see it very likely.

Ideally no.

> I see that OperationContext extends this interface, so maybe ClientContext might be a reasonable choice.

OperationContext, OperationSupport, and Client all extend/implement this interface.  OperationSupport just unwraps/holds an OperationContext, so for clarity we could not extend there and just pass the OperationContext as needed.

I'll update to this new name.

manusa (approved):
LGTM
Awesome job, thank you very much!

manusa:
A basic check with JKube worked OK, everything appears to work as intended :rocket:

PR #3526, opened 2021-10-11
fix #3460: support for deserializing templates with non-string params

shawkins:
## Description
Took another look at the template deserialization issue and thought it would be worth proposing this solution - there does not seem to be a great way to trap this error using other jackson mechanisms.

## Type of change

 - [x] Bug fix (non-breaking change which fixes an issue)
 - [ ] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [ ] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [ ] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [ ] I Added [CHANGELOG](../CHANGELOG.md) entry regarding this change
 - [x] I have implemented unit tests to cover my changes
 - [ ] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/master/doc/CHEATSHEET.md) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [ ] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

Issue #3460, opened 2021-09-10
Openshift InvalidFormatException  Cannot deserialize value for Kind Template


jeromedemangel:
Hello,

With the openshift-client library

```
<dependency>
    <groupId>io.fabric8</groupId>
    <artifactId>openshift-client</artifactId>
    <version>5.7.3</version>
</dependency>
```

I try to get a template :

```
openShiftClient.templates()
    .inNamespace("foo")
    .withName("bar")
    .get()
```
The template contains an object with :

```
securityContext:
  runAsUser: ${USERID}
```

And a parameter :

```
parameters:
- name: USERID
  required: true
```

When executing the Java code, the error is thrown

```
Caused by: com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type `long` from String "${USERID}": not a valid `long` value
 at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: io.fabric8.openshift.api.model.TemplateList["items"]->java.util.ArrayList[5]->io.fabric8.openshift.api.model.Template["objects"]->java.util.ArrayList[0]->io.fabric8.kubernetes.api.model.batch.v1.Job["spec"]->io.fabric8.kubernetes.api.model.batch.v1.JobSpec["template"]->io.fabric8.kubernetes.api.model.PodTemplateSpec["spec"]->io.fabric8.kubernetes.api.model.PodSpec["securityContext"]->io.fabric8.kubernetes.api.model.PodSecurityContext["runAsUser"])
```

The runAsUser is effectivly a long in the destination object, but in the template It's a string refered by the parameter

sknot-rh:
Hello, I am having a similar issue. I am trying to load an object with the `Integer` placeholder however, it is failing. Minimal reproducer:

Run this test in `io.fabric8.openshift.client.server.mock.LoadAsTemplateTest`
```
  @Test
  void unmarshallTemplateObjectWithIntegerPlaceholder() {
    Template template = Serialization.unmarshal(getClass().getResourceAsStream("/template-with-json-params.yml"), Template.class);
  }
```
It fails with `Cannot deserialize value of type java.lang.Integer from String "${{CONTAINER_PORT}}": not a valid Integer value` as the validator expects `Integer` value but the placeholder `String` is provided.

shawkins:
[...]

Issue #1979, opened 2020-02-02
Error when load yaml with parameters


marlonpatrick:
To reproduce the error run the code below with an multi-document yaml. Parameters must have the format `${param1}`. 

    public static void main(String[] args) {
    	try(DefaultKubernetesClient client = new DefaultKubernetesClient()){
    		
        	Map<String, String> parameters = new HashMap<>();
        	parameters.put("param1", "value1");
        	parameters.put("param2", "value2");

            client.load(MyClass.class.getResourceAsStream("/my-multiple-document.yaml"))
                	.withParameters(parameters).inNamespace("myNamespace").get();
    	}
    }

Log:

```
io.fabric8.kubernetes.client.KubernetesClientException: An error has occurred.
        at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:64)
        at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:53)
        at io.fabric8.kubernetes.client.utils.Serialization.unmarshal(Serialization.java:134)
        at io.fabric8.kubernetes.client.utils.Serialization.unmarshal(Serialization.java:95)
        at io.fabric8.kubernetes.client.dsl.internal.NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.<init>(NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.java:228)
        at io.fabric8.kubernetes.client.dsl.internal.NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.withParameters(NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.java:185)
        at io.fabric8.kubernetes.client.dsl.internal.NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.withParameters(NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.java:64)

Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: No content to map due to end-of-input
 at [Source: (BufferedInputStream); line: 1, column: 1]
        at com.fasterxml.jackson.databind.exc.MismatchedInputException.from(MismatchedInputException.java:59)
        at com.fasterxml.jackson.databind.DeserializationContext.reportInputMismatch(DeserializationContext.java:1442)
[...]

Issue #959, opened 2017-12-21
Template doesn't support ${{ }}


ckanner:
A Template has a number parameter ```PORT```. If i use it with ```${{PORT}}```,  the kubernetes client doesn't support. But When i use it with ```${PORT}```, it success.

novakov-alexey:
Latest release still has this issue, no matter what I use `{{}}` or `{}` braces

Version:
"io.fabric8" % "openshift-client" % "4.4.1"

```
Caused by: com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type `java.lang.Integer` from String "${BROKER_REPLICAS}": not a valid Integer value
 at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: io.fabric8.openshift.api.model.Template["objects"]->java.util.ArrayList[4]->io.fabric8.kubernetes.api.model.apps.StatefulSet["spec"]->io.fabric8.kubernetes.api.model.apps.StatefulSetSpec["replicas"])
  at com.fasterxml.jackson.databind.exc.InvalidFormatException.from(InvalidFormatException.java:67)
  at com.fasterxml.jackson.databind.DeserializationContext.weirdStringException(DeserializationContext.java:1549)
  at com.fasterxml.jackson.databind.DeserializationContext.handleWeirdStringValue(DeserializationContext.java:911)
  at com.fasterxml.jackson.databind.deser.std.NumberDeserializers$IntegerDeserializer._parseInteger(NumberDeserializers.java:522)
  at com.fasterxml.jackson.databind.deser.std.NumberDeserializers$IntegerDeserializer.deserialize(NumberDeserializers.java:474)
  at com.fasterxml.jackson.databind.deser.std.NumberDeserializers$IntegerDeserializer.deserialize(NumberDeserializers.java:452)
  at com.fasterxml.jackson.databind.deser.impl.MethodProperty.deserializeAndSet(MethodProperty.java:129)
  at com.fasterxml.jackson.databind.deser.BeanDeserializer.vanillaDeserialize(BeanDeserializer.java:288)
  at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:151)
  at com.fasterxml.jackson.databind.deser.impl.MethodProperty.deserializeAndSet(MethodProperty.java:129)
  at com.fasterxml.jackson.databind.deser.BeanDeserializer.vanillaDeserialize(BeanDeserializer.java:288)
[...]

PR #3574, opened 2021-11-08
UnmatchedFieldTypeModule to deal with mismatched target types for fields

manusa:
## Description

This is a partial implementation of what's described in https://github.com/fabric8io/kubernetes-client/issues/3460#issuecomment-925729235

The new behavior is as follows (edited after https://github.com/fabric8io/kubernetes-client/pull/3574#issuecomment-963379719):
1. A resource is deserialized into a KubernetesResource instance.
2. The resource can contain unknown fields and/or fields with mismatched types (e.g. DesploymentSpec#replicas with a String instead of an Integer)
3. Properties matching any of the categories mentioned in the previous step can stored in the `additionalProperties` Map (if present)
  a. By default, this unmatched types deserialization is enabled only for Templates
  b. Behavior can be enabled for all types by `Serialization.UNMATCHED_FIELD_TYPE_MODULE.setRestrictToTemplates(false);`
4. The resource is deserialized as a valid KubernetesResource instance (if applicable), no need to fallback into a GenericKubernetesResource
5. The object instance can now be processed as usual.
6. The object instance is serialized into a YAML/JSON representation
7. The object fields are persisted **unless** a property matching the field name is defined in the `additionalProperties` Map. In that case the field value is **ignored**.
  a. A warning will be logged if the situation is detected, indicating that a field is being ignored.
  b. Log can be opted-out by: `Serialization.UNMATCHED_FIELD_TYPE_MODULE.setLogWarnings(false);` 
8. **All** properties defined in the `additionalProperties` Map are serialized.

All of this logic is implemented in the  `UnmatchedFieldTypeModule` Jackson module that sets a serializer and a deserializer modifier. The second commit enables this module by default.

The only drawback when enabling the module is that Exceptions when deserializing input resources that have properties that don't match the target field type  won't be thrown. Any user relying on this to validate an Input will lose this functionality (I'm not sure if this is an extended approach).

I would like to know if we should preserve the second commit (where the module is automatically enabled), or revert it and just provide the module as an opt-in feature. (@shawkins, @metacosm, @rohanKanojia, @oscerd, @iocanel) 

The second commit also reverts the fallback implemented in #3526 which should no longer be necessary.

### Relates to:
- #448 
- #959
- #1979
- https://github.com/eclipse/jkube/issues/778

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 #### Improvements
 * Fix #3562: Kubernetes Mock Server improvements
+* Fix #3574: support for deserialization of properties that don't match the target field's type
 
 #### Dependency Upgrade
 * Fix #3562: Bump MockWebServer
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,10 @@
 * Fix #3430: Support Vertical Pod Autoscaler
 
 #### _**Note**_: Breaking changes in the API
-
+##### Tools Changes:
+- Serialization: Those KubernetesResources that include entries in the additionalProperties Map that override a field
+  of the resource instance, will no longer be duplicated. The values present in the additionalProperties Map take
+  precedence.
 
 ### 5.9.0 (2021-10-14)
 
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/Serialization.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/Serialization.java
@@ -60,14 +60,47 @@ public class Serialization {
 
   private static final String DOCUMENT_DELIMITER = "---";
 
+  /**
+   * {@link ObjectMapper} singleton instance used internally by the Kubernetes client.
+   *
+   * <p> The ObjectMapper has an {@link UnmatchedFieldTypeModule} module registered. This module allows the client
+   * to work with Resources that contain properties that don't match the target field type. This is especially useful
+   * and necessary to work with OpenShift Templates.
+   *
+   * <p> n.b. the use of this module gives precedence to properties present in the additionalProperties Map present
+   * in most KubernetesResource instances. If a property is both defined in the Map and in the original field, the
+   * one from the additionalProperties Map will be serialized.
+   */
   public static ObjectMapper jsonMapper() {
     return JSON_MAPPER;
   }
 
+  /**
+   * {@link ObjectMapper} singleton instance used internally by the Kubernetes client.
+   *
+   * <p> The ObjectMapper has an {@link UnmatchedFieldTypeModule} module registered. This module allows the client
+   * to work with Resources that contain properties that don't match the target field type. This is especially useful
+   * and necessary to work with OpenShift Templates.
+   *
+   * <p> n.b. the use of this module gives precedence to properties present in the additionalProperties Map present
+   * in most KubernetesResource instances. If a property is both defined in the Map and in the original field, the
+   * one from the additionalProperties Map will be serialized.
+   */
   public static ObjectMapper yamlMapper() {
     return YAML_MAPPER;
   }
 
+  /**
+   * Returns a JSON representation of the given object.
+   *
+   * <p> If the provided object contains a JsonAnyGetter annotated method with a Map that contains an entry that
+   * overrides a field of the provided object, the Map entry will take precedence upon serialization. Properties won't
+   * be duplicated.
+   *
+   * @param object the object to serialize.
+   * @param <T> the type of the object being serialized.
+   * @return a String containing a JSON representation of the provided object.
+   */
   public static <T> String asJson(T object) {
     try {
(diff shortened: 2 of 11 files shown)


oscerd (approved):

oscerd:
I would preserve the second commit

shawkins:
> or revert it and just provide the module as an opt-in feature.

For serialization the only practical difference compared to existing behavior is that if you set the field and additional properties it gets serialized twice - which should be invalid yaml.  It might be good to throw an exception in this case - that will make it clear there's an ambiguity.

> Any user relying on this to validate an Input will lose this functionality (I'm not sure if this is an extended approach).

That's a valid concern for deserialization.  Seems like this behavior would be best enabled only for templates.

manusa:
>  It might be good to throw an exception in this case - that will make it clear there's an ambiguity.

The custom serializer prevents exactly that. The value in additionalProperties takes precedence over any other and it only gets serialized **only once**. 

This circumstance could happen before (in fact we've seen it in some issue reports #3195). The new behavior will at least provide a valid YAML when serializing.

Maybe we can log a warning or an info message if the circumstance is detected (when the field is ignored).

manusa:
> Seems like this behavior would be best enabled only for templates.

The DeserializerModifier is applied per bean. There is no easy way to detect if the current bean (or its parents) is part of a list of Objects within a Template when updating the Deserializer builder.

We can only do something like this at the SettableBeanPropertyDelegate deserializeAndSetMethod and check the JsonParser tree. But it might not look very nice. I'll give it a try.

manusa:
Hi @shawkins
Both of your suggestions should be addressed by the 3rd (logs) commit and the 4th commit (only enabled for templates).

shawkins:
> Both of your suggestions should be addressed by the 3rd (logs) commit and the 4th commit (only enabled for templates).

For a warning by default should the additional properties be java-doc'd as having primacy over regular properties? - perhaps added to an interface like strimzi defines https://github.com/strimzi/strimzi-kafka-operator/blob/79e28149b6dd8dd7b61e0ed60aae6d26fb93b1b7/api/src/main/java/io/strimzi/api/kafka/model/UnknownPropertyPreserving.java

Also this logic for isInTemplate won't work - it needs to preserve the state top down, if you have a template like:

```
---
apiVersion: v1
kind: Template
metadata:
  name: foo
parameters:
- name: PORT
  value: 80
objects:
- apiVersion: v1
  kind: Service
  metadata:
    name: foo
  spec:
    ports:
    - port: 8080
    selector:
      cheese: edam
    type: ExternalName
- apiVersion: v1
  kind: Service
  metadata:
    name: bar
  spec:
    ports:
    - port: ${PORT}
    selector:
      cheese: edam
    type: ExternalName
```
Where you hit another object before the one with the parameters, then the flag will get flipped prematurely.

manusa:
> For a warning by default should the additional properties be java-doc'd as having primacy over regular properties?

Well, the circumstance before was worst because the user ended up with an invalid JSON or YAML with duplicate entries. Now the user gets a warning in case the ambiguity is detected and ends up with a valid resource (but maybe unexpected).

This _default_ is only applicable if the user is actually using the ObjectMappers provided by the [Serialization](https://github.com/fabric8io/kubernetes-client/blob/09252821b7d12ecb818aa8ddc36ef16c84501b48/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/Serialization.java) class, so I'm not sure if the Javadoc should include that entry. If it did we should add the conditional.

In any case, I'm not sure how we could add this Javadoc easily to the project, since not all KubernetesResources might include the additional properties. Any ideas?

> Where you hit another object before the one with the parameters, then the flag will get flipped prematurely.

Should be fixed now :heavy_check_mark:

shawkins:
> Well, the circumstance before was worst because the user ended up with an invalid JSON or YAML with duplicate entries. Now the user gets a warning in case the ambiguity is detected and ends up with a valid resource (but maybe unexpected).

The two concerns are that it's a behavioral change and it's subtle.  Historically there's a lot of warnings generated by fabric8, so some people will have that logging turned off entirely.

Just in case it should be documented somehow and in the release notes.

> In any case, I'm not sure how we could add this Javadoc easily to the project, since not all KubernetesResources might include the additional properties. Any ideas?

It would be up to the generation logic right?  The interface would have to be added into the javaInterfaces list when additionalProperties is true.  But that does seem like a lot of work.

manusa:
> It would be up to the generation logic right? The interface would have to be added into the javaInterfaces list when additionalProperties is true. But that does seem like a lot of work.

This is the only way I can think of, besides something similar to https://github.com/fabric8io/kubernetes-client/issues/3570#issuecomment-963200519. But both approaches imply extra work, that I really don't think adds any value.

I'll add an entry to the changelog.

Maybe some notes on the marshal/unmarhsal methods of the Serializer make sense too.

shawkins (approved):
LGTM

rohanKanojia (approved):

PR #3576, opened 2021-11-09
fix #3570: using lombok to add the additionalProperties setter

### LINKED ISSUE #3570: Builders do not preserve additionalProperties
When using a builder and you use a withXXX to associate another object - under the covers the a copy is made via a constructor that omits the additional properties.

shawkins:
## Description
fix for #3570 - since we have a simple mechanism for adding annotations, we'll rely on lombok for the additionalProperties setter.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Bugs
 * Fix #3408: quote pod upload file paths to support special chars
 * Fix #3561: ensure okhttp resources are closed
+* Fix #3570: added a setter for additionalProperties for proper builder support
 
 #### Improvements
 * Fix #3562: Kubernetes Mock Server improvements
--- a/extensions/camel-k/model-v1/src/main/java/io/fabric8/camelk/v1/TraitConfiguration.java
+++ b/extensions/camel-k/model-v1/src/main/java/io/fabric8/camelk/v1/TraitConfiguration.java
@@ -15,8 +15,6 @@
  */
 package io.fabric8.camelk.v1;
 
-import java.util.HashMap;
-import java.util.Map;
 import com.fasterxml.jackson.annotation.JsonAnyGetter;
 import com.fasterxml.jackson.annotation.JsonAnySetter;
 import com.fasterxml.jackson.annotation.JsonIgnore;
@@ -34,7 +32,12 @@ import io.fabric8.kubernetes.api.model.VolumeMount;
 import io.sundr.builder.annotations.Buildable;
 import io.sundr.builder.annotations.BuildableReference;
 import lombok.EqualsAndHashCode;
+import lombok.Setter;
 import lombok.ToString;
+import lombok.experimental.Accessors;
+
+import java.util.HashMap;
+import java.util.Map;
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonPropertyOrder({
@@ -43,6 +46,11 @@ import lombok.ToString;
 @JsonDeserialize(using = com.fasterxml.jackson.databind.JsonDeserializer.None.class)
 @ToString
 @EqualsAndHashCode
+@Setter
+@Accessors(prefix = {
+    "_",
+    ""
+})
 @Buildable(editableEnabled = true, validationEnabled = false, generateBuilderPackage = false, builderPackage = "io.fabric8.kubernetes.api.builder", refs = {
     @BuildableReference(ObjectMeta.class),
     @BuildableReference(LabelSelector.class),
--- a/extensions/certmanager/model-v1/src/main/java/io/fabric8/certmanager/api/model/acme/v1/ACMEChallengeSolverHTTP01Ingress.java
+++ b/extensions/certmanager/model-v1/src/main/java/io/fabric8/certmanager/api/model/acme/v1/ACMEChallengeSolverHTTP01Ingress.java
@@ -15,8 +15,6 @@
  */
 package io.fabric8.certmanager.api.model.acme.v1;
 
-import java.util.HashMap;
-import java.util.Map;
 import com.fasterxml.jackson.annotation.JsonAnyGetter;
 import com.fasterxml.jackson.annotation.JsonAnySetter;
 import com.fasterxml.jackson.annotation.JsonIgnore;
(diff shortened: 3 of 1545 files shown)


shawkins:
@manusa this seems to work - the sundrio processing is after lombok adds the setter.  I had to use the Accessors annotation as well to strip the _ prefix, which is used for any field names that are reserved words.

manusa:
:tada: I can't believe this worked. So simple.

Could you please add a test to verify that the initial issue is fixed?

The extension models should also be annotated. These also include persisted model classes that aren't generated, so they need to have the setter added manually.

shawkins:
> Could you please add a test to verify that the initial issue is fixed?

It's buried in there https://github.com/fabric8io/kubernetes-client/blob/b1a520bb204ed0c1bb320f90ea2b5608a0350915/kubernetes-model-generator/kubernetes-model-core/src/test/java/io/fabric8/kubernetes/api/model/AdditionalPropertiesTest.java

I don't think sonar cloud understands how tests are relevant to annotations.

> The extension models should also be annotated. These also include persisted model classes that aren't generated, so they need to have the setter added manually.

Which ones are those?

manusa:
> It's buried in there...

Sorry, I promise I try to check all of the files.

> Which ones are those?

 `find . -type f -name "*.java" -not -path '*/generated/*' -exec grep -q -e '@Buildable(' {} \; -print`

They are scattered everywhere.

shawkins:
> They are scattered everywhere.

Thanks.  It's not elegant, but the full list should match:

```
find . -type f -name "*.java" -not -path '*/generated/*' ! -path '*/target/*' -exec grep -q -e '@Buildable(' {} \; -print | while read line ; do
  grep additionalP $line -l
done
```

I just manually updated from there, and also updated the other model annotator.

The .json files got checked in along with everything else in the initial commit - just let me know if you want to exclude those.

manusa:
For some reason the current changes are modifying the order of fields when persisting (I run that test 10,000 times locally to actually verify this):
```
 Error:  Failures: 
Error:    SerializationAdditionalPropertiesTest.marshalWithAdditionalPropertiesOverridingFields:141 
expected: "{"apiVersion":"v1","kind":"ConfigMap","metadata":{"name":"name","annotations":"${annotations}"},"immutable":"${immutable}","unknownField":"unknownValue"}"
 but was: "{"apiVersion":"v1","kind":"ConfigMap","metadata":{"name":"name","annotations":"${annotations}"},"unknownField":"unknownValue","immutable":"${immutable}"}"
Error:    UnmatchedFieldTypeModuleTest.writeValueAsStringWithAdditionalPropertiesOverridingFields:164 
expected: "{"apiVersion":"v1","kind":"ConfigMap","metadata":{"labels":{"lKey":"value"},"name":"name","annotations":"${annotations}"},"immutable":"${immutable}","unknownField":"unknownValue"}"
 but was: "{"apiVersion":"v1","kind":"ConfigMap","metadata":{"labels":{"lKey":"value"},"name":"name","annotations":"${annotations}"},"unknownField":"unknownValue","immutable":"${immutable}"}"
```
I don't think that the order of the fields is that important, but I'm really curious about that changed.

shawkins:
> I don't think that the order of the fields is that important, but I'm really curious about that changed.

Should be a difference in the jackson logic using field reflection vs method reflection.

I've pulled in the master changes to the branch so I can look at this locally.

shawkins:
> Should be a difference in the jackson logic using field reflection vs method reflection.

That doesn't make sense... Hadn't thought through it's actually getting the properties through the AnyGetter method regardless.

As best as I can tell this ordering is JRE dependent - the additional properties is just a HashMap and the Jackson logic under the covers is creating a MapSerializer with sortKeys=false.  

@manusa the workaround would be to use a serialization feature on the json/yaml serializer:

```
private static final ObjectMapper JSON_MAPPER = new ObjectMapper().enable(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS);
```
The only downside is that will also be new behavior for anyone using the Serialization utility and LinkedHashMaps in their custom objects.

manusa:
> As best as I can tell this ordering is JRE dependent 

My guess when I implemented that test (and verified it always passed) is that Jackson was initializing the map with a LinkedHashMap or something like that in order to preserver ordering.

I'll change the test first thing tomorrow and try to include this in 5.10

shawkins:
> My guess when I implemented that test (and verified it always passed) is that Jackson was initializing the map with a LinkedHashMap or something like that in order to preserver ordering.

That's on the right track.  Here's what's happening -  Prior to this change ConfigMapBuilder.build did nothing to the additionalProperties.  It was left to initialize as a HashMap via the default assignment in the class.  When Jackson serialized the map it was entirely up to the JRE order.

After this change ConfigMapBuilder.build causes the additionalProperties to be set with a LinkedHashMap.  Now the order that you add the properties is meaningful.

manusa:
> After this change ConfigMapBuilder.build causes the additionalProperties to be set with a LinkedHashMap. Now the order that you add the properties is meaningful.

So if we switch the test expectation/assertion, it should pass 100% of times?

shawkins:
> So if we switch the test expectation/assertion, it should pass 100% of times?

Yes after this change if we reorder the puts or the expected result, it will pass 100% of the time.

With the existing state of master there's a chance that the expected result will fail on different JREs.

Beyond these tests though there will still be an issue in general - if you don't use a builder, the order won't be defined.  Either the class generation needs to be update to use LinkedHashMap instead of HashMap or the map order serialization feature mentioned above would need to be used.

manusa:
> Yes after this change if we reorder the puts or the expected result, it will pass 100% of the time.

For completeness:

- `additionalProperties` are set as HashMap when a resource is initialized (e.g. new ConfigMap()).
- When a new Builder is initialized, the field value for fluent instance gets the `additionalProperties` initialized as a `LinkedHashMap`
  - When any *Fluent.withAdditionalProperties is invoked, the original Map is copied into a new LinkedHashMap
    `new LinkedHashMap(this.additionalProperties)`
  - The *Builder **always** calls the `withAdditionalProperties` method in the constructor
    `fluent.withAdditionalProperties(instance.getAdditionalProperties())`

All of this corresponds to Sundrio standard behavior.

In a nutshell, whenever we use a Builder to initialize a resource, there's a guarantee that the additionalProperties field is always post-initialized as a LinkedHashMap, so input order is preserved.

manusa (approved):
LGTM, thx!

rohanKanojia (approved):

PR #3594, opened 2021-11-15
Support auto generation of interfaces by providing implementations

Sgitario:
Given the Golang model:

Go type `Tracing_CustomTag` that uses a field `Type` which is an interface `isTracing_CustomTag_Type`:
```go
type Tracing_CustomTag struct {
	Type                 isTracing_CustomTag_Type `protobuf_oneof:"type"`
}
```

Where the interface is defined in:

```go
type isTracing_CustomTag_Type interface {
	isTracing_CustomTag_Type()
	MarshalTo([]byte) (int, error)
	Size() int
}
```

The implementations of this interface are defined in:

```go
type Tracing_CustomTag_Literal struct {
	Literal *Tracing_Literal `protobuf:"bytes,1,opt,name=literal,proto3,oneof" json:"literal,omitempty"`
}
type Tracing_CustomTag_Environment struct {
	Environment *Tracing_Environment `protobuf:"bytes,2,opt,name=environment,proto3,oneof" json:"environment,omitempty"`
}
type Tracing_CustomTag_Header struct {
	Header *Tracing_RequestHeader `protobuf:"bytes,3,opt,name=header,proto3,oneof" json:"header,omitempty"`
}
```

At the moment, the generator won't inspect all the types to find the implementations, but users would need to specify the implementations in the generator golang program like:

```go
// types for interfaces
	interfacesMapping := map[string][]reflect.Type{
		"istio.io/api/telemetry/v1alpha1/isTracing_CustomTag_Type":     {reflect.TypeOf(api_telemetry_v1alpha1.Tracing_CustomTag_Literal{}), reflect.TypeOf(api_telemetry_v1alpha1.Tracing_CustomTag_Environment{}), reflect.TypeOf(api_telemetry_v1alpha1.Tracing_CustomTag_Header{})},
	}
```

Additionally, as json2pojo does not support the generation of interfaces, we need to use an custom registry. In the pom.xml of the model:

```xml
      <plugin>
        <groupId>org.jsonschema2pojo</groupId>
        <artifactId>jsonschema2pojo-maven-plugin</artifactId>
        <version>${jsonschema2pojo.version}</version>
        <configuration>
          <sourceDirectory>${project.basedir}/src/main/resources/schema</sourceDirectory>
          <targetPackage>io.fabric8.istio.api</targetPackage>
          <includeConstructors>true</includeConstructors>
          <includeJsr303Annotations>false</includeJsr303Annotations>
          <includeToString>false</includeToString>
          <includeHashcodeAndEquals>false</includeHashcodeAndEquals>
          <outputDirectory>${project.build.directory}/generated-sources</outputDirectory>
          <customRuleFactory>io.fabric8.kubernetes.jsonschema2pojo.Fabric8RuleFactory</customRuleFactory><!-- HERE -->
```

Finally, to use the new API, we need to use a new method to generate the schema with all the options:

```go
json := schemagen.GenerateSchemaWithAllOptions("http://fabric8.io/istio/IstioSchema#", crdLists, typesDescriptors, providedPackages, manualTypeMap, packageMapping, mappingSchema, providedTypes, constraints, interfacesMapping, "io.fabric8")
```

This has been tested using this branch: https://github.com/Sgitario/kubernetes-client/tree/add_istio. Concretely, in the new extension for Istio: https://github.com/Sgitario/kubernetes-client/tree/add_istio/extensions/istio

### DIFF
--- a/generator/pkg/schemagen/generate.go
+++ b/generator/pkg/schemagen/generate.go
@@ -34,6 +34,14 @@ type PackageInformation struct {
 	JavaPackage string
 }
 
+// Name strategy mapping for auto discovered classes
+type JavaNameStrategyMapping struct {
+	// To provide a custom generic rule that applies to all java, interface, enum
+	CustomJavaNameRule func(packageName *string, javaName *string)
+	// To manually map a golang class to a Java class
+	NameMapping map[reflect.Type]string
+}
+
 type schemaGenerator struct {
 	crdLists             map[reflect.Type]CrdScope
 	types                map[reflect.Type]*JSONObjectDescriptor
@@ -44,6 +52,7 @@ type schemaGenerator struct {
 	providedTypes        []ProvidedType
 	constraints          map[reflect.Type]map[string]*Constraint // type -> field name -> constraint
 	interfacesMapping    map[string][]reflect.Type               // interface -> list of implementations
+	javaNameStrategy     JavaNameStrategyMapping                 // java name strategy by extension
 	generatedTypesPrefix string
 }
 
@@ -61,10 +70,10 @@ const (
 )
 
 func GenerateSchema(schemaId string, crdLists map[reflect.Type]CrdScope, providedPackages map[string]string, manualTypeMap map[reflect.Type]string, packageMapping map[string]PackageInformation, mappingSchema map[string]string, providedTypes []ProvidedType, constraints map[reflect.Type]map[string]*Constraint, generatedTypesPrefix string) string {
-	return GenerateSchemaWithAllOptions(schemaId, crdLists, make(map[reflect.Type]*JSONObjectDescriptor), providedPackages, manualTypeMap, packageMapping, mappingSchema, providedTypes, constraints, make(map[string][]reflect.Type), generatedTypesPrefix)
+	return GenerateSchemaWithAllOptions(schemaId, crdLists, make(map[reflect.Type]*JSONObjectDescriptor), providedPackages, manualTypeMap, packageMapping, mappingSchema, providedTypes, constraints, make(map[string][]reflect.Type), JavaNameStrategyMapping{}, generatedTypesPrefix)
 }
 
-func GenerateSchemaWithAllOptions(schemaId string, crdLists map[reflect.Type]CrdScope, typesDescriptors map[reflect.Type]*JSONObjectDescriptor, providedPackages map[string]string, manualTypeMap map[reflect.Type]string, packageMapping map[string]PackageInformation, mappingSchema map[string]string, providedTypes []ProvidedType, constraints map[reflect.Type]map[string]*Constraint, interfacesMapping map[string][]reflect.Type, generatedTypesPrefix string) string {
+func GenerateSchemaWithAllOptions(schemaId string, crdLists map[reflect.Type]CrdScope, typesDescriptors map[reflect.Type]*JSONObjectDescriptor, providedPackages map[string]string, manualTypeMap map[reflect.Type]string, packageMapping map[string]PackageInformation, mappingSchema map[string]string, providedTypes []ProvidedType, constraints map[reflect.Type]map[string]*Constraint, interfacesMapping map[string][]reflect.Type, javaNameStrategy JavaNameStrategyMapping, generatedTypesPrefix string) string {
 	g := &schemaGenerator{
 		crdLists:             crdLists,
 		types:                typesDescriptors,
@@ -76,6 +85,7 @@ func GenerateSchemaWithAllOptions(schemaId string, crdLists map[reflect.Type]Crd
 		constraints:          constraints,
 		generatedTypesPrefix: generatedTypesPrefix,
 		interfacesMapping:    interfacesMapping,
+		javaNameStrategy:     javaNameStrategy,
 	}
 	schema, err := g.generate(schemaId, crdLists)
 
@@ -186,6 +196,13 @@ func (g *schemaGenerator) javaType(t reflect.Type) string {
 
 	// follow pointer
 	t = g.resolvePointer(t)
+
+	if g.javaNameStrategy.NameMapping != nil {
+		found, ok := g.javaNameStrategy.NameMapping[t]
+		if ok {
+			return found
+		}
+	}
 
 	manualType, ok := g.manualTypeMap[t]
 	if ok {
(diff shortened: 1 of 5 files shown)


iocanel (approved):
LGMT

oscerd (approved):

Sgitario:
Updated PR as the classes using an interface field needs to be annotated with `@JsonUnwrapped`.

model-annotator/src/main/java/io/fabric8/kubernetes/ModelRuleFactory.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Model rule factory to support mapping of interfaces.
+ */
+public class ModelRuleFactory extends RuleFactory {
+
+  @Override
+  public Rule<JPackage, JType> getObjectRule() {
+    return new ModelObjectRule(this, new ParcelableHelper(), getReflectionHelper());
+  }
+}
```
manusa, line 34: We already have a [RuleFactory](https://github.com/fabric8io/kubernetes-client/blob/4eee56eb02cc3880435bcf54aacd86c735205013/kubernetes-model-generator/kubernetes-model-jsonschema2pojo/src/main/java/io/fabric8/kubernetes/jsonschema2pojo/Fabric8RuleFactory.java) and a [custom rule](https://github.com/fabric8io/kubernetes-client/blob/b0890fa0535818f066a76c67311b0b6c5a7b6f92/kubernetes-model-generator/kubernetes-model-jsonschema2pojo/src/main/java/io/fabric8/kubernetes/jsonschema2pojo/Fabric8NameHelper.java) for the Name Helper.

Maybe we could add the new rule to the existing Fabric8RuleFactory and avoid the creation of another Factory.
i.e. All Fabric8 jsonschema rule customizations:
- Fabric8RuleFactory
  - Fabric8NameHelper
  - Fabric8ObjectRule

Sgitario: +1 will do it

Sgitario: updated

Sgitario:
PR updated to also support Golang structs using interfaces like:

```go
type isTracing_CustomTag_Type struct {
    Tag interface{}
}
```

This is now mapped to:

```java
public class isTracing_CustomTag_Type {
    // ...
    private Object tag;
}
```

metacosm:
This looks like a good first step but I think we should think about adopting some sort of convention to rename Go types to more standard Java type names in particular because this will affect how natural the fluent interfaces will look like…

Sgitario:
> isTracing_CustomTag_Type

I could certainly rename the interface class from `isTracing_CustomTag_Type` to `IsTracing_CustomTag_Type`, but not to `CustomTagType` as this is a rule only for Istio, not for all. 

About removing the `_` characters, is there some in place to remove them or apply some common name convention? @manusa

generator/pkg/schemagen/generate.go
```diff
@@ -588,7 +660,17 @@ func (g *schemaGenerator) isPartOfOwnPackage(t reflect.Type) bool {
 }
 
 func (g *schemaGenerator) resolveJavaClassUsingOwnPackages(t reflect.Type) string {
-	return g.packageMapping[t.PkgPath()].JavaPackage + "." + t.Name()
+	return g.adaptJavaClassName(g.packageMapping[t.PkgPath()].JavaPackage + "." + t.Name())
+}
+
+func (g *schemaGenerator) adaptJavaClassName(className string) string {
```
Sgitario, line 666: @metacosm @manusa I've created a new method where all the java classes (including the interfaces) will use to adapt the final java name.
The only rule in place will modify names like `is_Something` to `Is_Something`. 
I would happy to also add a new rule to also change from `Is_Something` to `IsSomething` but I'm afraid that this can bring breaking changes to existing extensions. Wdyt?

Sgitario:
I've updated the PR with the following changes:
- Added common method `adaptJavaName` to adapt all the java classes names (including interfaces)
- Add Java Name rule to rename things like `is_Class` to `Is_Class`
- Allow to add custom java name rules in extensions. For example:

```go
// custom name rules
	javaNameRules := []schemagen.JavaNameRule{
		func(packageName *string, className *string) {
			*className = strings.Replace(className, "_", "")
		},
	}

	json := schemagen.GenerateSchemaWithAllOptions("http://fabric8.io/istio/IstioSchema#", crdLists, typesDescriptors, providedPackages, manualTypeMap, packageMapping, mappingSchema, providedTypes, constraints, interfacesMapping, javaNameRules, "io.fabric8")
```

generator/pkg/schemagen/generate.go
metacosm: We shouldn't break existing extensions at this point but maybe that's something to consider for 6.0?

metacosm (approved):

generator/pkg/schemagen/generate.go
manusa: If you're talking about a jsonschema rule here, I think it's better not too.

We should try to avoid changes to the current Extensions APIs.

Also we need to keep in mind, in the (near) future, both generators should be merged, so every change added to this generators should be considered when merging both.

Sgitario: @manusa at the moment, the only rule is to parse java names from is_Class to Is_Class. The rest is to allow extensions adapt their model with custom java names rule (something we need for the Istio extension). So, I think it's ok and it won't break anything. Anyway, let me know if you want to change something from this PR before being merged.

Sgitario:
FYI: There are still two gaps to be addressed in this PR:
- The new field `javaNameRules` in the generator does not need to be an array
- There is still no way to rename a golang class to a java class. Example: `networking.VirtualService` to `io.any.VirtualServiceSpec`

Sgitario:
New changes related to the java name strategy rule. Usage:

```go
javaNameStrategyMapping := schemagen.JavaNameStrategyMapping{
		NameMapping: map[reflect.Type]string{
			reflect.TypeOf(api_telemetry_v1alpha1.Telemetry{}):           "io.fabric8.istio.api.telemetry.v1alpha1.TelemetrySpec",
		},
		CustomJavaNameRule: func(packageName *string, className *string) {
			// remove underscores
			*className = strings.Replace(*className, "_", "", -1)
		},
	}
```

manusa (approved):
LGTM, thx!

rohanKanojia (approved):

PR #3601, opened 2021-11-16
feat: add owner reference support to HasMetadata

### LINKED ISSUE #3600: `HasMetadata` should provide utility method to deal with owner references
In particular to create and add owner references directly from another `HasMetadata` object.

metacosm:
## Description
Fixes #3600

## Type of change

## Checklist

### DIFF
--- a/kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
+++ b/kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
@@ -194,6 +194,7 @@ public interface HasMetadata extends KubernetesResource {
     return getFinalizers().contains(finalizer);
   }
 
+  @JsonIgnore
   default List<String> getFinalizers() {
     return optionalMetadata().map(ObjectMeta::getFinalizers).orElse(Collections.emptyList());
   }
--- a/kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
+++ b/kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
@@ -339,7 +339,7 @@ public interface HasMetadata extends KubernetesResource {
       .withName(metadata.getName())
       .withKind(owner.getKind())
       .build();
-    return addOwnerReference(ownerReference);
+    return addOwnerReference(sanitizeAndValidate(ownerReference));
   }
 
   /**
--- a/kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
+++ b/kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
@@ -257,14 +257,32 @@ public interface HasMetadata extends KubernetesResource {
     return finalizers.contains(finalizer) && finalizers.remove(finalizer);
   }
 
+  /**
+   * Checks whether the provided {@code HasMetadata} is defined as an owner for this {@code HasMetadata}.
+   *
+   * @param owner the {@code HasMetadata} to check for potential ownership
+   * @return {@code true} if the provided {@code HasMetadata} is an owner of this instance
+   */
   default boolean hasOwnerReferenceFor(HasMetadata owner) {
     return getOwnerReferenceFor(owner).isPresent();
   }
 
+  /**
+   * Checks whether the provided UID identifies an owner for this {@code HasMetadata}.
+   *
+   * @param ownerUid the UID of a {@code HasMetadata} to check for potential ownership
+   * @return {@code true} if the provided {@code HasMetadata} is an owner of this instance
+   */
   default boolean hasOwnerReferenceFor(String ownerUid) {
     return getOwnerReferenceFor(ownerUid).isPresent();
   }
 
+  /**
+   * Retrieves the {@link OwnerReference} associated with the specified owner if it's part of this {@code HasMetadata}'s owners.
+   *
+   * @param owner the potential owner of which we want to retrieve the associated {@link OwnerReference}
+   * @return an {@link Optional} containing the {@link OwnerReference} associated with the specified owner if it exists or {@link Optional#empty()} otherwise.
+   */
   default Optional<OwnerReference> getOwnerReferenceFor(HasMetadata owner) {
     if (owner == null) {
       return Optional.empty();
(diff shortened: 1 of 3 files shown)


kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
```diff
@@ -232,4 +235,109 @@ default boolean isFinalizerValid(String finalizer) {
...
+    if (ownerUid == null || ownerUid.isEmpty()) {
+      return Optional.empty();
+    }
+
+    return getMetadata().getOwnerReferences().stream()
+      .filter(o -> ownerUid.equals(o.getUid()))
+      .findFirst();
+  }
+
+  default OwnerReference addOwnerReference(HasMetadata owner) {
```
iocanel, line 266: With a quick glance, it looks as if the purpose of this code is validation. 
However, these are validations that won't be performed when the user is creating objects using the builders.

So, I think that a more suitable place for these validations is in the constructor of `OwnerReference` and `ObjectMeta`.
And yes, I think that OwnerRefernce is generated, so way may consider managing it ourselves.

iocanel (changes requested):

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
metacosm: Sure, however, the same could have been said of the finalizer support that got added a while ago… so is your opinion that we should that code to `ObjectMeta` as well?

metacosm: Another issue with moving the validation to `OwnerReference`'s constructor is the fact that there needs to be a no-arg constructor for serialization meaning that we cannot ensure that the "appropriate" constructor will be called. 
I agree that design-wise it would make more sense to move the code elsewhere, however that would mean managing manually two additional classes: `ObjectMeta` and `OwnerReference`.

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
```diff
@@ -232,4 +235,109 @@ default boolean isFinalizerValid(String finalizer) {
...
+  default boolean hasOwnerReferenceFor(String ownerUid) {
+    return getOwnerReferenceFor(ownerUid).isPresent();
+  }
+
+  default Optional<OwnerReference> getOwnerReferenceFor(HasMetadata owner) {
+    if (owner == null) {
+      return Optional.empty();
+    }
+
+    final String ownerUID = owner.getMetadata().getUid();
```
shawkins, line 252: Calls to getMetadata() on passed in objects and even this object can be null, is that assumed to be a usage error?

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
```diff
@@ -232,4 +235,109 @@ default boolean isFinalizerValid(String finalizer) {
...
+    return getOwnerReferenceFor(ownerUID);
+  }
+
+  default Optional<OwnerReference> getOwnerReferenceFor(String ownerUid) {
+    if (ownerUid == null || ownerUid.isEmpty()) {
+      return Optional.empty();
+    }
+
+    return getMetadata().getOwnerReferences().stream()
```
shawkins, line 261: The owner references could be null.  Granted it would require someone calling setOwnerReferences(null).  Maybe the code generator could guard against that.

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
```diff
@@ -232,4 +235,109 @@ default boolean isFinalizerValid(String finalizer) {
...
+    final OwnerReference ownerReference = new OwnerReferenceBuilder()
+      .withUid(metadata.getUid())
+      .withApiVersion(owner.getApiVersion())
+      .withName(metadata.getName())
+      .withKind(owner.getKind())
+      .build();
+    return addOwnerReference(ownerReference);
+  }
+
+  default OwnerReference addOwnerReference(OwnerReference ownerReference) {
```
shawkins, line 281: It seems like there are two (possibly static) methods that could be created:

createOwnerReference(HasMetadata)
validateOwnerReference(OwnerReference)

I'm honestly not sure if addOwnerReference needs to perform validation - does kube not preform these validations?  And wouldn't the user already be using addOwnerReference(HasMetadata)?  And couldn't the object be modified after this?

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
metacosm: How often does that occur, though? We could shield against this case but we already have code that assumes the metadata to exist and it hasn't caused issues so far (that I'm aware of, at least). If we decide to shield against `null` metadata, we should fix the other usages as well.

shawkins: The rest of the fabric8 code is pretty mixed with checking for null metadata in many places it's not needed as the object is coming from the api-server.  But anytime it can be passed in there's a chance it could be null, which is typically guarded against in BaseOperation methods.  It's of course optional to add checks here, it's just a question of whether it's worth the effort to have something other than a NPE.

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
iocanel: Regarding the no-arg constructor, there is nothing to validate there, given it's `no-arg` so that doesn't change anything.
Regarding the finalizer, I missed that change,  so I can't really comment on that.
Regarding, managing two classes, I say it's a fair trade.

rohanKanojia (approved):

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
metacosm: I would suggest we (you?) do that change in a different PR.

shawkins (approved):
LGTM.  Marked the older comments as resolved.  The additional consideration would be to add validation hooks to our operation support so that visitors or other mechanisms could be used to validate the resource before it's sent to the server.

manusa (approved):

Issue #3592, opened 2021-11-15
Golang generator does not support Enumerators


Sgitario:
The golang generator for extensions used to map Golang model to Java model does not work properly when the Golang model contains consts (or enums). For example, given the Golang model:

Golang type `PeerAuthentication_MutualTLS` that uses a const:
```go
type PeerAuthentication_MutualTLS_Mode int32

type PeerAuthentication_MutualTLS struct {
	// Defines the mTLS mode used for peer authentication.
	Mode                 PeerAuthentication_MutualTLS_Mode `protobuf:"varint,1,opt,name=mode,proto3,enum=istio.security.v1beta1.PeerAuthentication_MutualTLS_Mode" json:"mode,omitempty"`
}
```

Where the const is:

```go
type PeerAuthentication_MutualTLS_Mode int32

const (
	// Inherit from parent, if has one. Otherwise treated as PERMISSIVE.
	PeerAuthentication_MutualTLS_UNSET PeerAuthentication_MutualTLS_Mode = 0
	// Connection is not tunneled.
	PeerAuthentication_MutualTLS_DISABLE PeerAuthentication_MutualTLS_Mode = 1
	// Connection can be either plaintext or mTLS tunnel.
	PeerAuthentication_MutualTLS_PERMISSIVE PeerAuthentication_MutualTLS_Mode = 2
	// Connection is an mTLS tunnel (TLS with client cert must be presented).
	PeerAuthentication_MutualTLS_STRICT PeerAuthentication_MutualTLS_Mode = 3
)
```

The resulting Java model contains:

```
// ...
public class PeerAuthentication_MutualTLS implements KubernetesResource
{

    @JsonProperty("mode")
    private Integer mode;
    // ...
}
```

And the expected output should be:
```
// ...
public class PeerAuthentication_MutualTLS implements KubernetesResource
{

    @JsonProperty("mode")
    private PeerAuthentication_MutualTLS_Mode mode;
    // ...
}
```

```java
public enum PeerAuthentication_MutualTLS_Mode {
        // Inherit from parent, if has one. Otherwise treated as PERMISSIVE.
	UNSET,
	// Connection is not tunneled.
	DISABLE,
	// Connection can be either plaintext or mTLS tunnel.
	PERMISSIVE,
	// Connection is an mTLS tunnel (TLS with client cert must be presented).
	STRICT;
}
```

PR #3602, opened 2021-11-17
Support enums in generator

Sgitario:
Given the Golang model:

Golang type `PeerAuthentication_MutualTLS` that uses a const:
```go
type PeerAuthentication_MutualTLS_Mode int32

type PeerAuthentication_MutualTLS struct {
	// Defines the mTLS mode used for peer authentication.
	Mode                 PeerAuthentication_MutualTLS_Mode `protobuf:"varint,1,opt,name=mode,proto3,enum=istio.security.v1beta1.PeerAuthentication_MutualTLS_Mode" json:"mode,omitempty"`
}
```

Where the const is:

```go
type PeerAuthentication_MutualTLS_Mode int32

const (
	// Inherit from parent, if has one. Otherwise treated as PERMISSIVE.
	PeerAuthentication_MutualTLS_UNSET PeerAuthentication_MutualTLS_Mode = 0
	// Connection is not tunneled.
	PeerAuthentication_MutualTLS_DISABLE PeerAuthentication_MutualTLS_Mode = 1
	// Connection can be either plaintext or mTLS tunnel.
	PeerAuthentication_MutualTLS_PERMISSIVE PeerAuthentication_MutualTLS_Mode = 2
	// Connection is an mTLS tunnel (TLS with client cert must be presented).
	PeerAuthentication_MutualTLS_STRICT PeerAuthentication_MutualTLS_Mode = 3
)
```

The resulting Java model now contains an auto generated enum:

```
// ...
public class PeerAuthentication_MutualTLS implements KubernetesResource
{

    @JsonProperty("mode")
    private PeerAuthentication_MutualTLS_Mode mode;
    // ...
}
```

```java
public enum PeerAuthentication_MutualTLS_Mode {
        // Inherit from parent, if has one. Otherwise treated as PERMISSIVE.
	UNSET("UNSET"),
	// Connection is not tunneled.
	DISABLE("DISABLE"),
	// Connection can be either plaintext or mTLS tunnel.
	PERMISSIVE("PERMISSIVE"),
	// Connection is an mTLS tunnel (TLS with client cert must be presented).
	STRICT("STRICT");
}
```

Note that all the enumerators are auto discovered when the target type is a String, which means that if we want to achieve the following enumerator:

public enum PeerAuthentication_MutualTLS_Mode {
        // Inherit from parent, if has one. Otherwise treated as PERMISSIVE.
	UNSET(1),
	// Connection is not tunneled.
	DISABLE(2),
	// Connection can be either plaintext or mTLS tunnel.
	PERMISSIVE(3),
	// Connection is an mTLS tunnel (TLS with client cert must be presented).
	STRICT(4);
}
```

Then, the extension need to provide the enum values manually:

```go
enumMapping := map[reflect.Type]schemagen.EnumDescriptor{
  reflect.TypeOf(type.MyEnum{}):schemagen.EnumDescriptor{
    Type: "integer"
    Values: []schemagen.EnumValueDescriptor{
        schemagen.EnumValueDescriptor{ Name: "STRICT", Value: 1},
        schemagen.EnumValueDescriptor{ Name: "PERMISSIVE", Value: 2},
....
    }
}
}

	json := schemagen.GenerateSchemaWithAllOptions("http://fabric8.io/istio/IstioSchema#", crdLists, typesDescriptors, providedPackages, manualTypeMap, packageMapping, mappingSchema, providedTypes, constraints, interfacesMapping, javaNameStrategyMapping, enumMapping, "io.fabric8")
```

Fix https://github.com/fabric8io/kubernetes-client/issues/3592

### DIFF
--- a/generator/pkg/schemagen/generate.go
+++ b/generator/pkg/schemagen/generate.go
@@ -18,10 +18,22 @@ package schemagen
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"log"
 	"reflect"
+	"strconv"
 	"strings"
 )
+
+type EnumDescriptor struct {
+	Type   string
+	Values []EnumValueDescriptor
+}
+
+type EnumValueDescriptor struct {
+	Name  string
+	Value interface{}
+}
 
 type ProvidedType struct {
 	GoType    reflect.Type
@@ -53,6 +65,7 @@ type schemaGenerator struct {
 	constraints          map[reflect.Type]map[string]*Constraint // type -> field name -> constraint
 	interfacesMapping    map[string][]reflect.Type               // interface -> list of implementations
 	javaNameStrategy     JavaNameStrategyMapping                 // java name strategy by extension
+	enumTypes            map[reflect.Type]EnumDescriptor         // enum -> json descriptor
 	generatedTypesPrefix string
 }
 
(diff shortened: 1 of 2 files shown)


Sgitario:
Example of usage in https://github.com/Sgitario/kubernetes-client/tree/good_istio
https://github.com/Sgitario/kubernetes-client/blob/good_istio/extensions/istio/generator/cmd/generate/generate.go

manusa (approved):
LGTM, thx!

oscerd (approved):

Issue #3610, opened 2021-11-24
Add Volcano related dependency to root pom


Yikun:
### Describe the bug

Volcano extension have been introduced In https://github.com/fabric8io/kubernetes-client/pull/3580 , but didn't add dependcy to kubernetes-client to root pom.

That means user still need do a extra volcano-client/model dependency in thier application pom besides kubernetes-client.

We should add the Volcano dependency like other extension.
[1] https://github.com/fabric8io/kubernetes-client/blob/de4468047e844ab4b5879bccf7be59ee6dd7ebfe/pom.xml#L372-L432

### Fabric8 Kubernetes Client version

next (development version)

### Steps to reproduce

Only import kubernetes-client dep, volcano related model cannot be import as transitive dep.

### Expected behavior

Only import kubernetes-client dep, volcano related model no need to import again.

### Runtime

Kubernetes (vanilla), other (please specify in additional context)

### Kubernetes API Server version

1.22.3@latest

### Environment

macOS

PR #3580, opened 2021-11-10
Fix #3579: Add Volcano extenstion

Yikun:
## Description
Volcano is a Kubernetes native system for high-performance workloads, which has been accepted by Cloud Native Computing Foundation (CNCF) as its first and only official container batch scheduling project.

This patch adds the volcano extension support, the user can use below code to list API using v1beta1 API:
```java
NamespacedVolcanoClient volcanoClient = new DefaultVolcanoClient()
// Current, will call v1beta1 by default
volcanoClient.podGroups().inNamespace("default").list()
// You can also sepecify the certion version
volcanoClient.v1beta1().podGroups().inNamespace("default").list()

```

Changes:
- Add `generate.go` to generate `volcano-schema.json` from [volcano-sh/apis](https://github.com/volcano-sh/apis).
- Generate Java Volcano model according `volcano-schema.json`, the v1beta1 include `PodGroup`, `Queue` related scheduling API.
- Add volcano client (default and v1beta1).
- Mock, test and examples.

Fix https://github.com/fabric8io/kubernetes-client/issues/3579

## Type of change

 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [x] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [ ] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [x] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [x] I Added [CHANGELOG](../CHANGELOG.md) entry regarding this change
 - [x] I have implemented unit tests to cover my changes
 - [x] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/master/doc/CHEATSHEET.md) accordingly
 - [x] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [x] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

PR #3611, opened 2021-11-24
Fix: Add volcano dep into root pom

Yikun:
## Description
Volcano extension have been introduced In https://github.com/fabric8io/kubernetes-client/pull/3580 , but didn't add dependcy to kubernetes-client to root pom.

It's used to align the version of volcano extension dependency with that of the kubernetes client

So, we should add the Volcano dependency like other extensions:

https://github.com/fabric8io/kubernetes-client/blob/de4468047e844ab4b5879bccf7be59ee6dd7ebfe/pom.xml#L372-L432

## Type of change

## Checklist

Closes: https://github.com/fabric8io/kubernetes-client/issues/3610

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -431,6 +431,22 @@
         <version>${project.version}</version>
       </dependency>
 
+      <dependency>
+        <groupId>io.fabric8</groupId>
+        <artifactId>volcano-model-v1beta1</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.fabric8</groupId>
+        <artifactId>volcano-client</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.fabric8</groupId>
+        <artifactId>volcano-server-mock</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+
       <!-- Jackson dependencies, imported as a BOM -->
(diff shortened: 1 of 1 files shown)


oscerd (approved):

manusa:
Just to be clear.

That's not really a transitive dependency declaration, but an entry on the dependency management.
It's used to align the version of the extension dependency with that of the client (so if you depend on some client, you only need to add the groupId and artifactId to the extension depenendency).

manusa (approved):

rohanKanojia (approved):

rohanKanojia:
@Yikun : ping, Is it ready for review?

Yikun:
@rohanKanojia @manusa Thanks

Yikun:
> That's not really a transitive dependency declaration, but an entry on the dependency management.
> It's used to align the version of the extension dependency with that of the client (so if you depend on some client, you only need to add the groupId and artifactId to the extension depenendency).

@manusa Thanks for your clarification, learned a lot. If a project are using `kubernetes client` and also `volcano client`, as your suggestion, we still need add these two dependencies into their pom rather than only kubernetes client.

Looks like there are some benfits we explicitly specify it (volcano-client). [1]

[1] https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#transitive-dependencies

PR #3617, opened 2021-11-25
fix #3615: always requesting bookmarks

### LINKED ISSUE #3615: Watch behavior with http gone
Just like the automatic re-connection logic supported by AbstractWatchManager and the retrywatcher https://github.com/kubernetes/client-go/blob/v0.22.3/tools/watch/retrywatcher.go it seems that we should be able to safely re-establish the watch at the last seen resource version if it's different than where we started the watch even after getting a 410 status.

Other than some logic that users have tried based upon parsing the too old exception to advance to the suggested value, I can't seem to find anything on an approach like this.

I'll see if I can make this work as expected unless anyone knows why it shouldn't work.

shawkins:
## Description
Draft fix for #3615 - always enable bookmarks since we always implicitly retry watches.

Since this affects a great number of test results, some of which are also touched by the okhttp refactoring, I'll pick this back up after that is committed.

## Type of change

## Checklist


shawkins:
@manusa @rohanKanojia other than needing to correct a lot of tests, is there any reason why we shouldn't automatically request bookmarks?  I see our watch logic as being the same as a retry watcher, so it really should support this by default.

shawkins:
@manusa after some discussion we decided that opt into bookmarks by default seems reasonable.  This can be considered for inclusion in 5.11, I'll deal with any test conflicts with the okhttp refactoring later.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
```diff
@@ -178,6 +185,10 @@ boolean isForceClosed() {
   }
   
   void eventReceived(Watcher.Action action, HasMetadata resource) {
+    if (!receiveBookmarks && action == Action.BOOKMARK) {
```
metacosm, line 188: Is that all that's required to support bookmarks? What is needed by watchers to handle bookmark events?

shawkins: If a user has not explicitly set allowWatchBookmarks=true in their ListOptions, then this will filter the events before they reach the Watcher.  Otherwise the built-in handling is already there in AbstractWatchManager.onMessage - we just need updateResourceVersion to be called on a bookmark.

shawkins: It's already there, this is what I was referring to: https://github.com/fabric8io/kubernetes-client/blob/e4d0a9fdc1beef595f922fe4e50aa4e67d709c87/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java#L251

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/LogWatchCallback.java
```diff
@@ -65,23 +65,16 @@ public LogWatchCallback(Config config, OutputStream out) {
...
-    } else {
-      this.out = out;
-      this.output = null;
-    }
-
-    //We need to connect the pipe here, because onResponse might not be called in time (if log is empty)
-    //This will cause a `Pipe not connected` exception for everyone that tries to read. By always opening
-    //the pipe the user will get a ready to use inputstream, which will block until there is actually something to read.
-    if (this.out instanceof PipedOutputStream && this.output != null) {
       try {
```
sunix, line 68: I am not sure if this is related to always enable bookmarks. Could you tell us more about this changes ?

shawkins: No, this came in unintentionally and are part of another pr.

kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/WatchTest.java
```diff
@@ -39,8 +39,8 @@
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
-import static io.fabric8.kubernetes.client.Watcher.Action.DELETED;
```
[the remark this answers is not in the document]
shawkins: I'd opt for leaving this change.  I know imports can be a touchy subject.  If devs are open to it I can add automatic import ordering using the the impsort-maven-plugin, just like is done with quarkus.

rohanKanojia (approved):

rohanKanojia:
There are some conflicts which need to be resolved

shawkins:
> There are some conflicts which need to be resolved

Should be good now.

rohanKanojia:
Thanks a lot, let's wait for Marc to approve then we can proceed with merge.

manusa (approved):
LGTM, thx!

PR #3619, opened 2021-11-25
feat: Get method for GenericKubernetesResource

manusa:
## Description
Allow the easy retrieval of field values from the additionalProperties Map.

The purpose of this method is to simplify the task when working with GenericKubernetesResources. With the proposed `GenericKubernetesResource#get` method, the following statements can be simplified:

```java
// before
int field = (Integer)gkr.getAdditionalProperties().get("field");
// after
int field = gkr.get("field");

// before
String value = ((Map)gkr.getAdditionalProperties().get("field")).get("nested").toString();
// after
String value = gkr.get("field", "nested");
// or
String value = gkr.get("field.nested");

// before
String value = ((Map)((List)((Map)gkr.getAdditionalProperties().get("field")).get("nested")).get(1)).get("inArray").toString();
// after
String value = gkr.get("field", "nested[1]", "inArray");
// or
String value = gkr.get("field.nested[1].inArray");
```

The more complicated the structure of the resource is, the more value this new method adds.

/cc @lordofthejars 

## Type of change

## Checklist

### DIFF
--- a/kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/GenericKubernetesResource.java
+++ b/kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/GenericKubernetesResource.java
@@ -28,8 +28,14 @@ import lombok.Getter;
 import lombok.Setter;
 import lombok.ToString;
 
+import java.util.Collection;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 @JsonDeserialize(using = com.fasterxml.jackson.databind.JsonDeserializer.None.class)
 @JsonInclude(JsonInclude.Include.NON_NULL)
@@ -45,6 +51,8 @@ import java.util.Map;
 @EqualsAndHashCode
 @Buildable(editableEnabled = false, validationEnabled = false, generateBuilderPackage = true, lazyCollectionInitEnabled = false, builderPackage = "io.fabric8.kubernetes.api.builder")
 public class GenericKubernetesResource implements HasMetadata {
+
+  private static final Pattern ARRAY_PROPERTY_PATTERN = Pattern.compile("^(.+)\\[(\\d+)]$");
 
   @JsonProperty("apiVersion")
   private String apiVersion;
@@ -69,4 +77,55 @@ public class GenericKubernetesResource implements HasMetadata {
     this.additionalProperties.put(name, value);
   }
 
+  /**
+   * Allows the retrieval of field values from this Resource for the provided path segments.
+   *
+   * <p> The following notation is supported
+   * <ul>
+   *   <li>
+   *     Dot Notation
+   *     <pre>{@code resource.get("path.of.the.nested.prop");}</pre>
+   *   </li>
+   *   <li>
+   *     Varargs Notation
+   *     <pre>{@code resource.get("path", "of", "the", "nested", "prop");}</pre>
+   *   </li>
+   *   <li>
+   *     Mixed Notation
+   *     <pre>{@code resource.get("path.of", "the.nested", "prop");}</pre>
+   *   </li>
+   * </ul>
+   *
+   * <p> In addition, fields that contain collections/arrays can also be traversed
+   * <pre>{@code resource.get("path.collectionField[1].nested.array[0].field");}</pre>
+   *
+   * @param path of the field to retrieve.
+   * @param <T> type of the returned object.
+   * @return the value of the traversed path or null if the field does not exist.
+   */
+  @SuppressWarnings("unchecked")
+  public <T> T get(String... path) {
+    Object current = null;
+    int firstPropertyMarker = 0;
+    final List<String> properties = Stream.of(path).map(p -> p.split("\\.")).flatMap(Stream::of)
+      .collect(Collectors.toList());
+    for (String p : properties) {
+      if (firstPropertyMarker++ == 0) {
+        current = getAdditionalProperties().get(p);
+      } else {
+        final Matcher arrayMatcher = ARRAY_PROPERTY_PATTERN.matcher(p);
+        if (current instanceof Map && arrayMatcher.find()) {
+          final String key = arrayMatcher.group(1);
+          final int index = Integer.parseInt(arrayMatcher.group(2));
+          final Map<String, Object> field = (Map<String, Object>) current;
+          current = ((Collection<Object>)field.get(key)).toArray()[index];
+        } else if (current instanceof Map) {
+          current = ((Map<String, Object>) current).get(p);
+        } else {
+          throw new IllegalArgumentException("Cannot get property " + String.join(".", properties) + " from " + getApiVersion() + " " + getKind());
+        }
+      }
+    }
+    return (T) current;
+  }
 }
--- a/kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/GenericKubernetesResource.java
+++ b/kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/GenericKubernetesResource.java
@@ -28,13 +28,12 @@ import lombok.Getter;
 import lombok.Setter;
 import lombok.ToString;
 
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.LinkedHashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 @JsonDeserialize(using = com.fasterxml.jackson.databind.JsonDeserializer.None.class)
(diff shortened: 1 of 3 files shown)


oscerd (approved):

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/GenericKubernetesResource.java
```diff
@@ -69,4 +77,54 @@ public void setAdditionalProperty(String name, Object value) {
...
+      .collect(Collectors.toList());
+    for (String p : properties) {
+      if (it++ == 0) {
+        current = getAdditionalProperties().get(p);
+      } else {
+        final Matcher arrayMatcher = ARRAY_PROPERTY_PATTERN.matcher(p);
+        if (current instanceof Map && arrayMatcher.find()) {
+          final String key = arrayMatcher.group(1);
+          final int index = Integer.parseInt(arrayMatcher.group(2));
+          current = ((Collection<Object>)((Map<String, Object>) current).get(key)).toArray()[index];
```
lordofthejars, line 120: Maybe I will split this line into two lines to avoid multiple casts in the same line, making it hard to read and debug.

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/GenericKubernetesResource.java
```diff
@@ -69,4 +77,54 @@ public void setAdditionalProperty(String name, Object value) {
...
+   * <pre>{@code resource.get("path.collectionField[1].nested.array[0].field");}</pre>
+   *
+   * @param path of the field to retrieve.
+   * @param <T> type of the returned object.
+   * @return the value of the traversed path or null if the field does not exist.
+   */
+  @SuppressWarnings("unchecked")
+  public <T> T get(String... path) {
+    Object current = null;
+    int it = 0;
```
lordofthejars, line 109: Maybe could be changed to `boolean isFirstTime;` I understand that an integer is easy to manage as you do a `++` in the conditional. Maybe leave it as integer but with meaningful name?

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/GenericKubernetesResource.java
```diff
@@ -69,4 +77,54 @@ public void setAdditionalProperty(String name, Object value) {
...
+      } else {
+        final Matcher arrayMatcher = ARRAY_PROPERTY_PATTERN.matcher(p);
+        if (current instanceof Map && arrayMatcher.find()) {
+          final String key = arrayMatcher.group(1);
+          final int index = Integer.parseInt(arrayMatcher.group(2));
+          current = ((Collection<Object>)((Map<String, Object>) current).get(key)).toArray()[index];
+        } else if (current instanceof Map) {
+          current = ((Map<String, Object>) current).get(p);
+        } else {
+          throw new IllegalArgumentException("Cannot get property " + String.join(".", properties) + " from " + getApiVersion() + " " + getKind());
```
metacosm, line 124: Maybe it'd be interesting to say which part of the path is missing, here.

shawkins:
One approach here would be to utilize json node:

Serialization.jsonMapper().convertValue(resource, JsonNode.class).get("child").get("name").asText();

There might be a way to directly construct an ObjectNode from a map.

Adding a path language makes things more complicated - at the very least there are additional considerations such as needing to escape reserved characters such as . [ ] you'd probably want to reference full json path specs to see what other features may eventually be requested and what else could need reserved.

It may not be worth the dependency, but the alibaba fastjson JSONPath can act directly on simple collection java mappings of json structures.

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/GenericKubernetesResource.java
manusa: No longer applicable after considering https://github.com/fabric8io/kubernetes-client/pull/3619#pullrequestreview-816316524

manusa:
> One approach here would be to utilize json node:...

To give the full context, this feature began when Alex asked about retrieving a given field from a GenericKubernetesResource and the resulting ugliness of the current approach.

The current method evolved from:
1. Simple access to nested properties `get("first", "second", "third")`
2. Support for lazyness ;) and more compact notation `get("first.second.third")`
3. Support for nested Array/Collection access `get("first.second[1].third")`

> Adding a path language makes things more complicated

I intentionally avoided the mention of the word [JSONPath](https://tools.ietf.org/id/draft-goessner-dispatch-jsonpath-00.html) in order to limit the expectations of the method.

I don't think that we should rely on an additional only dependency for this purpose (traditionally the client has tried to limit the number of external dependencies). The current GenericKubernetesResource still exposes the `getAdditionalProperties` method, so users are free to reprocess that Map with whatever they have available in their projects.

Removing `#2` in the current feature list would maybe restrict the user expectations, since we will no longer be introducing a path language. However, IMHO, I still think this adds value and as long as its clear this is not a JSONPath parser, it should be fine.

> One approach here would be to utilize json node:

I don't see what do you mean. What's the added benefit of converting the Map/Collection structure into JSONNode.



The point of the PR is to raise the discussion and see if something like this might be useful.

I think the current approach already tackles most of the requirements that users interacting with GKR resources might find (but I'm probably the least indicated person to decide this since I haven't worked that much with them).

So the answer to these questions would be good in order to see if we should proceed with this PR or leave it out.
- Is the current approach useful?
- Is the API extensible in the future?
- Are there any better alternatives?

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/GenericKubernetesResource.java
```diff
@@ -69,4 +77,57 @@ public void setAdditionalProperty(String name, Object value) {
...
+          final String key = arrayMatcher.group(1);
+          final int index = Integer.parseInt(arrayMatcher.group(2));
+          final Map<String, Object> field = (Map<String, Object>) current;
+          current = ((Collection<Object>)field.get(key)).toArray()[index];
+        } else if (current instanceof Map) {
+          current = ((Map<String, Object>) current).get(p);
+        } else {
+          throw new IllegalArgumentException("Cannot get property " + String.join(".", properties) +
+            " from " + getApiVersion() + " " + getKind() +
+            " (missing segment " + String.join(".", properties.subList(it - 1, properties.size())) + ")");
```
metacosm, line 127:
```suggestion
            " (missing segment '" + String.join(".", properties.subList(it - 1, properties.size())) + "')");
```
Make the missing part clearer…

metacosm (approved):

shawkins:
> I don't see what do you mean. What's the added benefit of converting the Map/Collection structure into JSONNode?

A core issue is that to traverse the additional properties you need a lot of casting:

```
Long value = (Long)((List)((Map)additionalPropertiesMap.get("x")).get("y")).get(1);
```
vs
```    
Long value = additionalPropertiesNode.get("x").get("y").get(1).asLong();
```
As a node that traversal is a lot cleaner.

> However, IMHO, I still think this adds value and as long as it's clear this is not a JSONPath parser, it should be fine.

Yes, I was guessing that another dependency was not what you were looking for.   Part The thinking behind offering or borrowing from a full jsonpath is that sooner or later someone will request more expressiveness especially for collection access - it's often not the case that you know the index ahead of time and want a predicate or other type of search.

Other considerations:
- Reserved characters - at least documenting that it's not supported.  It's possible for example for crd to have a field named "field.name" 
- multidimensional array access is not supported - prop[0][0].   
- validation of array access.  If you do something like get("prop[0]") it looks like the array access is ignored.  
- to address both of those and related issues, use of the array access should be treated like an operation just like the . operator.

shawkins:
Here's possible changes to address the other bullet points https://github.com/shawkins/kubernetes-client/commit/65687bd0d379353ea03fc3a0d815ceed39ad1bd9

manusa:
I pushed new changes that should easily solve most issues.

> Reserved characters - at least documenting that it's not supported. It's possible for example for crd to have a field named "field.name"

Given the nature of the query syntax for fields in the HTTP API, I assumed Kubernetes didn't allow for dots in the field names.

I haven't checked all possibilities, but given the simplicity of the current approach, I don't think there should be issues with other characters besides dots.

> multidimensional array access is not supported - prop[0][0].

Addressed in second commit (plus yours [shawkins@65687bd](https://github.com/shawkins/kubernetes-client/commit/65687bd0d379353ea03fc3a0d815ceed39ad1bd9), still haven't checked)

> validation of array access. If you do something like get("prop[0]") it looks like the array access is ignored.

I'm not sure we want to go through the validation path. Users should know the structure of their resources. Receiving an unexpected value or a Class Cast exception should be good enough (IMHO). If we start adding validations this can get very complex very soon.

> Here's possible changes to address the other bullet points [shawkins@65687bd](https://github.com/shawkins/kubernetes-client/commit/65687bd0d379353ea03fc3a0d815ceed39ad1bd9)

Hi, thanks! I had the solution with a test pending on Friday, I just pushed the changes. I will check yours and see If I can merge.

manusa:
At this point, I'm not even sure we want to perform any sort of validation or throw any kind of exception.

e.g. If a user is querying a resource that might be processed by a controller, he/she might be querying a nested field that might have not been initialized. Dealing with an exception instead of just handling a null value might be an extra-hassle.

I think it might be better not to throw an Exception at all (besides Class Cast Exceptions).

shawkins:
> I think it might be better not to throw an Exception at all (besides Class Cast Exceptions).

I agree.  json path usually has options for this, like to return null on missing or even to suppress all exceptions https://javadoc.io/doc/com.jayway.jsonpath/json-path/2.2.0/com/jayway/jsonpath/Option.html#DEFAULT_PATH_LEAF_TO_NULL

One last thought is to simply side-step all string parsing and start with supporting get(Object... path) - and expect only String or Integer path parts.

manusa:
After discussing the topic with @shawkins and a longer analysis, the shorthand/dot-array-notation/compact has some ambiguities that cannot be resolved ease.

Given the following structure:
```
{
  "field": {
    "nested": {
      "array[0]": {"entry": 1},
      "array": [{"entry": 2]
    }
  }
```

Will return an ambiguous result when invoking `get("field", "nested", "array[0]")` or `get("field.nested.array[0]")`.

This ambiguity can be prevented by forcing the escape of `[`, `]` characters just like we introduced with dots. However, this is yet another problem because we might eventually need a way to figure out how to retrieve the value of a field with a name such as `dotted\.dot` or  `escaped\[bracket`.

One of the suggestions from @shawkins  was to replace altogether with a method signature such as `get(Object... segment)` if segment is `int` we assume is a position parameter for an array, else it can be converted to string and used to retrieve the fiel from the Map.

Given the number of current ambiguities, I think it's no longer worthy having a shorthand/compact notation. Still `get("field", "array", 0, "entry")` seems compact enough.

shawkins (commented):
Thanks @manusa I think this will definitely help stave off future problems.  

Some additional thoughts - should this only be for GenericKubernetesResource?  We talked about perhaps KubernetesResource exposing getAdditionalProperties and this method, or should there be utility method <T> T get(Object root, Object... path)?

Related to this, a common issue is just getting labels or annotations from an object:

resource.getMetadata().getAnnotations().get("key")

You always need to check for a null map.  You could use  get(resource.getMetadata().getAnnotations(), "key") instead.  In our projects we have typically ended up adding a "get or default" variant, which seems more concise that using an Optional:  ofNullable(resource.getMetadata().getAnnotations()).map(m -> m.get("key")).orElse(default)

manusa:
> Some additional thoughts - should this only be for GenericKubernetesResource? We talked about perhaps KubernetesResource exposing getAdditionalProperties and this method, or should there be utility method T get(Object root, Object... path)?

If this ends up being something useful useful we can move this up the chain later on (just like the exposure of a getAdditionalProperties default method).

1. Move logic to utility method
2. Default implementation makes use of the Utility class method.
3. The Utility method can be exposed in an appropriate package.class so that it can be reused for other purposes (such as the one you mention)

Additionally a `HasMetadata`#`T getOrDefault(T defaultValue, Object... path)` could also be created for cases such as the one you mention for annotations. Or even a `Optional<T> get???(Object... path)` so that it could easily be consumed like `get("map-holder", "map", "field").orElse(42);`.

Anyway, since the declared API in this PR is easily extensible or can be moved to parent classes/interfaces if applicable, I would defer those to a future PR.

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/GenericKubernetesResource.java
```diff
@@ -69,4 +74,55 @@ public void setAdditionalProperty(String name, Object value) {
     this.additionalProperties.put(name, value);
   }
 
+  @JsonIgnore
```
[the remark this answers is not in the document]
manusa: I don't think this is an issue, why would it be?

We make heavy use of Jackson everywhere, from custom serialization to annotations, etc.

IMHO it is fine, I can't really think of any potential problems derived from this.

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/GenericKubernetesResource.java
```diff
@@ -69,4 +74,55 @@ public void setAdditionalProperty(String name, Object value) {
...
+   *
+   * @param path of the field to retrieve.
+   * @param <T> type of the returned object.
+   * @return the value of the traversed path or null if the field does not exist.
+   */
+  @SuppressWarnings("unchecked")
+  public <T> T get(Object... path) {
+    Object current = getAdditionalProperties();
+    for (Object segment : path) {
+      if (segment instanceof Integer && current instanceof Collection && ((Collection<?>)current).size() > (int)segment) {
```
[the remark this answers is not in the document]
manusa: I gave a thought to this while implementing.

I finally opted for the lenient approach since it had less branching in the logic implementation and I assumed it wouldn't be a problem.

However, the fact that you raised this might indicate the opposite.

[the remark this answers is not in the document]
shawkins: Yes, we just need to say in the javadoc that only strings and integers are expected.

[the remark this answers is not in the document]
manusa: :sweat_smile: I think everything should be covered now.

shawkins (approved):
LGTM

rohanKanojia (approved):

PR #3624, opened 2021-11-29
fix #3620: throw a meaningful exception if no kind/plural, default plural

### LINKED ISSUE #3620: Throw a meanful exception error instead of NPE when no plurals are set

When you create manually a `ResourceDefinitionContext` and you don't set the plurals field a NPE exception is thrown  when you get it using the Kubernetes client 

```
final ResourceDefinitionContext resourceDefinitionContext = new ResourceDefinitionContext.Builder()
            .withGroup("networking.istio.io")
            .withVersion("v1alpha3")
            .withKind("VirtualService")
            .build();

        System.out.println("Namespace: " + namespace);

        final GenericKubernetesResource virtualResource = kubernetesClient.genericKubernetesResources(resourceDefinitionContext).inNamespace(namespace).withName(virtualServiceName).get();

```

```
Caused by: java.lang.NullPointerException
	at java.base/java.net.URI$Parser.parse(URI.java:3129)
	at java.base/java.net.URI.<init>(URI.java:623)
	at io.fabric8.kubernetes.client.utils.URLUtils.join(URLUtils.java:50)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.getNamespacedUrl(OperationSupport.java:164)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.getNamespacedUrl(OperationSupport.java:169)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.getCompleteResourceUrl(BaseOperation.java:818)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.getMandatory(BaseOperation.java:188)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.get(BaseOperation.java:156)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.get(BaseOperation.java:89)
```

Inform why it is failing.

_No response_

### Additional context

_No response_

shawkins:
## Description
fix #3620: throw a meaningful exception if no kind/plural, and default plural based upon kind.

It looks like the code is tolerant to the other fields being missing - including group.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/ResourceDefinitionContext.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/ResourceDefinitionContext.java
@@ -45,6 +45,15 @@ public class ResourceDefinitionContext {
   public boolean isNamespaceScoped() {
     return namespaced;
   }
+  
+  protected void validate() {
+    if (plural == null) {
+      if (kind == null) {
+        throw new IllegalArgumentException("Neither kind nor plural was set, at least one is required");
+      }
+      plural = Utils.getPluralFromKind(kind);
+    }
+  }
 
   public static ResourceDefinitionContext fromResourceType(Class<? extends KubernetesResource> resource) {
     return new Builder()
@@ -89,6 +98,7 @@ public class ResourceDefinitionContext {
     }
     
     public ResourceDefinitionContext build() {
+      this.resourceDefinitionContext.validate();
       return this.resourceDefinitionContext;
     }
   }
--- a/kubernetes-client/src/test/java/io/fabric8/kubernetes/client/dsl/base/ResourceDefinitionContextTest.java
+++ b/kubernetes-client/src/test/java/io/fabric8/kubernetes/client/dsl/base/ResourceDefinitionContextTest.java
@@ -24,7 +24,7 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 class ResourceDefinitionContextTest {
 
   @Test
-  void testMissing() {
+  void missingRequiredKindShouldFail() {
     ResourceDefinitionContext.Builder builder = new ResourceDefinitionContext.Builder();
     assertThrows(IllegalArgumentException.class, () -> builder.build());
     builder.withKind("Kind");
--- a/kubernetes-client/src/test/java/io/fabric8/kubernetes/client/dsl/base/ResourceDefinitionContextTest.java
+++ b/kubernetes-client/src/test/java/io/fabric8/kubernetes/client/dsl/base/ResourceDefinitionContextTest.java
@@ -1,0 +1,35 @@
+/**
+ * Copyright (C) 2015 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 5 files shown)


rohanKanojia (approved):

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/ResourceDefinitionContext.java
```diff
@@ -45,6 +45,15 @@ public String getKind() {
   public boolean isNamespaceScoped() {
     return namespaced;
   }
+  
+  protected void validate() {
+    if (kind == null) {
+      throw new IllegalArgumentException("kind is null");
```
lordofthejars, line 51: How about "kind has not been provided"?

shawkins:
kind is not strictly needed.  updated the pr to relax that.

rohanKanojia:
Sonar is complaining about the use of `@Deprecated` method `Utils.getPluralFromKind` . Is it possible to use something else for calculating plural here?

shawkins:
>  Is it possible to use something else for calculating plural here?

I'm a fan of not duplicating code - I know this is somewhat trivial, but anything else would simply be a copy of that.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/ResourceDefinitionContext.java
```diff
@@ -45,6 +45,15 @@ public String getKind() {
   public boolean isNamespaceScoped() {
     return namespaced;
   }
+  
+  protected void validate() {
+    if (plural == null) {
+      if (kind == null) {
+        throw new IllegalArgumentException("Neither kind nor plural was set, at least one is required");
+      }
+      plural = Utils.getPluralFromKind(kind);
```
metacosm, line 54: This part is tricky because the target class might use a different plural version than the one computed here.

manusa:
When Alex reported that issue, I was thinking more of an Objects.requireNonNull wherever the plural was consumed to compute the URL. However, I haven't analyzed and maybe it's a better approach to validate the definition context, or both.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/ResourceDefinitionContext.java
shawkins: If the context is constructed from a class it will use the annotation to populate this field.  This will only occur in case where someone is manually constructing the context.  Every method the specifies both a context and a class has been deprecated - it's not expected that a user will manually construct the context when a class is used.

shawkins:
> However, I haven't analyzed and maybe it's a better approach to validate the definition context, or both.

Since these contexts are the only place that the plural can come from and are immutable it's sufficient to validate them as they are created.  That will also provide the most meaningful feedback.

manusa (approved):

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/dsl/base/ResourceDefinitionContextTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+class ResourceDefinitionContextTest {
+
+  @Test
+  void testMissing() {
```
metacosm, line 27: A more informative method name would be useful. Maybe something like:
```suggestion
  void missingRequiredKindShouldFail() {
```

metacosm (changes requested):

metacosm (approved):

manusa (approved):

Issue #3651, opened 2021-12-15
CRD generation fails with StackOverflow in case of cyclic references


andreaTP:
### Describe the bug

Using the CRD generation preview module and having cyclic references in the POJOs breaks the generator.

Discovered trying to produce a CRD from the Keycloak Real Representation:
https://github.com/andreaTP/repro_crd_gen_stackoverflow/blob/e398e95c265c9bd2fe992e517b6b554382ae863b/src/main/java/org/keycloak/operator/crds/KeycloakRealmSpec.java#L23

### Fabric8 Kubernetes Client version

5.10.1@latest

### Steps to reproduce

```
git clone https://github.com/andreaTP/repro_crd_gen_stackoverflow
cd repro_crd_gen_stackoverflow
mvn clean compile -e
```

### Expected behavior

In order of preference:

1. the cyclic reference is correctly handled
2. the cyclic reference can be somehow excluded (e.g. with an additional annotation? passing settings to the generator?)
3. in presence of a cyclic reference an appropriate error is thrown to the user

### Runtime

other (please specify in additional context)

### Kubernetes API Server version

other (please specify in additional context)

### Environment

Linux, macOS

### Fabric8 Kubernetes Client Logs

```shell
ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project foo-sample: Fatal error compiling: java.lang.StackOverflowError -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project foo-sample: Fatal error compiling
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:215)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:156)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:148)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
    at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:56)
[...]

PR #3652, opened 2021-12-15
Avoid a StackOverflow and properly fail on cyclic references in CRD generation

andreaTP:
## Description
This is an initial improvement for #3651 , it mainly adds the failing test and output some human friendly information.
I'm still eager to discuss a proper solution for the underlying issue.

## Type of change

## Checklist

### DIFF
--- a/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
+++ b/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
@@ -416,20 +416,7 @@ public abstract class AbstractJsonSchema<T, B> {
               .toArray(JsonNode[]::new);
             return enumProperty(enumValues);
           } else {
-            if (!resolving) {
-              visited.clear();
-              resolving = true;
-            } else {
-              String visitedName = name + ":" + classRef;
-              if (!def.getFullyQualifiedName().startsWith("java") && visited.contains(visitedName)) {
-                throw new IllegalArgumentException("Found a cyclic reference involving the field " + name + " of type " + classRef.getFullyQualifiedName());
-              }
-              visited.add(visitedName);
-            }
-
-            T res = internalFromImpl(def, visited);
-            resolving = false;
-            return res;
+            return resolveNestedClass(name, def, visited);
           }
 
         }
@@ -440,6 +427,23 @@ public abstract class AbstractJsonSchema<T, B> {
 
   // Flag to detect cycles
   private boolean resolving = false;
+
+  private T resolveNestedClass(String name, TypeDef def, Set<String> visited) {
+    if (!resolving) {
+      visited.clear();
+      resolving = true;
+    } else {
+      String visitedName = name + ":" + def.getFullyQualifiedName();
+      if (!def.getFullyQualifiedName().startsWith("java") && visited.contains(visitedName)) {
+        throw new IllegalArgumentException("Found a cyclic reference involving the field " + name + " of type " + def.getFullyQualifiedName());
+      }
+      visited.add(visitedName);
+    }
+
+    T res = internalFromImpl(def, visited);
+    resolving = false;
+    return res;
+  }
 
   /**
    * Builds the schema for specifically handled property types (e.g. intOrString properties)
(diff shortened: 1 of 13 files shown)


crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
```diff
@@ -361,12 +361,16 @@ private String extractUpdatedNameFromJacksonPropertyIfPresent(Property property)
    * @return the structural schema associated with the specified property
    */
   public T internalFrom(String name, TypeRef typeRef) {
+    return internalFromImpl(name, typeRef, new LinkedList<String>());
+  }
+
+  private T internalFromImpl(String name, TypeRef typeRef, List<String> visited) {
```
metacosm, line 367: If the insertion order is not relevant (and I don't think it is in this instance), it's probably better to use a `Set` instead of a `List`. This will also simplify the lookup part which would end up being simply a call to `contains`…

metacosm (approved):
This is a good first step but I think that the issue should be addressed at the `sundrio` level…

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
andreaTP: Good suggestion, switched to a `Set` for the signature and an `HashSet` for the implementation.

andreaTP:
@metacosm I'm not sure that the underlying issue is in `sundrio`, Kubernetes API and OpenAPI v3 are not allowing cyclic references but it's extremely common to have those in POJOs.
I think we need to come up with a design decision on how to move forward in this case.

kubernetes-tests/src/test/java/io/fabric8/crd/generator/zookeeper/v1alpha1/Zookeeper.java
```diff
@@ -24,7 +24,6 @@
 @Version(value="v1alpha1", storage=false)
 @Group("io.zookeeper")
 public class Zookeeper extends CustomResource<ZookeeperSpec, ZookeeperStatus> implements Namespaced {
-  private ZookeeperSpec spec;
```
andreaTP, line 27: is this a pattern that you need / want to support?
In the e2e tests this was throwing the cycle exception (e.g. the fields are parsed twice if they are explicitly declared in this way).

andreaTP:
Moving this PR to Draft since I found more edge cases.

crd-generator/api/src/test/java/io/fabric8/crd/example/nocyclic/NoCyclicStatus.java
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.fabric8.crd.example.nocyclic;
+
+public class NoCyclicStatus {
+  private String message;
+  // private Ref ref1;
```
andreaTP, line 20: This doesn't pass yet, and I'm not sure how to implement it without breaking the signature of public methods, any hint? Or any suggestion?
@iocanel maybe?

andreaTP: fixed

andreaTP:
@metacosm this should be ready for a final review now.
Tested also locally against the [Keycloak model](https://github.com/keycloak/keycloak/tree/main/core/src/main/java/org/keycloak/representations/idm) and added to the local tests all the edge cases I found, now is good to go 🙂

andreaTP:
cc. @manusa @rohanKanojia this should e ready to merge AFAIU.

manusa:
From the code, I understand that the fix consists on detecting cyclic references and throwing a human readable exception with the identification of the problematic field/entity in case on  is found.

manusa (approved):
LGTM, thx!

rohanKanojia (approved):

Issue #3645, opened 2021-12-10
split kubernetes-client-project into an api and an impl


shawkins:
### Is your task related to a problem? Please describe

Rather than just creating an api module specific to the http api, we can consider splitting the whole kubernetes-client into a separate api module.

### Describe the solution you'd like

This is related to #3549 and #2829 - beyond CustomResource and CustomResourceList, it would be a good idea for the dsl, client package interfaces, etc. to be in their own api module.  As much as possible the compile dependencies like common-*, generex, bouncy castle, slf4j, and ideally jackson would be eliminated there.

This would also somewhat alleviate the need for the internal package convention - only classes from the api would be intended for public use.

Ideally this will add some flexibility in terms of matching version across dependencies - if you depend on a module that uses fabric8 6.0.1 api, that should work with any fabric8 6.0.x impl, etc.

shawkins:
Changes (some breaking) that are needed for this:

- separation of dsl and other interfaces from io.fabric8.kubernetes.client from *Client impls, *ExtensionAdapters, and other logic.  This will require a new package to avoid a split package warning.
- " io.fabric8.kubernetes.client.dsl.base
- Some of the io.fabric8.kubernetes.client.internal is still needed on the api side with the current incarnation of the logic in Config and other classes - either they'll need to stay, or we'll have to do some refactoring
- KubernetesClient.customResource returns a RawCustomResourceOperationsImpl - that needs to be an interface or removed
- SharedInformerFactory needs to be an interface - ideally removing the deprecated exposure of OperationContext
- some of io.fabric8.kubernetes.client.informers.cache will move to informers.impl - Cache.NAMESPACE_INDEX though is effectively part of the public api, so we may need to create a separate CacheImpl.
- MetricAPIGroupDSL directly expose PodMetricOperationsImpl and NodeMetricOperationsImpl, that will require interfaces
- RunOperations needs to be an interface and/or have a different way getting a PodOperationsImpl
- Readiness based upon another issue should move to the api and out of an internal package - a deprecated/legacy version (just calling the new one) could be left in the original package if desired.
- ManagedKubernetesClient - not sure where that needs to live
- utils stuff would likely need to move to the api side, but things like PodOperationUtil would need to go to the impl side.

Based upon the above the only dependencies needed will be slf4j (which could be marked as provided) and jackson (which we could minimize somewhat.  It would take quite a bit more work to refactor / break things to not have those.

shawkins:
[...]

Issue #2838, opened 2021-02-25
Readiness utility should not be flagged as internal


iocanel:
Currently Readiness singleton / utility is under `internal` package and as such is treated as internal.
However, its not internal in practice as there are already downstream projects using it.

So, it would make sense for the next major release to move it somewhere else and consider it `public`.

Issue #1285, opened 2018-12-07
Why is KubernetesDeserializer in a package named "internal"?


seglo:
You must use `KubernetesDeserializer.registerCustomKind` in order to successfully watch `CustomResourceDefinitions`, as implemented in [`CRDExample`](https://github.com/fabric8io/kubernetes-client/blob/v4.1.0/kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/CRDExample.java).  Yet this type is in a package named "internal".  Does this imply it shouldn't be used by library users?  Should it be moved?

rohanKanojia:
I think it won't be a problem if we move it out of internal package.

awaj1234:
هل سعر البرنامج 5000دولار

PR #3654, opened 2021-12-16
WIP creating an api module

shawkins:
## Description
Initial start at #3645 - adding a kubernetes-client-api
removes operationcontext from sharedinformerfactory
removes rawoperation... completely
creates interfaces for metrics
does not yet address OpenShift, DefaultKubernetesClient, and uses reflection for loading okhttp client.

## Type of change

## Checklist


shawkins:
The next set of changes more fully refine the new interfaces - the node/pod metrics logic is a little more regular and includes a workaround for #3662 

The only really big change here is the removal of RawCustomResource logic - but I have not yet converted the main test over yet, so there are build failures in kubernetes-tests.

This also removed the mifmif dependency - I don't see that it's directly needed by the client logic.

As you can see the api module is effectively an api + utils/commons.  However as mentioned above we likely need to leave a lot of the common/util logic in place and/or it's too much effort to separate it.

Additional cleanups needed:
- both the api and client modules are generating a Version class
- there are a few test resources that can be removed from the client
- this does not yet address packages, classes are still in their original location.  As much as possible the classes in the client in packages that overlap with the api, should be moved.  The big exception to this is DefaultKubernetesClient.
- openshift-client has not yet been done.
- Readiness has not been moved to a non-internal package.

shawkins:
The refactoring of the metrics support now uses the default namespacing logic - thus the failing tests.  If it's desirable to keep it the existing, which does not use the current namespace by default, that can easily be added back.

shawkins:
The basic separation of both kubernetes-client and openshift-client is complete.  This does not yet address repackaging things left in the client modules nor readiness, and does not yet propose an alternative way to create clients.

If there's general agreement on how this works I'll continue to flesh it out.  If we want to opt instead for just a http client module, let me know.

shawkins:
> and does not yet propose an alternative way to create clients.

Added a Kubernetes/OpenShiftClientBuilder(s) using reflection - if that is not a good choice, we can switch to a service loader, or even do like slf4j and create/remove dummy classes in the api modules.

metacosm:
I have a question regarding the informers' cache… If we want to change the cache directly (for optimisation purposes), how do we do it with this new API? Right now, we get the store and cast it to `Cache<T>` but this doesn't look feasible anymore (unless we cast to `CacheImpl`?) Is there no API to modify the cache anymore (granted, this might not be a good idea to start with)?
The use case here is that when our code deletes or creates a resource, we'd like to remove/put it in the cache directly so that we're sure it's already available without having to wait for the server to send the corresponding event.

shawkins:
> I have a question regarding the informers' cache… If we want to change the cache directly (for optimisation purposes), how do we do it with this new API? 

If we move toward the model of an informer only saving what state is necessary for level event handling, then modifying the cache directly shouldn't be needed.  You'd maintain a separate application level cache of the full resources.

> Right now, we get the store and cast it to Cache<T> but this doesn't look feasible anymore (unless we cast to CacheImpl?) Is there no API to modify the cache anymore (granted, this might not be a good idea to start with)?

It wasn't directly available before these changes - the Store/Indexer interfaces did not allow for modifications, you already have to cast to Cache.

> The use case here is that when our code deletes or creates a resource, we'd like to remove/put it in the cache directly so that we're sure it's already available without having to wait for the server to send the corresponding event.

This seems like it would need to be done very carefully.  If you don't lock modifications on the resourceVersion, locking on the cache is not sufficient:

| Application Thread  | Informer Websocket Thread |
| ------------- | ------------- |
|  | add event for v1 processed |
| locks the cache  | modification event of v2 can't proceed |
| modification of v1 - via unlocked patch/edit/replace |  |
| adds the result object v3 to the cache |  |
| unlocks the cache  |  |
| | modification event of v2 now proceeds - this looks like a revert |
| | modification event of v3 |

The cache is eventually consistent, but there can be incorrect intermediate events.  I'm honestly not sure if you can simply compare resourceVersions to discard events like these, but if you did then the event stream would have events missing - there would be no proper emitting of the update from v1 to v2.  

With resourceVersion locking:
| Application Thread  | Informer Websocket Thread |
| ------------- | ------------- |
|  | add event for v1 processed |
| modification of v1 - via unlocked patch/edit/replace fails |  |
|  | modification event for v2 processed |
| retry modification now against v2 |  |
|  | there's still a window here for a modification event for v4 processed |
| adds the result object v3 to the cache |  |

It seems like the logic adding v3 to the cache would need to track what the expected entry revision is.  If not's not v2, then you can't add v3.

It seems like there are a lot of application level concerns and assumptions about consistency that the informer must take on or expose.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/impl/DefaultSharedIndexInformer.java
```diff
@@ -58,7 +58,7 @@
   private final Reflector<T, L> reflector;
   private final Class<T> apiTypeClass;
   private final ProcessorStore<T> processorStore;
-  private final Cache<T> indexer;
+  private final CacheImpl<T> indexer;
```
manusa: Is it necessary to declare the impl instead of the interface? i.e. are we using internal methods in the DefaultSharedInformer?

manusa:
### _WIP_ Review comments/notes

#### Commit `#1` [creating an api module with new interfaces](https://github.com/fabric8io/kubernetes-client/pull/3654/commits/998e0388b27c4800821f5f4a53f0a86ba4ea7338) :heavy_check_mark: 

- Removes deprecated `RawCustomResourceOperationsImpl customResource(CustomResourceDefinitionContext customResourceDefinition);`
  We need to document this in the migration guide (reminder for next PR)
- Introduces interface for `RunOperations`. Maybe add some Javadoc for the interface declaration stating that this is used for a similar to `kubectl run` operation (reminder for next PR)
- No use of constants for Media/MIME/Content type
- Initial entry-point for HttpClient instantiation (HttpClientUtils#createHttpClient)
- Improvement of Metric related impls

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/impl/DefaultSharedIndexInformer.java
shawkins: At this point yes.  I slimmed down the Cache interface based upon the assumption that will not be used as something the user can plug in - it mainly now exists in the api module for backwards compatibility as there are static references it provides.

shawkins:
> Introduces interface for RunOperations. Maybe add some Javadoc for the interface declaration stating that this is used for a similar to kubectl run operation (reminder for next PR)

That interface simply takes the place of a class with the same name.  There's nothing new from an api perspective - but yes new/additional docs never hurt.

shawkins:
> Despite my personal preference for an SPI approach to achieve this, wouldn't this reflection procedure be problematic for Quarkus Native?

For injection Quakus wouldn't need to use the builder, it would just directly reference the DefaultOpenShiftClient.

The complication with the service loader:
- we'll need to introduce another class into the mix, let's call it ClientFactory, that provides the constructor access.  
- do you consider the clients the same service or different?  If they are the same, then inclusion of the openshift client dependency should mean you can only create openshift clients (similar to what quarkus is doing now).  If they are different, then there's little benefit to using the service loader mechanism, it might be better to do like slf4j and introduce dummy instances into the api that compiler will bind to, but then we remove from the jar.

manusa:
I haven't give it a thought. I think the confusion here lies on what we're calling a "client" and what does the SPI provide. Probably we are referring to different things

In this PR you introduce the Builder pattern for creating instances of KubernetesClient (DefaultKubernetesClient/DefaultOpenShiftClient).

I still need to understand how does this play with https://github.com/shawkins/httpclient

What I mean is that the underlying implementation of the **HttpClient** should be retrievable via ServiceLoader. A valid implementation should then be loaded depending on what there is in the runtime classpath.

shawkins:
> What I mean is that the underlying implementation of the HttpClient should be retrievable via ServiceLoader. A valid implementation should then be loaded depending on what there is in the runtime classpath.

Yes they are somewhat separate concerns.  Here's the thoughts on the HttpClient:
-  two new modules will be created - okhttp-client, and jdkhttp-client - they will both depend on the kubernetes-client-api module.  The okhttp-client module will in the initial release be a dependency of kubernetes-client.  The logic in HttpClientUtils will select what factory to use via a service loader - we'll keep the selection logic simple, either use the first implementation found, or default to okhttp (which will require the factory to provide an identifier).  
- If you are a legacy user and want to keep using okhttp - nothing should change for you.  You just have a dependency on kubernetes-client, you access DefaultKubernetesClient just as you were, and the logic in HttpClientUtils will select the okhttp-client implementation.
- If you want to use a different http client implementation you have two options.  Either exclude the okhttp dependency and include whatever alternative you want and rely on the selection logic in HttpClientUtils, or do whatever you need to do with the dependencies and use something like the ClientBuilder(s) proposed to directly associate an HttpClient.Factory with the builder.

The plan was to address this as part of the work to bring in the jdk implementation.

manusa:
Sounds good, more or less aligned with what I was thinking. Anyway, maybe we could sync  on this (I'll reach out to you internally) because there are things that I'm not fully grasping.

shawkins:
Repacking issues: #1285 #2838

manusa:
> @manusa PodIT.evict seems to be a flaky test, I can get it to fail locally as well - but only sometimes.

I'm aware 😅. We need to do something with this test and the Karat, I hate this situation.

Issue #3636, opened 2021-12-07
LogWatch throws IOException


cmdjulian:
### Describe the bug

In versions prior to version `5.0.0` especially `4.13.3` the following code works fine collecting and printing logs: 
 
```kotlin
private val logger = KotlinLogging.logger { }

private fun streamLogs(jobId: UUID, linesToSkip: Long) {
    val reader = client.batch().jobs().withName("$jobId").watchLog().output.bufferedReader()
    streamLogs(reader, linesToSkip)
}

private fun streamLogs(reader: BufferedReader, linesToSkip: Long) {
    if (linesToSkip > 0) reader.skip(linesToSkip)
    while (!Thread.currentThread().isInterrupted) {
        // On exit null is read and method therefore terminates
        val line = reader.readLine()
        if(line.isNullOrBlank()) break else logger.info { line }
    }
}
```

After the latest update from Spring which is now using `KubernetesClient 5.9.0` under the hood I'm getting the error attached at the bottom

The problem is not the Exception, the problem lays in when the Exception occurs. it is thrown very unpredictable. I always results in loss of log lines so its not thrown at the end of the pipe getting closed.  
Sometimes a few lines from the beginning are missing, sometimes a few lines at the end are missing and sometimes it even crashes directly after the first readLine occurred and returning zero read lines.  
If I have long running Jobs the problem doesn't occur, it only occurs with Jobs lasting only 1 to 2 seconds.  

Before calling the LogWatch I make sure that the Pod of the Job is running because when I don't do that I get an error that the Pod is not ready for log extraction. If I use the `withLogWaitTimeout(5000)` method the Job is maybe already `failed` and therefore the method is never actually reaching the pod success state and always is running in the timeout.

If I screwed up here and there is a better way to achieve that please let me now.  
What I need is waiting before the Job Pod gets ready --> stream logs (all logs line by line) and return from the method if the pod terminates --> extract exit code.  

I tried it with the latest version as well with the same outcome.

### Fabric8 Kubernetes Client version

5.10.1@latest

### Steps to reproduce

Run Job in Kubernetes with a runtime of one to two seconds and see the log watch failing

### Expected behavior

Same behavior as in version `4.13.0`.  
Just returning `null` if the stream is closed and don't throw an IOException.

### Runtime

other (please specify in additional context)

### Kubernetes API Server version

1.21.6

### Environment

Linux

### Fabric8 Kubernetes Client Logs

```shell
java.io.IOException: Pipe closed
[...]

PR #3753, opened 2022-01-15
fix #3587: adding inform support for limit/batch fetching

### LINKED ISSUE #3587: Add a construct in between a Watch and an Informer

shawkins:
## Description
Related to #3587 #3616 - adds limit / batch fetching support to informers.  This can cut down on the size of the responses from the server and eventually the memory footprint of an informer.  To support the latter this will not fall back to fetching the entire list like the go client - this may need some refinement as currently it will just keep retrying.

If the limit is not specified, the default is to fetch everything in a single list.

A behavioral difference introduced here is that the affect of a relist is no longer atomic on the cache - rather each item in the relist will perform a cache update and create an event as it is seen.  Since the cache is still eventually consistent either way, I think this change is not a concern.

The javadocs call out a performance warning - it seems that pagination hits etc.d directly and so there's a concern about performance and scaling.

These changes to conflict of course with the api module creation, but that will be easy to resolve.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 #### Improvements
 * Fix #3674: allows the connect and websocket timeouts to apply to watches instead of a hardcoded timeout
 * Fix #3651: Introduce SchemaFrom annotation as escape hatch (CRD Generator)
+* Fix #3587: adding inform support for limit/batch fetching
 * Fix #3734: extract static finalizer validation method
 
 #### Dependency Upgrade
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Informable.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/Informable.java
@@ -38,6 +38,21 @@ public interface Informable<T> {
   Informable<T> withIndexers(Map<String, Function<T, List<String>>> indexers);
   
   /**
+   * Set the limit to the number of resources to list at one time.  This means that longer
+   * lists will take multiple requests to fetch.
+   * <p>If the list fails to complete it will be re-attempted with this limit, rather than
+   * falling back to the full list.  You should ensure that your handlers are either async or 
+   * fast acting to prevent long delays in list processing that may cause results to expire
+   * before reaching the end of the list.
+   * <p>WARNING As noted in the go client: "paginated lists are always served directly from
+   * etcd, which is significantly less efficient and may lead to serious performance and
+   * scalability problems."
+   * @param limit of a items in a list fetch
+   * @return the current {@link Informable}
+   */
+  Informable<T> withLimit(Long limit);
+  
+  /**
    * Similar to a {@link Watch}, but will attempt to handle failures after successfully started.
    * and provides a store of all the current resources.
    * <p>This returned informer will not support resync.
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/base/BaseOperation.java
@@ -103,7 +103,9 @@ public class BaseOperation<T extends HasMetadata, L extends KubernetesResourceLi
   protected String apiVersion;
 
   protected Class<L> listType;
+  // informable state
   private Map<String, Function<T, List<String>>> indexers;
+  private Long limit;
 
   protected BaseOperation(OperationContext ctx) {
     super(ctx);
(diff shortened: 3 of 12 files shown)


oscerd (approved):

shawkins:
@metacosm @attilapiros Some additional thoughts - it seems to be a pretty advanced case to consider using the limit option with an informer.  There are quite a few comments around the behavior in the go client.  

I don't see much in the literature about it being a built-in case, but it seems like you'd also look towards a sharding pattern - an overall/coordinating operator would create something like a statefulset of controllers, where each has responsibility for a subset of the resources - driven by an annotation (something that is filterable by informers/kubernetes) added by the overall operator.

manusa:
https://github.com/kubernetes/kubernetes/blob/6f896dec4f45ffb82491bc9ce2393e7452886562/staging/src/k8s.io/client-go/tools/cache/reflector.go#L92

manusa:
> I don't see much in the literature about it being a built-in case, but it seems like you'd also look towards a sharding pattern - an overall/coordinating operator would create something like a statefulset of controllers, where each has responsibility for a subset of the resources - driven by an annotation (something that is filterable by informers/kubernetes) added by the overall operator.

I can't find now the source, but I recall that operators should be singletons (or work with some sort of leader-election mechanism). (Which is something that disappointment given the nature of K8s and its overall purpose) 

Is this new _sharding_ approach something documented?

I understand that this has nothing to do with the scope of the current PR, or does it?

manusa (approved):

shawkins:
> Is this new sharding approach something documented?

No, just a musing on how you would approach it assuming a "singleton" coordinating operator.

> I understand that this has nothing to do with the scope of the current PR, or does it?

So far the only request for this limit/batching logic is to work towards scaling operators.  If there are other applicable approaches we should determine what is best to ultimately peruse.  The next step from here is to control the memory footprint of the cache - from #3636 that either looks like a mapping/pruning function or a store that's not fully in memory.  I don't think the latter seems like the right direction. 

As for this change it does seem like it can stand alone based upon the go client.  There are situations where it is beneficial to form smaller paginated results.

shawkins:
> No, just a musing on how you would approach it assuming a "singleton" coordinating operator.

To elaborate further - you'd have the coordinating operator set a hash on the resource as a label in the range of the number of statefulset replicas.  The statefulset pods would be setup to provide the pod name via an env property, which means they could parse out the pod index to determine which bucket they were responsible for.  The individual informers would then be setup to only look for resources with that index label.  Any dependent resources would need to have that label transitively set as well for the sharding pattern to work all the way down.

manusa:
> To elaborate further [...]

Yes, I figured some mechanism like this. However, I might still see some flaws depending on the use-case (from my undetailed, first-glance perspective). For some other uses, it might not be necessary to have an orchestrating process/controller (i.e. considering a StatefulSet, the Pod's ordinal name might be enough to devise a constant strategy to self-assign the bucket).

shawkins:
> the Pod's ordinal name might be enough to devise a constant strategy to self-assign the bucket

To self-assign the hash/index label would already need to be present on the resources.

> Anyway, I guess this is really off topic for the PR.

Yes, it was just to add some more to the discussion about operator scaling.  I'll add a reference to this to the other on-going issue.

manusa:
> the hash/index label would already need to be present on the resources

Well, in my imaginary example the hash (or input for the resource assignment) could be computed from the resource UID or something similar.

shawkins:
> Well, in my imaginary example the hash (or input for the resource assignment) could be computed from the resource UID or something similar.

Something (which I'm referring to as the overall operator) has to compute a hash to the index and apply it as a label to the resource - it needs to be filterable by an informer.  I tried to summarize all the extra work this would entail at https://github.com/fabric8io/kubernetes-client/issues/3587#issuecomment-1014652601

manusa:
> > Well, in my imaginary example the hash (or input for the resource assignment) could be computed from the resource UID or something similar.
> 
> Something (which I'm referring to as the overall operator) has to compute a hash to the index and apply it as a label to the resource - it needs to be filterable by an informer. I tried to summarize all the extra work this would entail at [#3587 (comment)](https://github.com/fabric8io/kubernetes-client/issues/3587#issuecomment-1014652601)

:sweat_smile: I forgot about the filtering part :) 

From your detailed explanations I get the following:
- The Orchestrating Pod (Overall Operator) is responsible to assign a filterable property (in this case a label) to the resource to be processed. _bottleneck_
- 1-N worker Pods with shared informers that filter the assigned resources and perform the actual work. I guess that scaling might impose a challenge in order to rehash the existing resources(I think you expand on this on your comment).

Support for some sort of server-side dynamic labels or wildcard filtering would be of help to completely remove the orchestrating part (https://github.com/kubernetes/kubernetes/issues/30755)

rohanKanojia (approved):

amydentremont:
Just wanted to say thanks for adding this feature! My team just ran into needing the ListerWatcher to paginate in the Informer's Reflector and we realized the ability was something we just got from upgrading versions 🙌

shawkins:
> Just wanted to say thanks for adding this feature!

No problem, hope it works out well.

Issue #3785, opened 2022-01-25
Cannot use untyped informers along with typed API


andreaTP:
### Describe the bug

Using informers for `GenericKubernetesResources` along with accesses to the typed API result in deserialization errors (of the `GenericKubernetesResource`) showing that the typed `Deserializer` has been invoked.

### Fabric8 Kubernetes Client version

next (development version)

### Steps to reproduce

(with access to a cluster e.g. minikube):

1. Clone [https://github.com/andreaTP/repro_GenericKubernetesResource_deser](https://github.com/andreaTP/repro_GenericKubernetesResource_deser)
2. `mvn clean compile`
3. `kubectl apply -f target/classes/META-INF/fabric8/examples.org.example-v1.yml`
4. `mvn clean compile exec:java`

### Expected behavior

The informer should deserialize the `GenericKubernetesResource` without trying to use the "typed" deserializer.

### Runtime

Kubernetes (vanilla), minikube, other (please specify in additional context)

### Kubernetes API Server version

next (development version)

### Environment

macOS

### Fabric8 Kubernetes Client Logs

```shell
2022-01-25 14:42:22,600 ERROR [AbstractWatchManager] - Unhandled exception encountered in watcher event handler
io.fabric8.kubernetes.client.KubernetesClientException: An error has occurred.
        at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:103)
        at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:97)
        at io.fabric8.kubernetes.client.utils.Serialization.unmarshal(Serialization.java:319)
        at io.fabric8.kubernetes.client.utils.Serialization.unmarshal(Serialization.java:240)
        at io.fabric8.kubernetes.client.utils.Serialization.unmarshal(Serialization.java:226)
        at io.fabric8.kubernetes.client.dsl.internal.AbstractWatchManager.readWatchEvent(AbstractWatchManager.java:237)
        at io.fabric8.kubernetes.client.dsl.internal.AbstractWatchManager.onMessage(AbstractWatchManager.java:262)
        at io.fabric8.kubernetes.client.dsl.internal.WatcherWebSocketListener.onMessage(WatcherWebSocketListener.java:68)
        at io.fabric8.kubernetes.client.okhttp.OkHttpWebSocketImpl$BuilderImpl$1.onMessage(OkHttpWebSocketImpl.java:96)
[...]

PR #3786, opened 2022-01-25
Deserialize WatchEvents using the specific object type

andreaTP:
## Description

This is a proposal to fix the specific issue reproduced in #3785 

Instead of relying on the `KubernetesDeserializer` fallback mechanism here we are trying to deserialize the `WatchEvent`s using the specific deserializer for the type.

## Type of change

## Checklist

cc. @shawkins

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
@@ -244,7 +244,12 @@ public abstract class AbstractWatchManager<T extends HasMetadata> implements Wat
       }
 
       WatchEvent watchEvent = Serialization.jsonMapper().treeToValue(json, WatchEvent.class);
-      T object = Serialization.jsonMapper().treeToValue(objectJson, baseOperation.getType());
+      KubernetesResource object = null;
+      try {
+        object = Serialization.jsonMapper().treeToValue(objectJson, baseOperation.getType());
+      } catch (JsonProcessingException e) {
+        object = Serialization.jsonMapper().treeToValue(objectJson, KubernetesResource.class);
+      }
       watchEvent.setObject(object);
       return watchEvent;
     } catch (JsonProcessingException e) {
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
@@ -240,10 +240,8 @@ public abstract class AbstractWatchManager<T extends HasMetadata> implements Wat
       try {
         JsonNode json = Serialization.jsonMapper().readTree(messageSource);
         JsonNode objectJson = null;
-        if (json instanceof ObjectNode) {
-          if (json.has("object")) {
-            objectJson = ((ObjectNode) json).remove("object");
-          }
+        if (json instanceof ObjectNode && json.has("object")) {
+          objectJson = ((ObjectNode) json).remove("object");
         }
 
         WatchEvent watchEvent = Serialization.jsonMapper().treeToValue(json, WatchEvent.class);
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
@@ -235,25 +235,25 @@ public abstract class AbstractWatchManager<T extends HasMetadata> implements Wat
 
   private WatchEvent contextAwareWatchEventDeserializer(String messageSource) {
     try {
-      JsonNode json = Serialization.jsonMapper().readTree(messageSource);
-      JsonNode objectJson = null;
-      if (json instanceof ObjectNode) {
-        if (json.has("object")) {
-          objectJson = ((ObjectNode)json).remove("object");
-        }
-      }
-
-      WatchEvent watchEvent = Serialization.jsonMapper().treeToValue(json, WatchEvent.class);
-      KubernetesResource object = null;
+      return Serialization.unmarshal(messageSource, WatchEvent.class);
+    } catch (Exception ex1) {
       try {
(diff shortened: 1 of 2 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
```diff
@@ -232,9 +232,31 @@ public void close() {
     closeRequest();
     cancelReconnect();
   }
+
+  private WatchEvent contextAwareWatchEventDeserializer(String messageSource) {
+    try {
+      return Serialization.unmarshal(messageSource, WatchEvent.class);
```
andreaTP, line 238: If the current deserialization mechanism fails we fallback to try to deserialize the more specific class

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
```diff
@@ -261,12 +283,10 @@ protected void onMessage(String message) {
     try {
       WatchEvent event = readWatchEvent(message);
       Object object = event.getObject();
-      if (object instanceof HasMetadata) {
-        @SuppressWarnings("unchecked")
-        T obj = (T) object;
-        updateResourceVersion(obj.getMetadata().getResourceVersion());
-        Action action = Action.valueOf(event.getType());
-        eventReceived(action, obj);
+      if (object instanceof Status) {
```
andreaTP, line 286: This is just a little bit of reordering:
 - deserializing the specific type is "more likely" to end up with something that `extends HasMetadata`
 - `Status` will be checked first

kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/DefaultSharedIndexInformerTest.java
```diff
@@ -927,6 +928,23 @@ void testGenericKubernetesResourceSharedIndexInformerWithNamespaceConfigured() t
     assertEquals(0, foundExistingAnimal.getCount());
   }
 
+  @Test
+  void testGenericKubernetesResourceSharedIndexInformerWithAdditionalDeserializers() throws InterruptedException {
+    // Given
+    setupMockServerExpectations(Animal.class, "ns1", this::getList, r -> new WatchEvent(getAnimal("red-panda", "Carnivora", r), "ADDED"), null, null);
+
+    // When
+    KubernetesDeserializer.registerCustomKind("jungle.example.com/v1","Animal", CronTab.class);
```
andreaTP, line 937: This is a little "extreme", instead of registering an incompatible version I'm registering a completely wrong deserializer that will take precedence; but it shows the behaviour described in #3785

andreaTP:
@shawkins this is ready for review now, managed to reproduce the issue in a test case.

shawkins:
> @shawkins this is ready for review now, managed to reproduce the issue in a test case.

My concern is that this is an invalid situation to start off with, so it seems like this handling is not generally needed.

andreaTP:
I can argue that watching `GenericKubernetesResources` events should not break if you register a random serializer for totally different purpose.
This PR specifically fixes this issue.

If you want I can include a more idiomatic test showing this situation.

shawkins:
> I can argue that watching GenericKubernetesResources events should not break if you register a random serializer for totally different purposes.

This would not occur if the resources you were getting back conformed to their expected schema.  The general question - is that an expected case?  If the consensus is that we need to guard against it - then we need to think more broadly.  Does that only apply to the generic inform case?  Because from what I can see once you create the v2 resource, any usage of the v1 typed api (or subsequent) generic will be broken.  For example in your original test case, once the value exists try doing a list operation using the typed v1 api - it will fail.

andreaTP:
> This would not occur if the resources you were getting back conformed to their expected schema.

This is not completely correct, when you change the `storageVersion` the Kubernetes API is going to return "invalid" objects, e.g. with `version` inconsistent with the payload.

This is the use-case that I was trying to support:
https://github.com/java-operator-sdk/java-operator-sdk/pull/874

and this is the example project that I'm targeting with those modifications:
https://github.com/andreaTP/josdk-multiple-crd-versions-example

My expectation is not having two working controllers, the old controller can e.g. simply log the fact that a CR has been modified (and the situation will be eventually handled by WebHooks).
But I'd expect the underlying layers (JOSDK. and kubernetes-client) to be not break in this condition.

shawkins:
> This is not completely correct, when you change the storageVersion the Kubernetes API is going to return "invalid" objects, e.g. with version inconsistent with the payload.

Only if you choose not to have a conversion strategy.

> (and the situation will be eventually handled by WebHooks).

You're referring to the conversion webhooks right?  https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#overview - if they were in place wouldn't this change be moot?  Wouldn't it be a best practice to have that in place anytime there is a breaking schema modfication?

andreaTP:
> You're referring to the conversion webhooks right? https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#overview - if they were in place wouldn't this change be moot? Wouldn't it be a best practice to have that in place anytime there is a breaking schema modfication?

It's best practice, but, you cannot guarantee the webhooks to be in place.
I do believe that this library should be agnostic and support this valid use-case/situation gracefully.

This comes extremely useful for handling controlled upgrades procedures, such as Canary releases etc.

shawkins (approved):
> I do believe that this library should be agnostic and support this valid use-case/situation gracefully.

Sorry I wasn't seeing that all other generic usage is fine because it's rooted by parsing from a class that doesn't reference the KubernetesDeserializer.  The only places that will have problems are from classes, like the WatchEvent, which do reference the KubernetesDeserializer and no way to control the type of contained resources.  WatchEvent is probably the only affected place.  I don't think within fabric8 we'll try to independently parse things like ControllerRevision or other resources that contain references to HasMetadata or KubernetesResource under the generic api.

So the understanding here is that any continued usage of the typed api will break (which is not a fabric8 specific problem), but you'd like the generic watches/inform to continue to work even if the values are invalid for the reported version.  

I think your solution is fine for that.  There's probably a way to override the derserializer with a mixin or something, but nothing that jumps out quickly with a couple of attempts.

PR #3796, opened 2022-02-02
Limit usage of YAML Serializer

geoand:
## Description

The YAML serializer is mainly used in order to parse the kubernetes config. 
This change should not have any functional effect whatsoever, but it paves the way for what I have in mind as a follow-up, which is to provide a public method that allows for clearing out the YAMLMapper (the reasoning being that this is a very heavy object that most likely is only used when the client is first booted).


metacosm:
One usage of the YAML parser that needs to be taken into account is that people very often load an initial/skeleton resource from YAML and then populate it with values at runtime, so this use case needs to be taken into account as well.

oscerd (approved):

manusa (approved):

PR #3802, opened 2022-02-04
[java-gen] Create classes with unique names

andreaTP:
## Description

The generated class names are, sometimes, colliding.
Since we are generating everything in a single package we have been emitting a warning in case of name clash, but this is not good enough.

This is a somewhat simplistic, but effective, approach to the problem, adding trailing numbers to the class names.
The names are going to be uglier but the output is going to be correctly typed and checked.

I'm open to alternative ways to mangle the class names! 🙏 
One "obvious" possibility is to generate nested packages(names needs to be mangled anyhow), but, at the moment, I still enjoy the simplicity of a single package and a flat structure. ( :change_my_mind: 😃  )

_side note:_
I took the opportunity to go ahead with a few small refactoring here and there.

## Type of change

## Checklist

### DIFF
--- a/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
+++ b/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
@@ -20,6 +20,8 @@ import static io.fabric8.java.generator.nodes.Keywords.JAVA_KEYWORDS;
 import com.github.javaparser.ast.CompilationUnit;
 import io.fabric8.java.generator.exceptions.JavaGeneratorException;
 import io.fabric8.kubernetes.api.model.apiextensions.v1.JSONSchemaProps;
+import java.util.Set;
+import java.util.TreeSet;
 import java.util.function.Function;
 
 public abstract class AbstractJSONSchema2Pojo {
@@ -64,6 +66,26 @@ public abstract class AbstractJSONSchema2Pojo {
         }
 
         return sanitized;
+    }
+
+    private static Set<String> classNames = new TreeSet<>();
+
+    public static void resetClassNames() {
+        classNames.clear();
+    }
+
+    public static String uniqueClassName(String name) {
+        String finalTypeName = uniqueClassName(name, name, 0);
+        classNames.add(finalTypeName);
+        return finalTypeName;
+    }
+
+    private static String uniqueClassName(String base, String name, int num) {
+        if (classNames.contains(name)) {
+            return uniqueClassName(base, base + num, num + 1);
+        } else {
+            return name;
+        }
     }
 
     public static AbstractJSONSchema2Pojo fromJsonSchema(
--- a/java-generator/core/src/main/java/io/fabric8/java/generator/CRGeneratorRunner.java
+++ b/java-generator/core/src/main/java/io/fabric8/java/generator/CRGeneratorRunner.java
@@ -62,6 +62,7 @@ public class CRGeneratorRunner {
         List<WritableCRCompilationUnit> writableCUs = new ArrayList<>(crSpec.getVersions().size());
         for (CustomResourceDefinitionVersion crdv : crSpec.getVersions()) {
             CompilationUnit cu = new CompilationUnit();
+            AbstractJSONSchema2Pojo.resetClassNames();
 
             String version = crdv.getName();
 
--- a/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
+++ b/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
@@ -15,14 +15,12 @@
  */
 package io.fabric8.java.generator.nodes;
 
-import static io.fabric8.java.generator.nodes.Keywords.ADDITIONAL_PROPERTIES;
-import static io.fabric8.java.generator.nodes.Keywords.JAVA_UTIL_MAP;
+import static io.fabric8.java.generator.nodes.Keywords.*;
 
 import com.github.javaparser.ast.CompilationUnit;
 import com.github.javaparser.ast.Modifier;
 import com.github.javaparser.ast.NodeList;
 import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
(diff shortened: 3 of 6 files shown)


andreaTP:
cc. @metacosm @manusa @iocanel

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
```diff
@@ -66,6 +68,26 @@ public static String sanitizeString(String str) {
         return sanitized;
     }
 
+    private static Set<String> classNames = new TreeSet<>();
```
manusa, line 71: I'm not sure I like the approach of using a static variable to keep track of the generated class names.

I think it might be better initialize the variable in the CRGeneratorRunner and then pass it in. (consider the option of running this in parallel for multiple CRDs)

manusa:
> adding trailing numbers to the class names.

I personally dislike it, but can't propose any other alternative right now.
Having a real case might help understanding better the problem.

andreaTP:
@manusa thanks for the feedback!
#3817 show some potential in generating more "idiomatic" sources, I move this one back to draft.

OneCricketeer:
I understand this issue is about class names, but what about duplicate _content_? 

e.g.

image

andreaTP:
@OneCricketeer 

> what about duplicate content?

What would you expect the generator to do in such case?

OneCricketeer:
I suppose not create the duplicate classes, by referring to field-name equality... Maybe provide a way to rename such a class; for my shown case `HostPortInfoSpec`. 

Dunno, really, maybe provide some config flag options for edge cases such as these?

iocanel (commented):
Why are generated classes colliding in the first place?

metacosm:
> Why are generated classes colliding in the first place?

I'm actually confused by this as well because I would have expected classes to be generated in packages named after the group / version… but I guess reading the issue description that isn't the case.

andreaTP:
> Why are generated classes colliding in the first place?

It did't realised it at first, but, we are naming Types after the field names so a CRD containing a field e.g. `items` that contains another field `items` generates name clashes.

> I'm actually confused by this as well because I would have expected classes to be generated in packages named after the group / version… but I guess reading the issue description that isn't the case.

The package is generated after the group / version, but nested fields in the same version can have clashing names in different branches of the hierarchy, I think that it's pretty evident looking at the attached image here: https://github.com/fabric8io/kubernetes-client/pull/3817#issuecomment-1033767171 e.g. `MatchExpressions`

andreaTP:
> I suppose not create the duplicate classes, by referring to field-name equality... Maybe provide a way to rename such a class; for my shown case HostPortInfoSpec.
> Dunno, really, maybe provide some config flag options for edge cases such as these?

I think that we can safely implement an "optimizer" to remove clashing classes when the field name and the implementation is equal.
Not sure yet how to make this mechanism tunable by the user, @OneCricketeer would you mind to open a separate issue where we can discuss this in depth?

metacosm:
> > Why are generated classes colliding in the first place?
> 
> @iocanel It did't realised it at first, but, we are naming Types after the field names so a CRD containing a field e.g. `items` that contains another field `items` generates name clashes.
> 
> > I'm actually confused by this as well because I would have expected classes to be generated in packages named after the group / version… but I guess reading the issue description that isn't the case.
> 
> @metacosm the package is generated after the group / version, but nested fields in the same version can have clashing names in different branches of the hierarchy, I think that it's pretty evident looking at the attached image here: [#3817 (comment)](https://github.com/fabric8io/kubernetes-client/pull/3817#issuecomment-1033767171) e.g. `MatchExpressions`

OK, I understand better now. I agree with @OneCricketeer that it'd be better if same types could be shared but I'd argue that this work should be reserved for later (if we decide to undertake it at all). 

Regarding the name conflict avoidance, we could take the Go route and use the name of the enclosing class as a prefix, though the package method used in #3817 might be more Java-like (albeit with quite ugly package names 😅).

andreaTP:
> Regarding the name conflict avoidance, we could take the Go route and use the name of the enclosing class as a prefix, though the package method used in https://github.com/fabric8io/kubernetes-client/pull/3817 might be more Java-like (albeit with quite ugly package names 😅).

Thanks for this feedback @metacosm ! Exactly the discussion I wanted to have on this subject 🙂 
The enclosing class as a prefix (only) doesn't cover all the use cases e.g. you can have `items` nested 3+ times, not sure what Go does in such case ...

Yeah the package names are ugly, but, hopefully, they will be mostly in the `import`s section.
Another alternative I've evaluated is to use nested classes, but the resulting generated code becomes really impenetrable ...

andreaTP:
> I agree though the package names might rear their ugly head again if you ever need to support several versions of the CR with the same code.

@metacosm the different versions are already scoped under different packages already; but yes, having to deal with the same object in the same file from two different versions is going to be pretty messy ... but I don't have any better solution / proposal ...

metacosm:
> > I agree though the package names might rear their ugly head again if you ever need to support several versions of the CR with the same code.
> 
> @metacosm the different versions are already scoped under different packages already; but yes, having to deal with the same object in the same file from two different versions is going to be pretty messy ... but I don't have any better solution / proposal ...

That might also happen if you need to deal with fields of another object in a class that has the same field type, which would be a not-so-uncommon scenario… but yes, I don't have a better solution either… 😞

andreaTP:
@metacosm if we agree that, as of today, the solution with the packages is the less ugly and more Java friendly would you mind to review #3817 ? 🙂

OneCricketeer:
> Why are generated classes colliding in the first place

I have one example from the project I am using (class names have been changed) - 

```
public class Custom extends io.fabric8.kubernetes.client.CustomResource<CustomSpec, CustomStatus> ...
```

Meanwhile, `CustomSpec` defines 

```
@com.fasterxml.jackson.annotation.JsonProperty("custom")
private CustomSpec custom;
```

**However**, this inner `custom` field is actually a different object, as defined by the OpenAPI schema in the YAML, not the CRD Spec (which would result in a recursive class definition).

andreaTP:
Closing in favour of #3817

PR #3801, opened 2022-02-04
[java-gen] Adding support for basic Javadoc comments generation

fabiobrz:
## Description

This PR aims at improving the Java (from CRD) generator feature in its baseline implementation, by adding support for generating basic Javadoc comments for a POJO fields as part of the model, by using an YAML element `description` as the source of information.

As discussed in the [issue for the baseline implementation](https://github.com/fabric8io/kubernetes-client/issues/3642), it seems reasonable to track this enhancement separately, hence this is a draft and a review by @andreaTP as the original contributor of the baseline implementation would help in evaluating whether it could be accepted.

Fixes #3762

## Type of change

 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [x] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [ ] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [x] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [ ] I Added [CHANGELOG](../CHANGELOG.md) entry regarding this change
 - [x] I have implemented unit tests to cover my changes
 - [ ] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/master/doc/CHEATSHEET.md) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [ ] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

PR #3817, opened 2022-02-08
[java-gen] Nested structure to avoid collision of class names

andreaTP:
## Description

This is an alternative to #3802 that generates a nested structure instead of appending numbers to names.

Happy to receive feedback now!

## Type of change

## Checklist


java-generator/core/src/test/java/io/fabric8/java/generator/CompilationTest.java
```diff
@@ -91,7 +91,7 @@ void testKeycloakCRDCompiles() throws Exception {
 
         // Assert
         assertTrue(compilation.errors().isEmpty());
-        assertEquals(32, compilation.sourceFiles().size());
+        assertEquals(50, compilation.sourceFiles().size());
```
andreaTP, line 94: 🎉  even with two completely different implementations, now, numbers are matching up!
https://github.com/fabric8io/kubernetes-client/pull/3802/files#diff-67c4ca64a3cd2e04012103c0094fe36edeee771de5011016ec635670735cc8c2R110
🙂

andreaTP:
In this change:
 - unflattened the `Compilation Unit`, now each object is responsible for creating and managing it's own
 - each object fields relevant classes are now generated in a package named after the object itself (effectively building a hierarchy)
 - use fully qualified class names everywhere (avoid cyclic dependencies in case a field class name has the same name as the containing object)
 
 not covered (maybe at some point in the future?):
 - deduplicate matching classes if they are actually the same
 
 I'm pretty much happy with this encoding now!
I ask our fellow early adopters to check the result on their CRDs and provide a feedback, cc. @OneCricketeer @fabiobrz @manusa 

For the rest of the reviewers I attach an example generated structure (of the Keycloak Go operator CRD, of course 🙂  )
Screenshot 2022-02-09 at 13 31 58

java-generator/core/src/main/java/io/fabric8/java/generator/WritableCRCompilationUnit.java
```diff
@@ -28,44 +27,35 @@
...
+    public List<GeneratorResult.ClassResult> getClassResults() {
+        return classResults;
     }
 
-    public void writeAllJavaClasses(File basePath) {
+    WritableCRCompilationUnit(List<GeneratorResult.ClassResult> classResults) {
+        this.classResults = classResults;
+    }
+
+    public void writeAllJavaClasses(File basePath, Optional<String> basePackage) {
```
fabiobrz, line 40: Just sayin' here... 
According to a [comment](https://github.com/fabric8io/kubernetes-client/pull/3801#discussion_r801389053) in a recent PR review, using `Optional`s as method parameters is not recommended.
BTW I can spot one between `createFolders()` arguments, too.

java-generator/core/src/main/java/io/fabric8/java/generator/WritableCRCompilationUnit.java
```diff
@@ -28,44 +27,35 @@
...
 public class WritableCRCompilationUnit {
 
-    private final CompilationUnit cu;
-    private final List<String> classNames;
+    private final List<GeneratorResult.ClassResult> classResults;
 
-    WritableCRCompilationUnit(CompilationUnit cu, List<String> classNames) {
-        this.cu = cu;
-        this.classNames = classNames;
+    public List<GeneratorResult.ClassResult> getClassResults() {
```
fabiobrz, line 32: This is just about coding style & me, but I'd keep this property after the ctor definition.

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
```diff
@@ -37,7 +37,12 @@
 
     public abstract String getType();
 
-    public abstract GeneratorResult generateJava(CompilationUnit cu);
+    public abstract GeneratorResult generateJava();
+
+    /** Takes a string and return the corresponding package name */
+    public static String packageName(String str) {
+        return str.toLowerCase(Locale.ROOT);
```
fabiobrz, line 44: Should some kind of additional validation happen here in order to be sure that what is returned is compliant with Java naming conventions for packages? Or can this be ignored safely? WDYT?

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
```diff
@@ -47,44 +44,44 @@
...
 
     public JObject(
+            String pkg,
             String type,
             Map<String, JSONSchemaProps> fields,
             List<String> required,
-            JObjectOptions options) {
-        this.options = options;
+            boolean preserveUnknownFields,
+            String prefix) {
```
fabiobrz, line 59: Could `prefix` be renamed so that is clear it relates to the class? Maybe just add a Javadoc comment here?

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
```diff
@@ -67,9 +72,9 @@ public static String sanitizeString(String str) {
     }
 
     public static AbstractJSONSchema2Pojo fromJsonSchema(
-            String key, JSONSchemaProps prop, String prefix, String suffix) {
+            String key, JSONSchemaProps prop, String parentPkg, String prefix) {
```
fabiobrz, line 75: Maybe it's just me but here again, I'd name `prefix` as `classPrefix`, or something like that. WDYT?

fabiobrz (commented):
Hi @andreaTP - I just had a quick look to the sources and dropped some comments, feel free to let me know what you think.

I'd like to look at the tests too, maybe later. In the meantime, though, I've run the generator both with WildFly and Infinispan CRDs, and used them in a local branch of some internal tooling that we have, successfully :smiley:

java-generator/core/src/main/java/io/fabric8/java/generator/WritableCRCompilationUnit.java
andreaTP: done

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
andreaTP: `packageName` gets called on `className`s that are already sanitized ... we should have a better story overall about mangling strings (e.g. a more principled, tested and better thought throw set of utilities), I would say that this can be a follow up PR.

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
andreaTP: `classPrefix`

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
andreaTP: `classPrefix`

andreaTP:
Moving back to Draft waiting to have #3812 in first before doing a final rebase.

java-generator/maven-plugin/src/main/java/io/fabric8/java/generator/maven/plugin/JavaGeneratorMojo.java
```diff
@@ -42,9 +42,31 @@
...
+    @Parameter(property = "fabric8.java-generator.suffix-strategy", required = false)
+    Config.Suffix suffixStrategy = null;
+
+    @Parameter(property = "fabric8.java-generator.always-preserve-unknown", required = false)
+    Boolean alwaysPreserveUnknown = null;
+
+    @Parameter(property = "fabric8.java-generator.extra-annotations", required = false)
+    Boolean extraAnnotations = null;
+
+    @Parameter(property = "fabric8.java-generator.code-structure", required = false)
```
andreaTP, line 57: How do we hide some configurations?

andreaTP:
This should be ready for review again 🙂  :

- rebased on latest `master`
- made the nested structure optional but default (e.g. you can generate code using the flat structure)
- added more configurations to mangle class names etc.
- added an option to always preserve unknown fields
- added more tests to check different configurations

cc. @manusa 🙏

andreaTP:
on https://github.com/fabric8io/kubernetes-client/pull/3817/commits/288f85aac468b981d077785dd7e4bb97de9be3eb I started to use `parallelStreams` to improve performance, and well, generation time of the full `cert-manager` on my machine went from ~60 seconds to ~2 🙂

java-generator/core/src/main/java/io/fabric8/java/generator/CRGeneratorRunner.java
```diff
@@ -56,23 +55,26 @@ public void run(File source, File basePath) {
...
-
-                    for (WritableCRCompilationUnit w : writables) {
-                        w.writeAllJavaClasses(basePath);
-                    }
-                } else {
-                    LOGGER.warn(
-                            "Not generating nothing for resource of kind: " + resource.getKind());
-                }
-            }
+            resources.parallelStream()
```
manusa, line 58: This is the simplest approach to divide the work. However it will use the number of available cores in your machine to spawn that many threads.
We should probably explore other options (in the future) such as a queue and a configurable executor service. This jobs are IO intensive, so it's very likely that the performance can be improved even more.

java-generator/maven-plugin/src/main/java/io/fabric8/java/generator/maven/plugin/JavaGeneratorMojo.java
manusa: What do you mean by "hide"?

java-generator/core/src/main/java/io/fabric8/java/generator/CRGeneratorRunner.java
andreaTP: Yup, this was extremely simple to get in and already provided great value, but I'm open(and eager) to look for alternatives 😄 .

I'm not 100% sure about the "IO intensive" part, a lot of time is spent in the `generateJava` function which is pure in-memory computation (e.g. spawning javaparser `CompilationUnit` etc. etc.), but we will measure next time we iterate over performance.

java-generator/maven-plugin/src/main/java/io/fabric8/java/generator/maven/plugin/JavaGeneratorMojo.java
andreaTP: Like we do in the CLI:
https://github.com/fabric8io/kubernetes-client/pull/3817/files#diff-1c647fc1e9204b845f5024b4ad0a74a5d58ded1e458286d577a163c1ab8d5dd1R67

It's an available option but is removed from documentation etc.

manusa: I'm not sure there's an equivalent option for Maven plugins.
You either want to expose the option or not. There is also no help CLI message generated (AFAIK).

manusa: OK; maybe I'm wrong:
```shell
mvn help:describe -Dplugin=$pluginName -Dmojo=$pluginMojo -Dfull=true
```

But if you're doing this, you probably want to see all of the parameters.

java-generator/core/src/main/java/io/fabric8/java/generator/CRGeneratorRunner.java
andreaTP: Talking about CI pipelines, AFAIK we are using [default GH runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) and they have 2 cores each, so I would expect an improvement in performance even there (although it is still not the "target machine").

java-generator/maven-plugin/src/main/java/io/fabric8/java/generator/maven/plugin/JavaGeneratorMojo.java
andreaTP: Can confirm I can see all the parameters using:
```
mvn help:describe -Dplugin=io.fabric8:java-generator-maven-plugin -Dmojo=io.fabric8.java.generator.maven.plugin.JavaGeneratorMojo -Dfull=true -Ddetail 
```

Unfortunately I cannot find any way to hide one, should we add a follow-up issue?
Should we name the "hidden" properties as "unsupported" or something?

java-generator/core/src/main/java/io/fabric8/java/generator/CRGeneratorRunner.java
manusa: I wasn't talking about ours, but about what users might have in place (especially those running on Kubernetes).
Anyway, I will probably experiment with this to get a better feeling.

java-generator/maven-plugin/src/main/java/io/fabric8/java/generator/maven/plugin/JavaGeneratorMojo.java
manusa: If the reasoning is because these options are not supported (what we talked about), I would just prefix them with `preview`, or something like this.

```suggestion
    @Parameter(property = "fabric8.java-generator.preview.code-structure", required = false)
    Config.Prefix previewPrefixStrategy = null;
```

andreaTP: Using the [`readonly` property](https://maven.apache.org/plugin-tools/maven-plugin-annotations/apidocs/org/apache/maven/plugins/annotations/Parameter.html#readonly--) seems to do job @manusa , and the integration tests are still working, so the user-defined properties are actually taken into account.

andreaTP: ouch @manusa , concurrent update on the same subject ☝️  🙂

manusa (approved):

PR #3830, opened 2022-02-10
[java-gen] feat: Maven Plugin + Integration test infrastructure

manusa:
## Description
- Spiked a Maven Plugin to invoke the generator.
- Created an integration test infrastructure based on the plugin
  - TODO: Test generated sources compared to our actual extensions (the end goal would be to be able to replace our current code generation with this --> Configuration options should be added for package, additionalProperties generation, etc.)
  - TODO: Create a proper verification infrastructure

Other comments:
- Current implementation is slow. We can probably improve the performance by running the generation in multiple threads for each individual CRD.
- CRD Generator only admits passing in a File. We could support reading complete directories (makes a lot of sense for the Maven Plugin workflow)

/cc @andreaTP 

## Type of change

## Checklist

### DIFF
--- a/java-generator/core/pom.xml
+++ b/java-generator/core/pom.xml
@@ -99,8 +99,6 @@
     <dependency>
       <groupId>com.approvaltests</groupId>
       <artifactId>approvaltests</artifactId>
-      <version>13.3.0</version>
-      <scope>test</scope>
     </dependency>
   </dependencies>
   <build>
--- a/java-generator/it/src/it/cert-manager/verify.groovy
+++ b/java-generator/it/src/it/cert-manager/verify.groovy
@@ -1,0 +1,23 @@
+/**
+ * Copyright (C) 2015 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import static org.junit.jupiter.api.Assertions.assertEquals
+
+// TODO: This should be modified to test against the current extension sources
+assertEquals(
+  new File(basedir, "/expected/AuthSpec.java").getText("UTF-8").replace("\r\n", "\n"),
+  new File(basedir, "/target/generated-sources/java/io/cert_manager/v1/AuthSpec.java").getText("UTF-8").replace("\r\n", "\n")
+)
+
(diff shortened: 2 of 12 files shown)


andreaTP:
This is awesome, an extremely appreciated and amazing addition! 🙏 

Do you think that we should, eventually, leverage this infrastructure instead of using the `maven-invoker` for integration tests? ( [ref](https://github.com/fabric8io/kubernetes-client/pull/3812/files#diff-3da10905fd06ededa34a8e3bb75f061210de4f1e799ddb2d55f87511965382e4R139) )

Regarding:

> Current implementation is slow. We can probably improve the performance by running the generation in multiple threads for each individual CRD.

I think this is a great starting point and we can get there in the future.

> CRD Generator only admits passing in a File. We could support reading complete directories (makes a lot of sense for the Maven Plugin workflow)

Agreed, maybe we should scan specific directories at some point e.g. `kubernetes`

rohanKanojia (approved):

oscerd (approved):

java-generator/core/pom.xml
```diff
@@ -99,8 +99,6 @@
     <dependency>
       <groupId>com.approvaltests</groupId>
       <artifactId>approvaltests</artifactId>
-      <version>13.3.0</version>
-      <scope>test</scope>
```
andreaTP, line 103: this way we are adding `approvaltests` to the transitive dependencies, is it intended?

java-generator/it/src/it/cert-manager/verify.groovy
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import static org.junit.jupiter.api.Assertions.assertEquals
+
+// TODO: This should be modified to test against the current extension sources
+assertEquals(
```
andreaTP, line 19: should we use `approvaltests` here?

andreaTP (commented):
A couple of questions, but the bulk of it is great work! 👍

java-generator/it/src/it/cert-manager/verify.groovy
manusa: I tried to, but it didn't work (Approval tests relies on a JUnit execution environment).

Anyway, considering the way we are using Approvaltests, this is practically the same (minus the dependency)

java-generator/core/pom.xml
andreaTP: 👍

java-generator/it/src/it/cert-manager/verify.groovy
andreaTP: There is a port for Groovy:
https://github.com/rzabini/ApprovalTests.Spock

What I'm enjoining from the Approval tests is the workflow with `approved` and `received` files ... but now fussy on it, especially at this stage.

andreaTP (approved):

PR #3836, opened 2022-02-10
fix #3832 #1883 simplifying the isHttpsAvailable check

### LINKED ISSUE #1883: isHttpsAvailable isn't reliable
Doing a GET on https://$KUBERNETES_MASTER/ to decide whether isHttpsAvailable is problematic

* Most roles without full cluster permissions are unable to GET / - whereas all authenticated users are in the `system:discovery` group which could call GET /api for example
```
$ curl k -H "Authorization: Bearer $KUBERNETES_TOKEN" https://$KUBERNETES_SERVICE_HOST:$KUBERNETES_SERVICE_PORT/
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {
    
  },
  "status": "Failure",
  "message": "forbidden: User \"system:serviceaccount:default:default\" cannot get path \"/\"",
  "reason": "Forbidden",
  "details": {
    
  },
  "code": 403
}
```

```
$ curl k -H "Authorization: Bearer $KUBERNETES_TOKEN" https://$KUBERNETES_SERVICE_HOST:$KUBERNETES_SERVICE_PORT/api
{
  "kind": "APIVersions",
  "versions": [
    "v1"
  ],
  "serverAddressByClientCIDRs": [
    {
      "clientCIDR": "0.0.0.0/0",
      "serverAddress": "ip-10-0-167-167.ap-southeast-2.compute.internal:443"
    }
  ]
}
```

* There is no distinction made between an unauthorized response (log an error and bomb out) and being unable to talk HTTPS to an HTTP server

### LINKED ISSUE #3832: SSLUtils.isHttpsAvailable should use an HttpURLConnection instead of an HttpClient

isHttpsAvailable bypasses any HttpClient.Factory associated with a higher level construct such as builder for creating the clients.

Directly use HttpURLConnection, rather than constructing and immediately destroying a client.

_No response_

### Additional context

See #1883

shawkins:
## Description

Fix #1883
Fix #3832

This is a bit of a corner case, but I don't want to construct a client instance just for this check - it is done without a reference to the HttpClient.Factory, which is a problem if it's possible to pass one in from a higher level. 

What is proposed here is to just use an HttpsURLConnection - that also addresses #1883 because all it's checking for is a successful handshake.  This assumes that you'll of course still use the Config to setup your https related properties.

Another direction to take would be to have the HttpClient.Factory available here - such as setting it on the Config.

And the last option is to as much as possible not allow for the programmatic association of an HttpClient.Factory / HttpClient, so that all calls use what will be the service loader.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ### 6.0-SNAPSHOT
 
 #### Bugs
+* Fix #3832 #1883: simplifying the isHttpsAvailable check
 * Fix #3848: Supports Queue (cluster) API for Volcano extension
 * Fix #3582: SSL truststore can be loaded in FIPS enabled environments
 
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/internal/SSLUtils.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/internal/SSLUtils.java
@@ -16,15 +16,11 @@
 package io.fabric8.kubernetes.client.internal;
 
 import io.fabric8.kubernetes.client.Config;
-import io.fabric8.kubernetes.client.ConfigBuilder;
-import io.fabric8.kubernetes.client.http.HttpClient;
-import io.fabric8.kubernetes.client.http.HttpRequest;
-import io.fabric8.kubernetes.client.http.HttpResponse;
-import io.fabric8.kubernetes.client.utils.HttpClientUtils;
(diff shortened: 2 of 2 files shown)


rohanKanojia (approved):

oscerd (approved):

manusa (approved):

manusa:
Tests need to be updated

PR #3838, opened 2022-02-11
Allow applications to clear out YAML ObjectMapper

geoand:
This is useful because in a lot of cases the YAML mapper
is only need at application startup when the client
is created, so there is no reason to keep the very heavy
(in terms of memory) mapper around indefinitely.


oscerd (approved):

metacosm (approved):
That looks reasonable to me. At worse, people will never call `clearYamlMapper` and the behavior should be the same as now.

manusa (approved):

PR #3846, opened 2022-02-11
fix #3831: creating a separate version class for the client

### LINKED ISSUE #3831: Version class after module split

Do we want just a single version class or multiple version classes - one for the api, and one for the impl?

I don't have a strong preference, but if the expectation is for the impl to provide its version, then it should do so in a way that doesn't require referencing a class in the impl - for example Client.getVerion

_No response_

### Additional context

_No response_

shawkins:
## Description
fix for #3831
This class is currently not used.  This is to avoid having the same class in multiple jars.

## Type of change

## Checklist


rohanKanojia (approved):

manusa:
Why don't we just remove it?

If this is moved to an internal package, and it's actually not used internally, what's the point of having it?

shawkins:
> If this is moved to an internal package, and it's actually not used internally, what's the point of having it?

To be there proactively once someone mismatches the api and impl versions and wants a way to find the version of the impl at runtime.  Yes, just like any other internal usage it would be considered unsupported, so they would still need to request an enhancement moving forward.

> Why don't we just remove it?

If we discount the above case, then removing it is fine.

manusa:
If the purpose is to be able to identify module versions, I'd at least, rename this one (or the other one) to something else.
- ApiVersion + Version
- Version + KubernetesClientVersion

shawkins:
> If the purpose is to be able to identify module versions, I'd at least, rename this one (or the other one) to something else.

I had forgotten you could get the version information from the maven info in the jar - so there is a workaround already.  I'll just remove the -client class for now.

manusa (approved):

PR #3857, opened 2022-02-14
fix #3855 adding modules for okhttp and jdkhttp

### LINKED ISSUE #3855: Add modules for okhttp and jdkhttp

Now that we have an api module, we can create separate modules for okhttp and jdkhttp

New modules and a service loader mechanism to choose the implementation.

_No response_

### Additional context

_No response_

shawkins:
## Description
Adds modules fork okhttp and jdkhttp.  The jdkhttp module is conditionally included when using java 11 or up to build.  okhttp remains a dependency of kubernetes-client - so that nothing changes by default.

A service loader is used to pick up the implementation.  The simple logic initially proposed there is warn if more than 1 impl, and use anything other than okhttp.  That of course is open for discussion - it could be priorities on the factories, or other mechanism such as a Config property declaring the preferred impl.  My thought was that if you had multiple impls in your classpath you were doing something advanced and would need to make use of something like the KubernetesClientBuilder - to pragmatically associate the HttpClient.Factory.  

We would likely document the jdkhttp client as experimental for the first release - see it's readme for unsupported config options and remaining work.  I did not add tests specifically for the http client logic, rather I locally manipulated the kubernetes-client to use the jdkhttp to determine coverage.  We'll need to consider if an additional test run through kubernetes-tests or itests per httpclient impl make sense.

There are two changes which could be seen as breaking
- making the okhttp client a runtime dependency of kubernetes-client (if you still want to do custom okhttp things, you'll be responsible for adding the compile dependency in your project to kubernetes-httpclient-okhttp).
- removing the DefaultKubernetesClient constructor that referenced OkHttp (follows from the first change)

That can be walked back if needed, or if approved it will be added to the release notes / migration guide.

## Type of change

## Checklist


shawkins:
This seems to be working now with some more refinements to the osgi stuff.  The migration docs will be fleshed out more after #3865

The last question is how this will affect existing osgi users - it seems like they will need to install one of the new bundles for things to keep working - what documentation is needed for that?

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/HttpClientUtils.java
```diff
@@ -136,16 +142,26 @@ public static String basicCredentials(String username, String password) {
...
   public static HttpClient createHttpClient(Config config) {
-    // TODO: replace with service load
-    try {
-        return ((HttpClient.Factory) Class.forName("io.fabric8.kubernetes.client.okhttp.OkHttpClientFactory")
-                .getDeclaredConstructor()
-                .newInstance()).createHttpClient(config);
-    } catch (InstantiationException | IllegalAccessException | IllegalArgumentException | InvocationTargetException
-            | NoSuchMethodException | SecurityException | ClassNotFoundException e) {
-        throw KubernetesClientException.launderThrowable(e);
+    ServiceLoader<HttpClient.Factory> loader = ServiceLoader.load(HttpClient.Factory.class);
```
iocanel, line 148: This is something that will work flat class loaders, but may fail for other cases (e.g. OSGi, APT and so on).

I find it safer, to to perform SPI lookup using layers:
1. Plain `ServiceLoader.load(clazz)`
2. Using the current class classloader: `ServiceLoader.load(clazz, getClass().getClassLoader())`
3. Using TCCL: `ServiceLoader.load(clazz,  Thread.currentThread().getContextClassLoader())`

This is how we deal with adapters two: https://github.com/kameshsampath/kubernetes-client/blob/solo-extension/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Adapters.java#L79

platforms/karaf/features/src/main/resources/feature.xml
```diff
@@ -59,6 +59,7 @@
     <bundle>mvn:io.fabric8/kubernetes-model-scheduling/${project.version}</bundle>
     <bundle>mvn:io.fabric8/kubernetes-model-storageclass/${project.version}</bundle>
     <bundle>mvn:io.fabric8/openshift-model/${project.version}</bundle>
+    <bundle>mvn:io.fabric8/kubernetes-httpclient-okhttp/${project.version}/jar/bundle</bundle>
```
iocanel, line 62: This is tricky, so consider it more like thinking out loud rather than a change request.

I would move the `kubernetes-httpclient-okhttp` to it's own feature, along with it's dependencies. I would also provide a feature for `kubernetes-httpclient-okhttp`.

Then I would either decouple `kubernetes-client` from `kubernetes-httpclient-okhttp` and let users decide what to use or just refer to it (but most probably the first). The former adds an extra step for the user, the latter is less flexible. In both cases the idea is that we communicate this way which are the pluggable parts.

Again, this is something that we can do afterwards. I prefer merging this faster and add this touches later on.

iocanel (approved):
LGTM

Added a couple of remarks. These can be added in followup pull requests.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/HttpClientUtils.java
shawkins: 1 and 3 are the same according to the ServiceLoader javadocs.  In adapters 2 is being called with a class from the extension.  Here it would be called with a class from the api.   Other than to guard against the case where someone has set an inappropriate TCCL, I don't see why you'd need 2 here.

platforms/karaf/features/src/main/resources/feature.xml
shawkins: Opening as #3878

manusa:
> That of course is open for discussion - it could be priorities on the factories, or other mechanism such as a Config property declaring the preferred impl. My thought was that if you had multiple impls in your classpath you were doing something advanced and would need to make use of something like the KubernetesClientBuilder - to pragmatically associate the HttpClient.Factory.

IMO we should provide a simplistic approach for users (that can be improved by configuration).

_Consider a scenario where you want to use the Fabric8 KubernetesClient in a JBang script (or any other Java scripting platform) to perform some chores for you (Java rul3z ShellScript)._

I see that we should provide an HttpClient implementation by default. For traditional and compatibility reasons, this client should of course be OkHttp 3. But if as a user, I want to replace this by some other implementation, it should be sufficient to add the new dependency to the mix, the client initialization should do the rest for me with no further configuration.

As I see it, the OkHttp3 implementation should have the lowest score/priority, any other implementation should have a higher score/priority and should be automatically picked up instead. No exclusion, or manual configuration should be required. If 2 implementations with the same score are provided, an error or at least a warning would be convenient.

Coming back to the JBang scenario, I would add the 2 dependencies, and would be in business straight away, no need for extra configurations.

shawkins:
> As I see it, the OkHttp3 implementation should have the lowest score/priority, any other implementation should have a higher score/priority and should be automatically picked up instead.

Are you ok with what has been implemented?  We default to okhttp because of the runtime dependency, but will use any other dependency if found and warn that there are multiple.

manusa:
> Are you ok with what has been implemented? We default to okhttp because of the runtime dependency, but will use any other dependency if found and warn that there are multiple.

I haven't checked the implementation yet, sorry. I only read the PR description and added my comments to the discussion topic. 

The current behavior sounds aligned with my PoV, not sure if it's guaranteed that OkHttp will be the one with the least priority though.

shawkins:
> The current behavior sounds aligned with my PoV, not sure if it's guaranteed that OkHttp will be the one with the least priority though.

https://github.com/fabric8io/kubernetes-client/pull/3857/files#diff-55d324938196e4cf9617880b531b9976c45f90f9abcaead29638bb8ef6e1b036R150 - it's checking by name and will use anything other than the okhttp impl first.  This seemed like the simplest mechanism that didn't involve introducing an explicit priority, or name and config property, to select from multiple.

shawkins:
Testing concerns separated as #3887

httpclient-jdk/pom.xml
```diff
@@ -0,0 +1,151 @@
...
+                <Export-Service>${osgi.export.service}</Export-Service>
+                <Include-Resource>
+                  /META-INF/services/io.fabric8.kubernetes.client.http.HttpClient$Factory=target/classes/META-INF/services/io.fabric8.kubernetes.client.http.HttpClient$Factory,
+                </Include-Resource>
+              </instructions>
+              <classifier>bundle</classifier>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
```
manusa, line 147: This should be configured via properties, <plugin> definition should be inherited from the parent module and not redefined.

httpclient-okhttp/pom.xml
```diff
@@ -0,0 +1,159 @@
...
+                <Export-Service>${osgi.export.service}</Export-Service>
+                <Include-Resource>
+                  /META-INF/services/io.fabric8.kubernetes.client.http.HttpClient$Factory=target/classes/META-INF/services/io.fabric8.kubernetes.client.http.HttpClient$Factory,
+                </Include-Resource>
+              </instructions>
+              <classifier>bundle</classifier>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
```
manusa, line 155: https://github.com/fabric8io/kubernetes-client/pull/3857#pullrequestreview-890822440

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/HttpClientUtils.java
```diff
@@ -136,16 +142,26 @@ public static String basicCredentials(String username, String password) {
...
+    HttpClient.Factory factory = null;
+    for (Iterator<HttpClient.Factory> iter = loader.iterator(); iter.hasNext();) {
+      HttpClient.Factory possible = iter.next();
+      if (factory != null && WARNED.compareAndSet(false, true)) {
+        LOGGER.warn("There are multiple httpclient implementation in the classpath, "
+            + "choosing the first non-okhttp implementation. "
+            + "You should exclude dependencies that aren't needed or use an explicit association of the HttpClient.Factory.");
+      }
+      if (factory == null || !possible.getClass().getName().equals("io.fabric8.kubernetes.client.okhttp.OkHttpClientFactory")) {
```
manusa, line 157: As discussed during triage call, a less fragile approach to determine if this is the fallback/default client would be better.

kubernetes-examples/pom.xml
```diff
@@ -39,7 +39,12 @@
       <groupId>io.fabric8</groupId>
       <artifactId>openshift-client</artifactId>
     </dependency>
-
+    
+    <dependency>
+      <groupId>com.squareup.okhttp3</groupId>
+      <artifactId>okhttp</artifactId>
+    </dependency>
+
```
manusa, line 47: Isn't this provided transitively by the kubernetes-client dependency (kubernets-client <- kubernetes-httpclient-okhttp <- okhttp3)?

httpclient-okhttp/pom.xml
shawkins: Yes I had noted that above https://github.com/fabric8io/kubernetes-client/pull/3857#issuecomment-1039453862

I haven't updated this pr since those changes went in.

kubernetes-examples/pom.xml
shawkins: Only at a runtime scope.  There are compile references in the examples as well.

httpclient-okhttp/pom.xml
manusa: oh sorry, I missed that comment

kubernetes-examples/pom.xml
manusa: OK, shall we mark this as `provided`?

shawkins: No, PortForwardExample is creating it's own client, so it's a hard dependency.

manusa: OK, I see now.
I guess that to avoid confusion, we can refactor (in the future) the PodForwardExample to use something else (maybe just a plain URL connection.

httpclient-okhttp/pom.xml
shawkins: Removed the local plugin override and changed the packaging to bundle.

shawkins: Noticed that this didn't actually work - it was just because the feature wasn't updated and I had a previous local version of the bundle artifact.

I've tried several things locally and none have worked, so I'm reverting the prior changes.  This will need to be resolved under #3878 instead.

In looking at the osgi logs I think that #3865 may not be working as expected.

```
2022-02-23T09:26:20,773 | WARN  | features-3-thread-1 | BaseActivator                    | 63 - org.apache.aries.spifly.dynamic.bundle - 1.3.0 | Could not load provider io.fabric8.openshift.client.NamespacedOpenShiftExtensionAdapter of service io.fabric8.kubernetes.client.ExtensionAdapter
java.lang.NullPointerException: null
	at org.apache.aries.spifly.BaseActivator.lambda$registerProviderBundle$3(BaseActivator.java:268) ~[?:?]
	at java.util.Map.compute(Map.java:1171) ~[?:?]
	at java.util.Collections$SynchronizedMap.compute(Collections.java:2692) ~[?:?]
	at org.apache.aries.spifly.BaseActivator.registerProviderBundle(BaseActivator.java:261) ~[?:?]
	at org.apache.aries.spifly.ProviderBundleTrackerCustomizer.addingBundle(ProviderBundleTrackerCustomizer.java:173) ~[?:?]
	at org.apache.aries.spifly.ProviderBundleTrackerCustomizer.addingBundle(ProviderBundleTrackerCustomizer.java:64) ~[?:?]
	at org.osgi.util.tracker.BundleTracker$Tracked.customizerAdding(BundleTracker.java:475) ~[osgi.core-7.0.0.jar:?]
	at org.osgi.util.tracker.BundleTracker$Tracked.customizerAdding(BundleTracker.java:420) ~[osgi.core-7.0.0.jar:?]
	at org.osgi.util.tracker.AbstractTracked.trackAdding(AbstractTracked.java:256) ~[osgi.core-7.0.0.jar:?]
	at org.osgi.util.tracker.AbstractTracked.track(AbstractTracked.java:229) ~[osgi.core-7.0.0.jar:?]
	at org.osgi.util.tracker.BundleTracker$Tracked.bundleChanged(BundleTracker.java:450) ~[osgi.core-7.0.0.jar:?]
	at org.apache.felix.framework.EventDispatcher.invokeBundleListenerCallback(EventDispatcher.java:915) ~[?:?]
	at org.apache.felix.framework.EventDispatcher.fireEventImmediately(EventDispatcher.java:834) ~[?:?]
	at org.apache.felix.framework.EventDispatcher.fireBundleEvent(EventDispatcher.java:516) ~[?:?]
	at org.apache.felix.framework.Felix.fireBundleEvent(Felix.java:4817) ~[?:?]
	at org.apache.felix.framework.Felix.activateBundle(Felix.java:2393) ~[?:?]
	at org.apache.felix.framework.Felix.startBundle(Felix.java:2308) ~[?:?]
	at org.apache.felix.framework.BundleImpl.start(BundleImpl.java:998) ~[?:?]
	at org.apache.felix.framework.BundleImpl.start(BundleImpl.java:984) ~[?:?]
	at org.apache.karaf.features.internal.service.BundleInstallSupportImpl.startBundle(BundleInstallSupportImpl.java:165) ~[?:?]
	at org.apache.karaf.features.internal.service.FeaturesServiceImpl.startBundle(FeaturesServiceImpl.java:1160) ~[?:?]
	at org.apache.karaf.features.internal.service.Deployer.deploy(Deployer.java:1041) ~[?:?]
	at org.apache.karaf.features.internal.service.FeaturesServiceImpl.doProvision(FeaturesServiceImpl.java:1069) ~[?:?]
	at org.apache.karaf.features.internal.service.FeaturesServiceImpl.lambda$doProvisionInThread$13(FeaturesServiceImpl.java:1004) ~[?:?]
	at java.util.concurrent.FutureTask.run(FutureTask.java:264) [?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) [?:?]
	at java.lang.Thread.run(Thread.java:829) [?:?]
```

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/HttpClientUtils.java
shawkins: This has been updated to use isDefault, but I'm thinking we'll likely want to refine this again when we get around to #3887 - so that we can just have a single module with all client dependencies and system or config property to drive the selection.

httpclient-okhttp/pom.xml
shawkins: Spawned this as #3890

manusa (approved):

Issue #3845, opened 2022-02-11
Provide a way to write extensions without a compile dependency on kubernetes-client


shawkins:
### Is your task related to a problem? Please describe

There are a couple of follow-on issues after the split:
extensions use -client internal stuff, and in general have a compile dependency on the client.  Affected classes: 
./service-catalog/client/src/main/java/io/fabric8/servicecatalog/client/internal/ClusterServiceBrokerOperationsImpl.java
./service-catalog/client/src/main/java/io/fabric8/servicecatalog/client/internal/ClusterServiceClassOperationsImpl.java
./service-catalog/client/src/main/java/io/fabric8/servicecatalog/client/internal/ClusterServicePlanOperationsImpl.java
./service-catalog/client/src/main/java/io/fabric8/servicecatalog/client/internal/ServiceBindingOperationsImpl.java
./service-catalog/client/src/main/java/io/fabric8/servicecatalog/client/internal/ServiceInstanceOperationsImpl.java
./volumesnapshot/client/src/main/java/io/fabric8/volumesnapshot/client/internal/VolumeSnapshotClassOperationsImpl.java
./volumesnapshot/client/src/main/java/io/fabric8/volumesnapshot/client/internal/VolumeSnapshotContentOperationsImpl.java
./volumesnapshot/client/src/main/java/io/fabric8/volumesnapshot/client/internal/VolumeSnapshotOperationsImpl.java

### Describe the solution you'd like

The extensions are making use of HasMetadataOperation, but it requires a lot of boiler plate.  

VolumeSnapshotResource and VolumeSnapshotContentResource are just marker interfaces for later use.  The rest seem to just mix-in behavior, which could be handled by a proxy based framework.  The only part of that which isn't straight forward is the withIem on the context.  As far as I can see all of the calls using withItem(null) aren't actually needed.  ClusterServicePlanOperationsImpl.instantiateAnd does create a ServiceInstanceResource scoped to that particular item.  There is no public way to do that.

shawkins:
Wrt to the extensions there are two options:
- split them as well.  Their client code is using BaseClient and Handlers - this requires the fewest code changes.
- make them work off of just the api - this would require introducing a mechanism for operation mixin behavior with the ability to register the handler, and the client code would need to change to be based off of the KubernetesClient - for example rather than calling Handlers.getOperation, you would use KubernetesClient.resource.

@manusa @rohanKanojia do you have a preference?

shawkins:
[...]

Issue #3407, opened 2021-08-16
design of resource and resourceList methods


shawkins:
After #3364 the implementation logic for these methods aligns well with everything else.  However they are still not aligned completely with the rest of the api.

There are several possibilities:
1. Deprecate the methods fully.  Advise users to use resources instead:

```
client.resource(resource).createOrReplace();
```
Would be
```
client.resources(resource.getClass()).inNamespace(resource.getMetadata().getNamespace()).createOrReplace(resource);
```
The interface for resourceList would be minimized to just a handful of operations, as any operation could be implemented by 
```
resources.forEach(r -> client.resources(r.getClass())....);
```

2. Have resource return a Namespaceable<Resource> / Resource - this will be a breaking change.  Similar to the above you would minimize the resourceList methods.   That would make the namespace behavior more concise than above.

shawkins:
2 - is not a simple change.  BaseOperation/HasMetadataOperation can't implement two Namespaceable interfaces.  

The simplest thing we could do is change resource to return Resource<T>, so

client.resource(resource)

would be implemented by something like:
```
HasMetadataOperation resources = resources(item.getClass(), null);
BaseOperation op = resources.withItem(item);
String namespace = KubernetesResourceUtil.getNamespace(item);
if (namespace != null) {
   op = resources.inNamespace(namespace);
}
return op;
```
However that means you would loose the ability to later call inNamespace as that is not provided on the Resource interface.  The only drawback is that means you have to directly manipulate the resource object (or convert the string to object) to change the namespace.

shawkins:
Also relates to https://github.com/fabric8io/kubernetes-client/discussions/3860 - if you use the KubernetesClient.load methods they are different that the Loadable.load methods - they will default to inferring the namespace from the item, and allow for the namespace to be changed.  I'd like to make everything uniform if possible for 6.0

shawkins:
@manusa @rohanKanojia #3858 proposes to replace convert the resource method into one that returns NamespaceableResource - a Resource that also exposes an inNamespace call.  The implementation is roughly doing what is described in https://github.com/fabric8io/kubernetes-client/issues/3407#issuecomment-924905591  

Other than the interface name changes, the biggest difference in support is deletingExisting, which is discussed in the pr comments. 

[...]

PR #3858, opened 2022-02-14
Extension development using only api and resource/resourceList implementations

shawkins:
## Description
See #3845 - this shows what the ServiceCatalog and Camel-K extensions could look like written only against the api.

The core of the changes are:
- to provide Client and Resource adapters to extend from rather than BaseClient and HasMetadataOperation
- to make the client, kubernetes client, and namespaced client highly available
- add resources and another variant to Client that allows you to specify the expected Resource class type
- to add a withItem method (not properly represented yet, it's just tacked in for now) - in several places I'm inferring this was to be called, for example https://github.com/fabric8io/kubernetes-client/pull/3858/files#diff-0c1cdf2689ac477b41f476cf50f20f9f0a5c2f44ea420ea5a8579dbe56a4111cR69 seems to be missing associating the item with the context.

These changes by themselves are not breaking as you could still use a dependency on kubernetes-client to use the existing functionality.

There are some additional cleanup to be pursued:
- how we are holding / cleaning up references to handlers

## Type of change

## Checklist


shawkins:
The next commit more fully realizes the resource changes for #3407 - it had seemed like I need to get Resource usage straightened out prior to proceeding with #3845, but now it looks like it's more separable than I thought.  In any case I'll go back to that now and clean up the Handler logic.  I'll leave switching over the other extension until a later pr.

shawkins:
@manusa @rohanKanojia @iocanel What's also not addressed here (which can be spawned to its own issue) are the interfaces for the new methods:

NonNamespaceOperation.withItem
Operation.withItem
ListVisitFromServerGetDeleteRecreateWaitApplicable.getResources

This is because some of the choices here overlap with the concerns brought up wrt Namespace handling in https://github.com/fabric8io/kubernetes-client/issues/3407#issuecomment-1042989018

KubernetesClient.resource will use the namespace of the item by default, but Loadable.load will not.  I think users may find this confusing - see #3860 for example

Options include:

1. getResources - I don't think this needs a separate interface.
1.a. Have just the ListVisitFromServerGetDeleteRecreateWaitApplicable.getResources method
1.b. Have NamespaceListVisitFromServerGetDeleteRecreateWaitApplicable.getResources return a NamespaceableResource list

2. load / withItem
2.a. Document the current behavior and have withItem functioning the same as Loadable.load - that is not take any action based upon the item namespace (which is how it currently is in the pr).
2.b. The behavior of kubectl is to use the explicit namespace (-n) if provided, otherwise use the item namespace.  
The rules would be:
- if you ever call inNamespace, it wins
- if you haven't called inNamespace, use the namespace from the item if it's set
- use the default namespace

This could include up to the client level itself:
```
// Existing behavior - exception if item has a namespace that is not monitoring.  
// Proposed behavior - the item namespace is monitoring
this.kubernetesClient.inNamespace("monitoring").configMaps().load(item); 

// Existing behavior - exception if item has a namespace that is not monitoring.  
// Proposed behavior - the item namespace is monitoring
this.kubernetesClient.configMaps().inNamespace("monitoring").load(item); 

// Existing behavior - context namespace is changed to match the item.  
// Proposed behavior - the item namespace is monitoring
this.kubernetesClient.inNamespace("monitoring").resource(item); 

// Existing behavior - exception if item has a namespace that does not match the context.  
// Proposed behavior - the context namespace will match the item
this.kubernetesClient.configMaps().load(item); 
```

The problem with 2.b.:
- tracking that an explicit namespace has been set - we are not currently doing
- it's a breaking change, there are quite a few cases that have behavioral changes
- if you pass around just a reference to a Loadable (or similar) it won't be clear what the expected behavior is to downstream methods - will the resource be in the item namespace or in Loadable context namespace?  Ideally it shouldn't matter - it should be up to the caller to pass an appropriately contexted Loadable.  You could entertain putting methods on Loadable/KubernetesClient, and even Resource to get the default and the explicit namespace - there is already KubernetesClient.getNamespace - but there's no way to tell if that's been explicitly set. 
 Based upon the namespacing rules this.kubernetesClient.inNamespace("default") - would be different than this.kubernetesClient even though the namespace is the same.

3. withItem
3.a. Loadable.withItem - small breaking change as Loadable will need two type parameters
3.b. A new interface, such as Itemable and they both implement Itemable<R>
3.c. Just leave the method on Operation and NonNamespaceOperation 0 such that like 1.b Operation can return NamespaceableResource.

Other possible changes:
- deprecation of KubernetesClient.lists
- deletingExisting replacement
- Removal of deprecated methods / interfaces from ListVisitFromServerGetDeleteRecreateWaitApplicable (depends on what is being done about deletingExisting)

shawkins:
I have the changes ready that implement the previous comment (and related cleanups) should we want to go that direction: https://github.com/shawkins/kubernetes-client/pull/6

shawkins:
@manusa this should be ready for review again.  after the merge from master this also includes the removal of KubernetesClient.lists

After this is committed I'll open a pr with the work from https://github.com/shawkins/kubernetes-client/pull/6 and the additional changes to namespace handling.

shawkins:
Thought it might be better just to show the totality of changes, this commit should fail on things that expect an exception on a namespace mismatch.  I'll correct that and add migration notes in another commit.

shawkins:
The test failures https://github.com/fabric8io/kubernetes-client/runs/5300256746?check_suite_focus=true#step:6:680 are being caused by a new sequence of events such that the wrong apiGroupName is being used.  Essentially OperationContext.setApiGroupName is implemented by looking at the item first if non null, then will take the passed in value.  The test resources, like the BuildConfig for BuildConfigIT have apiVersion: "v1" - so if you create the context by first associating the invalid item then doing any other operation, it will revert to a null apiGroupName.

My workaround in the next commit is to call updateApiVersion(item) in withItem, but it does look like some of this needs changed - because it appears in the context that the item is the primary place for the apiGroupName, then in the Operation logic it appears that the context (typically without an item the value is coming from the class/rdc or is hardcoded in the XXXOperationsImpl constructor) is primary.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesResourceUtil.java
```diff
@@ -84,6 +84,21 @@ public static void setResourceVersion(HasMetadata entity, String resourceVersion
...
+   * @param namespace the new namesapce
+   */
+  public static void setNamespace(HasMetadata entity, String namespace) {
+    if (entity != null) {
+      ObjectMeta metadata = entity.getMetadata();
+      if (metadata != null) {
+        metadata.setNamespace(namespace);
+      }
+    }
+  }
```
manusa, line 100: I haven't checked the purpose of this util.
But was wondering if it might be interesting to initialize metadata instead of just ignoring if it's null:
```suggestion
  public static void setNamespace(HasMetadata entity, String namespace) {
    if (entity != null) {
      if (entity.getMetadata() == null) {
        entity.setMetadata(new ObjectMeta());
      }
      entity.getMetadata().setNamespace(namespace);
    }
  }
```

manusa: It makes sense to me that in scope of this class, both methods should initialize the entity's metadata if necessary.
Maybe we can consider this as a future change.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/OperationSupport.java
```diff
@@ -216,24 +212,33 @@ public URL getResourceURLForPatchOperation(URL resourceUrl, PatchContext patchCo
...
+      return item;
+    }
+    String itemNs = KubernetesResourceUtil.getNamespace((HasMetadata)item);
+    
+    if (Utils.isNotNullOrEmpty(namespace) && Utils.isNotNullOrEmpty(itemNs) && !namespace.equals(itemNs)) {
+      item = Serialization.clone(item);
+      KubernetesResourceUtil.setNamespace((HasMetadata)item, namespace);
+    }
+    return item;
+  }
```
manusa, line 227: Not sure if we have an explicit test for this, but full coverage and documentation of conditions should be provided since it's determinant for the Namespace resolution logic.

I'm not sure if this PR also includes the `default` namespace fallback logic discussed in some meetings.

shawkins: > Not sure if we have an explicit test for this, but full coverage and documentation of conditions should be provided since it's determinant for the Namespace resolution logic.

There's an addition to the migration guide https://github.com/fabric8io/kubernetes-client/pull/3858/files#diff-6cf6f0d1132c6d642456eda0b231e404d9b3743c0c0ff86346055c5f7c78c268R15

And https://github.com/fabric8io/kubernetes-client/blob/cfd501a75f96c8e02e2e554ecdfb4d3dd56f161c/kubernetes-tests/src/test/java/io/fabric8/openshift/client/server/mock/NamespacedItemTest.java highlighting the usage of the explicit namespace.

> I'm not sure if this PR also includes the default namespace fallback logic discussed in some meetings.

It does not, that issue is if the Config has no default namespace, and still in the 6.0.0 bucket.  It is a little related to the discussion about what should inNamespace(null) mean.

manusa: > And https://github.com/fabric8io/kubernetes-client/blob/cfd501a75f96c8e02e2e554ecdfb4d3dd56f161c/kubernetes-tests/src/test/java/io/fabric8/openshift/client/server/mock/NamespacedItemTest.java highlighting the usage of the explicit namespace.

This is what I was looking for :+1: :smiley:

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.java
```diff
@@ -51,6 +68,20 @@
...
+
+    ChangeNamespace(String explicitNamespace) {
+      this.explicitNamespace = explicitNamespace;
+    }
+
+    @Override
+    public void visit(ObjectMetaBuilder builder) {
+      builder.withNamespace(explicitNamespace);
+    }
+  }
```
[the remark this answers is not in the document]
manusa: @metacosm @iocanel

shawkins:
@manusa I've updated with the changes discussed in the meeting.  Feel free to squash this.

manusa (approved):

PR #3872, opened 2022-02-16
[java-gen - chore] Automate code formatting on compile

andreaTP:
## Description

I found a solution for automating the code formatting on `compile`, to avoid spurious diffs in the future.
I'm not a Maven expert, so, any hint or suggestion is very welcome, this is the only configuration I got to work.
The problem is that the `license:format` goal should run from the root and `spotless` is slightly changing the headers :-S

## Type of change

## Checklist


manusa:
From your PR, I understand you're talking about formatting the project code and not the generated code.

This probably belongs to a git hook or CI check (maybe checkstyle), and not a compile goal.

In addition, the current formatting for this modules is completely out of format as compared with the rest of the code-base (we really need to update the CONTRIBUTING.md file).

The first step should be to create some `ide-config` Eclipse config files that can be reused across most tools.

andreaTP:
@manusa what you are saying is correct.
But there are codebases that are already using automatic formatting on `compile` (e.g. https://github.com/java-operator-sdk/java-operator-sdk or even Quarkus itself).

Since we are about to have a breaking release, if you want to enforce a style globally (and not in a single sub-module) checking it in CI I like it!

I don't have super strong opinions on the style to be used as long as it's consistent, if you can provide a configuration I can try to integrate it 🙂

OneCricketeer:
Can't you bind `spotless:apply` to `prepare-package` lifecycle rather than a separate exec action / maven subprocess, essentially?

andreaTP:
@OneCricketeer the problem is that running `spotless:apply` messes up the headers and we need to run a following `license:format` from the root of the project to fix it 😞 

Anyhow, not sure how to proceed on this one, @manusa would you like to apply a consistent style on the whole codebase?
Would you be able to provide a configuration file to start with?

Issue #3625, opened 2021-11-30
HasMetadata should have convenience methods for annotations and labels


shawkins:
### Is your enhancement related to a problem? Please describe

Similar to #3600 if we are adding utility methods for owner references it would be good to have similar methods for labels and annotations.

### Describe the solution you'd like

This would cover at least for getting:

resource.getLabel("key");
resource.getOrDefaultLabel("key", "default");
etc.

It could also include setting labels/annotations as well.

### Describe alternatives you've considered

The code generation could be updated to use a default map for annotations and labels that ideally would also be omitted from serialization when empty.  I'm not sure how involved this would be, so it seems easier to put new methods on HasMetadata.

manusa:
How about making some sort of convenience method in HasMetadata that delegates to GenericKubernetesResource#get.

Example:
```java
  default <T> T get(Object... path) {
    return Serialization.jsonMapper().convertValue(this, GenericKubernetesResource.class).get(path);
  }
  default <T> T getOrDefault(T defaultValue, Object... path) {
    final T value = get(path);
    return value == null ? defaultValue : value;
  }
```

This would also require a minor change in `GenericKubernetesResource` to consider the metadata field (`if (path[0].equals("metadata")`)

shawkins:
This probably has a similar cost as a builder, which it seems expensive for just reads.

> This would also require a minor change in GenericKubernetesResource to consider the metadata field (if (path[0].equals("metadata"))

More generally there would be some utility in having everything parsed as just abstract model (map of maps - the old generic), and have the ObjectMeta aspect just applied as a view.  But I can see that doesn't work well with building logic - https://github.com/fabric8io/kubernetes-client/pull/3931/files#diff-68db8ad66510b2e36592ac1c814508958f9dfbc42ea19a3c2bcc2dda89305697R51 - and is also less performant for normal read access.

If it's not doable to have a default metadata, we at a could add HasMetadata methods - Optional<Metadata> getOptionalMetadata() and getOrCreateMetadata.  
But even with a guarantee that getAnnotations returns non-null

resource.getOptionalMetadata().map(m -> m.getAnnotations().get("key")).orElse(null)

that's only a little better than what you'd have to do today

Optional.ofNullable(resource.getMetadata()).map(ObjectMeta::getAnnotations).map(m -> m.get("key")).orElse(null)

So I think we should at least add:

resource.getAnnotations().get("key")
resource.getLabels().get("key")

[...]

PR #3898, opened 2022-02-24
Initialize labels and annotations on ObjectMeta

amydentremont:
## Description

Initialize the `labels` and `annotations` maps on ObjectMeta so they're empty Maps instead of null if there are no labels/annotations on a resource. 
The `additionalProperties` map is already initialized in ObjectMeta, as are all the List fields, so I don't see a reason to not initialize the `labels` and `annotations` too. 

## Type of change

## Checklist

It says to open the PR as a draft if you haven't completed the checklist, but most of this doesn't seem warranted for such a simple change, and I'm afraid this won't be looked at if I don't just open for review

### DIFF
--- a/kubernetes-model-generator/kubernetes-model-core/src/generated/java/io/fabric8/kubernetes/api/model/ObjectMeta.java
+++ b/kubernetes-model-generator/kubernetes-model-core/src/generated/java/io/fabric8/kubernetes/api/model/ObjectMeta.java
@@ -53,7 +53,7 @@ public class ObjectMeta implements KubernetesResource
 {
 
     @JsonProperty("annotations")
-    private Map<String, String> annotations;
+    private Map<String, String> annotations = new HashMap<>();
     @JsonProperty("clusterName")
     private java.lang.String clusterName;
     @JsonProperty("creationTimestamp")
@@ -70,7 +70,7 @@ public class ObjectMeta implements KubernetesResource
     @JsonProperty("generation")
     private Long generation;
     @JsonProperty("labels")
-    private Map<String, String> labels;
+    private Map<String, String> labels = new HashMap<>();
     @JsonProperty("managedFields")
     @JsonInclude(JsonInclude.Include.NON_EMPTY)
     private List<ManagedFieldsEntry> managedFields = new ArrayList<ManagedFieldsEntry>();


shawkins:
Linking to #3625 - The issue is that ObjectMetadata, labels, and annotations are all nullable (even with this change as the setters or getters are not guarding against null) and that complicates a lot of basic logic.  One option is to go further with this change and enforce they are all non-null.  The issue with that is that the serialization will change - that is instead of omitting metadata, you'll see metadata: {}, and similar changes for annotations and labels.  Changing the generated code to add @JsonInclude(JsonInclude.Include.NON_EMPTY) on annotations and labels could help.  

Another option is to have convenience methods - HasMetadata.getOrCreateMetadata(), etc. that will guarantee non-null results.

manusa:
What's the end purpose of this?

Is it to be able to `resource.getMetedata().getAnnotations().put("annotation", "is-safe-to-out")`.

Considering this use-case, there's what Steven already mentioned.

But there's also the immutable approach where you `new ResourceBuilder(resource).editOrNewMetadata().addToAnnotations("annotation", "this-is-safe-too").endMetadata().build()`

shawkins:
> What's the end purpose of this?

It could be for reads or writes.  For writes a builder can make sense.  For reads it's an awfully heavy weight approach.

manusa:
> For reads it's an awfully heavy weight approach.

#3625 (getters/ getOrDefault) should fix reads.

This more or less highlights the point of my question, can Amy's use-case be fixed/improved by #3625, or is there something else.

shawkins:
> This more or less highlights the point of my question, can Amy's use-case be fixed/improved by https://github.com/fabric8io/kubernetes-client/issues/3625, or is there something else.

Yes it can.  These are all driving at the same things.  Here's the options:

- convenience methods such as HasMetadata.getAnnotations, HasMetadata.getLabels that would be documented as to the behavior of auto-creating the ObjectMetadata, annotations, and labels.  Other methods like HasMetadata.getAnnotation wouldn't have to create anything.

- change the generation logic so that ObjectMetadata, annotations, and labels are non-null by default.  That especially makes sense for annotations and labels because it's done for other collections (arrays/lists), but would be inconsistent for ObjectMetadata as we don't do that for other nested object state.  If we don't do it for ObjectMetadata, you still would likely add a similar set of convenience as above.

- Doing https://github.com/fabric8io/kubernetes-client/issues/3625#issuecomment-1057714508 for reads is syntactically shorter that using a builder, but would be even more heavy weight.  I'll respond with some more thoughts over there.

amydentremont:
Sorry I missed these replies! And yeah I totally missed that this was a generated class 😅 

My purpose is for reading. I submitted this PR after fixing a bug where we got a NPE when no labels existed on a resource.
The code has been:
```
Optional<String> deployGroup = Optional.ofNullable(
   javaApp.getMetadata().getLabels().get("deployGroup")
);
```
And I needed to update it to first check that labels weren't null:
```
Optional<String> deployGroup;
if (javaApp.getMetadata().getLabels() == null) {
    deployGroup = Optional.empty();
} else {
    deployGroup =
       Optional.ofNullable(javaApp.getMetadata().getLabels().get("deployGroup"));
}
```

Having the convenience methods for getting labels and annotations would definitely help; I could use just `getLabel` for this example. Though personally I still just wouldn't expect `getLabels` and `getAnnotations` to return `null` instead of empty maps, especially when `additionalProperties` is initialized to an empty map in the same class.

PR #3899, opened 2022-02-24
fix #3859: refinements to how a deserialization class is chosen

### LINKED ISSUE #3859: Listing / loading a CAPI Cluster resource throws an exception
### Describe the bug

The `KubernetesDeserializer` does a lookup for internally defined classes when deserializing JSON/YAML.

I have a CRD where the kind is `Cluster` ([definition can be found here](https://github.com/kubernetes-sigs/cluster-api/blob/main/config/crd/bases/cluster.x-k8s.io_clusters.yaml)). This gets resolved as `class io.fabric8.kubernetes.api.model.Cluster`, which is used internally to describe a cluster as part of the kubeconfig (why is this even implementing `KubernetesResource`?). This class does not match my resource at all, so I expected the deserializer to return a `GenericKubernetesResource`. Right now this loading behaviour results in all kind of weird exceptions later on.

**Code analysis**
I noticed that in `io.fabric8.kubernetes.internal.KubernetesDeserializer.Mapping#getInternalTypeForName`, there is actually a check that the complete `apiVersion + kind` matches the found class. This check is however only done if there are more than 1 possible classes found, which is not the case here. I would suggest to always do this check to be on the safe side. Part of the code:
```
            // If only one class found, return it
            if (possibleResults.size() == 1) {
                return possibleResults.get(0);
            } else if (possibleResults.size() > 1) {
              // Compare with apiVersions being compared for set of classes found
              for (Class<? extends KubernetesResource> result : possibleResults) {
                String defaultKeyFromClass = getKeyFromClass(result);
                if (key.equals(defaultKeyFromClass)) {
                  return result;
                }
              }
              return possibleResults.get(0);
            } else {
              return null;
            }
```

next (development version)

### Steps to reproduce

When listing CAPI cluster resources as generic kubernetes resources, like so
```
kubernetesClient.genericKubernetesResources("cluster.x-k8s.io/v1beta1", "Cluster").list()
```
The following exception is thrown:
```
java.lang.NullPointerException: Cannot invoke "String.length()" because "this.input" is null
	at java.base/java.net.URI$Parser.parse(URI.java:3165)
	at java.base/java.net.URI.<init>(URI.java:623)
	at io.fabric8.kubernetes.client.utils.URLUtils.join(URLUtils.java:89)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.getRootUrl(OperationSupport.java:146)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.getNamespacedUrl(OperationSupport.java:153)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.getNamespacedUrl(OperationSupport.java:164)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.list(BaseOperation.java:414)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.list(BaseOperation.java:402)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.list(BaseOperation.java:83)
	at dev.ruddur.operator.clusterapi.SideroClient.clusterExists(SideroClient.kt:19)
```

Same goes for loading a resource of this kind:
```kotlin
val s = """
apiVersion: cluster.x-k8s.io/v1beta1
kind: Cluster
metadata:
  name: test
spec:
  clusterNetwork:
    pods:
      cidrBlocks:
        - 10.244.0.0/16
    services:
      cidrBlocks:
        - 10.96.0.0/12
  infrastructureRef:
    apiVersion: infrastructure.cluster.x-k8s.io/v1alpha3
    kind: MetalCluster
    name: test
  controlPlaneRef:
    apiVersion: controlplane.cluster.x-k8s.io/v1alpha3
    kind: TalosControlPlane
    name: test-cp
"""
val r = kubernetesClient.resource(s).get()
println(r)
```
Which throws:
```
Exception in thread "main" java.lang.ClassCastException: class io.fabric8.kubernetes.api.model.Cluster cannot be cast to class io.fabric8.kubernetes.api.model.HasMetadata (io.fabric8.kubernetes.api.model.Cluster and io.fabric8.kubernetes.api.model.HasMetadata are in unnamed module of loader 'app')
	at io.fabric8.kubernetes.client.BaseKubernetesClient.resource(BaseKubernetesClient.java:235)
	at dev.ruddur.operator.Main.main(Main.kt:55)
```

### Expected behavior

That I would be able to list and/or create `Cluster` resources using the `GenericKubernetesResource` APIs.

### Runtime

Kubernetes (vanilla)

1.22.3@latest

### Environment

Linux

_No response_

### Additional context

Current workaround is to explicitly register this resource as a GenericKubernetesResource:
```
KubernetesDeserializer.registerCustomKind("cluster.x-k8s.io/v1beta1", "Cluster", GenericKubernetesResource.class)
```

shawkins:
## Description
This updates the search strategy used by the KubernetesDeserializer - to not use classes with conflicting apiGroups.  

~~It goes further to not use classes with conflicting versions - but that leads to several test failures.  That was a concern of @andreaTP when dealing with multiple versions - if they are custom resources which do not implement additionalProperties, then it could be a lossful conversion (if ignoring unmatched fields) or an exception when fields don't match.~~  

Sorry I wasn't thinking straight, the internal types are only going to be from what's in the PACKAGES, so any laxity made there on version matching won't affect custom types.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Fix #3848: Supports Queue (cluster) API for Volcano extension
 * Fix #3582: SSL truststore can be loaded in FIPS enabled environments
 * Fix #3818: adding missing throws to launderThrowable
+* Fix #3859: refined how a deserialization class is chosen to not confuse types with the same kind
 
 #### Improvements
 * Fix #3811: Reintroduce `Replaceable` interface in `NonNamespaceOperation`
--- a/doc/MIGRATION-v6.md
+++ b/doc/MIGRATION-v6.md
@@ -2,6 +2,7 @@
 
 ## Contents:
 - [API/Impl split](#api-impl-split)
+- [Deserialization Resolution](#deserialization-resolution)
 - [Deprecation Removals](#deprecation-removals)
 - [IntOrString changes](#intorstring-changes)
 - [Object sorting](#object-sorting)
@@ -30,7 +31,11 @@ To use it, exclude the kubernetes-httpclient-okhttp dependency and add the kuber
 - client.utils classes including Base64, CreateOrReplaceHelper, DeleteOrCreateHelper, OptionalDendencyWrapper, etc. are not in the -api jar, they are still in the -client jar under utils.internal.
 - Some other effectively internal classes in dsl.base and other packages were moved to corresponding internal packages - it is unlikely this will affect you unless you developed a custom extension.
 
-### Deprecation Removals
+## Deserialization Resolution
+
+The group on the object being deserialized is not required to match the prospective class - even for built-in types.  This prevents the unintentional parsing of custom types without a registered class as a built-in type of the same name.  This also means that you must ensure the apiVersion values are correct on the objects you are deserializing as they will no longer resolve to built-in type of the same name when there is a mistake.
+
+## Deprecation Removals
 
 - Removed KubernetesClient.customResource / RawCustomResourceOperationsImpl, please use the generic resource api instead 
 - Removed HttpClientUtils.createHttpClient(final Config config, final Consumer<OkHttpClient.Builder> additionalConfig), please use the OkHttpClientFactory instead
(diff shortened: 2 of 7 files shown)


shawkins:
With the latest commit there's still a failure - we're trying to parse a Template, but it's class does not have the expected Group template.openshift.io annotation.  I can either further relax the parsing, but it seems like that and any other annotation issue should get fixed.

shawkins:
Just so this pr could be standalone, I updated the Template class group.  Also updated the apiVersion for the pod test that was failing - and added to the migration guide that the api group name is now strict.

manusa (approved):

PR #3905, opened 2022-02-25
fix #3628 removing sorting from lists and templates

### LINKED ISSUE #3628: Remove sorting from Template#getObjects
Template sorts objects in when retrieved through the getter using `HasMetadataComparator`.

_Originally posted by @shawkins in https://github.com/fabric8io/kubernetes-client/issues/3606#issuecomment-976753173_

> @rohanKanojia @manusa this may be a silly question, but why is sorting required?

> No idea, I guess that at some point the order on which the items was applied was important.
> Traced it back to fabric8io/kubernetes-model@78ab3a6#diff-4a305396a3d2548c17df3329d5251eda2dca65c060f858588d99068cc7d205e9

> If a reference can't be found, I'd say this behavior seems undesirable. Based upon HasMetadataComparator (which seems somewhat arbitrary) it look like a pretty print option that was added to KubernetesList and Template. Ideally it should optional or the user's responsibility.

shawkins:
## Description
Removing built-in object sorting, which allows for the Template class to be generated.

This also address one of the errors from #3899 since the Template class now has the appropriate group.

With this change and changes from #3858 (removal of lists) I think the KubernetesList class only be a public construct for Templates. 

I think the rationale for this sorting may have been to help with the creation order of items in a list - but that is not something that kubectl does and it does not seem expected that things get reordered. 

## Type of change

## Checklist


manusa:
> I think the rationale for this sorting may have been to help with the creation order of items in a list - but that is not something that kubectl does and it does not seem expected that things get reordered.

I don't know about why it's in this project (I remember trying to track this down when fixing a bug, but there was nothing significant on the first implementation). In JKube, this ordering is reused precisely for that. Making sure that required/dependency resources get created prior to the rest of dependent resources.

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/KubernetesList.java
```diff
@@ -71,8 +71,6 @@ public String getApiVersion() {
 
     @Override
     public List<HasMetadata> getItems() {
-        List<HasMetadata> sortedItems = new ArrayList<>(super.getItems());
-        Collections.sort(sortedItems, new HasMetadataComparator());
-        return sortedItems;
+      return super.getItems();
     }
 }
```
manusa, line 76: Why don't we just remove the method?

shawkins: Because of the other overrides: https://github.com/fabric8io/kubernetes-client/blob/9025af6260443ddb4d6f5613f66eb0ad6c8c0905/kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/KubernetesList.java#L61 - I can't imagine that's still relevant though.  I'm also going to remove these classes eventually #3906

PR #3913, opened 2022-03-02
fix #3912: simplifying extension mocking

### LINKED ISSUE #3912: Simplify extension mocking

Extension mocking currently requires specialized classes and a separate extension module.

To introduce more general mocking logic that requires only the base mocking dependency.  See https://github.com/fabric8io/kubernetes-client/issues/3876#issuecomment-1053729405

After this change there may be additional refinements depending upon the outcome of https://github.com/fabric8io/kubernetes-client/issues/3876#issuecomment-1054317024

_No response_

### Additional context

_No response_

shawkins:
## Description
fix for #3912 - a general approach to mock support.  For now everything that a user could be using was deprecated, rather
than removed

As mentioned on the other comment, I did not try to reduce the duplication between the extension classes because the other implementation should go away in a subsequent release.

## Type of change

## Checklist


oscerd (approved):

iocanel (approved):
Looks amazing!

shawkins:
Do we want to go one step further and add this functionality into the existing EnableKubernetesMockClient rather than introducing the new EnableKubernetesMock?

manusa:
> Do we want to go one step further and add this functionality into the existing EnableKubernetesMockClient rather than introducing the new EnableKubernetesMock?


I'm missing the point of the `EnableKubernetesMock` annotation. IMHO our end goal should be to provide a single JUnit5 annotation and JUnit4 rule and prescribe those. We should probably keep the old annotation+rule (if renamed) for compatibility reasons.

So in this case (and if my assumptions are correct), we should keep both annotations, but provide only a single extension that covers everything.

kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/KubernetesMockExtension.java
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  protected void initializeKubernetesClientAndMockServer(Class<?> testClass) {
+    EnableKubernetesMock a = testClass.getAnnotation(EnableKubernetesMock.class);
+    final Map<ServerRequest, Queue<ServerResponse>> responses = new HashMap<>();
+    mock = a.crud()
+      ? new KubernetesMockServer(new Context(), new MockWebServer(), responses, new KubernetesMixedDispatcher(responses), a.https())
+      : new KubernetesMockServer(a.https());
+    mock.init();
+    client = new DefaultKubernetesClient(mock.getMockConfiguration());
```
manusa, line 104: Why is an explicit client instantiation needed?

shawkins: Because we need the functionality that is below the KubernetesClient, see the next call.  We could adapt(BaseClient.class) instead if you want to avoid what will be a deprecated call for a while.

shawkins:
> So in this case (and if my assumptions are correct), we should keep both annotations, but provide only a single extension that covers everything.

I've updated the pr with using just a single annotation / extension.  The only difference is that by default all Clients seen in the test class will be created the same way as a KubernetesClient.  If that seems like an acceptable change, I'd just reuse the existing annotation.  The purpose of the extensions class may shift depending on the resolution of #3922

kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/KubernetesMockExtension.java
manusa: Then maybe we should change the return type of KubernetesMockServer#createClient to BaseClient. Then change that in the future if possible.

shawkins: Unfortunately createClient effectively became a user facing method and should not expose -client classes.

shawkins: This implies that usage of the ExternalResource won't work correctly in adapting scenarios.  We need KubernetesMockServerExtension and KubenetesServer to share the same KubernetesMockServer and client creation logic for that.

shawkins:
@manusa the next commit addressed your concerns and puts everything using the same mechanisms.

extensions/istio/tests/src/test/java/io/fabric8/istio/test/v1beta1/VirtualServiceTest.java
```diff
@@ -106,83 +104,84 @@ void testCreate() throws InterruptedException {
...
+   * ownerReferences: []
+   * spec:
+   * hosts:
+   * - "details"
+   * http:
+   * - route:
+   * - destination:
+   * host: "details"
+   * subset: "v1"
+   */
```
manusa, line 165: Indentation was broken here by spotless :sob: 
Need to wrap with `<pre>{@code` ... `}</pre>`

extensions/istio/tests/src/test/java/io/fabric8/istio/test/v1beta1/VirtualServiceTest.java
```diff
@@ -218,52 +217,52 @@ public void checkBasicVirtualService() throws Exception {
...
+   * rewrite:
+   * uri: "/newcatalog"
+   * route:
+   * - destination:
+   * host: reviews.prod.svc.cluster.local
+   * subset: v2
+   * - route:
+   * - destination:
+   * host: reviews.prod.svc.cluster.local
+   * subset: v1
```
manusa, line 242: https://github.com/fabric8io/kubernetes-client/pull/3913/files#r820488251

extensions/istio/tests/src/test/java/io/fabric8/istio/test/v1beta1/VirtualServiceTest.java
```diff
@@ -305,48 +304,48 @@ public void checkVirtualServiceWithMatch() throws IOException {
...
+   * port:
+   * number: 9090
+   * subset: "v2"
+   * - route:
+   * - destination:
+   * host: "reviews.prod.svc.cluster.local"
+   * port:
+   * number: 9090
+   * subset: "v1"
+   */
```
manusa, line 327: https://github.com/fabric8io/kubernetes-client/pull/3913/files#r820488251

extensions/istio/tests/src/test/java/io/fabric8/istio/test/v1beta1/VirtualServiceTest.java
```diff
@@ -391,24 +390,24 @@ public void checkVirtualServiceWithPortSelector() throws IOException {
...
+     * http:
+     * - route:
+     * - destination:
+     * host: ratings.prod.svc.cluster.local
+     * subset: v1
+     * fault:
+     * abort:
+     * percent: 10
+     * httpStatus: 400
+     */
```
manusa, line 410: https://github.com/fabric8io/kubernetes-client/pull/3913/files#r820488251

extensions/istio/tests/src/test/java/io/fabric8/istio/test/v1beta1/VirtualServiceTest.java
```diff
@@ -431,19 +430,19 @@ public void loadingFromYAMLIssue48() throws Exception {
...
+     * hosts:
+     * - recommendation
+     * http:
+     * - match:
+     * - headers:
+     * baggage-user-agent:
+     * regex: .*DarkLaunch.*
+     * 
+     * ...
+     */
```
manusa, line 445: https://github.com/fabric8io/kubernetes-client/pull/3913/files#r820488251

extensions/istio/tests/src/test/java/io/fabric8/istio/test/v1beta1/VirtualServiceTest.java
shawkins: The config has 

```<setting id="org.eclipse.jdt.core.formatter.comment.format_block_comments" value="true"/>```

On something like this, I'd opt for just removing from the file - we generally don't have comments with the raw yaml anywhere else.

manusa: I understand that <pre> enclosed Javadoc should not be altered.

Anyway, just remove it then. One way or the other, we shouldn't have invalid YAML.

shawkins: I've updated this to use formatter:off / on

manusa (approved):

PR #3925, opened 2022-03-03
fix #3745: throw better exceptions when a namespace is missing

shawkins:
## Description
This differs from what we talked about in that it addresses #3745 without introducing a "default" namespace.  This will simply produce better error messages.

## Type of change

 - [x] Bug fix (non-breaking change which fixes an issue)
 - [ ] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [ ] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [x] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [x] I Added [CHANGELOG](../CHANGELOG.md) entry regarding this change
 - [x] I have implemented unit tests to cover my changes
 - [ ] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/master/doc/CHEATSHEET.md) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [ ] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

Issue #3922, opened 2022-03-03
Adapting api and checks performed


shawkins:
ExtensionAdapter/ ExtensionAdapterSupport / APIGroupExtensionAdapter: 
- APIGroupExtensionAdapter is broken - could be made to work using calls to getApiResources(apiVersion) - but the getAPIGroupName as an apiVersion is not consistently specified.  In some classes it's just group, others just version, and others the full apiVersion.
- ExtensionAdapterSupport caching strategy is not generally safe, there could be a timing issue wrt the initial ask for support or you could later remove the extension.  Could just be a check of getApiGroup(name) rather than investigating the rootPaths
- In either of the above cases, a similar strategy as https://github.com/shawkins/kubernetes-client/compare/extension-api...shawkins:extension-mock?expand=1 could be used to intercept or even provide implementations for getApiGroup and getApiResources to provide mock support.
- OpenShift is the odd man out as it checks across everything for at least one openshift.  That makes it more involved to provide alternative BaseClient logic (not simply based upon the client class type).

My general question is how much of this is useful to implement / retain?  Possibilities include:
- make this work as originally envisioned - isAdaptable and adapt will both check for support at both the overall client and api level.  Caching of support may need to be refined to have a timeout.  It may be better to have an exclusion list on EnableKubernetesMock of supported clients, rather than an inclusion list.
- Expanding on the previous comment - separate the notion of isAdaptable and adapt.  adapt could do no checks (neither in ExtensionAdapter nor BaseClient) of support, and isAdaptable could be deprecated/renamed to isSupported - and potentially do no caching.  The idea would be that clients would be making only infrequent calls to isSupported, but adapt should generally be expected to work and is called both internally and by clients.

_Originally posted by @shawkins in https://github.com/fabric8io/kubernetes-client/issues/3876#issuecomment-1054317024_

manusa:
As agreed, we should deprecate `Client#isAdaptable` method and point to the new `isSupported` method. The new behavior of `isAdaptable` checks if the **client** can be converted/adapted to the Extension specific client implementation. With the current code-base isAdaptable resulting in false would only be possible in the case of OpenShift: a user has a dependency to the OpenShift Client API, but not to the OpenShift Client module (the call client.isAdaptable(OpenShiftClient.class) or client.adapt(OpenShiftClient.class) would fail).

[...]

PR #3931, opened 2022-03-04
fix #3865: using just a single builder

### LINKED ISSUE #3865: How to create clients via api

After #3654 DefaultKubernetesClient and DefaultOpenShiftClient ended up in their respective -client modules.  So we need a way to construct clients via just the api.

An initial proposal was https://github.com/fabric8io/kubernetes-client/pull/3654/commits/2d22886a71419d0fa5c35a20b752ade204973619 - to introduce client specific builders.  

They are using reflection - if that is not a good choice (for example for quarkus), we can switch to a service loader, or even do like slf4j and create/remove dummy classes in the api modules.

Whatever the outcome of this is, the highest level construct should be used to hold state for things like the Handlers - so that we can avoid static references.

Alternatives could include just a single client KubernetesClientBuilder:

new KubernetesClientBuilder().withConfig(...).build().adapt(OpenShiftClient.class)

### Additional context

_No response_

shawkins:
## Description
Removes the use of the OpenShiftClientBuilder and documents the use of KubernetesClientBuilder.  Uses the same reflection strategy as KubernetesResourceUtil, but I didn't create a utility method for it just yet.

So the only odd man out is HttpClientUtils.createHttpClient since it only uses the thread context classloader - but I believe that is ok for now as at worst all someone would need to do is set an appropriate context class loader or set the httpclient.factory on the client builder.  We should be able in subsequent releases move the logic from HttpClientUtils to the Kubernetes client factory to keep all in one place.

Depending upon the outcome of #3922 we can probably add a build(Class) shortcut for build().adapt(Class).

This also builds upon https://github.com/fabric8io/kubernetes-client/pull/3925/commits/0ca13258cd52ac174d10c6bccd4326be874a2e74 to change OpenShiftConfig to just a decorator on a base config - no state can be lost if an openshiftconfig is inadvertently converted to base config.

## Type of change

## Checklist


kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/Config.java
```diff
@@ -251,11 +255,14 @@ private Config(Boolean autoConfigure) {
...
-   *                               .build();
-   * }</pre>
+   * <pre>
+   * {
+   *   &#64;code
+   *   Config configFromBuilder = new ConfigBuilder(Config.empty())
+   *       // ...
+   *       .build();
+   * }
+   * </pre>
```
manusa, line 265: So my assumption is incorrect... 
https://github.com/fabric8io/kubernetes-client/pull/3913#discussion_r820702410

We need to change the formatter configuration to prevent this.

manusa: The correct option is `org.eclipse.jdt.core.formatter.comment.format_source_code`, which is working now: https://github.com/fabric8io/kubernetes-client/runs/5462546383?check_suite_focus=true

manusa (approved):

Issue #1601, opened 2019-06-23
question: use kubectl with KubernetesServer mock


palmerabollo:
Is it possible to start a `KubernetesServer` mock this way:

```
  import io.fabric8.kubernetes.client.server.mock.KubernetesServer;

  KubernetesServer server = new KubernetesServer(true, true);
```

and then use a standard `kubectl` client to make requests? Thank you for you help.

palmerabollo:
Not stale.

rohanKanojia:
@palmerabollo : have you figured this out yet? I think you're right you can use it like done here:
https://github.com/fabric8io/kubernetes-client/blob/38dc6e3237e9c94f2b143ff781962f3e57cdfe0e/kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/PodCrudTest.java#L37

palmerabollo:
@rohanKanojia I wasn't able to make it work back in June, but I'll double-check it as soon as I can (after Xmas) with the example you provide. Thanks for your help.

manusa:
I see there is a lot of potential in this feature and that it can be very useful if MockServer mocks could be configured by using JSON/YAML files and distributed as a binary. This way, the MockServer could be used for different technologies (JavaScript, Go, Python...)

For the scope of this issue it would be nice to actually test if running the server standalone is possible and creating a documented quick start/example on how to achieve this to test a couple of `kubectl` commands. If the community is interested we could work on the exposed additional features in the future.

palmerabollo:
@rohanKanojia I've tried it. The code below works. However, I'm still not able to access the server using kubectl or any other k8s client because afaik it's not possible to get a valid kubeconfig.

As @manusa pointed out, I wanted to access the k8s mock server from client that uses a different technology.

```java
import io.fabric8.kubernetes.api.model.Pod;
import io.fabric8.kubernetes.api.model.PodBuilder;
import io.fabric8.kubernetes.api.model.PodList;
import io.fabric8.kubernetes.client.KubernetesClient;
import io.fabric8.kubernetes.client.server.mock.KubernetesServer;

import java.io.IOException;

public class KubernetesMock {
    public static void main(String[] args) throws IOException {
        KubernetesServer server = new KubernetesServer(true, true);
        server.before(); // mock init
        KubernetesClient client = server.getClient();

        Pod pod = new PodBuilder().withNewMetadata().withName("pod").endMetadata().build();
[...]

PR #3936, opened 2022-03-05
mock: generation is an Integer field

petitlapin:
## Description

The generation field in the Object Metadata should be an Integer (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#objectmeta-v1-meta).
I'm using the Mock to test the api from https://github.com/kubernetes-client/c and it fails to parse the generation field because it is actually a string.

## Type of change

## Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -18,7 +18,7 @@ What types of changes does your code introduce? Put an `x` in all the boxes that
 
 ## Checklist
  - [ ] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
- - [ ] I Added [CHANGELOG](../CHANGELOG.md) entry regarding this change
+ - [ ] I Added [CHANGELOG](https://github.com/fabric8io/kubernetes-client/blob/master/CHANGELOG.md) entry regarding this change
  - [ ] I have implemented unit tests to cover my changes
  - [ ] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/master/doc/CHEATSHEET.md) accordingly
  - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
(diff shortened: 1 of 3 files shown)


petitlapin:
Hi,

I can open an issue to reference if needed. Also, in the template, the CHANGELOG should refer to https://github.com/fabric8io/kubernetes-client/blob/master/CHANGELOG.md, not https://github.com/fabric8io/kubernetes-client/CHANGELOG.md

I ran the compilation and found no bug with the existing tests.

petitlapin:
I reformated the whole file to follow the code style. The only change is 
```java
    metadata.put(GENERATION, getOrDefault(existingMetadata, GENERATION, "1"));
```
to
```java
    metadata.put(GENERATION, Integer.parseInt(getOrDefault(existingMetadata, GENERATION, "1")));
```

manusa:
> I'm using the Mock to test the api from https://github.com/kubernetes-client/c and it fails to parse the generation field because it is actually a string.

This is very interesting, is your setup available or documented somewhere?

We had similar request a while back: #1601

manusa (approved):

manusa:
> To unit test, I've created a small Java executable that starts the mock in CRUD mode and I'm connecting to it to create and list services (using boost in the c++ part to start the java executable and stop it at the end of the test).

Thanks for sharing.

This is the part I was interested on, thanks for sharing. So if I understood correctly, basically your testing process runs a separate JVM process with a dummy MockServer and probably prints the HTTP port. The C++ unit test process orchestrates this part, picks up the port, runs the tests, and finally stops the JVM process.


If I may, why did you choose to use Fabric8 for this purpose instead of [Microcks](https://microcks.io/) or other alternatives?

petitlapin:
> This is the part I was interested on, thanks for sharing. So if I understood correctly, basically your testing process runs a separate JVM process with a dummy MockServer and probably prints the HTTP port. The C++ unit test process orchestrates this part, picks up the port, runs the tests, and finally stops the JVM process.
> 

The port is hardcoded on both sides (using the same KUBERNETES_MASTER env variable), the ip is set to localhost so no need to have a cluster running, but yes, that's the principle.

> If I may, why did you choose to use Fabric8 for this purpose instead of [Microcks](https://microcks.io/) or other alternatives?

I'm not the one who did the research on the tool, but I guess we are already using the library so it's the same repo and one less tool to install aside :). And this one was easy to set up and use directly.

manusa:
Jut one last question :sweat_smile:, so you mind if I share this thread publicly?

petitlapin:
I'm not sure by what you mean publicly, but this discussion is already here so yes, you can share what I wrote here :)

PR #3937, opened 2022-03-05
fix #3334: adding basic support for server side apply

### LINKED ISSUE #3334: Add support for Server Side Apply
# Server Side Apply
 
Starting from kubernetes 1.16, [Server Side Apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) it's enabled by default.
 
This feature is very useful from a operator development point of view,
 since it would require fewer requests to the Kubernetes API during the reconciliation cycle.
 
Using Server Side Apply has also the advantage that inform to users
 who owns which field of a given resource.
 
# Proposed Change
 
Add a boolean field to the io.fabric8.kubernetes.client.dsl.base.PatchContext named serverSide. That if it's true, sets the request content type to `application/apply-patch+yaml`.

shawkins:
## Description
Adds basic support for server side apply via the additional patch type requested with #3334 - any higher level support for apply will be addressed under the other apply / createOrReplace issues.

This does ensure the fieldManager is set (fabric8 by default) and that the managedFields are cleared.

ServerSideIT shows the usage - typically all creates/updates would be done as server side apply.  If you don't create the resource with server side apply any subsequent usage of server side apply will throw an exception - which you can workaround by including the force flag.

We may want to consider defaulting the user agent to fabric8 - if you create a resource using create, it will set the fieldManager to okhttp.

## Type of change

## Checklist


shawkins:
On the UserAgent.  It looks like we are looking for a property key of "fabric8-kubernetes-client/" + Version.clientVersion() - that seems more like it should be the default value if not set in the Config.

oscerd (approved):

manusa:
> We may want to consider defaulting the user agent to fabric8 - if you create a resource using create, it will set the fieldManager to okhttp.

>  It looks like we are looking for a property key of "fabric8-kubernetes-client/" + Version.clientVersion() - that seems more like it should be the default value if not set in the Config.

Agree on the second, configurable value with default to `fabric8-kubernetes-client/v$version`

shawkins:
> Agree on the second, configurable value with default to fabric8-kubernetes-client/v$version

Applied that change as part of this pr - fabric8-kubernetes-client/$version

rohanKanojia (approved):

rohanKanojia:
Thanks, please add some doc about this in CHEATSHEET as well. This has been requested multiple times

shawkins:
> Thanks, please add some doc about this in CHEATSHEET as well. This has been requested multiple times

Sure.  Also updated some other cheatsheet information related to creating clients and generic types.

manusa (approved):

PR #3959, opened 2022-03-10
ci: Pipeline to run E2E tests for each httpclient implementation

manusa:
## Description

ci: Pipeline to run E2E tests for each httpclient implementation

Runs the tests based on the latest available SNAPSHOT builds.

## Type of change

## Checklist

### DIFF
--- a/.github/workflows/e2e-httpclient-tests.yml
+++ b/.github/workflows/e2e-httpclient-tests.yml
@@ -1,0 +1,84 @@
+#
+# Copyright (C) 2015 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


oscerd (approved):

rohanKanojia (approved):

shawkins (approved):
LGTM

Are you good with the dependency approach, or would you rather property drive the client selection?

manusa:
> or would you rather property drive the client selection?

I'm not following on this, could you elaborate?

With the current approach, for each new client we'll need to create a new profile and add the entry in the CI pipeline matrix. However, the rest of the workflows are completely unaffected by these changes (for example, the JDK dependency is only added when needed).

manusa:
I think this moves away from what we discussed other times. I thought that the agreement was:
- httpclient-okhttp is provided as a default with the kubernetes|openshift-client modules
- If the classpath contains an additional entry for a single httpclient (e.g. httpclient-jdk), this one is selected and a warning is logged
- If the classpath contains 2 or more additional entries for alternate httpclient, an error is thrown

If the previous spec is fulfilled, the exclusion in the test is not really necessary. However, for the sake of "unitness" and peace of mind, I like to keep the redundant exclusion. Other tests can take care of validating the httpclient module selection spec.

manusa:
> such as in a given kubernetes-test being able to declare that it's a HttpClient regression test

If the purpose is solely for internal testing, I don't think that this should be implemented in our code. Maven profile configuration should be enough for this use-case and would prevent coupling production code with testing environment preparation.

Anyway, let's move forward with the current pipeline and reevaluate later on.

Issue #3923, opened 2022-03-03
Remove the need for KubernetesResourceMappingProvider


shawkins:
I think we can also do away with the need for KubernetesResourceMappingProvider - basically the same mechanism as what is registering handlers that is proposed by #3845 but via the ExtensionAdapter.  All it needs to do is declare the packages of the models.  

It all depends on the expectation of if the user can make use of KubernetesDesializer / Serialization independently of the client as that will complicate this simplification.

_Originally posted by @shawkins in https://github.com/fabric8io/kubernetes-client/issues/3876#issuecomment-1051269384_

shawkins:
Relates to #3993

PR #3966, opened 2022-03-14
Moving static state to non-static

shawkins:
## Description
This is currently layered on #3955 so it's not quite as large as it seems.

In moving the static Adapters and Handlers logic to non-static and per client, that allowed me to further refine when and how calls were made to register clients/resources.  A similar change could be made for #3923  The end result is that almost all extensionadapters can be removed - they are only for the root of extensions now.  

There are only 2 real client implementations with this change - the DefaultKubernetes and DefaultOpenShift.  Everything else just wraps the client.  Rather than expanding the ClientContext, we just pass the client around.  Any modification to the client state - the config or httpclient - will make use of a copy constructor to copy over the adapters, handlers, and other state (such as the supported check overrides set by the mocking logic).  Because of this change we can now move/remove mocking logic off of the annotation and onto the KubernetesServer - as any modification that we make there can be applied to any client (root or derived).  

Also a lot of logic surrounding the creation of openshift clients has been greatly simplified - as most was no longer applicable.

The general strategy of Adapters / Handlers has not changed.  Everything is registered as a peer.  If we do want to support unregistering an extension, or doing fancier client conversion, this will need to change.

Because of the initial usage model in the extensions, I only made the override point a Resource, rather than a whole operation.  There are 5 openshift operations that provide any real customization - 3 can be refactored to the existing api as they contribute only Resource methods.  2 (ProjectOperationsImpl and TemplateOperationsImpl) contribute operation level methods - which would require additional extension api changes to accommodate.  So for now I added an internal interface to allow for their registration.  A stretch goal would be to turn the whole openshift client into just an api extension - but I don't think there is time for that.

Also several openshift operation impls were removed by correcting whether the resource is namespaced and its plural.

## Type of change

## Checklist


shawkins:
I'll rebase and add onStart to the KubernetesMockServer reset.

shawkins:
As for moving openshift to an api only implementation, beyond the operation refactoring the things that block that are:
- the client logic that switches the httpclient / configuration to openshift specific.  OpenShiftConfig, Readiness, and the OpenShiftOAuthInterceptor could all be moved to client to facilitate this.
- OpenShiftNamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl - it's a whole other abstraction
- CreateOnlyResourceOperationsImpl

shawkins:
Marking this as ready for review.

https://github.com/fabric8io/kubernetes-client/pull/3966#issuecomment-1075415793 - can be split as a follow up issue.  It's a non-trivial amount of work to support the openshift client as api only.

I'll also follow up on the static nature of deserialization with #3972

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/extension/ClientAdapter.java
```diff
@@ -36,19 +35,15 @@
 /**
  * To be used as the base class for creating extension clients
  */
-public abstract class ClientAdapter<C extends Client> implements Client {
+public abstract class ClientAdapter<C extends ClientAdapter<C>> implements Client {
```
manusa, line 38: In OperationContext a new method
```java
  <T extends HasMetadata, V extends VisitableBuilder<T, V>> ResourceHandler<T, V> getHandler(T item) {
    return getClient().adapt(BaseClient.class).getHandlers().get(item, getClient());
  }
```

Is this something we might want to replicate here so that we can reuse in the DSL clients?

```java
  Handlers getHandlers() {
    return getClient().adapt(BaseClient.class).getHandlers();
  }
```

or even:

```java
... getNamespacedHasMetadataCreateOnlyOperation(...) {}
... getNonListingOperation(...) {}
```

Or maybe it's just not worth it?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/osgi/ManagedKubernetesClient.java
```diff
@@ -155,9 +158,9 @@
     @Reference(referenceInterface = ExtensionAdapter.class, cardinality = ReferenceCardinality.OPTIONAL_MULTIPLE, policy = ReferencePolicy.DYNAMIC, bind = "bindExtensionAdapter", unbind = "unbindExtensionAdapter"),
     @Reference(referenceInterface = OAuthTokenProvider.class, cardinality = ReferenceCardinality.OPTIONAL_UNARY, policyOption = ReferencePolicyOption.GREEDY, bind = "bindOAuthTokenProvider", unbind = "unbindOAuthTokenProvider")
 })
-public class ManagedKubernetesClient extends BaseClient implements NamespacedKubernetesClient {
+public class ManagedKubernetesClient implements NamespacedKubernetesClient {
 
-  private NamespacedKubernetesClient delegate;
+  private DefaultKubernetesClient delegate;
```
manusa, line 163: Maybe it might make sense to expose the adapters or registration capabilities through an interface and we don't have to depend on the specific type.

kubernetes-client/src/main/resources/META-INF/services/io.fabric8.kubernetes.client.extension.ExtensionAdapter
```diff
@@ -13,46 +13,3 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
```
manusa, line 16: Does it make sense to keep this file and SPI?

kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/KubernetesMockServer.java
```diff
@@ -109,7 +110,26 @@ public NamespacedKubernetesClient createClient() {
...
+   * To exclude a specific apiVersion, you would fully specify
+   * route.openshift.io/v1
+   * <p>
+   * NOTE this affects calls to {@link Client#hasApiGroup(String, boolean)}
+   * and {@link Client#supports(Class)}. Other calls to get the full root path or other
+   * api group metadata will not return valid results in mock scenarios.
+   *
+   * @param unsupported apiGroup patterns
+   */
+  public void unsupported(String... unsupported) {
```
manusa, line 132: Are we sure we want to change the method name from setUnsupported to unsupported?

For me, this naming suggests that I might add new unsupported patterns to the ones I might have already added previously.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/extension/ClientAdapter.java
shawkins: > Is this something we might want to replicate here so that we can reuse in the DSL clients?

I'm trying to not expose Handlers to the api.  There are only 2 usages of getHandler - to get a Builder, and to create an operation in the same context as the current.

For the Builder I'd much prefer updating the model generation logic to do something similar to lombok - add a asBuilder() method / interface to each class.  Only if the resource does not implement Buildable, would you try to lookup the builder via the classpath.

Creating an operation in the same context as the current - that is only currently needed by the resourceList operations as you can do resourceList(...).dryRun().resources().  With the ExtensibleResourceAdapter that would only be needed if you override the behavior of a mutating method (create, edit, etc.) in a way that wanted to do a modification on an outside resource (none of the logic does this yet).  If you just add a new method they are not available to be called after a dsl method like dryRun.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/osgi/ManagedKubernetesClient.java
shawkins: Since Client methods are currently re-exposed on every Client subclass, I was wanting to keep the methods there to a minimum.  I'll update the managed openshift one though to just use OpenShiftClient.

kubernetes-client/src/main/resources/META-INF/services/io.fabric8.kubernetes.client.extension.ExtensionAdapter
shawkins: I'm fine with removing it.

kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/KubernetesMockServer.java
shawkins: Sorry, I had copied straight from the annotation.

manusa (approved):

kubernetes-client/src/main/resources/META-INF/services/io.fabric8.kubernetes.client.extension.ExtensionAdapter
manusa:
```
Error:  Bundle io.fabric8:kubernetes-client:jar:6.0-SNAPSHOT : Input file does not exist: target/classes/META-INF/services/io.fabric8.kubernetes.client.extension.ExtensionAdapter
Error:  Error(s) found in bundle configuration
```
Reference needs to be remved too

shawkins: Updated now.  I look forward to the day when osgi support is dropped, or at least pushed out as a concern of the core project.

manusa: :pensive:

linnjason678:
Shawkins, 
I understand some items , 
Are you trying to help" a person like myself " why would you want to make so many change s so in one 
Day to see if your correct or 2 see if 
What if we did this or what if we did that ? I m learning and not being
Smart butt towards you and anyone 
Else  of their teaching s.

PR #3981, opened 2022-03-18
Several informer related changes

### LINKED ISSUE #3968: The Store API is not writable
### Describe the bug

I am updating the kubernetes client in my application from 5.4.1 to the latest 5.12.1. And the API changed, I cannot write in the store of the informer anymore. The workaround I have is to explicietly cast the Store instance into a Cache, to be able to use the add function.
My use case is that we have an external database where is stored all the history of the statuses of the pods launched by Argo. To be sure that we don't miss any status update, we use the informer pattern. And in case of error where our application is not up, but pods are deleted during that time, before starting the informer we prepopulate the cache with the last known list of running pods. This way, at startup we get notified of any missed deletion. This pattern was found described in this exemple in the go client:
https://github.com/kubernetes/client-go/blob/2f52a105e63e9ac7affb1a0f533c62883611ba81/examples/workqueue/main.go#L202

Looking at the history of the fabric8 client, I see that the APIs have been removed by #3091

### Steps to reproduce

Use the Store API.

### Expected behavior

A writable Store API, like the go client has.

### Runtime

Kubernetes (vanilla)

1.21.6

### Environment

_No response_

### Additional context

_No response_

### LINKED ISSUE #3969: onUpdate is called on re-list even if resourceVersion is unchaged
### Describe the bug

When a re-list happens, an onUpdate event is received even if the object resourceVersion didn't change.
5.10.1

5.10.1@latest

### Steps to reproduce

create a test having a relist to a quite short version
wait for the relist to happen
see that an onUpdate is triggered for even a not changed object

### Expected behavior

the onUpdate should not be triggered if resourceVersion is unchanged

### Runtime

OpenShift

1.22.3@latest

### Environment

Amazon

_No response_

### Additional context

_No response_

shawkins:
## Description
This addresses a host of informer related requests:

Fix #3969 - stop emitting sync events on relist - it was simplest to make this change to support #3968
Fix #3968 - added SharedInformer.initialState to set

Also took the parts of #3943 that were most general purpose, such as being able to add/remove indexers at any time, along with deprecating Informable.withIndexers

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/impl/cache/CacheImpl.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/impl/cache/CacheImpl.java
@@ -15,6 +15,7 @@
  */
 package io.fabric8.kubernetes.client.informers.impl.cache;
 
+import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.ObjectMeta;
 import io.fabric8.kubernetes.client.informers.cache.Cache;
 import io.fabric8.kubernetes.client.utils.ReflectUtils;
@@ -29,7 +30,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.function.BooleanSupplier;
 import java.util.function.Function;
 
 /**
@@ -37,7 +37,7 @@ import java.util.function.Function;
  *
  * @param <T> type for cache object
  */
-public class CacheImpl<T> implements Cache<T> {
+public class CacheImpl<T extends HasMetadata> implements Cache<T> {
   // Defines how to map objects into indices
   private Function<T, String> keyFunc;
 
@@ -48,25 +48,18 @@ public class CacheImpl<T> implements Cache<T> {
   private final Map<String, Function<T, List<String>>> indexers = new HashMap<>();
 
   // items stores object instances
-  private volatile ConcurrentHashMap<String, T> items = new ConcurrentHashMap<>();
+  private final Map<String, T> items = new ConcurrentHashMap<>();
 
   // indices stores objects' key by their indices
   private final Map<String, Map<String, Set<String>>> indices = new HashMap<>();
-  
-  private BooleanSupplier isRunning = () -> false;
 
   public CacheImpl() {
     this(NAMESPACE_INDEX, Cache::metaNamespaceIndexFunc, Cache::metaNamespaceKeyFunc);
   }
 
   public CacheImpl(String indexName, Function<T, List<String>> indexFunc, Function<T, String> keyFunc) {
-    this.indexers.put(indexName, indexFunc);
     this.keyFunc = keyFunc;
-    this.indices.put(indexName, new HashMap<>());
-  }
-  
-  public void setIsRunning(BooleanSupplier isRunning) {
-    this.isRunning = isRunning;
+    addIndexFunc(indexName, indexFunc);
   }
 
   /**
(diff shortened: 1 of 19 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/impl/cache/CacheImpl.java
```diff
@@ -111,10 +97,27 @@ public synchronized T put(T obj) {
...
+    }
     this.updateIndices(oldObj, obj, key);
     return oldObj;
   }
 
+  static <T extends HasMetadata> int compareResourceVersions(T oldObj, T obj) {
+    if (oldObj == null) {
+      return -1;
+    }
+    return Long.compare(Long.parseLong(oldObj.getMetadata().getResourceVersion()),
```
hibnico, line 117: Even if I have never seen anything other than increasing numbers as resource versions, I think this is making too much assumption regarding the specification.
As declared in the official API, the resourceVersion is `an opaque value that represents the internal version of this object`, cf https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#objectmeta-v1-meta
As far as I can tell, the only thing we can do to tell if two versions are different, not order them.

shawkins: I agree that it's not inline with the official docs, but this is how resourceVersion will function until etcd is no longer being utilized.  You can see similar logic in the go client https://github.com/kubernetes/client-go/blame/45fb834fedea5c753501ea8ea597e2dc54080510/tools/cache/mutation_cache.go#L242

Obviously for your usage, you don't need this checking - if modifications are restricted to before the informer is running there's no consistency issue.  However I've been asked by the operator sdk team about also supporting proactive updates to the cache as well - and I really can't come up with any other to do that consistently.  My hope was to just address both of those cases together and providing a writable store in general.

The options are:
1. Offer an option to, or just allow, unsafe updates - I don't believe that's a good approach though
2. Restrict this checking to only when someone has made mutations after calling run and document that it may not always be supported
3. Only support modifications before run and still don't try to support consistent modifications after run has been called.

One more thing that I would need to add is that if this checking is done, is that relist needs to force updates.  In cases like this https://github.com/kubernetes/kubernetes/issues/107457 the restart of the api server will cause a relist and we'll see older resourceVersions - this is obviously an edge case which currently does't work even with the go client informers that are relisting with a resourceVersion.

shawkins: I'll update this to number 2 and also correct the update/delete to be atomic and then we can see what everyone thinks of that.

hibnico: Ok, I didn't know that resource versions were number due to etcd, nice to know. I am just a user of Kubernetes, so I can only relay, but I have a heard that for scaling kubernetes some companies have setup other kind of storage like postgres. Considering the code in the go client you pointed, I guess that even with other storage they ensured that resources version are still incresing numbers.
My biggest worry was about `Long.parseLong` raising exception and crashing all that.
Thank you for sharing.

shawkins: Looking at the additional complexity of accounting for forcing updates on relist and cases like https://github.com/kubernetes/kubernetes/issues/107457 - I'm having second thoughts as it's still not possible to ensure that your consistent with the api server.  So I'll have to keep maintaining it's not a good idea to manipulate the store on a running informer.  I'll refine this pr to 3 instead.

shawkins: Linking to #https://github.com/fabric8io/kubernetes-client/pull/3943#issuecomment-1067962930 and resolving this comment.  If the operator sdk want to pursue this logic, it will need to do so with an adjunct cache.

manusa (approved):

Issue #3982, opened 2022-03-18
BaseOperation.deleteList swallows return value making the BaseOperation.delete not behave as expected


XComp:
### Describe the bug

I wanted to test code where `client.configMaps().withLabels(labels).delete()` is called but an 404 HTTP error is returned. The expected behavior would be that the `delete()` call returns `false`. But it doesn't because of [BaseOperation.deleteList()](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/BaseOperation.java#L541) swallowing the return value of the internally used delete method. This results in not returning the right return value in [BaseOperation.delete:456](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/BaseOperation.java#L456).

### Fabric8 Kubernetes Client version

next (development version)

### Steps to reproduce

1. Create ConfigMap with labels
2. Mock the server response to return a 404:
```
final String path = String.format("/api/v1/namespaces/%s/configmaps/%s", NAMESPACE, configMapName);
server.expect().delete().withPath(path).andReturn(404, "").once();
```
3. Call `client.configMaps().withLabels(labels).delete()` on that label and evaluate the response.

### Expected behavior

`false` should be returned, but `true` was observed.

### Runtime

Kubernetes (vanilla)

### Kubernetes API Server version

next (development version)

### Environment

other (please specify in additional context)

### Additional context

This is an internal bug independent of any environment as far as I can see. That's why I labeled the Environment as `other`

shawkins:
I'm letting this issue take over for #2931 as well.  I know it's not strictly the same thing, but it should get addressed / resolved along with this.

Essentially the resolution here is to have deleteList return a boolean and for that to be returned by delete as well.

The next issue is what should be the return value for delete(empty list) - currently it's true.  There is no documentation on MultiDeleteable as to what the behavior should be.

The last is that the delete attempt will short circuit at the first failure, including a 404.  You could also make the case for continuing the delete attempts against the other items.

Possible alternatives:

List<T> deleteList() - return the list of items that were successfully deleted.  404 of an individual item would effectively be ignored (it's just a concurrency issue between when you obtained the list, and when you process the deletes).  For any other exception the list of items already deleted should be attached.

[...]

PR #3983, opened 2022-03-18
Delete issues

shawkins:
## Description
Addressing various delete issues from #3982
- adding deleteList methods that provide more information about exactly what has been deleted
- clarified the javadocs of existing delete methods
- have delete() of a list return false when an item can't be deleted
- added these changes to the resourceList interfaces - it was easiest to remove what had already been deprecated

## Type of change

## Checklist


rohanKanojia (approved):

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/dsl/MultiDeleteable.java
```diff
@@ -16,11 +16,47 @@
 
 package io.fabric8.kubernetes.client.dsl;
 
+import java.util.Arrays;
 import java.util.List;
 
-public interface MultiDeleteable<T> {
+public interface MultiDeleteable<T> extends ListDeletable<T> {
```
XComp, line 22: I'm not convinced that we should extend `ListDeletable` here. It feels that both interfaces serve different purposes: `ListDeletable` has an internal context whereas `MultiDeleteable` expects to get the resources passed as parameters.

Keeping these two separate doesn't harm their usage. Any implementing class can still implement both interfaces. IMHO, interfaces should be a collection of methods that serve the same purpose. That doesn't apply to these two methods in my opinion.

XComp, line 22:
```suggestion
public interface MultiDeletable<T> extends ListDeletable<T> {
```
I guess, without the `e` is the right version analogous to `ListDeletable`

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/BaseOperation.java
```diff
@@ -448,8 +449,7 @@ public boolean delete() {
       }
     } else {
       try {
-        deleteList();
-        return true;
+        return delete(list().getItems());
```
XComp, line 452: why not use `deleteList()` to remove redundant code?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/BaseOperation.java
```diff
@@ -1009,4 +1003,24 @@ public static URL appendListOptionParams(URL base, ListOptions listOptions) {
     return urlBuilder.build();
   }
 
+  @Override
+  public List<T> deleteList(List<T> items) throws ListDeleteException {
```
XComp, line 1007: There is some redundancy between the `delete(List)` and the `deleteList(List)` here. Maybe, we could align these two in a single private method?
In addition to that, `delete(List)` does a `null` check on the parameter. `deleteList(List)` doesn't.

XComp (commented):
Thanks for addressing my comments, @shawkins . I only looked at the list deletion code I described in #3982 . It fixes the overall issue. I have a few comments, though. Please have a look below.

shawkins:
@XComp this pr will change based upon the comments #3982

shawkins:
@manusa here's what this looks like with keeping the boolean return.

shawkins:
After chatting with @manusa we'll update to returning void for now, and capture a new issue for implementing collection delete against the api server (should be a small change) with the eventual goal of processing the return from single item and collection delete to return a list of objects or ids.

shawkins:
The big issue with changing to void rather than boolean is fixing up all of the tests - there are a couple hundred places where we are checking for that response.  I'll separate that off as a different issue.

manusa (approved):

Issue #3973, opened 2022-03-15
Further consolidate and expand resource methods


shawkins:
### Is your task related to a problem? Please describe

There are quite a few places where getting a hold of a Resource makes sense, we should do this as consistently as possible.

### Describe the solution you'd like

Instead of Itemable.withItem, we'll use resource.  Everywhere else that uses Nameable (except for pod and node metrics) could also use a resource method:

client.configMaps().resource(item) - instead of client.configMaps().withItem(item)
client.configMaps().resource(name) - instead of client.configMaps().withName(name)

For lists:

client.configMaps().resources() - instead of client.configMaps().list().getItems().stream()
client.resourceList(list).resources() - instead of client.resourceList(list).getResources()

From @manusa -
client.configMaps()...stream() or similar (vanilla or Reactive library specific return type), which might pave the way for a future async implementation
	- Blocking impl: list().getItems().stream() / Multi.createFrom / Observable.create / …
	- Non-blocking impl: watch/inform -> stream / reactive fwk specific streaming impl

Unfortunately inform will still be blocking.  The first operation it needs to do is a list.  A list could be non-blocking, but the result parsing will be tricky as non-blocking as the root object is a KubernetesList.  A watch could be non-blocking, but you won't know when the initial add events are done.

Issue #3875, opened 2022-02-17
Move static state to non-static


shawkins:
### Is your task related to a problem? Please describe

Adapters and Handlers are currently static, which creates problematic references to things like classloaders.

### Describe the solution you'd like

They should be moved to state on the KubernetesClientBuilder or whatever we eventually settle on for that construct.

manusa:
After 3858 the change can be achieved by switching Adapters and Handlers from utility to regular instantiable classes.

shawkins:
Switching Adapters to be non-static is straight-forward - especially if we ignore being able to directly call register/unregister.

Handlers is much more involved - it has a lot more usage and from places that don't currently have a reference to the client.

manusa:
TBH I only checked the Adapters part. Maybe we can split the issue in 2.

PR #3994, opened 2022-03-23
Consistent/expanded usage of the resource term

shawkins:
## Description
This implements #3973, it is layered on top of the changes for #3875 to minimize conflicts - so only the last commit is of interest and this will get rebased prior to inclusion.

The change here is instead of withItem, we'll use resource, then instead getResources off of a resourceList, we'll use resources, and finally there's the addition of a convenience method resources, which serves the same purpose as the resourceList method.

This requires making watchlistdeletable more generic as we need the resource type.  It also makes podresource non-generic as that is not needed.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/BaseOperation.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/BaseOperation.java
@@ -73,6 +73,7 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.function.UnaryOperator;
+import java.util.stream.Stream;
 
 public class BaseOperation<T extends HasMetadata, L extends KubernetesResourceList<T>, R extends Resource<T>>
     extends CreateOnlyResourceOperation<T, T>
@@ -257,7 +258,7 @@ public class BaseOperation<T extends HasMetadata, L extends KubernetesResourceLi
 
   @Override
   public R load(InputStream is) {
-    return withItem(unmarshal(is, type));
+    return resource(unmarshal(is, type));
   }
 
   @Override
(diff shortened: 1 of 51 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/BaseOperation.java
```diff
@@ -1011,4 +1012,9 @@ public static URL appendListOptionParams(URL base, ListOptions listOptions) {
     return urlBuilder.build();
   }
 
+  @Override
+  public Stream<R> resources() {
+    return list().getItems().stream().map(this::resource);
```
manusa, line 1017: What's the purpose of the map/peek? It will set the last listed item as the current item.

kubernetes-client/src/main/resources/META-INF/services/io.fabric8.kubernetes.client.extension.ExtensionAdapter
```diff
@@ -0,0 +1,15 @@
+#
```
manusa, line 1: ?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/BaseOperation.java
shawkins: map does not perform a peek, it will convert from a stream of HasMetadata to a stream of Resources

kubernetes-client/src/main/resources/META-INF/services/io.fabric8.kubernetes.client.extension.ExtensionAdapter
shawkins: That looks unintentional.  I'll back that out as this file is going away in the other pr.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/BaseOperation.java
manusa: OK, GH interface is playing dirty again.

manusa (approved):

Issue #3995, opened 2022-03-23
Implement test to verify #3957 // #3958


manusa:
### Description

#3958 provides a fix for #3957. The fix basically consist in calling `future.complete(value)` after `listener.onOpen(value)`. However, no test was implemented that makes sure this ordering is not reversed in the future.

https://github.com/fabric8io/kubernetes-client/blob/655f4de83da450dd20dc0a889dce7b42f794cc47/httpclient-okhttp/src/main/java/io/fabric8/kubernetes/client/okhttp/OkHttpWebSocketImpl.java#L86-L87

A unit test should be implemented to make sure that this bug is not reintroduced in future changes.

PR #3958, opened 2022-03-10
fix #3957: Lister `onOpen` should be called before marking the connection open

AdrianFarmadin:
## Description
Fix #3957 

In `PodOperationsImpl` class the `buildAsync` method is called with `ExecWebSocketListener`.
The `ExecWebSocketListener` is opening `outputPipe` and `errorPipe` in `onOpen`.
If the `future` is completed before the `outputPipe` and `errorPipe` are open, then the consumer can read from the pipes before  they are open. 

The pipes should be open first and then the future should be marked as complete. 

## Type of change

 - [X] Bug fix (non-breaking change which fixes an issue)
 - [ ] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [ ] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [X] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [X] I Added [CHANGELOG](https://github.com/fabric8io/kubernetes-client/blob/master/CHANGELOG.md) entry regarding this change
 - [ ] I have implemented unit tests to cover my changes
 - [ ] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/master/doc/CHEATSHEET.md) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [ ] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

PR #3997, opened 2022-03-23
Changes for client tests hanging and testability

shawkins:
## Description
The first code change needed is to address hung nightly tests.  Essentially one thread holding the reflector lock was waiting for the watch to start.  Another thread via onOpen was trying to stop the reflector.  I don't believe this happened before #3958 the The fix is to use an async watch start.  It does not yet move list to async, but that should be done at some point.

The next issue is that the Arquillian test framework will create a new KubernetesClient for each test method - and it does not call close on them.  Even if it did call close, since we aren't taking responsibility for the creation of the executor we are relying on garbage collection to cleanup the http client resources.  I introduced a partial change along these lines that nulls out the httpclient on close, but we may want to take ownership of a default executor such that we can shut it down as well.  That coupled with the tests calling close would ensure timely cleanup.

The last part of this will be to include a regression test suite for http clients - that will address #3995

## Type of change

## Checklist


shawkins:
This moves us much closer to having a non-blocking core for the client.  Should be back to passing with the tests.
- it removes the usage of the synchronous HttpClient.send
- adds async versions of list/watch to support informers and addresses the deadlock that was hanging tests
- the code takes more responsibility for cleaning up jdk client thread - if close is called.  If close is not called, then it still relies on garbage collection
- some of the integration tests will use a base class to close client created for each test method.  I didn't change everything, as we'll move away from arquillian #4008
- to still support withRequestConfig we can no longer use a threadlocal, so a new method is needed to create a client specifically for that RequestConfig.  This also requires httpclients to provide a reference to the factory - alternatively we could pass the factory into the DefaultKubernetesClient when it is constructed.
- consolidates several usages of ExecListener that should be waiting on results to explicitly wait for the exit code via a new ExecWatch.exitCode future.
- created a new module to test httpclients using a common base class.

Not done:
- We can eliminate the need to use the Informer executorservice to start informers, since there is a non-blocking start method now provided.  This can be handled as a separate issue.
- ExecWebSocketListener writes to output streams, which may be blocking - this should be refined.
- There are probably more usages of listeners that could use javadocs explaining not to use blocking logic.

shawkins:
It is possible to use a non-blocking json parser (as shown below), but it doesn't buy us much - the parameter replacement logic acts only on an in-memory string, so it needs the whole input anyway.  And we don't have a way to use individual list items as they are parsed.  So there's very little need to go to this level at this time.

```
JsonParser asyncParser = JSON_MAPPER.getFactory().createNonBlockingByteArrayParser();
    ByteArrayFeeder feeder = ((ByteArrayFeeder) asyncParser.getNonBlockingInputFeeder());

    CompletableFuture<HttpResponse<AsyncBody>> future = retryWithExponentialBackoff(new AtomicInteger(), client, request, (l, a) -> {
      l.forEach(b -> {
        try {
          if (b.hasArray()) {
            feeder.feedInput(b.array(), b.arrayOffset(), b.remaining());  
          } else {
            byte[] bytes = new byte[b.remaining()];
            b.get(bytes);
            feeder.feedInput(bytes, 0, bytes.length);
          }
          a.consume();
        } catch (IOException e) {
        }
      });
    });

    return future.thenCompose(r -> {
      boolean success = false;
      try {
        assertResponseCode(request, r);
        success = true;
      } finally {
        if (!success) {
          r.body().cancel();
        }
      }
      r.body().consume();
      return r.body().done();
    }).thenApply(v -> {
      feeder.endOfInput();

      try {
        return JSON_MAPPER.readValue(asyncParser, type);
      } catch (IOException e) {
      }
    });
```

shawkins:
This next commit further consolidate threading concerns - which corrects NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl mistakenly using the fork join pool.  

The thread usage breaks down as follows:
- 1 self shutting down scheduling thread
- a common thread pool that is not bounded in terms of thread or work queue (these are for tasks that shouldn't be run in the scheduling thread or may be blocked)
- Dedicated threads for use by PortForwarderWebsocket (a socket listener and a pumper), and ExecWebSocketListener (a pumper)

Ideally we could remove the pumpers threads, but there's no guarantee on what the respective incoming ReadableByteChannel and InputStream support - in fact with the way the code is now there's a problem with ExecWebSocketListener - if the stream does not report anything as available, nothing will be read.  We may want to make it the user responsibility for calling InputStreamPumper.asInterruptible or do a specifically check to see if the inputsteam is System.in.

If we take the next step the scheduler and common thread pool could be non-static such that we could bound the common pool - and shut it down when the client closes.

shawkins:
@manusa this should be ready for review

manusa (approved):

Issue #3869, opened 2022-02-16
[java-gen] Support default attributes


OneCricketeer:
### Describe the bug

`default` fields are not converted into the Java models. 

Example: https://github.com/fabric8io/kubernetes-client/blob/1d7d869681faafe166e5581bcde0f928499ef765/java-generator/core/src/test/resources/akka-microservices-crd.yml#L44-L47

This includes all types

- `integer`
- `boolean`
- `string`
- `object`
- etc.

### Fabric8 Kubernetes Client version

5.10.1@latest

### Steps to reproduce

1. Have CRD with `default` attributes
2. Use `java-gen` util
3. Inspect Java model output to see defaults are not applied via field attributes or in getter methods

### Expected behavior

For the above example, I would expect

```java
@JSONProperty("replicas")
int replicas = 1;
```

Or 

```java
@JSONProperty("replicas")
Integer replicas;

public Integer getReplicas() {
  return Optional.ofNullable(this.replicas).orElse(1); // example
}
```

I understand that object types will be more complicated, but maybe some `FooSpec_DEFAULT` instance could be created. 

### Runtime

Kubernetes (vanilla), minikube

### Kubernetes API Server version

1.21.6

### Environment

Linux, macOS

andreaTP:
Just to make sure we take the specification into account when we implement this (it's rather tricky):
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting
and
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting-and-nullable

Additional note on the need to tweak Jackson as well, probably something on those lines:
```java
@JsonSetter(nulls=Nulls.SKIP)
@JSONProperty("replicas")
Integer replicas = 1;
```

fabiobrz:
I was playing a bit with this, and have a potential - although not complete - candidate. 
While working on the implementation I was wondering whether it would be good to base the changes on top of https://github.com/fabric8io/kubernetes-client/pull/3817, since that would take the nested structure package into account, which could be key for properly setting object-type defaults.

WDYT @andreaTP? (CC @manusa in case you'd want to assign)

andreaTP:
In #3817 we are making a lot of improvements toward a more correct generation.

At this point almost anything else will conflict with it, I hope we can prioritize a review there 🙂 .

[...]

PR #3918, opened 2022-03-02
[java-gen] Generate Builder pattern for CustomResource

andreaTP:
## Description

Revamped this branch to follow up on the discussion we had here:
https://github.com/fabric8io/kubernetes-client/pull/3932

This should be ready for review 🙂  cc. @manusa 

#### Old description:

E2E tests to compare the output of the "curated" extensions with the output of the code generated using the `java-gen` tool.
Sharing to discuss about a few interesting facts:

 - @manusa this is a starting implementation of what we discussed, but those tests are slightly hard to write and slow to execute 😞 
 - writing the code to use the java generated code is painful (e.g. slow, I do it in another project and copy-paste the result ...) as the IDE doesn't help in the Maven integration tests and we are relying on generated code + sundrio annotations, maybe @manusa knows some tricks that can be used in this case?
 - now the "slow" part of the integration tests is now the generation of classes done by `sundrio`(for the camel CRDs we are generating > 500 classes), should we look into improving performance there using std lib parallel collections (or [something more nifty](https://github.com/pivovarit/parallel-collectors) ) ? wdyt @iocanel ?

## Type of change
 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [ ] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [X] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [X] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [ ] I Added [CHANGELOG](../CHANGELOG.md) entry regarding this change
 - [ ] I have implemented unit tests to cover my changes
 - [ ] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/master/doc/CHEATSHEET.md) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [ ] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

PR #4019, opened 2022-03-29
[java-gen] Adding support for default values (Serialization::unmarshal based)

fabiobrz:
## Description

Draft/attempt to support default values generation in a different way than in https://github.com/fabric8io/kubernetes-client/pull/3944. 

The approach here is to delegate the actual deserialization of the default value (as a `JsonNode` content) to the `Serialization::unmarshal` method, which is used already extesilvely in the client.

Consequently, the generated code will depend on `io.fabric8.kubernetes.client` classes and the actual deserialization will take place at runtime.

Fixes https://github.com/fabric8io/kubernetes-client/issues/3869

FYI @andreaTP 

## Type of change

## Checklist

### DIFF
--- a/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
+++ b/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
@@ -15,9 +15,15 @@
  */
 package io.fabric8.java.generator.nodes;
 
+import com.fasterxml.jackson.databind.JsonNode;
+import com.github.javaparser.ast.expr.Expression;
+import com.github.javaparser.ast.expr.NameExpr;
+import com.github.javaparser.ast.stmt.BlockStmt;
+import com.github.javaparser.ast.stmt.ReturnStmt;
 import io.fabric8.java.generator.Config;
 import io.fabric8.java.generator.exceptions.JavaGeneratorException;
 import io.fabric8.kubernetes.api.model.apiextensions.v1.JSONSchemaProps;
+import io.fabric8.kubernetes.client.utils.Serialization;
 
 import java.util.Locale;
 import java.util.function.Function;
@@ -40,6 +46,7 @@ public abstract class AbstractJSONSchema2Pojo {
   protected final String description;
   protected final Config config;
   protected final boolean isNullable;
+  protected final JsonNode defaultValue;
 
   public abstract String getType();
 
@@ -54,10 +61,19 @@ public abstract class AbstractJSONSchema2Pojo {
     return description;
   }
 
-  protected AbstractJSONSchema2Pojo(Config config, String description, final boolean isNullable) {
+  protected JsonNode getDefaultValue() {
+    return defaultValue;
+  }
+
+  protected String getClassType() {
+    return getType();
+  }
+
+  protected AbstractJSONSchema2Pojo(Config config, String description, final boolean isNullable, JsonNode defaultValue) {
     this.config = config;
     this.description = description;
     this.isNullable = isNullable;
+    this.defaultValue = defaultValue;
   }
 
   /** Takes a random string and manipulate it to be a valid Java identifier */
@@ -169,7 +185,7 @@ public abstract class AbstractJSONSchema2Pojo {
     final boolean isNullable = Boolean.TRUE.equals(prop.getNullable());
     switch (nt.getType()) {
       case PRIMITIVE:
-        return new JPrimitive(nt.getName(), config, prop.getDescription(), isNullable);
+        return new JPrimitive(nt.getName(), config, prop.getDescription(), isNullable, prop.getDefault());
       case ARRAY:
         return new JArray(
             fromJsonSchema(
@@ -181,7 +197,8 @@ public abstract class AbstractJSONSchema2Pojo {
                 config),
             config,
             prop.getDescription(),
-            isNullable);
+            isNullable,
+            prop.getDefault());
       case MAP:
         return new JMap(
             fromJsonSchema(
@@ -193,9 +210,10 @@ public abstract class AbstractJSONSchema2Pojo {
                 config),
             config,
             prop.getDescription(),
-            isNullable);
+            isNullable,
+            prop.getDefault());
       case OBJECT:
-        boolean preserveUnknownFields = Boolean.TRUE.equals(prop.getXKubernetesPreserveUnknownFields());
+        final boolean preserveUnknownFields = Boolean.TRUE.equals(prop.getXKubernetesPreserveUnknownFields());
         return new JObject(
             parentPkg,
             key,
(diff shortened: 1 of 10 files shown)


java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
```diff
@@ -212,11 +215,20 @@ public GeneratorResult generateJava() {
         if (!prop.isNullable) {
           // from https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting-and-nullable :
           // "_null values for fields that either don't specify the nullable flag, or give it a false
-          // value, will be pruned before defaulting happens..._"
+          // value, will be pruned before defaulting happens. If a default is present, it will be applied_"
           objField.addAnnotation(
               new SingleMemberAnnotationExpr(
                   new Name("com.fasterxml.jackson.annotation.JsonSetter"),
                   new NameExpr("nulls = com.fasterxml.jackson.annotation.Nulls.SKIP")));
+          if (prop.defaultValue != null) {
```
andreaTP, line 223: Have you attempted to move this block out from the `if` for the nullable?
If we have set all the Jackson annotations properly this might work according to spec even defining always the body block.

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
```diff
@@ -212,11 +215,20 @@ public GeneratorResult generateJava() {
...
+            // always add an "init_*" method which will be expanded by statements needed to create the default value
```
andreaTP, line 224: The actual body of the method is trivial, I would suggest to avoid generating additional methods that can possibly cause collisions in edge cases. Can you directly assign the block expression generated by `generateDefaultInitializerBody` ?

java-generator/core/src/test/java/io/fabric8/java/generator/GeneratorTest.java
```diff
@@ -504,4 +511,60 @@ void testClassNamesDisambiguationWithPackageNesting() {
     assertTrue(clzO2.get().getFieldByName("o2").isPresent());
     assertTrue(clzO2.get().getFieldByName("o3").isPresent());
   }
+
+  @Test
+  void testObjectWithDefaultFields() {
```
andreaTP, line 516: this test looks pretty complicated, as we already have compilation and approval tests to verify the actual "content" of the default, I would be happy to only test that a default value exists/doesn't exists when is defined/when is not.

I would be actually more interested into using the integration tests with the Maven plugin to verify that the behaviour is according to specification e.g.: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting-and-nullable

andreaTP (changes requested):
Thanks a lot @fabiobrz !

This is definitely going in the right direction!
The implementation is simple and straightforward (and can be slightly simplified even more 🙂  )

I added a few comments but I really like this PR already, thanks a lot!

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
```diff
@@ -54,10 +60,31 @@ public String getDescription() {
...
+  }
+
+  /**
+   * This method is responsible for creating a block statement that
+   * represents the body of a default value initialization method.
+   *
+   * @return a {@link com.github.javaparser.ast.stmt.BlockStmt} instance that contains all the statements that code the
+   *         default value initialization.
+   */
+  protected BlockStmt generateDefaultInitializerBody() {
```
andreaTP, line 74: I understand that this request might sounds a little "odd", but I would prefer if we can keep this encoding in the "parent" class (e.g. probably is sufficient a `private` method in JObject) and expose the `defaultValue` using a Getter.

The reason is:
currently the recursion is leveraging a single code "brach" that is implemented using `generateJava`, keeping it uniform makes it easier to refactor in the future and improve performance on the hot path.

Do you see any problem in implementing it in this way?

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
```diff
@@ -54,10 +60,31 @@ public String getDescription() {
...
+    BlockStmt body = new BlockStmt();
+    Expression unmarshallExpression = new NameExpr(
+        "io.fabric8.kubernetes.client.utils.Serialization.unmarshal(\"" + this.defaultValue.toString().replace("\"", "\\\"")
```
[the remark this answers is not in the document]
fabiobrz: Fixed, let me know whether this fits in with what you meant.

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
fabiobrz: Right, fixed.

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
fabiobrz: Indeed, fixed.

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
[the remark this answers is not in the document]
fabiobrz: Yep, done as per your suggestion.

java-generator/core/src/test/java/io/fabric8/java/generator/GeneratorTest.java
fabiobrz: Test is now way more simple. Integration testing can be done in a different PR.

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
fabiobrz: I do agree, fixed.

manusa:
Hi @andreaTP, could you please review again :pray:

andreaTP:
Thanks again @fabiobrz , this implementation is really excellent! (especially enjoyed the use of `StringEscapeUtils`!)

I approve this since it's good to go, it would be nice if we can add (even in a follow up PR) an integration tests (maybe based on https://github.com/fabric8io/kubernetes-client/pull/3918 ) covering the actual behavior and interaction with nullable (e.g. taken from the kubernetes docs).

andreaTP (approved):

fabiobrz:
> Thanks again @fabiobrz , this implementation is really excellent! (especially enjoyed the use of `StringEscapeUtils`!)

Thanks you as well, for providing valuable feedback n the changes.
 
> I approve this since it's good to go, it would be nice if we can add (even in a follow up PR) an integration tests (maybe based on #3918 ) covering the actual behavior and interaction with nullable (e.g. taken from the kubernetes docs).

Yes, since this belongs to a preview implementation I think that integration tests combining defaulting and nullable should be provided in a follow up PR. I'll log the issue, and maybe take care of it, since that's something I am interested in as well.

fabiobrz:
Tiny code format change to trigger CI checks again.

manusa (approved):

oscerd (approved):

andreaTP:
Nice! 🎉  Thanks!

Issue #4012, opened 2022-03-28
Design of operations with varargs


shawkins:
### Is your enhancement related to a problem? Please describe

With

CreateOrReplaceable.createOrReplace
InOutCreateable.create

Supporting varargs is hiding a design issue.  There should be separate methods:

CreateOrReplaceable.createOrReplace()
CreateOrReplaceable.createOrReplace(T item)

The no-arg version should be supported off of Resource, resourceList, etc.

The single arg version should be supported off of collection context.  The no-arge will simply throw an exception when it's called off of a collection context:

configMaps().createOrReplace() -- exception

We can even consider deprecating the single arg version.  It's just a short-cut for:

configMaps().resource(value).createOrReplace()

Deletable/MultiDeletable has similar issues.  The meaning of Delete with a boolean return changes between a single item and a collection context.  The proposals to either return void or a list of deleted items would make it consistent.

delete(item... ) is only available off of MultiDeletable.   That means you can do:
configMaps.delete(configMap), but not configMaps.dryRun().delete(configMap) - as the latter is only Deletable.

### Describe the solution you'd like

deprecate / remove all of the xxx(item) methods.

The deprecation / removal has follow-on interface changes.  Things like DryRunable / WritableOperation would need cleanups as well. 

@manusa @rohanKanojia what would you be in favor of for 6.0?

PR #4027, opened 2022-03-31
DSL refinements for #4012

shawkins:
## Description
This is an intermediate state for #4012 - it does not yet deprecate or remove the single argument functions like create(item).  After seeing the impact it seems best that we start with deprecation, rather than full removal.  As described on #4012 - it's a little hard to make all of this work the way we want, but after working around the dryRun issue this is my best representation.  There's still nothing we can do easily to differentiate between named and item based resource.  What this mostly does is remove the vararg method signatures like create(T... items) and refines the collection context to not expose any no-arg functions - configMaps().create()

The result is a more focused set of classes that is still api compatible with the old - with just a few already deprecated things removed removed, and a few additions such as resourceList(...).inAnyNamespace()

It's just a draft, so the functions aren't fully documented, spotless has not been run, etc.

Removed:
Applicable
CascadingEditReplacePatchDeletable
CreateFromserverGettable
EditReplacePatchDeletable
Operation
VersionWatchAndWaitable
WatchListDeletable
StatusUpdatable
Cascading

New/moved:
FilterWatchListMultiDeletable -> AnyNamespaceOperation
ItemWritableOperation
ItemReplacable 

The ItemReplacable and ItemWritableOperation now represent where the single argument function are called - both for collection and resource contexts.  There are a few more methods that could be handled this same way - including patchStatus.

Given that this is already so large, I'd probably shoot for cleaning this up before proceeding with additional deprecations.

Let me know if there are any concerns with this before I proceed further.

## Type of change

## Checklist


kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/dsl/ItemReplacable.java
```diff
@@ -16,14 +16,26 @@
 
 package io.fabric8.kubernetes.client.dsl;
 
-public interface Cascading<T> {
+public interface ItemReplacable<T> {
+
+  /**
+   * Similar to calling resource(item).replaceStatus()
```
metacosm, line 22: Should we aim to remove functionality duplication? More specifically, can we make it so that there's only one way to perform an operation? It would reduce the cognitive load of users.

shawkins: Yes that is the eventual goal.  There are several concerns.  The first is that it will be a major api change, which I'm now leaning towards using deprecations initially.  Based upon the changes in this pr it will be relatively easy to introduce CollectionItemReplacable and CollectionItemWritableOperation which are deprecated - so that calling configMaps().create(x) would be deprecated and the user would be pointed to configMaps().resource(x).create().

The second is that there are other points of inconsistency in the dsl.  patchStatus(item), as mentioned above, does not currently perform a diff, so it could be offered as patchStatus().  ProjectOperation.createProjectAndRoleBindings(name, ...) - does not belong as a collection level method, and instead should be projects().withName(name).createProjectAndRoleBindings(...).  Before we fully remove things we should probably identify what all of these are.

The last is that to be done completely will be a lot of work.  The core issue is the differentiation between calling configMaps().withName("x") and calling configMaps().resource(x) - they both return the same Resource, or Resource subclass.  However you shouldn't be able to call configMaps().withName("x").create(), nor does configMaps().resource(x).create(y) make sense.  In theory you'd like to split the Resource class, such that when withName is called you'd logically have:
```
NamedResource<T> extends
    FromServerGettable<T>,
    Lockable<? extends Deletable>,
    WatchAndWaitable<T>, Versionable<WatchAndWaitable<T>>,
    Deletable,
    DryRunable<? extends Deleteable>,
    Requirable<T>, Readiable, Informable<T>
```

Lockable and DryRunable would need to be converted to non-generic, but other than that it's representative.  Then

```
Resource<T> extends NamedResource<T>
    Lockable<ReplaceDeletable<T>>,
    WritableOperation<T>,
    DryRunable<WritableOperation<T>>
```

However the ability to extend a Resource - (PodResource, ServiceInstanceResource, etc.) is based only off of extending Resource.  You would instead need both NamedPodResource and PodResource.  So something like we currently have as:

```
public interface ServiceInstanceResource extends Resource<ServiceInstance> {

    ServiceBinding bind(String secretName);
}
```

You would actually want to be a mixin that is applied to both the NamedResource and the Resource - 
serviceInstances().withName("x").bind("secret") or serviceInstances().resource(x).bind("secret")

Alternatively you could view resource(x) as a way to mixin Lockable<ReplaceDeletable<T>>, WritableOperation<T>, DryRunable<WritableOperation<T>>

While there are relatively few Resource subclasses left, the code is setup to only know about one Resource type.  So the simplest solution for now is to just allow configMaps().withName("x").create() and similar methods to exist, but throw exceptions.

shawkins:
@manusa let's determine whether to proceed with this, then I can address #4028

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/BaseOperation.java
```diff
@@ -216,6 +214,9 @@ public R withName(String name) {
 
   @Override
   public ExtensibleResource<T> lockResourceVersion(String resourceVersion) {
+    if (resourceVersion == null) {
```
manusa, line 217: `Utils.isNotNullOrEmpty` instead?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.java
```diff
@@ -242,6 +229,9 @@ public List<HasMetadata> get() {
 
   @Override
   public ListVisitFromServerGetDeleteRecreateWaitApplicable<HasMetadata> inNamespace(String explicitNamespace) {
+    if (explicitNamespace == null) {
```
manusa, line 232: `Utils.isNotNullOrEmpty` instead?

We might also want to implement an alternative `Utils.isNotBlank`

manusa (approved):
LGTM, thx!

I really like where this is going. Can we merge it then?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.java
shawkins: It can be - but we aren't doing that check with the other inNamespace locations.  Ideally we would be consistent about not accepting empty / null anywhere a name was expected - so maybe a separate issue to cover that.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/BaseOperation.java
shawkins: https://github.com/fabric8io/kubernetes-client/pull/4027#discussion_r848315325 - should apply here as well.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/NamespaceVisitFromServerGetWatchDeleteRecreateWaitApplicableListImpl.java
manusa: Creating an issue

manusa: https://github.com/fabric8io/kubernetes-client/issues/4057

shawkins:
> I really like where this is going. Can we merge it then?

I believe so.  There will be more javadocs added in the follow-on issue to add deprecations.

PR #4043, opened 2022-04-06
Ensure that e2e cron jobs don't run on forks

geoand:


manusa (approved):

manusa:
LGTM, thx!

As an additional note, newer forks should not be affected by this:

https://github.com/laravel/framework/issues/34356#issuecomment-718831832
https://news.ycombinator.com/item?id=24655458

Also, users should be responsible for their fork setup
image

geoand:
> Also, users should be responsible for their fork setup

The problem is that if I just do a PR here and there, I am not really aware of the CI setup. Moreover, if I do sporadic PRs to various projects, then I need to disable jobs for each project, when I would rather not have to worry about any of that :)

manusa:
> The problem is that if I just do a PR here and there, I am not really aware of the CI setup. Moreover, if I do sporadic PRs to various projects, then I need to disable jobs for each project, when I would rather not have to worry about any of that :)

I'm happy with the change :) It was just a reminder that we should be more careful and responsible. I'm one to be blamed for wasting resources in tons of forks that I just haven't configured to disable the pipelines due to laziness. :sweat_smile:

geoand:
I would argue it's a wrong default on Github's part :)

PR #4060, opened 2022-04-12
deprecating single argument operations and removing extra interfaces

shawkins:
## Description
Addressing #4012

@manusa this also more aggressively removing interfaces - as long as we don't think we'll ever generate the dsl from metadata (the supported verbs, if there's a status subresource, etc.), then we can remove more of the specialized interfaces like Createable.  I'll update the migration doc if we agree on this.

~~Some of the new interfaces may also be temporary - for example once ItemReplacable is removed, AnyNamespaceOperation will just collapse down to FilterWatchListDeletable.~~ - Actually we need to still differentiate between the top-level operations, like load, and the filterwatch...

## Type of change

## Checklist


shawkins:
This also adds a no-arg patchStatus.  I did not yet address projects().createProjectAndRoleBindings nor patch(context, item).

For patching anything that is not a JSON patch the item is not necessary when using an actual resource - as long as the resource has metadata.name specified.  For example:

```
resource(x).patch(PatchContext.of(JSON_MERGE)) 
```

will work as long as x has a name.

```
resource(x).patch(PatchContext.of(SERVER_SIDE_APPLY)) 
```

is of course expected to work as x should be the full resource.

However if don't want the name as part of the patch, then you would still want to do:

```
withName("name").patch(PatchContext.of(JSON_MERGE), x)
```

So for now I've just left the single item argument signatures for patch.

manusa (approved):

PR #4068, opened 2022-04-14
fix #4028 changing delete return type to list of statusdetails

### LINKED ISSUE #4028: Delete return type

Delete returning boolean is semantically unclear as noted in #3983 - however it's used extensively in our own test logic, so it was separated off as it's own issue.  

We have two choices.

Return void - the simplest option, but will break how we currently test delete behavior.

```List<T>``` - easy to implement for now since our deletes are based upon individual or lists of items - but may be more difficult to implement for collection deletes.  To account for this we may want to use ```Optional<List<T>>``` instead or a full blown DeleteResult that would allow for use to evolve the contract more as needed.

_No response_

### Additional context

_No response_

shawkins:
## Description
Addresses #4028 and adds support for delete collection.

These changes are not finalized as I'm waiting on the arquillian refactoring to update the itests.

## Type of change

## Checklist


shawkins:
On the implementation of delete - if the resource is not namespaced, or if there is a name, or a namespace, it will use a delete / deletecollection depending upon whether the name is specified.  For a namespaced item with no name / namespace, the logic will first do a list to obtain the namespaces the resource is in, then issue a deletecollection against each namespace.

@manusa in looking at the verb support there's definitely a bug in the kubernetes metadata - almost every resource that advertises delete support also supports deletecollection - even if that is not listed in the verbs.  The only examples I could find where deletecollection wasn't supported, but delete is were with things like RoleBinding - so rather than the logic attempting to compensate with a list delete we'll allow that exception to bubble up.

manusa (approved):
LGTM, thx!
Please fix the merge conflicts

manusa (approved):

PR #3806, opened 2022-02-07
refactor (kubernetes-model-core) : Refactor IntOrString class to avoid invalid field combinations

rohanKanojia:
## Description
    
Right now IntOrString determines whether the wapper object contains an
integer or string by the value of integer field `kind` - `0` for integer
and `1` for string.
    
This field is exposed via setters so it's also possible that user can
mutate object to produce invalid combinations. For example:
```
IntOrString i1 = new IntOrString("23");
    
i1.setIntVal(32);
// Would print "23"
System.out.println(Serialization.jsonMapper().writeValueAsString(i1));
```
    
Refactor IntOrString class to remove these `intVal`,`strVal`,`kind`
fields and just rely on one `Object` field; use `instanceOf` to
distinguish whether enclosing object is integer or string.
    
We faced an issue due to this behavior in
https://github.com/eclipse/jkube/issues/1126

Signed-off-by: Rohan Kumar <rohaan@redhat.com>

## Type of change

 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [ ] Feature (non-breaking change which adds functionality)
 - [X] Breaking change (fix or feature that would cause existing functionality to change
 - [ ] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [X] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [x] I Added [CHANGELOG](../CHANGELOG.md) entry regarding this change
 - [x] I have implemented unit tests to cover my changes
 - [X] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/master/doc/CHEATSHEET.md) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [ ] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

PR #4078, opened 2022-04-21
fix: IntOrString supports null values

manusa:
## Description
fix: IntOrString supports null values

- For backwards compatibility, IntOrString should still support
null values.

Relates to:
- #3806 

## Type of change

## Checklist


oscerd (approved):

rohanKanojia:
Why are we doing this? Did someone reported this issue? Was it breaking some deserialization ?

manusa:
> Why are we doing this? Did someone reported this issue? Was it breaking some deserialization ?

It breaks lots of stuff in JKube. In addition, it also makes sense that if you have a public empty constructor which leaves the value as null, that you can actually pass a null argument to the other constructor too.

So considering that this is harmless, the resulting object instance can also be achieved via the other constructor, and that it makes the changes backwards-compatible, IMHO we should keep it this way.

Issue #4009, opened 2022-03-25
Consider Deployments & StatefulSets ready when desired replica count is 0


amydentremont:
### Is your enhancement related to a problem? Please describe

Scaling StatefulSets and Deployments down to 0 desired replicas is a real use case for my company. We especially scale java application deployments down to 0 replicas temporarily to "pause" them in a kubernetes cluster, so that when we run java apps locally in Intellij for testing they're not completing with a java app running in the kubernetes cluster to process/update resources. 
The issue is, [`Readiness.isDeploymentReady`](https://github.com/fabric8io/kubernetes-client/blob/v5.8.0/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/readiness/Readiness.java#L147-L154) always returns `false` for deployments with a desired replica count of 0, because k8s sets `availableReplicas` to `null` instead of `0` in this case on the deployment resource. (The `Readiness.isDeploymentReady` logic linked always returns false if `availableReplicas` is `null`.)
For StatefulSets, it's the same issue but with `readyReplicas` being null instead of 0 when the desired replica count is 0, and [`Readiness.isStatefulSetReady`](https://github.com/fabric8io/kubernetes-client/blob/v5.8.0/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/readiness/Readiness.java#L128-L134) not accounting for this. 

We have implemented an extension of `Readiness` where `isReady` does return true for deployments and StatefulSets with spec.replicas=0, but we can't override the `isStatefulSetReady` and `isDeploymentReady` methods since they're static. We could copy them into another class with our own behavior instead, but I'm afraid someone would still be bound to use `Readiness.isDeploymentReady` and be confused by the inconsistent behavior.

### Describe the solution you'd like

Ideally, `isStatefulSetReady` and `isDeploymentReady` would check if spec.replicas is 0, and if so, consider the statefulSet/deployment ready if `readyReplicas`/`availableReplicas` is null or 0 (with the current version of k8s, they will only be null not 0, but might as well check for 0 too so it's handled if k8s changes this at some point). 

For example [the statefulSet case](https://github.com/fabric8io/kubernetes-client/blob/v5.8.0/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/internal/readiness/Readiness.java#L128-L134) would be like this: 
```
public static boolean isStatefulSetReady(StatefulSet ss) {
[...]

Issue #3996, opened 2022-03-23
Spurious access to deployment.extensions API from deployments.apps


andreaTP:
### Describe the bug

This is an issue that is really puzzling me.

In the Keycloak operator, only when deploying it with OLM [this call](https://github.com/keycloak/keycloak/blob/3ebfc91b75c3ca6c48bd54380fe4ffe891829aad/operator/src/main/java/org/keycloak/operator/v2alpha1/KeycloakDeployment.java#L110-L117):

```java
client
    .apps()
    .deployments()
    .inNamespace(getNamespace())
    .withName(getName())
    .get();
```

throws an exception:
```
cannot get resource "deployments" in API group "extensions"
```

[We are not giving such permissions](https://github.com/keycloak/keycloak/blob/3ebfc91b75c3ca6c48bd54380fe4ffe891829aad/operator/src/main/kubernetes/kubernetes.yml#L6-L9) but the very same docker image deployed without OLM works without any issue.

I checked the environment variables, and, as far as I can tell, there are no differences.

Is the `openshift-client` doing any additional call to `extensions.deployments` in case it detects some specific CRDs?
I'm running out of ideas and any help in debugging this issue is very welcome 🙏 .

### Fabric8 Kubernetes Client version

other (please specify in additional context)

### Steps to reproduce

Deploy the Keycloak operator using OLM instead of plain resources.

### Expected behavior

The Kubernetes client should behave consistently.

### Runtime

minikube

### Kubernetes API Server version

1.22.3@latest

### Environment

macOS, other (please specify in additional context)

### Fabric8 Kubernetes Client Logs

```shell
2022-03-23 18:27:14,557 ERROR [org.key.ope.v2a.KeycloakController] (EventHandler-keycloakcontroller) --- Error reconciling: io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://10.96.0.1/apis/apps/v1/namespaces/default/deployments/example-kc. Message: Forbidden!Configured service account doesn't have access. Service account may have been revoked. deployments.extensions "example-kc" is forbidden: User "system:serviceaccount:default:keycloak-operator" cannot get resource "deployments" in API group "extensions" in the namespace "default".
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.requestFailure(OperationSupport.java:683)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.requestFailure(OperationSupport.java:662)
[...]

PR #4080, opened 2022-04-21
Minor changes to consider 0 replicas and disable the backwards compatibility interceptor by default

shawkins:
## Description
Addresses #4009 and #3996 

## Type of change

## Checklist


manusa (approved):

oscerd (approved):

rohanKanojia (approved):

amydentremont:
Thanks for this change!

This doesn't matter as much, but it would be more consistent with other cases if `status.readyReplicas` was still taken into account when `spec.replicas` is 0, such that a deployment/statefulSet with `spec.replicas` 0 wouldn't be considered ready if `readyReplicas` was still 1 or more, since the desired state wouldn't really have been met yet in that case

shawkins:
@amydentremont when I watched deployments scaling I didn't see any intermediate readyReplicas - the controller will update/remove the unavailableReplicas, then the next update will have null readyReplicas, so I discounted the case where there still may be readyReplicas set with replicas=0 - but you are correct it would be more consistent with the existing code.  

However I'm still not sure about the notion of readiness here - there are three possibilities:  
- ready replicas == replicas
- ready replicas >= replicas
- condition Available=True, which should the message MinimumReplicasAvailable

Only with the first one is a Deployment not ready with 0 replicas in what should normally be a very narrow window of time.  It will always be ready with the other two.  

Please open a pr if you want a further refinement.

PR #4083, opened 2022-04-21
fix #4081 #4082: correcting/removing versionable, also improving informOnCondition

### LINKED ISSUE #4081: withResourceVersion off of collection problems with waitable
waitable for mixedoperations is currently only implemented for a single resource context.  However it is available off of a collection context via:

client.configMaps().withResourceVersion("x").waitUntilCondition...

Either this should be disallowed, or the implementation should apply the condition / readiness to each item - that is the behavior of those methods when called from a resourceList - client.resourceList(...).waitUntilCondition(...)

### LINKED ISSUE #4082: informOnCondition should not test initial add events

If you are using informOnCondition to check for non-existence the initial add events will provide a partial list and therefore potentially a misleading state.

Test the state initially based upon the full list.

_No response_

### Additional context

_No response_

shawkins:
## Description

Fix #4081
Fix #4082

For #4081 we don't want waitable under versionable - waitable requires a list, which has "no older than" not exact resourceVersion semantics.

With kubernetes 1.19+ there is support for resourceVersionMatch, which includes resourceVersionMatch=Exact, but that should probably be supported with another method - withResourceVersion(version, match) that would return something listable / informable / waitable.

The condition logic in BaseOperation was also changed to check each item, rather than assume a single to support collection contexts.

@manusa noted a related issue that is appears at least projects do not support a fieldSelector of metadata.name, which is how we implement withName when watching/informing.  This will need another issue to capture this in documentation and/or add a check for multiple items when only 1 is expected.

For #4082 the BaseOperation logic will perform / test the result of a list operation, then use that as the informer initial state - that ensures that initial add events won't be processed individually.

## Type of change

## Checklist


shawkins:
Considering the possibility of setting a limit, we can't simply call list - it would need to be paginated.  So I took the fix a different direction - as much as possible we'll defer add events until the store is complete.  The only case where this is not possible is when using limit and a non full state store - which doesn't apply to informOnCondition and related calls.

manusa:
Once this is merged we can update the ProjectIT delete assertion

manusa (approved):

manusa (approved):

PR #4133, opened 2022-05-10
fix (kubernetes-client) : Throw exception when `HasMetadata` is used in `resources(...)` API (#3990)

### LINKED ISSUE #3990: kubernetesClient.resources(HasMetadata.class).inNamespace("default").withLabels(Map.of("foo", "bar").list() throws NullPointerException
### Describe the bug

When trying to obtain a list of any resource matching a certain set of labels, a NullPointerException is thrown.

```java
foo = kubernetesClient.resources(HasMetadata.class).inNamespace("default").withLabels(Map.of("foo", "bar").list()
```

The NPE is thrown here io.fabric8.kubernetes.client.dsl.base.OperationSupport#getRootUrl:
```
      return new URL(URLUtils.join(config.getMasterUrl().toString(), "api", apiGroupVersion));
```

`apiGroupVersion` is null at this point.

I'm sure I'm invoking the API incorrectly, but I couldn't find how to do this in a different way.

I'm using the operator, and I think it's pulling in the client version 5.12, but I can't select that below.

5.10.0

### Steps to reproduce

Run Java code as shown above and observe the NPE.

### Expected behavior

A list of resources is returned.

### Runtime

1.22.3@latest

### Environment

macOS

_No response_

### Additional context

v1.22.6+k3s1 (k3d)

rohanKanojia:
## Description
Fix #3990

Access to `resources(...)` DSL method with generic HasMetadata interface
shouldn't be allowed.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 #### Bugs
 * Fix #3832 #1883: simplifying the isHttpsAvailable check
 * Fix #3745: the client will throw better exceptions when a namespace is not discernible for an operation
+* Fix #3990: Throw exception when `HasMetadata` is used in `resources(...)` API
 * Fix #4106: removed listing from projectrequests
 * Fix #4140: changed StatefulSet rolling pause / resume to unsupported.  Also relying on default rolling logic to Deployments and StatefulSets
 * Fix #4081: moving Versionable.withResourceVersion to a method on WatchAndWaitable and removing Waitable from the return type
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/BaseClient.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/BaseClient.java
@@ -229,6 +229,9 @@ public abstract class BaseClient implements Client {
     if (GenericKubernetesResource.class.equals(resourceType)) {
       throw new KubernetesClientException("resources cannot be called with a generic type");
     }
+    if (resourceType.isInterface()) {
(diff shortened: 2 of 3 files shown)


shawkins:
This was somewhat fixed by other upstream changes - you no longer get an npe, but an invalid url /api/v1/hasmetadatas

We could be doing a similar check for CustomResource or in general not allow any interface or abstract class.  Generic is different in that it's a concrete class.  It's hard in general for us to determine that a particular concrete class is invalid as a resource - the underlying logic is tolerant to both the group and version annotations being missing in which case the Config.getApiVersion is used.

rohanKanojia:
I had pointed this out in https://github.com/fabric8io/kubernetes-client/issues/3990#issuecomment-1122134337

Shall I add an additional check to throw IllegalArgumentException in case any interface is provided in `resources(..)` method?

shawkins:
> Shall I add an additional check to throw IllegalArgumentException in case any interface is provided in resources(..) method?

Yes a check for interfaces or abstract would cover the HasMetadata case as well.

manusa:
> This was somewhat fixed by other upstream changes - you no longer get an npe, but an invalid url /api/v1/hasmetadatas

The fix should address the comment https://github.com/fabric8io/kubernetes-client/issues/3990#issuecomment-1087552099

which should mostly deal with UX. Users who perform such a call should get an explicit message stating that the operation is forbidden, and the suggested fix/alternative operation

shawkins (approved):
LGTM

manusa (approved):

PR #4148, opened 2022-05-15
fix #3569: removing all api references to customResource

### LINKED ISSUE #3569: Parent issue for removal of deprecated items from fabric8 6
Just wanted to capture some proposals of what could be removed in 6.  Items will be spawned off as separate issues as needed.

- Any method dealing with OperationContext on SharedInformerFactory.  This will also for additional cleanup of the OperationContext, such as encapsulating the filter logic.
- At a lower priority any of the other deprecated on SharedInformerFactory
- RawCustomResourceOperationsImpl customResource(CustomResourceDefinitionContext customResourceDefinition)

shawkins:
## Description
Addresses #3569 by removing all api references to CustomResource

It also further minimizing SharedInformerFactoryImpl concerns and removes the use of the executor - before I add the migration guide for this, we should have consensus on the approach:

Options include:
1. Leave it up to the user to user to handle their own threading concerns wrt Watchers and ResourceEventHandlers - the main  pain point with this change is for users of the SharedInformerFactory who were relying on the async nature of the associated executor, but that is not consistent with the other apis (creating a watch or Informable informer).  There has also been at least one issue where it seems that there is confusion over the role of the factory executor.
2. Provide wrappers or convenience methods to create async versions of the Watchers and ResourceEventHandlers - e.g. Watcher.async(Watcher watcher, Executor executor).  This means we don't have to introduce additional dsl methods and it's very flexible as to what executor is used.  We do have the common executor via Utils.getCommonExecutor - but we really shouldn't be promoting static usage.  Another goal of the static refactoring to scope the common executor per client, if that change were worked, you could have have client methods available as well - client.configMaps().watch(client.async(watcher)).
3. Make it annotation driven - if the watcher or handler is annotated with Async (spring has a similar annotation) then use the common pool (which as mentioned above will eventually be configurable per client), otherwise call it directly.  This tries to encapsulate things more than 2.

Any other thoughts?

## Type of change

## Checklist


doc/FAQ.md
```diff
@@ -16,4 +16,19 @@ Further more you will also have choices in the HttpClient that is utilized.
...
+
+With the JDK http client it will only maintain a selector thread and a small worker pool which will be based upon your available processors per client.  It does not matter how many Informers or Watches you run, the same worker pool is shared.
+
+If it recommended with either client that your Watchers and ResourceEventHandlers do not execute long running tasks.  When using the OkHttp client this inhibits Watch event processing that can timeout the ping for that particular watch.  When using the JDK http client the long running task will inhibit the use of that thread for ALL http processing.  Note that calling other KubernetesClient operations, especially those with waits, can be long-running.  We are working towards providing non-blocking for may of these operations, but until that is available consider using a separate task queue for such work.
```
metacosm, line 29:
```suggestion
It is recommended with either client that your Watchers and ResourceEventHandlers do not execute long running tasks.  When using the OkHttp client this inhibits Watch event processing that can timeout the ping for that particular watch.  When using the JDK http client the long running task will inhibit the use of that thread for ALL http processing.  Note that calling other KubernetesClient operations, especially those with waits, can be long-running.  We are working towards providing non-blocking mode for many of these operations, but until that is available consider using a separate task queue for such work.
```

metacosm (approved):
LGTM apart from minor wording issues.

shawkins:
I apologize that there are really two things being merged together here - trying to clarify the Executor usage turned out to be more involved than I would have hoped.

@metacosm @manusa the next set of changes remove the usage of the static Utils.getCommonExecutor and allow for an executor or executorservice to be scoped to the client via the KubernetesClientBuilder.  This can be a different executor than the one used for IO operations because this is for potentially blocking calls (although the default logic for creating the IO thread pools for OkHttp or the JDK client default to a cached / unbounded thread pool, which could be used for both).  

This client scoped executor will be used rather than allowing for an ExecutorService per SharedInformerFactory - which we didn't document well nor provide any examples of.  It is also no longer needed to start the informers, since we have a non-blocking way of doing that - it's real purpose is to decouple the ResourceEventHandler method calls from the IO threads.  For consistency we'll use the same executor under informers created by the Informable dsl.  When using the OkHttp client other than a few corner cases it was ok to use the IO thread to process events (that's why the DSL was using Runnable::run) because the IO threads are managed per websocket.  However with the JDK client that can be a problem - there is just a small dedicated IO thread pool, which you can quickly exhaust and stall / deadlock if you block.  The informers used under the covers for Waitable or informOnCondition I left as just using the IO thread as we already had it documented and/or it's unlikely that the user would use blocking logic there - we can of course change that to use the task executor if needed.

In general I wanted to consolidate / hide / document the threading concerns as much as possible - anything that we do api-wise like Watcher.async(Watcher watcher, Executor executor) is just a workaround.  The real solution is for the call back methods to return a CompletionStage or CompletableFuture - CompletionStage<Void> eventReceived(Action action, T resource) - or pass in a parameter that allows for triggering when to receive more events.

I'll refine the above into the migration doc if this all seems reasonable.

manusa:
I think it might be better if we have a sync review on this one, especially to think about the executor removal options.

doc/FAQ.md
```diff
@@ -16,4 +16,20 @@ Further more you will also have choices in the HttpClient that is utilized.
...
+
+### What threading concerns are there?
+
+There has been a lot of changes under the covers with thread utilization in the fabric8 client over the 5.x and 6.x releases.  So the exact details of what threads are created / used where will depend on the particular release version.
+
+At the core the thread utilization will depend upon the http client implementation.  Per client OkHttp maintains a pool of threads for task execution.  It will dedicate 2 threads out of that pool per WebSocket connection.  If you have a lot of WebSocket usage (Informer or Watches) with OkHttp, you can expect to see a large number of threads in use - which can potentially exhaust the OkHttp defaults.
+
+With the JDK http client it will only maintain a selector thread and a small worker pool which will be based upon your available processors per client.  It does not matter how many Informers or Watches you run, the same worker pool is shared.
+
+It is recommended with either http client that logic you supply via Watchers, ExecListeners, ResourceEventHandlers, Predicates, etc. do not execute long running tasks.
```
manusa, line 29: I'd try to highlight this as a callout note box:

```suggestion
> **Note:** It is recommended with either HTTP client implementation that logic you supply via Watchers, ExecListeners, ResourceEventHandlers, Predicates, etc. does not execute long running tasks. 
```

Maybe also suggest the appropriate approach too ---> Schedule those long-running tasks in a separate executor

doc/FAQ.md
```diff
@@ -16,4 +16,20 @@ Further more you will also have choices in the HttpClient that is utilized.
...
+
+On top of the http client threads the fabric8 client maintains a task thread pool for scheduled tasks and for potentially long-running tasks that are called from WebSocket operations, such as handling input and output streams and ResourceEventHandler call backs.  This thread pool defaults to an unlimited number of cached threads, which will be shutdown when the client is closed - that is a sensible default with either http client as the amount of concurrently running async tasks will typically be low.  If you would rather take full control over the threading use KubernetesClientBuilder.withExecutor or KubernetesClientBuilder.withExecutorSupplier - however note that constraining this thread pool too much will result in a build up of event processing queues.
+
+Finally the fabric8 client will use 1 thread per PortForward and an additional thread per socket connection - this may be improved upon in the future.
```
manusa, line 35: 1 thread or 1 extra thread?

I understand that the websocket connection uses a thread, and the WebSocket.Listener implementation uses an extra-thread to process the received messages. right?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/BaseClient.java
```diff
@@ -39,10 +40,30 @@
...
 public abstract class BaseClient implements Client {
 
+  /**
+   * An {@link ExecutorSupplier} that provides an unlimited thread pool {@link Executor} per client.
+   */
+  public static final ExecutorSupplier DEFAULT_EXECUTOR_SUPPLIER = new ExecutorSupplier() {
+
+    @Override
+    public Executor get() {
+      return Executors.newCachedThreadPool(Utils.daemonThreadFactory(this));
```
manusa, line 57: I'm concerned about this one.
When is the DEFAULT_EXECUTOR_SUPPLIER.get() method going to be invoked during the lifecycle of the application?
Should we just keep a single instance? Is there a use-case where we want to maintain more than one Executor?

doc/FAQ.md
shawkins: It should say instead that "a LocalPortForward uses 1 thread, and every PortForward or connection to the LocalPortForward uses an additional thread."

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/BaseClient.java
shawkins: It is called once per client.  With the way things are currently every call to build() creates a fresh client.  If the builder is only allowed to create a single client, then it could just be a single instance.

doc/FAQ.md
shawkins: > Maybe also suggest the appropriate approach too ---> Schedule those long-running tasks in a separate executor

A separate task queue is mentioned below.  We can move or elaborate on that if you want.

manusa: No, it's fine, I saw the additional explanation later.
The callout is good for diagonal readers, which might stop and pay closer attention to the surrounding paragraphs.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/BaseClient.java
manusa: Shall we switch this to create lazily loaded singleton instance?

manusa (approved):

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/BaseClient.java
shawkins: KubernetesClientBuilder is new, so whatever we do there won't affect existing users.  However that is functionally different than many of the other Builders, including the HttpClient.Builder.  Because you can still call the with methods after build it would probably be clearest if subsequent calls to build through an illegalstateexception.

shawkins:
@manusa assuming #4159 is going in first, I'll wait to rebase this until after that.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/BaseClient.java
manusa: > it would probably be clearest if subsequent calls to build through an illegalstateexception.

Yup, I think that's even better.

manusa:
> @manusa assuming #4159 is going in first, I'll wait to rebase this until after that.

OK, I'll merge #4159 because it's ready to merge and CI takes forever (If that wasn't the case, this one should go in first)

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/BaseClient.java
shawkins: Let's take that up as a follow-on task.

manusa:
The ExecWebSocketListenerTest never ends in the Java8 tests. Please take a look.

shawkins:
> I think this pool might be backed by the number of available cores, and I guess that in a CI env this must be probably just one.

It will default to Max(1, cores - 1).

> We should probably replace this by a pool with a fixed number of threads.

There are several things.  The first is to figure out what task is hanging around - it's not expected that something should indefinitely block.

The next step is that we can:
- explicitly set java.util.concurrent.ForkJoinPool.common.parallelism (possibly to 1 in the maven properties) - I can reproduce the hang if I run locally with java.util.concurrent.ForkJoinPool.common.parallelism=1
- or stop using the ForkJoinPool common pool because it's not really intended for arbitrary tasks.  Fully managing a threadpool per test is possible and would help localize lingering tasks.

manusa:
OK, I get your point now.

My assumption is that the commonPool might be used elsewhere (OkHttp) and we might be running into some lock. ---> If we replace the ForkJoinPool with a singlethreaded pool, does it still hang?

shawkins:
> My assumption is that the commonPool might be used elsewhere (OkHttp) and we might be running into some lock.

Turns out this is only occasionally reproducible on java 8 when the parallelism is 1.  No task is hanging.  The common pool simply won't execute the task.

> If we replace the ForkJoinPool with a singlethreaded pool, does it still hang?

It appears fine.  I've added another commit that removes the usage of the fork join common pool.

PR #4159, opened 2022-05-18
refactor: extracted PortForwarderWebsocketListener and added tests

manusa:
## Description
refactor: extracted PortForwarderWebsocketListener and added tests

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PortForwarderWebsocket.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PortForwarderWebsocket.java
@@ -21,7 +21,6 @@ import io.fabric8.kubernetes.client.http.HttpClient;
 import io.fabric8.kubernetes.client.http.WebSocket;
 import io.fabric8.kubernetes.client.utils.URLUtils;
 import io.fabric8.kubernetes.client.utils.Utils;
-import io.fabric8.kubernetes.client.utils.internal.SerialExecutor;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


manusa:
> It might be simpler if PortForwarderWebsocketListener implemented PortForward - that tends to be the way it's done for the other listeners.

I considered that option, however, I decided to proceed this way to properly separate concerns and keep things simple.

It also simplifies specifying the behaviors to be tested.

> This won't induce the failure mentioned with [#4115 (comment)](https://github.com/fabric8io/kubernetes-client/pull/4115#issuecomment-1129563049) - but it seems like your goal was just to have it be more testable in general.

I didn't try altering the method calls, but part of the goal should be to cover that scenario and make sure the issue won't get reintrduced. I'll give it another spin in case that's not prevented.

> There will be some rebasing to do depending on what is committed first given the changes in #4148

I couldn't check 4148, so I'm not sure if you decided on one of the suggested approaches. I'll check tomorrow and ping you.

shawkins:
> I'll give it another spin in case that's not prevented.

Causing it is based upon a timing issue with the WebSocket implementation.  See the breakup of partial writes occurring in PodTest - emitting "Hell" and "o world" separately.  From there if remote websocket close / done is processed in a racey way with the final write you can get truncated output.

manusa:
So I've checked moving the `webSocket.request();` around.

The current tests do verify that more data is only requested once we're ready to process it. So moving the `webSocket.request();` request invocation causes a few test failures.

There is no specific test for the race condition, but I'm not even sure that this is a valid behavior. Clients should be responsible of closing the websocket when they've received the requested data or when they meet any other condition. IMO the `PodTest#testPortForward` is not OK, since it assumes that the transmission is complete when the SocketChannel is empty, which might or might not be true. The sync nature of the onMessage tests do verify that the websocket is processed in an orderly fashion, which is also important.

The current onMessage implementation, where the data is processed asynchronously in the SerialExecutor would also suffer from moving the `webSocket.request();` out, since the "backpressure" is lost and we might get flooded with messages from the server (even if the message processing order is still guaranteed by the SerialExecutor). This is also verified.

In any case, I'll try to add a few more test to verify additional scenarios.

shawkins:
> So moving the webSocket.request(); request invocation causes a few test failures.

My initial mistake was this behavior is dependent on the websocket implementation.  Based upon the JDK logic my initial understanding is that it didn't matter where you called websocket.request in the listener - the events were being serially delivered because the current thread when it finished the listener method would trigger the processing.  With OkHttp it appears could deliver close via the other websocket thread - the one processing timed events and ping/pong - so that as soon as you call request a pending close could get delivered.

PR #4166, opened 2022-05-24
Avoid setting subresources to null

iocanel:
The existing code can can result in setting .spec.subresources to null if both versions have no subresources.
The issue is currently by the fact that withSubresources(null) is ignored.

### DIFF
--- a/crd-generator/api/src/main/java/io/fabric8/crd/generator/v1beta1/decorator/PromoteSingleVersionAttributesDecorator.java
+++ b/crd-generator/api/src/main/java/io/fabric8/crd/generator/v1beta1/decorator/PromoteSingleVersionAttributesDecorator.java
@@ -53,7 +53,7 @@ public class PromoteSingleVersionAttributesDecorator extends CustomResourceDefin
       spec.removeAllFromVersions(versions);
       spec.withVersions(newVersion);
     } else {
-      Set<CustomResourceSubresources> subresources = versions.stream().map(CustomResourceDefinitionVersion::getSubresources).collect(Collectors.toSet());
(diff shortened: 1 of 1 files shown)


crd-generator/api/src/main/java/io/fabric8/crd/generator/v1beta1/decorator/PromoteSingleVersionAttributesDecorator.java
```diff
@@ -53,17 +53,21 @@ public void andThenVisit(CustomResourceDefinitionSpecFluent<?> spec, ObjectMeta
...
       boolean hasIdenticalAdditionalPrinterColumns = additionalPrinterColumns.size() == 1;
       boolean hasIdenticalSchemas = schemas.size() == 1;
       
+      if (hasIdenticalSubresources) {
+        spec.withSubresources(subresources.iterator().next());
+
+      }
+
       spec
         .withValidation(hasIdenticalSchemas ? schemas.iterator().next() : null)
```
metacosm, line 70: Wouldn't the same or similar issue happen with this call as well? Or the one for printer columns?

Ananya2001-an:
Why can't we just include the definition of subresources in 1 of the versions or even both so that they are never disabled for custom resources? 
I would like to continue with this issue. I will implement all the requested changes.
Kindly assign this to me.

rohanKanojia:
There hasn't been much activity on this PR. Do we still need this fix or was this issue addressed in some other PR?

iocanel:
FWIW, the issue should have been addressed on the sundrio side of things so, I assume this is no longer an issue.

PR #4163, opened 2022-05-20
fix #3300: address the watchTest failures

shawkins:
## Description
There's a race condition with the watch reconnect - it's possible that the next reconnect call is happening before the current scheduled method flips the reconnect flag.  Since all of the code is now async I've switched to using just synchronization and changed the name of the methods from run to start.

## Type of change

 - [x] Bug fix (non-breaking change which fixes an issue)
 - [ ] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [x] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [x] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [ ] I Added [CHANGELOG](https://github.com/fabric8io/kubernetes-client/blob/master/CHANGELOG.md) entry regarding this change
 - [x] I have implemented unit tests to cover my changes
 - [x] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/master/doc/CHEATSHEET.md) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [ ] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

PR #4173, opened 2022-05-26
Making the calling of WebSocket.request consistent

shawkins:
## Description
I should have noticed this with #4163, but for consistency request should be called after the message is processed. Also adding a test to make sure that an exception won't cause message processing to stop - it won't because it's being caught and the call to request is in a finally block just in case.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatcherWebSocketListener.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatcherWebSocketListener.java
@@ -48,11 +48,14 @@ class WatcherWebSocketListener<T extends HasMetadata> implements WebSocket.Liste
 
   @Override
   public void onMessage(WebSocket webSocket, String text) {
-    webSocket.request();
     // onMesssage and onClose are serialized, but it's not specified if onError
     // may occur simultaneous with onMessage.  So we prevent concurrent processing
-    synchronized (this) {
-      manager.onMessage(text);
+    try {
+      synchronized (this) {
+        manager.onMessage(text);
+      }
+    } finally {
+      webSocket.request();
     }
   }
 
--- a/kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/WatchTest.java
+++ b/kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/WatchTest.java
@@ -59,8 +59,12 @@ class WatchTest {
 
   @BeforeEach
   void setUp() {
-    pod1 = new PodBuilder().withNewMetadata().withNamespace("test").withName("pod1")
-      .withResourceVersion("1").endMetadata().build();
+    pod1 = new PodBuilder().withNewMetadata()
+        .withNamespace("test")
+        .withName("pod1")
+        .withResourceVersion("1")
+        .endMetadata()
+        .build();
   }
 
   @Test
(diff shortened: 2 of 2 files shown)


shawkins:
Another follow up that could be part of this is to remove the test exclusion for watchtest from kubernetes-tests.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatcherWebSocketListener.java
```diff
@@ -48,11 +48,14 @@ public void onError(WebSocket webSocket, Throwable t) {
...
     // onMesssage and onClose are serialized, but it's not specified if onError
     // may occur simultaneous with onMessage.  So we prevent concurrent processing
-    synchronized (this) {
-      manager.onMessage(text);
+    try {
+      synchronized (this) {
+        manager.onMessage(text);
+      }
+    } finally {
+      webSocket.request();
```
manusa, line 58: Why the try-finally? Won't this force a new request even if the user interrupts the execution, or there's any other sort of runtime exception?

manusa: I see the new test verifies this, but I'm not really sure why we wouldn't want the message processing to stop in **any** event

shawkins: > Why the try-finally? Won't this force a new request even if the user interrupts the execution, or there's any other sort of runtime exception?
> but I'm not really sure why we wouldn't want the message processing to stop in any event

That's never been part of the contract.  Any Exception (including runtime) thrown by the Watcher is caught and the Watch continues.  You could make the case for an unchecked exception causing Watch termination, but that will be a minor breaking change.

This test is just ensuring that if that catch is inadvertently removed, we'll still continue with message processing.

manusa: For me the test is fine(, and also the behavior). I was just pointing this out because I'm unaware of the implementation parts of webSocket.request, and how might that be affected.

We should also provide (if not there), similar tests to those we implemented for the Portforward to verify that the message processing is performed in order + that a new requests are sent only when the previous message has been consumed.

shawkins: > I'm unaware of the implementation parts of webSocket.request, and how might that be affected.

A call to request is required to receive the next message or onClose (this matches the documented behavior of the jdk client).  If you call it after onClose or onError - it will be a no-op as those will be terminal events.

manusa: > If you call it after onClose or onError - it will be a no-op as those will be terminal events.

Should we add this (or a similar description) as part of the interface's Javadoc?

manusa (approved):

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatcherWebSocketListener.java
shawkins: > Should we add this (or a similar description) as part of the interface's Javadoc?

Sure I'll add something.

PR #4218, opened 2022-06-17
fix: patchStatus doesn't increment metadata.generation field

manusa:
## Description

fix #4217

patchStatus doesn't increment metadata.generation field in Kubernetes Mock Server (CRUD)

## Type of change

 - [x] Bug fix (non-breaking change which fixes an issue)
 - [ ] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [ ] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [x] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [x] I Added [CHANGELOG](https://github.com/fabric8io/kubernetes-client/blob/master/CHANGELOG.md) entry regarding this change
 - [x] I have implemented unit tests to cover my changes
 - [ ] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/master/doc/CHEATSHEET.md) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [ ] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

PR #4220, opened 2022-06-18
Update resource version when `replaceStatus()` is used with Mock Kubernetes Server

### LINKED ISSUE #4216: Predicate not called after resource status is updated with `replaceStatus()` on KubernetesMock in CRUD mode
### Describe the bug

In Strimzi we are using the Kubernetes Mock server in the CRUD mode to run mock tests. In some of the tests, we create a resource and wait for the operator to update it with certain status fields. We use for it the `waitUntilCondition(...)`, where we pass the condition predicate and timeout and wait for it to complete. The expectation is that the predicate will be called everytime the resource it waits for is updated. However, that does not seem to be the case in all situations. When the resource status is updated using the `replaceStatus()` method, the predicate is never triggered and as a result, the `waitUntilCondition` never completes.

Interesting enough, when using the `patchStatus()` method, it seems to work fine and the predicate is tested and if it is fulfilled, the `waitUntilCondition` call succeeds. However, the `patchStatus()` method incorrectly increments the resource generation (see [this comment](https://github.com/fabric8io/kubernetes-client/pull/4199#issuecomment-1158143187)), so it is causing other issues in our tests and we cannot really use it instead of `replaceStatus()` easily. I'm not sure whether incrementing or not incrementing the generation is in any way related to triggering or not triggering the predicate in `waitUntilCondition`.

Since `waitUntilCondition` is also used when the `scale(...)` method is called for example for a StatefulSet with the `wait` flag set to `true`, it causes the same problem here as well.

6.0.0-RC1

### Steps to reproduce

The following test case can be used to reproduce this issue. When you run it with `patchStatus()`, you can see that it will trigger the predicate for every change and complete successfully. When you run it with `replaceStatus()`, the predicate will be triggered only once and the subsequent status updates will never trigger it anymore. And the test will wait until it times out.

```java
package io.strimzi.test.mockkube2;

import io.fabric8.kubernetes.api.model.ContainerBuilder;
import io.fabric8.kubernetes.api.model.LabelSelector;
import io.fabric8.kubernetes.api.model.apps.StatefulSet;
import io.fabric8.kubernetes.api.model.apps.StatefulSetBuilder;
import io.fabric8.kubernetes.api.model.apps.StatefulSetStatus;
import io.fabric8.kubernetes.api.model.apps.StatefulSetStatusBuilder;
import io.fabric8.kubernetes.client.KubernetesClientBuilder;
import io.fabric8.kubernetes.client.Watch;
import io.fabric8.kubernetes.client.Watcher;
import io.fabric8.kubernetes.client.WatcherException;
import io.fabric8.kubernetes.client.server.mock.EnableKubernetesMockClient;

import java.util.Map;

@EnableKubernetesMockClient(crud = true)
public class Fabric8Testing {
    private final static String NAMESPACE = "my-namespace";

    // Injected by Fabric8 Mock Kubernetes Server
    private KubernetesClient client;

    @Test
    public void testWaitUntil() throws InterruptedException {
        final String statefulSetName = "my-sts";

        StatefulSet newSts = new StatefulSetBuilder()
                .withNewMetadata()
                    .withNamespace(NAMESPACE)
                    .withName(statefulSetName)
                .endMetadata()
                .withNewSpec()
                    .withReplicas(3)
                    .withSelector(new LabelSelector(null, Map.of("app", "my-sts")))
                    .withNewTemplate()
                        .withNewMetadata()
                            .withLabels(Map.of("app", "my-sts"))
                        .endMetadata()
                        .withNewSpec()
                            .withContainers(new ContainerBuilder().withName("nginx").withImage("nginx:1.14.2").build())
                        .endSpec()
                    .endTemplate()
                .endSpec()
                .build();

        client.apps().statefulSets().inNamespace(NAMESPACE).resource(newSts).create();

        Thread t1 = new Thread(new Runnable() {
            @Override
            public void run() {
                client.apps().statefulSets().inNamespace(NAMESPACE).withName(statefulSetName).waitUntilCondition(s -> {
                    System.out.println("Predicate called: " + s.getStatus());
                    return s.getStatus() != null && s.getStatus().getAvailableReplicas() == 5;
                }, 5, TimeUnit.MINUTES);

                return;
            }
        });
        t1.start();

        for (int i = 0; i < 6; i++) {
            int finalI = i;

            client.apps().statefulSets().inNamespace(NAMESPACE)
                    .resource(new StatefulSetBuilder(newSts)
                            .withNewStatus()
                                .withAvailableReplicas(finalI)
                            .endStatus()
                            .build())
                    //.patchStatus(); // <= This will make the test pass
                    .replaceStatus(); // <= This will make the test timeout
        }

        t1.join();
        System.out.println("DONE");
    }
}
```

### Expected behavior

The expectation is that both method provide the same result in this test.

### Runtime

### Environment

macOS

_No response_

### Additional context

This seems to happen only with the Kubernetes Mock Server. It seems to work fine with actual Kubernetes server.

scholzj:
## Description

Fix #4216 

When the Kubernetes Mock Server is used in CRUD mode and a resource status is updated using the `replaceStatus()` method, it does not seem to dispatch the update to clients and thus it does not trigger the `waitUntilCondition(...)` or `scale(N, true)` methods waiting for them.

This seems to be caused by the resource version field in metadata not being updated. This PR uses the `deepCopy()` of the existing object (in order to be able to diff it against the updated one) and calls the `setDefaultMetadata(...)` method on it. This now causes the resource version to be updated and that triggers the dispatching of the events tot he clients even when status is updated with `replaceStatus()`

This should close close #4216.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### 6.0-SNAPSHOT
 
 #### Bugs
+* Fix #4216: Update metadata when `replaceStatus()` is called
 
 #### Improvements
 * Fix #4006: Remove outdated shared test classes in `kubernetes-client/` and `openshift-client/` modules
--- a/junit/kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/KubernetesCrudDispatcher.java
+++ b/junit/kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/KubernetesCrudDispatcher.java
@@ -250,7 +250,6 @@ public class KubernetesCrudDispatcher extends CrudDispatcher implements Resetabl
         }
 
         setDefaultMetadata(updated, pathValues, source);
-
         String updatedAsString = Serialization.asJson(updated);
 
         return validateRequestBodyAndHandleRequest(updatedAsString, h -> {
@@ -414,8 +413,9 @@ public class KubernetesCrudDispatcher extends CrudDispatcher implements Resetabl
           if (isStatusPath(path)) {
             JsonNode status = removeStatus(updated);
             // set the status on the existing node
-            updated = existingNode;
+            updated = existingNode.deepCopy();
             setStatus(updated, status);
+            setDefaultMetadata(updated, pathValues, existingNode);
           } else {
             // preserve status and generated fields
             if (statusSubresource) {


manusa:
Hi @scholzj 

I was actually fixing this one too on #4218, apparently we came to the same conclusion :) see my changes in KubernetesCrudDispatcher.java#L437.

Anyway, I'll merge your changes and then continue working on my fixes after a rebase. I'm using the time to fix a few additional issues and add tests. I'm mentioning this in case you were spending time on the issue too.

manusa (approved):

scholzj:
> I was actually fixing this one too on https://github.com/fabric8io/kubernetes-client/pull/4218, apparently we came to the same conclusion :) see my changes in KubernetesCrudDispatcher.java#L437.
>
> Anyway, I'll merge your changes and then continue working on my fixes after a rebase. I'm using the time to fix a few additional issues and add tests. I'm mentioning this in case you were spending time on the issue too.

Thanks for looking into it. I had a quick fix for the other issue as well, but saw you were doing some bigger refactoring there, so I let it be.

PR #4230, opened 2022-06-24
fix: crd-gen includes only enum values and discard enum fields

### LINKED ISSUE #4225: [crd-gen] Enum fields written in generated crd yaml
### Describe the bug

When the crd-gen tool tries to generate schema properties from an enum, both values and fields are written in the generated crd yaml.

5.12.2

### Steps to reproduce

Create an enum with values and at least a field and put it into Spec. 
Example:

```
@Group("com.example")
@Version("mycrd")
@Plural("mycrds")
public class MyCRD extends CustomResource<MyCRDSpec, Void> implements Namespaced {

  public class MyCRDSpec {

    private Test test;

    public enum Test {
      A("test-a"),
      B("test-b");

      private final String field1;

      Test(String field1) {
        this.field1 = field1;
      }

      public String getField1() {
        return field1;
      }
    }
  }
}


```

Exec `mvn clean compile`

the generated file target/classes/META-INF/fabric8/mycrds.com.example-v1.yml will contain:
```
# Generated by Fabric8 CRDGenerator, manual edits might get overwritten!
apiVersion: apiextensions.k8s.io/v1
metadata:
  name: mycrds.com.example
spec:
  group: com.example
  names:
    kind: MyCRD
    plural: mycrds
    singular: mycrd
  scope: Namespaced
  versions:
  - name: mycrd
    schema:
      openAPIV3Schema:
        properties:
          spec:
            properties:
              test:
                enum:
                - A
                - B
                - field1
                type: string
            type: object
          status:
            type: object
        type: object
    served: true
    storage: true

```

### Expected behavior

A crd with just enum values and without enum fields:
```

# Generated by Fabric8 CRDGenerator, manual edits might get overwritten!
apiVersion: apiextensions.k8s.io/v1
metadata:
  name: mycrds.com.example
spec:
  group: com.example
  names:
    kind: MyCRD
    plural: mycrds
    singular: mycrd
  scope: Namespaced
  versions:
  - name: mycrd
    schema:
      openAPIV3Schema:
        properties:
          spec:
            properties:
              test:
                enum:
                - A
                - B
                type: string
            type: object
          status:
            type: object
        type: object
    served: true
    storage: true

```

### Runtime

Kubernetes (vanilla)

1.22.3@latest

### Environment

Windows

_No response_

### Additional context

_No response_

leoandrea7:
## Description

Fix #4225 
## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Bugs
 * Fix #5218: No export for `io.fabric8.tekton.triggers.internal.knative.pkg.apis.duck.v1beta1` in tekton v1beta1 triggers model
 * Fix #5224: Ensuring jetty sets the User-Agent header
+* Fix #4225: Enum fields written in generated crd yaml
 
 #### Improvements
 
--- a/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
+++ b/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
@@ -23,12 +23,29 @@ import io.fabric8.kubernetes.api.model.Duration;
 import io.fabric8.kubernetes.api.model.IntOrString;
 import io.fabric8.kubernetes.api.model.Quantity;
 import io.sundr.builder.internal.functions.TypeAs;
-import io.sundr.model.*;
+import io.sundr.model.AnnotationRef;
+import io.sundr.model.ClassRef;
+import io.sundr.model.Method;
+import io.sundr.model.PrimitiveRefBuilder;
+import io.sundr.model.Property;
+import io.sundr.model.TypeDef;
+import io.sundr.model.TypeRef;
 import io.sundr.utils.Strings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
 
 import static io.sundr.model.utils.Types.BOOLEAN_REF;
 import static io.sundr.model.utils.Types.DOUBLE_REF;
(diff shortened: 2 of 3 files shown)


andreaTP:
I took a look at this issue and cannot find a proper check to correctly include/exclude enum's instances.

Specifically given the test here:
https://github.com/sundrio/sundrio/blob/b7091f8fbe792596258d6bf3a10c05bc3a415bd4/adapters/api/src/test/java/io/sundr/adapter/testing/AbstractAdapterTest.java#L206

We include all the needed properties, but we don't know how to discriminate actual values of the Enum from other methods/accessos/constructors.

Maybe @metacosm or @iocanel can help.

Alternatively, it should be enough to "workaround" this issue by considering the `JsonIgnore` annotation on enum properties (I can open a separate issue for it if needed).

leoandrea7:
@andreaTP maybe you should add a filter [here ](https://github.com/sundrio/sundrio/blob/b7091f8fbe792596258d6bf3a10c05bc3a415bd4/adapters/api/src/test/java/io/sundr/adapter/testing/AbstractAdapterTest.java#L204)like the filter i added with this PR

andreaTP:
@leoandrea7 couldn't find any way to discriminate the actual enum values from other properties :-(

leoandrea7:
@andreaTP you are absolutely right...The problem is that TypeDef class does not have a method that return just enum values..I tried to fetch enum with `Class.forName(def.getFullyQualifiedName())` in order to use .values() directly but there is classloader problems for test..

leoandrea7:
@andreaTP After some thought, I think your test case with an enum containing a field of the same type as the enum itself is a very infrequent case. For now it would be better to merge this solution which solves most of the use cases and I could add in the docs of the crd-gen a new section (titled limitation or known problems) in which I specify that this particular case is not yet supported. Then if in the future the sundrio library will introduce a specific method to recover only the values of an enum through the TypeDef class then we will also solve the use case you have proposed. This would solve most of the use cases for enums containing fields. What do you think?

andreaTP:
I don't think this case is infrequent, and anyhow, unfortunately, it's semantically incorrect.
I would suggest that we move on fixing JsonIgnore behavior in enums so that there will be a viable workaround for your use case(possibly re-using the more generic Properties parser logic).

Meanwhile, we can add the necessary information in Sundrio cc. @iocanel and move on from there.

andreaTP:
Submitted #4294 should provide a workaround also for this issue.

manusa:
Hi #4294 has been merged. Is this PR providing something else?

andreaTP:
@manusa , yes, this PR is providing something different, but, to have a robust implementation we need to bubble up information changing `sundrio` first.

andreaTP:
@shawkins do you mind fixing the conflicts?
As discussed in the other issue, I have no objection to moving this forward at this point.

manusa (approved):

PR #4263, opened 2022-07-07
fix #4206: ensuring null won't be returned from kubernetesdeserializer

### LINKED ISSUE #4206: Throw an exception when a resource cannot be deserialised instead of passing `null` downstream

It sometimes happen that a resource cannot be deserialised but when that happens the downstream code gets a `null` object thus resulting in an NPE further downstream where the cause might not be obvious. Not actually sure if this behavior is present in JVM mode but this is definitely an issue when running the client in a native binary.

An exception should be thrown allowing to identify which resource couldn't be deserialised (and any exception that might help explain why along the way).

_No response_

### Additional context

_No response_

shawkins:
## Description
This is to address #4206.  We cannot throw an exception as it appears that the Config parsing has embedded HasMetadata (extensions) references, which lack type information.  So instead we'll return a generic resource - that should ensure round-trip parse/serialization is also correct.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/internal/KubernetesDeserializer.java
+++ b/kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/internal/KubernetesDeserializer.java
@@ -18,6 +18,7 @@ package io.fabric8.kubernetes.internal;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.databind.DeserializationContext;
 import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import io.fabric8.kubernetes.api.KubernetesResourceMappingProvider;
 import io.fabric8.kubernetes.api.model.GenericKubernetesResource;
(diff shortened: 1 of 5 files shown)


shawkins:
@manusa @rohanKanojia base upon our meeting we agree that mapping raw to HasMetadata is not good.  It also creates a potentially odd set of builder methods as sundrio will assume the only valid types are the concrete types in the module and in the BuildableReferences.  Raw should instead be any valid structured content.  

My motivation to use the GenericKubernetesResource however was that it was the easiest way to preserve this content.  if you run something like:

```
    IngressControllerSpec spec = new IngressControllerSpec();
    GenericKubernetesResource resource = new GenericKubernetesResource();
    resource.setAdditionalProperty("foo", "bar");
    spec.setUnsupportedConfigOverrides(resource);
    System.out.println(Serialization.asYaml(spec)); // contains unsupported
    IngressControllerSpec parsed = Serialization.unmarshal(Serialization.asYaml(spec), IngressControllerSpec.class);
    System.out.println(Serialization.asYaml(parsed)); // does not contain unsupported
```

You'll see that the unsupportedConfigOverrides are lost when they lack a kind/apiVersion.  

That should be occurring in all such places like NamedExtension.extension, which seems to already exist in the Configs used in our test cases.

So if we don't want to use GenericKubernetesResource to conform to the existing mapping, then we'll have to change the generator / mapping logic.

As for the original issue #4206 - that could be handled separately for now just by considering the top-level object only, but it will of course get fixed if we generally handle the raw case - the Serialization.unmarshall would return a map of maps or whatever we think is appropriate for representing raw content when the root object lacks kind/apiVersion information.

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/internal/KubernetesDeserializer.java
```diff
@@ -116,26 +117,26 @@ private KubernetesResource fromArrayNode(JsonParser jp, JsonNode node) throws IO
...
+        return jp.getCodec().treeToValue(node, resourceType);
+      } finally {
+        if (inTemplate) {
+          IN_TEMPLATE.remove();
         }
       }
     }
-    return null;
+    throw new JsonMappingException(jp, String.format(
+        "There's a class loading issue, %s is registered as a KuberentesResource, but is not an instance of KuberentesResource",
```
metacosm, line 138:
```suggestion
        "There's a class loading issue, %s is registered as a KubernetesResource, but is not an instance of KubernetesResource",
```

metacosm (approved):
LGTM apart from typo (see comment).

shawkins:
@manusa @metacosm here's an update to a hybrid / tactical solution.  The deserializer will always return a generic, but we'll make a top level check to ensure that only valid generics are returned - which will satisfy #4206.  This will allow for the HasMetadata mapping of raw to continue to work - then for a major release, or whenever we're comfortable with the breaking change, the mappings can be fixed.

manusa (approved):

rohanKanojia (approved):

PR #3105, opened 2021-05-12
Fixes #3097 : refresh token with autoconfigure even if authprovider is null

hypnoce:
## Description
Token refreshes happen after a 401 and retries to call to the k8s API. On a local machine using AWS EKS, token generation is done via an Exec task and not an authprovider. The token refresher was only refreshing in case authprovider was not null thus was throwing a kubernetesclientexception after a small timeout (10-15 min to my memory) imposed by AWS.
This PR makes sure to refresh token via autoconfigure in any case..

Fixes #3097 

## Type of change
 - [X] Bug fix (non-breaking change which fixes an issue)
 - [ ] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [ ] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [X] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [X] I Added [CHANGELOG](../CHANGELOG.md) entry regarding this change
 - [X] I have implemented unit tests to cover my changes
 - [ ] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/master/doc/CHEATSHEET.md) accordingly
 - [X] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [X] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

PR #4264, opened 2022-07-08
fix #2271: Refresh token every minute

### LINKED ISSUE #2271: Feature Request: Support projected bound service account tokens when run in-cluster
Kubernetes has a beta feature to increase service account token security that is planned to go to General Availability and will become the default at some point.

The general idea is that rather than using the automounted service account token at /var/run/secrets/kubernetes.io/serviceaccount/token, you can project a volume that contains a pod scoped token that is valid for the lifetime of the pod using it. The token can also be auto-rotated based on some TTL. This means accidental token leakage doesn't require a manual rotation of keys etc. 

See https://github.com/kubernetes/community/blob/master/contributors/design-proposals/auth/bound-service-account-tokens.md and https://github.com/mikedanese/community/blob/2bf41bd80a9a50b544731c74c7d956c041ec71eb/contributors/design-proposals/storage/svcacct-token-volume-source.md

It would be great if the kubernetes-client could handle the SA token being updated in the filesystem. Currently it stores the value of the token in the io.fabric8.kubernetes.client.Config object's oauthToken property. AFAIK this never gets reloaded.
Currently the downstream consumers of the client will need to reload their client objects to refresh the config when a token rotates. Whilst this is doable, it feels like it should be taken care of by the kubernetes-client.

It's also not possible (that I could see) to configure the SA token path, which could be projected to somewhere other than the default location of /var/run/secrets/kubernetes.io/serviceaccount/token.

Vlatombe:
## Description

Should address #2271.
Platforms such as Amazon EKS since 1.21 rely on `BoundServiceAccountTokenVolume` to have time-limited access tokens. It is expected from kubernetes client implementations to refresh the configuration periodically to get a new token.

From [kubernetes 1.21 release notes](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.21.md?rgh-link-date=2021-05-18T09%3A11%3A56Z#changelog-since-v1200)
> Clients should reload the token from disk periodically (once per minute is recommended) to ensure they continue to use a valid token. k8s.io/client-go version v11.0.0+ and v0.15.0+ reload tokens automatically.

Prior work from #3105 addressed it once the token expired, however it doesn't address the current situation where the token is actually valid for 90 days, but expected to be refreshed at least every hour. This triggers audit logs events in AWS which can lead users to think their client will break once the token really expires.

This introduces a proactive aproach of refreshing the token from disk if the current token has been retrieved more than one minute ago.

For oidc implementation, I'm not knowledgeable enough to determine whether a more optimal solution could be implemented using some alternate validity timestamp.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 #### Dependency Upgrade
 
 #### New Features
+* Fix #2271: Support periodic refresh of access tokens before they expire
 
 #### _**Note**_: Breaking changes in the API
 
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/TokenRefreshInterceptor.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/TokenRefreshInterceptor.java
@@ -52,7 +52,6 @@ public class TokenRefreshInterceptor implements Interceptor {
     }
   }
 
-
   @Override
   public CompletableFuture<Boolean> afterFailure(BasicBuilder headerBuilder, HttpResponse<?> response) {
     if (response.code() == HttpURLConnection.HTTP_UNAUTHORIZED) {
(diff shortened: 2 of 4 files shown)


Vlatombe:
@manusa Would it be possible to get a review on this work?

manusa (approved):

rohanKanojia (approved):

apiwoni:
I think this is more of a brute force approach to refresh service account token. Could we attempt to test for recommended `expires_in` before preemptively refreshing token or just let it fail and then refresh access token and retry?

manusa:
thread continues here: https://github.com/fabric8io/kubernetes-client/issues/2271#issuecomment-1255498383

PR #4284, opened 2022-07-19
fix #887: adding visiting logic for walking api resources

### LINKED ISSUE #887: How to search for resources
Is there a way to search for resources by label or annotation? For example, to search for all resources that include the label app=my-app. This would return Pods, Deployments, Services, StatefulSets, etc.

Thanks!

shawkins:
## Description
Fix #887

Added a visitor terminating visitor to walk api resources.

## Type of change

## Checklist


kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/ApiVisitor.java
```diff
@@ -0,0 +1,79 @@
...
+    SKIP
+  }
+
+  /**
+   * Visit the api. The core/legacy group will be an empty string.
+   * 
+   * @param api the group name
+   * @return the result to control future actions. {@link ApiVisitResult#SKIP} will skip visiting all versions of this api.
+   */
+  default ApiVisitResult visitApi(String api) {
```
manusa, line 52: Why was the word Api chosen?

Won't this be more confusing than calling it visitGroup?

Kubernetes coordinates:
- Group: API group the resource belongs to, or empty for the core resources
- Version
- Kind
- `apiVersion`: `$group/$version`
- Plural: HTTP resource name where GVK resources are exposed

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/ApiVisitor.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  /**
+   * Visit the apiVersion, which will be in the form of api/version or just v1 for the core.
+   * Use {@link ApiVersionUtil} to separate components if needed.
+   * 
+   * @return the result to control future actions. {@link ApiVisitResult#SKIP} will skip visiting all resources under this
+   *         apiVersion.
+   */
+  default ApiVisitResult visitApiVersion(String apiVersion) {
```
manusa, line 63: Again, just as a suggestion.
Would it be more user-friendly to change the signature to something like
```
default ApiVisitResult visitGroupVersion(String group, String version) {
```

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/ApiVisitor.java
```diff
@@ -0,0 +1,79 @@
...
+  /**
+   * Visit the resource.
+   * 
+   * @param apiVersion in the form of api/version or just v1 for the core
+   * @param apiResource can be used to look at applicable verbs and other resource information
+   * @param operation the {@link GenericKubernetesResource} operation for the current resource. Will be in the namespace of the
+   *        client by default.
+   * @return {@link ApiVisitResult#TERMINATE} to terminate. SKIP and CONTINUE will both continue visiting.
+   */
+  ApiVisitResult visitResources(String apiVersion, APIResource apiResource,
```
manusa, line 76: Maybe the `apiVersion` parameter is redundant, since all of its components should be available from the `apiResource`

manusa:
I think this is a wonderful solution to the problem, and a great way to deal with the lack of reactive options to provide a cancel/stop mechanism.

The only thing I'm not convinced about is the naming. I'm not sure if the concept of `api` as an alias for the group is something extended or not.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/ApiVisitor.java
shawkins: > Would it be more user-friendly to change the signature to something like

I'm fine either way - I'm sure some would want it concatenated, while others will want the components.  So I just left it with what we have already available in the logic.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/ApiVisitor.java
shawkins: > Why was the word Api chosen?

All of the other client entry points related to "apis" are api methods - getApiGroups, getApiGroup, getApiResources, hasApiGroup, so it seems like we should keep consistency with the api term.  If you prefer to use group in this method name that's fine - visitApiGroup or just visitGroup.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/ApiVisitor.java
shawkins: > since all of its components should be available from the apiResource

Unfortunately they are not.  The group and version are typically not populated by the api server - https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#apiresource-v1-meta - an alternative would be to set them ourselves when null.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/ApiVisitor.java
manusa: I understand that `apiGroup` -> `visitApiGroup` is what might both clearer and consistent, right?

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/ApiVisitor.java
manusa: :sob: That makes the problem even worse.

I can already see people complaining by mismatches about the `apiVersion` argument and the `apiResource` content.

Would it make sense to have:
- `default ApiVisitResult visitApiGroup(String group) {`
- `default ApiVisitResult visitApiGroupVersion(String group, String version) {`
- One of:
  - `ApiVisitResult visitApiResources(String group, String version, String kind,
      MixedOperation<GenericKubernetesResource, GenericKubernetesResourceList, Resource<GenericKubernetesResource>> operation);`
  - `ApiVisitResult visitApiResources(APIResource apiResource,
      MixedOperation<GenericKubernetesResource, GenericKubernetesResourceList, Resource<GenericKubernetesResource>> operation);` (where apiResource has been completed if necessary)

shawkins:
@manusa should be good now

manusa (approved):

Issue #4201, opened 2022-06-10
[refactor] Change JettyHttpClient#sendAsync method to not read full response in memory


manusa:
### Is your enhancement related to a problem? Please describe

Currently, the [`JettyHttpClient#sendAsync`](https://github.com/fabric8io/kubernetes-client/blob/27d1eafaa3f64088eb9d317aad3a5567f8586ab2/httpclient-jetty/src/main/java/io/fabric8/kubernetes/client/jetty/JettyHttpClient.java#L78-L96) method reads the response completely in-memory before completing the future, even for streaming return types (`Readable` or `InputStream`).

### Describe the solution you'd like

Provide alternative response listener implementation for InputStream and Readable so that the future is complete as soon as the response is received and we pipe the packets as an InputStream.

shawkins:
ExecWatchInputStream can be adapted for use with http logic as well.

Issue #4157, opened 2022-05-18
Why does it take thread sleep to get logs?


Code-WangShuai:
I want to get every new log through watchlog and compile it into an object to facilitate my subsequent operations. Can I only write it locally through outpustream and then read it?

rohanKanojia:
Could you please elaborate your question? I'm slightly confused by title of your issue and description

PR #4288, opened 2022-07-21
fix #4254: adding debug logging for exec stream messages

### LINKED ISSUE #4254: Log or expose standard error from internal exec

If you use pod file operations, such as copy, the stdErr is not requested / captured. 

We may not simply be able to use terminateOnError because a command like tar will report things on stdErr such as "stripping leading /" - although we may be able to workaround that in other ways like using -C.

At the very least it should be logged.

If a monitor like facility is added via #4236, the stdErr could be exposed there as well.

_No response_

### Additional context

_No response_

shawkins:
## Description

Fix #4254

Add debug logging for streams that don't have explicit handling.  This is mostly just an internal nicety, but can also be useful to users, when they are tweaking their commands and don't need/want explicit handling for stderr for example.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
@@ -100,17 +100,29 @@ public class ExecWebSocketListener implements ExecWatch, AutoCloseable, WebSocke
   private final class ListenerStream {
     private MessageHandler handler;
     private ExecWatchInputStream inputStream;
+    private String name;
+
+    public ListenerStream(String name) {
+      this.name = name;
+    }
 
     private void handle(ByteBuffer byteString, WebSocket webSocket) throws IOException {
       if (handler != null) {
         handler.handle(byteString);
       } else {
+        if (LOGGER.isDebugEnabled()) {
+          String message = ExecWebSocketListener.toString(byteString);
+          if (message.length() > 200) {
+            message = message.substring(0, 197) + "...";
+          }
+          LOGGER.debug("exec message received on channel {}: {}", name, message);
+        }
         webSocket.request();
       }
     }
   }
 
-  private static final Logger LOGGER = LoggerFactory.getLogger(ExecWebSocketListener.class);
+  static final Logger LOGGER = LoggerFactory.getLogger(ExecWebSocketListener.class);
   private static final String HEIGHT = "Height";
   private static final String WIDTH = "Width";
 
@@ -152,14 +164,14 @@ public class ExecWebSocketListener implements ExecWatch, AutoCloseable, WebSocke
     }
 
     this.terminateOnError = context.isTerminateOnError();
-    this.out = createStream(context.getOutput());
-    this.error = createStream(context.getError());
-    this.errorChannel = createStream(context.getErrorChannel());
+    this.out = createStream("stdOut", context.getOutput());
+    this.error = createStream("stdErr", context.getError());
+    this.errorChannel = createStream("errorChannel", context.getErrorChannel());
     this.serialExecutor = new SerialExecutor(executor);
   }
 
-  private ListenerStream createStream(StreamContext streamContext) {
-    ListenerStream stream = new ListenerStream();
+  private ListenerStream createStream(String name, StreamContext streamContext) {
+    ListenerStream stream = new ListenerStream(name);
     if (streamContext == null) {
       return stream;
     }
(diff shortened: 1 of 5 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
```diff
@@ -100,17 +100,29 @@ public interface MessageHandler {
...
+          }
+          LOGGER.debug("exec message received on channel {}: {}", name, message);
+        }
         webSocket.request();
       }
     }
   }
 
-  private static final Logger LOGGER = LoggerFactory.getLogger(ExecWebSocketListener.class);
+  static final Logger LOGGER = LoggerFactory.getLogger(ExecWebSocketListener.class);
```
manusa, line 125: Could we keep this private and retrieve the logger in PodOperationContext using the same approach?

manusa: Also, are we sure that this is the class/package/point where we want to add the flag to decide if these messages should be logged? Can we document how to enable this logging? Would it make sense to check for debug enabled at the base package?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/PodOperationContext.java
```diff
@@ -164,10 +164,11 @@ public void addQueryParameters(URLBuilder httpUrlBuilder) {
     if (in != null || redirectingIn) {
       httpUrlBuilder.addQueryParameter("stdin", "true");
     }
-    if (output != null) {
+    boolean debug = ExecWebSocketListener.LOGGER.isDebugEnabled();
```
manusa, line 167: This reveals an odd coupling between ExecWebSocketListener and this class.

The name of the method suggests a wider usage of these parameters, however, they are only ultimately consumed at the `public ExecWatch exec(String... command)` method. Maybe we should rename the addQueryParameters method.

shawkins: > This reveals an odd coupling between ExecWebSocketListener and this class.

It's partly an issue with the existing code structure.  The entry class is PodOperationsImpl, but it's in a different package from the rest of the logic - presumably because there could eventually be a v2.  That's how most of dsl.internal is used.

> Maybe we should rename the addQueryParameters method.

It's usage should be clear from the context :) You could also make the case that the classes in dsl.internal should move into core.v1 and have methods be made package private.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
shawkins: > Could we keep this private and retrieve the logger in PodOperationContext using the same approach?

You mean enable this off of two separate loggers?  I didn't like that approach, even if they share a common base.

> Also, are we sure that this is the class/package/point where we want to add the flag to decide if these messages should be logged? 

Yes, they should be logged in the ExecWebSocketListener, and we have to make an earlier decision to include the query flags.

> Can we document how to enable this logging?

Yes, we just have to agree on what it should be first.

> Would it make sense to check for debug enabled at the base package?

You could, it's just a bit more code to create / access a couple of additional loggers - call me lazy, but I'd rather reuse something that exists if possible.

manusa: > You mean enable this off of two separate loggers? I didn't like that approach, even if they share a common base.

I meant doing this from `PodOperationContext`:
```
boolean debug = // ExecWebSocketListener.LOGGER.isDebugEnabled();
  LoggerFactory.getLogger(ExecWebSocketListener.class);
``` 
It's still coupled, but at least we don't access the other class variable that should be used for internal purposes only.

However, I'm going over both https://github.com/fabric8io/kubernetes-client/pull/4288#discussion_r928781328 and https://github.com/fabric8io/kubernetes-client/pull/4288#discussion_r928777707 and as I see it's just that the structure is flawed.

The `PodOperationContext` is not only used for Exec operations, but for other operations too.

> It's usage should be clear from the context :)

For me it's not clear that a **public** `addQueryParameters` method is used only to add exec-related parameters. Maybe this method just doesn't belong here.

I guess that at this point it doesn't matter anyway.

manusa (approved):

shawkins:
@manusa based upon the approval, I'll add doc note about this.  

It's worth mentioning that an alternative is to directly associate a debug output stream.  However as an output stream we loose the ability to discern individual messages.  It would be worth considering at some point directly exposing the lower-level onMessage abstraction.  That somewhat relates to #4157 and #4201 as well.  The only place we use sendAsync with a Reader or InputStream is from log reading calls where we're directly handling the user back the stream abstraction.

manusa (approved):

PR #4034, opened 2022-04-04
decoupling unmatched handling from kubernetesdeserializer

shawkins:
## Description
As either a stand alone change or part of one of the other serialization issues #3972 or #4024 I'd like to remove the coupling between the KubernetesDeserializer and the unmatched property handling.  This change elevates that to the Serialization level.  This will further allow us to change the handling of parsing / processing parameters - that has been embedded down in the serialization layer as raw string processing, but it doesn't need to be.  Parsing with unmatched type handling on, or parsing as a map would allow for post processing of property substitution.

## Type of change

 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [ ] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [x] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [x] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [ ] I Added [CHANGELOG](https://github.com/fabric8io/kubernetes-client/blob/master/CHANGELOG.md) entry regarding this change
 - [ ] I have implemented unit tests to cover my changes
 - [ ] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/master/doc/CHEATSHEET.md) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [ ] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

Issue #3972, opened 2022-03-15
Static deserialization issues


shawkins:
### Is your task related to a problem? Please describe

To eventually remove the static type registration from the KubernetesDeserializer we can't have static entry points that rely on this implicit class resolution.

### Describe the solution you'd like

Deprecate and provide alternatives to any static Serialization.unmarshall routines that don't supply the type.

shawkins:
Specifically deprecate
```
UNMATCHED_FIELD_TYPE_MODULE
jsonMapper()
yamlMapper()
unmarshal(InputStream is)
unmarshal(InputStream is, Map<String, String> parameters)
unmarshal(InputStream is, ObjectMapper mapper)
unmarshal(InputStream is, ObjectMapper mapper, Map<String, String> parameters)
unmarshal(String str)
```

If we want the replacement to be available without completely building a client, then we'll need to add some method on the KubernetesClientBuilder.  Otherwise with the logic in #3966 it would be easier to obtain the deserializer from the client:

var serdes = client.serialization(); // type and method name tbd
serdes.unmarshal("something yaml");

Under the covers I'm assuming, but not knowledgeable enough with Jackson to known for sure, that we can associate an instance of the KubernetesDeserializer to use.

shawkins:
At a minimum we would introduce here an api for accessing what are now static concerns of serialization:

```
public interface KubernetesSerialization {

  <T> T unmarshal(InputStream is);

  <T> T unmarshal(InputStream is, Map<String, String> parameters);

  <T> T unmarshal(String str);

  <T> T clone(T resource);

  void setLogUnmatchedFieldWarnings(boolean logWarnings);

  void setRestrictUnmatchedFieldToTemplates(boolean restrictToTemplates);

  default void registerCustomKind(String kind, Class<? extends KubernetesResource> clazz) {
    registerCustomKind(null, kind, clazz);
  }

  void registerCustomKind(String apiVersion, String kind, Class<? extends KubernetesResource> clazz);

  // optionally we could include pathutils methods

  String withoutRuntimeState(KubernetesResource object, Format format, boolean omitStatus);

  String jsonDiff(KubernetesResource current, KubernetesResource updated, boolean omitStatus);

}
```

Assuming that we want users to obtain this without creating a client, there would likely need to be another factory:  KubernetesSerialization.Factory or similar.  It would also be obtainable off of the client: client.getKubernetesSerialization()

[...]

PR #4292, opened 2022-07-22
fix #4109: moving template handling to a custom deserializer

### LINKED ISSUE #4109: deserialize value of type `int` from String
` template = client.templates().inNamespace(NamespaceName).withName(templateName).get(); 
`
```
Caused by: com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type `int` from String "${{NUMBER_OF_INSTANCES}}": not a valid `int` value
 at [Source: UNKNOWN; byte offset: #UNKNOWN] (through reference chain: io.fabric8.openshift.api.model.Template["objects"]->java.util.ArrayList[3]->io.fabric8.kubernetes.api.model.apps.StatefulSet["spec"]->io.fabric8.kubernetes.api.model.apps.StatefulSetSpec["replicas"])
```
```
<dependency>
            <groupId>io.fabric8</groupId>
            <artifactId>openshift-client</artifactId>
            <version>5.12.2</version>
</dependency>
```

shawkins:
## Description
This addresses #4109 and is related to #3972.  It is also a replacement for #4034 - since the idea of subclassing ObjectMapper is unpalatable, the other way to do this is with a custom deserializer.  But to keep the same unmatched logic, we need to move that to model-common, rather than client-api.

If it's acceptable, then I can further add a changelog about moving the unmatched static methods, etc.  The unmatched test references thing from the api, so it can't be moved to model-common as is - but it could of course be further refactored to move the tests.

With this change we also don't need to have the parameter handling logic so deeply coupled with the Serialization logic - that is we don't have to do string substitution on the string prior to parsing.  That could easily be done as post processing logic in the TemplateOperationsImpl.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/Serialization.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/Serialization.java
@@ -25,7 +25,7 @@ import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import io.fabric8.kubernetes.api.model.GenericKubernetesResource;
 import io.fabric8.kubernetes.api.model.KubernetesResource;
 import io.fabric8.kubernetes.client.KubernetesClientException;
-import io.fabric8.kubernetes.client.utils.serialization.UnmatchedFieldTypeModule;
+import io.fabric8.kubernetes.model.jackson.UnmatchedFieldTypeModule;
 import org.yaml.snakeyaml.DumperOptions;
 import org.yaml.snakeyaml.Yaml;
 import org.yaml.snakeyaml.constructor.SafeConstructor;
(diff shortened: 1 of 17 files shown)


manusa:
I need to check, but I think this one would break the feature where we allow deserializing fields with unmatched types to the additionalProperties map.

shawkins:
> I need to check, but I think this one would break the feature where we allow deserializing fields with unmatched types to the additionalProperties map.

Other than changing the package, it does not.  You can of course leave the old ones in place, but deprecated if you want.

manusa:
> > I need to check, but I think this one would break the feature where we allow deserializing fields with unmatched types to the additionalProperties map.
> 
> Other than changing the package, it does not. You can of course leave the old ones in place, but deprecated if you want.

I just gave it a quick look and the changes in the Serialization class made me think of this. I'll try to review properly later

shawkins:
@sunix I mentioned something about backporting on the call - I was mixed up.  #4263 is the one that should be considered for backport, not this one.

manusa (approved):

PR #4298, opened 2022-07-22
Add support for Min, Max, Pattern and Nullable on CRD generator fields

### LINKED ISSUE #4224: [crd-gen] Support openAPIV3Schema fields: minimum, maximum, pattern, nullable

_(duplicate of #2959)_

At the moment there is no way to auto generate a crd (with crd-gen tool) that contains properties with validation fields like **_minimum_**, **_maximum_**, **_pattern_**, **_nullable_**.

CRD example:

```yaml
...

apiVersion: apiextensions.k8s.io/v1
metadata:
  name: tenants.example.com
spec:
  group: example.com
  names:
    kind: Tenant
    plural: tenants
    singular: tenant
  scope: Namespaced
  versions:
  - name: v1
    schema:
      openAPIV3Schema:
        properties:
          spec:
            properties:
              kafka:
                type: object
                properties:
                  port:
                    type: integer
                    minimum: 0
                    maximum: 65535
                    nullable: false
                  hostname:
                    type: string
                    pattern: '^([A-Za-z0-9-]{1,63}\.)+[[A-Za-z0-9-]{1,63}$'
                    nullable: false
                nullable: false
...
```

I'd like to use javax annotation (like `@min @max @pattern`) on a java model fields.

    private String protocol;

    @Pattern(regexp = "\\^\\(\\[A-Za-z0-9-]\\{1,63}\\\\\\.\\)\\+\\[\\[A-Za-z0-9-]\\{1,63}\\$")
    private String hostname;

    @Min(0)
    @Max(65535)
    private Integer port;

Currently @NotNull annotation generate the filed "**_required_**" on crd, but for "_**nullable**_" validation maybe we need of a new specific annotation.

_No response_

### Additional context

version: 5.12.2

andreaTP:
## Description
Resolves #4224

## Type of change

## Checklist

### DIFF
--- a/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
+++ b/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
@@ -17,7 +17,6 @@ package io.fabric8.crd.generator;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
-import io.fabric8.crd.generator.annotation.SchemaSwap;
 import io.fabric8.crd.generator.utils.Types;
 import io.fabric8.kubernetes.api.model.Duration;
 import io.fabric8.kubernetes.api.model.IntOrString;
@@ -32,14 +31,10 @@ import java.util.*;
 import java.util.stream.Collectors;
 
 import static io.sundr.model.utils.Types.BOOLEAN_REF;
-
-import static io.sundr.model.utils.Types.STRING_REF;
-
-import static io.sundr.model.utils.Types.INT_REF;
-
-import static io.sundr.model.utils.Types.LONG_REF;
-
 import static io.sundr.model.utils.Types.DOUBLE_REF;
+import static io.sundr.model.utils.Types.INT_REF;
+import static io.sundr.model.utils.Types.LONG_REF;
+import static io.sundr.model.utils.Types.STRING_REF;
 
 /**
  * Encapsulates the common logic supporting OpenAPI schema generation for CRD generation.
@@ -54,13 +49,12 @@ public abstract class AbstractJsonSchema<T, B> {
   protected static final TypeDef OBJECT = TypeDef.forName(Object.class.getName());
   protected static final TypeDef QUANTITY = TypeDef.forName(Quantity.class.getName());
   protected static final TypeDef DURATION = TypeDef.forName(Duration.class.getName());
-  protected static final TypeDef INT_OR_STRING =TypeDef.forName(IntOrString.class.getName());
+  protected static final TypeDef INT_OR_STRING = TypeDef.forName(IntOrString.class.getName());
 
   protected static final TypeRef OBJECT_REF = OBJECT.toReference();
   protected static final TypeRef QUANTITY_REF = QUANTITY.toReference();
   protected static final TypeRef DURATION_REF = DURATION.toReference();
   protected static final TypeRef INT_OR_STRING_REF = INT_OR_STRING.toReference();
-
 
   protected static final TypeDef DATE = TypeDef.forName(Date.class.getName());
   protected static final TypeRef DATE_REF = DATE.toReference();
@@ -75,7 +69,7 @@ public abstract class AbstractJsonSchema<T, B> {
   protected static final TypeRef P_LONG_REF = new PrimitiveRefBuilder().withName("long").build();
   protected static final TypeRef P_DOUBLE_REF = new PrimitiveRefBuilder().withName("double").build();
   protected static final TypeRef P_BOOLEAN_REF = new PrimitiveRefBuilder().withName(BOOLEAN_MARKER)
-    .build();
+      .build();
 
   private static final Map<TypeRef, String> COMMON_MAPPINGS = new HashMap<>();
   public static final String ANNOTATION_JSON_PROPERTY = "com.fasterxml.jackson.annotation.JsonProperty";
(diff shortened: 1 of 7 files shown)


crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
```diff
@@ -84,11 +78,53 @@
   public static final String ANNOTATION_JSON_ANY_GETTER = "com.fasterxml.jackson.annotation.JsonAnyGetter";
   public static final String ANNOTATION_JSON_ANY_SETTER = "com.fasterxml.jackson.annotation.JsonAnySetter";
   public static final String ANNOTATION_NOT_NULL = "javax.validation.constraints.NotNull";
+  public static final String ANNOTATION_MAX = "javax.validation.constraints.Max";
+  public static final String ANNOTATION_MIN = "javax.validation.constraints.Min";
+  public static final String ANNOTATION_PATTERN = "javax.validation.constraints.Pattern";
+  public static final String ANNOTATION_NULLABLE = "io.fabric8.crd.generator.annotation.Nullable";
```
manusa, line 84: Since we're using javax.validation, maybe we should consider jsr305 annotations or support for others (JetBrains, Spring) instead of creating our own.
WDYT?

JSR 305 is definitely dead, but providing yet another annotation for the same purpose doesn't sound great.

andreaTP: I proposed this implementation as stated in the relevant Issue, but, I definitely relate with your feelings @manusa .

Here are my considerations:
- Spinning an own annotation to emit a single field in the CRD is great, I agree
- The current implementation of `required` relies on `javax.validation.constraints.NotNull` and here I tried to stitch with the original code design
- Adding an additional dependency (on Spring, JetBrains or Micronaut) is something I would very likely avoid as we are, anyhow kind of "abusing" the API
- The implementation of `Min`/`Max` is [semantically incorrect](https://github.com/fabric8io/kubernetes-client/pull/4298/files#diff-0bfefceda9952f447cf1a5965a31e7467320a96f1a1fd8877c49978d98102116R399-R403) as we are converting `Long` to `Double` effectively prohibiting the user to use floating point numbers where it's actually allowed
- Those encodings will need a corresponding match in the `java-generator` which is currently relying on Jackson tricks to encode e.g. `Nullable`
- Parsing those annotations **does not** guarantee at all that they are respected during deserialization, on the contrary, without the proper Jackson annotations in place ppl are going to be extremely surprised by the behavior

I'm not sure how to solve all of the above, we have a few options with different tradeoffs though:
- continue with the current approach keeping in mind the limitation and possible tech debt we are accepting
- roll out our own annotations for everything `constraints` to avoid extra dependencies, give a clear semantic to the API (that cannot be mixed with other frameworks etc.), in this case, we should draft a deprecation path for `javax.validation.constraints.NotNull`

I'm looking for other opinions and ideas here, maybe @OneCricketeer or @fabiobrz wanna chip into this discussion.

andreaTP: cc. @csviri too, as this is going to transitively affect JOSDK too.

manusa: Let's discuss this point this afternoon. Maybe I wasn't explaining myself correctly, or I misunderstand the implications.

My understanding is that since we use a string to reference the annotation, there's no need for the generator to actually have a dependency on an artifact that provides those annotations.

My proposal was supporting a set of annotations that are equivalent, for example, for Not Null: `org.jetbrains.annotations.NotNull`, `org.springframework.lang.NonNull`, `javax.annotation.Nonnull`, and so on. Maybe or maybe not provide our own annotation too.

andreaTP: > My understanding is that since we use a string to reference the annotation, there's no need for the generator to actually have a dependency on an artifact that provides those annotations.

Correct, in fact, the generator doesn't depend on the [validation annotations library](https://github.com/fabric8io/kubernetes-client/blob/bfeafb28ec9ea364dd629eef058754080985b215/crd-generator/api/pom.xml#L75-L79)

> My proposal was supporting a __set__ of annotations that are equivalent

The tricky part about this approach is that every implementation has its own nuances and semantics but we are using it for a very different purpose. We should check for each available annotation semantic and, ideally, properly test both generation and runtime behavior.

andreaTP: The agreed approach is to spin out our own annotations, mentioning that they are only relevant for generating the corresponding fields in the CRD (e.g. no runtime guarantees for ser/deser).
The change will be made in a retro-compatible way emitting a warning when `javax.validation.constraints.*` is used.

iocanel: IMHO `javax.validation.contraints` should be the first thing supported. It's a standard, it's aligned with what we already have in place, it's supported by the builder validation ...

manusa: One of the issues we discussed about with these more generic annotations, was their lack of alignment with Kubernetes OpenApi Schema spec.

For example, the `@Max` annotation expects a long as a value, while the Kubernetes equivalent is a [float64](https://github.com/kubernetes/apiextensions-apiserver/blob/2184a8de754ca328bc92fa33a339307016f953f4/pkg/apis/apiextensions/types_jsonschema.go#L30).

andreaTP: @fabiobrz I agree with your analysis, please, feel free to submit a PR to improve/make more clear the current docs, it will be really appreciated!

andreaTP:
Superseded #4348

Issue #4318, opened 2022-08-05
Better Lease Object Handling on Leader Election Cancel


csviri:
### Is your enhancement related to a problem? Please describe

Will describe the problem in context of operators, considering the following situation:

The controllers are deployed in a way that there are multiple instances running with leader election on. Let's say a new version of the controller is released, in this case the old ones needs to be stopped and one of the new controllers becomes the leader.

In the current implementation when a leader election is cancelled it just stops renewing the lease. However this can cause a delay for the new instances to become a leader (the delay is  <=  [lease duration] + [retry period] ).

### Describe the solution you'd like

Would it be better to on cancel delete or update a Lease object to indicate that now there is no leader?

shawkins:
> Would it be better to on cancel delete or update a Lease object to indicate that now there is no leader?

Makes sense.

manusa:
The [release](https://github.com/kubernetes/kubernetes/blob/f0bd02ca5e1773396288149f4faab25d2f9c6dfe/staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go#L287-L290) + [releaseOnCancel](https://github.com/kubernetes/kubernetes/blob/f0bd02ca5e1773396288149f4faab25d2f9c6dfe/staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go#L149-L154) feature is missing and needs to be ported.

shawkins:
The pr has https://github.com/fabric8io/kubernetes-client/pull/4322/files#diff-46850c90b4117631201b452d6d1b2bc5cc7de7270ccda4d69e5ac2c829938419R120 - which takes the current record and sets the duration to 0.  It can of course be made to look like the go code, but I'm not sure why they would leave a 1 second duration or reset the acquireTime as the current is already the leader.

manusa:
> which takes the current record and sets the duration to 0

I'm not sure either, `0` might be interpreted as indefinite duration :shrug:?

In any case, I think we should port the exact behavior to avoid errors or future bugs due to inconsistent behaviors.

shawkins:
> I'm not sure either, 0 might be interpreted as indefinite duration shrug?

No, the time comparison is based upon "time + duration > now" in both the go client and java client.

> In any case, I think we should port the exact behavior to avoid errors or future bugs due to inconsistent behaviors.

The go implementation must have changed a good bit - for example they are no longer relying on timestamps in the record, rather there is a local observedTime instead.

[...]

Issue #2207, opened 2020-05-07
Missing cancellation support in `LeaderElector`


dgloeckner:
After calling run() on the leader it will eventually enter the loop for renewing its lease.
There does not seem to be an API for gracefully shutting down a leader elector. I imagine that there should be a way to make the LeaderElector break out of the renewal loop, e.g. calling cancel() on a leader elector.

Following code does _not_ cancel the leader elector

```
ExecutorService executorService = Executors.newSingleThreadExecutor();
		try (KubernetesClient kc = new DefaultKubernetesClient()) {
			LeaderElector leaderElector = kc.leaderElector().withConfig(
					new LeaderElectionConfigBuilder().withName("Sample Leader Election configuration").withLeaseDuration(Duration.ofSeconds(15L))
							.withLock(new LeaseLock(namespace, leaseName, identity)).withRenewDeadline(Duration.ofSeconds(10L))
							.withRetryPeriod(Duration.ofSeconds(2L)).withLeaderCallbacks(
							new LeaderCallbacks(() -> startLeading.countDown(), () -> log.info("STOPPED LEADERSHIP"),
									newLeader -> log.info("New leader elected {}", newLeader))).withReleaseOnCancel(true).build()).build();
			// The future takes care of re-newing the lease until cancelled.
			Future<?> makeMeTheLeader = executorService.submit(() -> leaderElector.run());
			startLeading.await();
			// We have the lock.
			...
			// Release the lock.
			makeMeTheLeader.cancel(true);
		}
```

Note: the original kubernetes Java client has a similar issue, see https://github.com/kubernetes-client/java/issues/944

manusa:
In your example you can achieve this in a "not so clean" way by killing the thread that is running the LeaderElector loop, exception handling in LeaderElector should take care of cleaning up local resources, cluster should get outdated in the provided period, so a new leader would take its place (this is demonstrated in [this example](https://github.com/fabric8io/kubernetes-client/blob/d3d659426b29f7ccbeee42a63ae5d7da7400c94a/kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/LeaderElectionExamples.java#L96)).

However, adding a cleaner way to stop the LeaderElector could be easily added, we should definitely consider it.

dgloeckner:
Hi there,

I guess by "killing the thread" you mean I should set its interrupted flag, right?

[...]

PR #4322, opened 2022-08-08
fix #4320 #4318: implementing release on cancel and fixing transitions

### LINKED ISSUE #4320: LeaderElection leader transitions
LeaderTransitions is resetting on a change of leadership, and incrementing while holding the lock.  I think this should instead be to increment on a change and to leave the value alone will renewing.

shawkins:
## Description
addresses #4320 #4318 for leader election.

@manusa @csviri in the previous refactoring I had already put in a call to onStopLeading when cancelled - however this is not strictly correct.  If you don't have release on cancel set you won't loose the leadership until after the duration has expired.  Should that be refined as well?

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,13 @@
 * Fix #4256: crd-generator-apt pom.xml includes transitive dependencies
 * Fix #4294: crd-generator respects JsonIgnore annotations on enum properties
 * Fix #4247: NO_PROXY with invalid entries throws exception
+* Fix #4320: corrected leader transitions field on leader election leases
 
 #### Improvements
 * Fix #4041: adding Quantity.getNumericalAmount with an explanation about bytes and cores.
 * Fix #4241: added more context to informer logs with the endpoint path
 * Fix #4250: allowing for deserialization of polymorphic unwrapped fields
+* Fix #4318: implemented LeaderElection releaseOnCancel
 * Fix #4259: Java Generator's CR should have Lombok's `@EqualsAndHashCode` with `callSuper = true`
 
 #### Dependency Upgrade
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/extended/leaderelection/LeaderElector.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/extended/leaderelection/LeaderElector.java
@@ -112,27 +112,34 @@ public class LeaderElector {
 
   private void stopLeading() {
     LeaderElectionRecord current = observedRecord.get();
-    if (current == null || !Objects.equals(current.getHolderIdentity(), leaderElectionConfig.getLock().identity())) {
+    if (current == null || !isLeader(current)) {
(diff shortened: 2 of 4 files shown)


manusa:
> @manusa @csviri in the previous refactoring I had already put in a call to onStopLeading when cancelled - however this is not strictly correct. If you don't have release on cancel set you won't loose the leadership until after the duration has expired. Should that be refined as well?

I discussed this internally with Attila. The Leader Election "extension" was implemented as a port of what was available in Client Go at the time.
The cancel functionality was added later on by you in scope of #2207 in #4125. I'm not sure if this was ported or not from Client-Go too.
- If the cancel behavior is available on Client-Go, then we should port **exactly** what's implemented there.
- If the cancel behavior is something exclusive of ours, then we can do as we please. It seems reasonable that if a leader actively cancels its leadership, it should also update the lease record (regardless of the Lock implementation).

OT: It would be nice to discuss what are good use-cases of this cancel behavior. Personally I don't see why a leader should decide to stop leading.

shawkins:
> I'm not sure if this was ported or not from Client-Go too.

No, since it's base upon the CompletableFuture.

> If the cancel behavior is available on Client-Go, then we should port exactly what's implemented there.

It is available.  As part of the initial port the leader election configuration included releaseOnCancel - which was left as doing nothing.  This pr makes that config property meaningful.

> Personally I don't see why a leader should decide to stop leading.

If the leader is in some way out-of-date and wants to cede leadership to others, then you have a reason to cancel with a release.

shawkins:
> Should that be refined as well?

To clarify this behavior is not new, it was carried over from the existing port.  Anytime the loop stopped it could call onStopLeading if it had been the leader.

The question is - without a successful call to release, this isn't necessarily true - so does it need to be refined?

manusa:
> If the leader is in some way out-of-date and wants to cede leadership to others, then you have a reason to cancel with a release.

Yes, this is the obvious reason for the _"why a leader should decide to stop leading."_. However, it's unclear to me what the bigger picture would be here and why this should be relevant and what are the implications to the whole ecosystem. It's more of a discussion topic.

A use-case based on the outdated version would be an Operator or Controller that has a new version available. However, probably the Deployment controller would be taking care of rolling out the newer version of the workload by tearing down the leader and the outdated replicas. So it might make sense that the outdated workloads no longer participate on the leadership battle, but this might bring some unwanted problems (and would also require them to be aware of their outdated version state).

shawkins:
> The question is - without a successful call to release, this isn't necessarily true - so does it need to be refined?

My reading of: https://github.com/kubernetes/kubernetes/blob/f0bd02ca5e1773396288149f4faab25d2f9c6dfe/staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go#L202

is that onStoppedLeading is always called - even if you never obtain the leadership, so I think we can just leave our current logic.

shawkins:
@manusa should be good now

manusa (approved):

PR #4328, opened 2022-08-10
Fix istio

Sgitario:
## Type of change

## Checklist

### DIFF
--- a/extensions/istio/model-v1alpha3/pom.xml
+++ b/extensions/istio/model-v1alpha3/pom.xml
@@ -113,6 +113,19 @@
               </artifacts>
             </configuration>
           </execution>
+          <execution>
+            <id>add-source</id>
+            <phase>process-resources</phase>
+            <goals>
+              <goal>add-source</goal>
+            </goals>
+            <configuration>
+              <sources>
+                <source>src/generated/java</source>
+                <source>target/generated-sources/annotations</source>
+              </sources>
+            </configuration>
+          </execution>
         </executions>
       </plugin>
       <plugin>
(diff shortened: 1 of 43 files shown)


Sgitario:
cc @manusa @Encounter77

By the way, @Encounter77 according to https://istio.io/latest/docs/ops/diagnostic-tools/istioctl-analyze/#enabling-validation-messages-for-resource-status, the level is provided as a text not an integer:

```yaml
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
...
spec:
  gateways:
  - bogus-gateway
  hosts:
  - ratings
...
status:
  observedGeneration: "1"
  validationMessages:
  - documentationUrl: https://istio.io/v1.14/docs/reference/config/analysis/ist0101/
    level: ERROR
    type:
      code: IST0101
```

Note that integers won't work and this is a limitation of the json schema generator that we might want to fix. wdyt @manusa ?

Sgitario:
> In this url : [https://istio.io/latest/docs/ops/diagnostic-tools/istioctl-analyze/#enabling-validation-messages-for-resource-status](url)
> 
> this is an `experimental function`, we did the transformation configuration for compatibility, yes?

I don't understand your question. My point is that this will only work whether the server returns exactly the text "ERROR", "WARNING", ..., not integers (as the example you shared in the issue) or "Error", "Info".

manusa (approved):

manusa (approved):
LGTM, thank you for taking care of this :heart:

rohanKanojia (approved):

Sgitario:
I've added a new commit to allow parsing enum values regardless what the format we receive from the server. 
So, if we receive the value 3, we'll parse it to ERROR because it's how it's defined in the Go model. 
Moreover, we relax the condition of strings, so we can also get values as string (no case sensitive). See updated test in: https://github.com/fabric8io/kubernetes-client/pull/4328/files#diff-8f86dd83c6f7430eb24af4c0df19053786fa894b34e21e2cbb66f2f95956894cR28 and https://github.com/fabric8io/kubernetes-client/pull/4328/files#diff-a75f540c99c4014aaf89fec369f79791c65b2f46891bd01e3552abf20bb38f82R79

@Encounter77 I think with the last commit, this will work for you. 
FYI @manusa @rohanKanojia

manusa (approved):

PR #4345, opened 2022-08-22
Lgass patch3

lgass:
## Type of change

## Checklist
 - [x] I tested my code in K3D

### DIFF
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/internal/CertUtils.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/internal/CertUtils.java
@@ -17,6 +17,7 @@ package io.fabric8.kubernetes.client.internal;
 
 import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.utils.Utils;
+import org.bouncycastle.jce.provider.BouncyCastleProvider;
 import org.bouncycastle.openssl.PEMKeyPair;
 import org.bouncycastle.openssl.PEMParser;
 import org.bouncycastle.openssl.jcajce.JcaPEMKeyConverter;
@@ -51,7 +52,8 @@ import java.util.concurrent.Callable;
 import java.util.stream.Collectors;
 
 public class CertUtils {
-  private CertUtils() { }
+  private CertUtils() {
+  }
 
   private static final Logger LOG = LoggerFactory.getLogger(CertUtils.class);
   private static final String TRUST_STORE_SYSTEM_PROPERTY = "javax.net.ssl.trustStore";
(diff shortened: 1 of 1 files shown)


lgass:
I am not sure how to proceed, the OpenShift failures appear to be unrelated?
In regards to the SonarCloud Code Analysis, I am not sure what exactly I am supposed to do, as there is not much to unit test without having the FIPS library in the runtime.

manusa:
Hi @lgass ,

I've seen you opened a few PRs to "solve" some kind of issue you might be experiencing, but never created an issue or discussion.

From what I've seen, your issue with FIPS should be solvable just by adding the Bouncy Castle dependencies. The code you're trying to remove is just used to make those dependencies optional and warn the users in case their runtime code tries to use them but haven't included them.

manusa:
Also note that FIPS should be configured externally: https://github.com/fabric8io/kubernetes-client/pull/2788/files#r571887528

lgass:
If I use the code without this modification, it will fail because it attempts to load the standard BouncyCastleProvider class even though BCFIPS is found. This is because it is still trying to resolve the class dependencies on the anonymous callable class. It is easy to reproduce if you have a runtime that has the BCFIPS provider, once it gets to the callable code in the original code, it still tries to resolve the standard class even though the line where it tries to add it does not get called.

lgass:
To be clear, my use case is that I have the BCFIPS provider. You can only have BC or BCFIPS, not both. So the EC key method is getting called because the K3D environment where this is ran is using that type of key. Without the modification it fails, even though BCFIPS is in the current runtime. Note that with my modifications, EC keys work fine with the BCFIPS provider.

manusa:
I'm not sure I'm following, but it might be better if you provide a reproducer project in an issue so that it can be properly discussed.

lgass:
Forget it man, this is simple class loading logic. I do not have time to prove such a simple claim. Ill close the PR and you can fix it for the next client of the library that actually uses BCFIPS and has an EC key scenario, which is probably not the case for 99% of the consumers of this library.

On the other hand, kudos for such a cool library, up till now it has been a breeze to use.

manusa:
> Forget it man, this is simple class loading logic. I do not have time to prove such a simple claim.

As I said, you don't need to prove anything, this is more about providing the most optimal solution (for you and for the rest of users). You have created several PRs and have done a lot of work already...

My suggestion is just to create an issue to discuss the best possible solution. Since right now, FIPS is supported, but must be configured externally. The point of this discussion would be to see if a better BC configurable option can be provided instead of preventing BCFIPS, especially considering seldom use cases like your own. A reproducer helps not only "proving" your case, but understanding what you're trying to achieve and how you are using the client in your specific context. This is also used to ease the creation of tests that verify that a fix won't be regressed in the future.

lgass:
The change took all of 10 minutes, the multiple PR's were due to me getting used to GITHUB (which I hate) and making a few mistakes on the way. I closed all of the other PR's, but this one at least is more or less clean. It simply moves the callable class to another method so that the primary method for EC keys won't get bombed in the BCFIPS use case.

I can appreciate your concerns with an open source project, but sometimes real users don't have time to create all the boiler plate infrastructure to demonstrate something that should be obvious. In any case I just wanted to give back to the community but it obviously is more complicated than my feeble attempt.

PR #4348, opened 2022-08-24
Introduce specific annotations for the generators

andreaTP:
## Description
Follow up from #4298 , introducing annotations for the fields that we want to support/map.

## Type of change

## Checklist

### DIFF
--- a/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
+++ b/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
@@ -17,7 +17,6 @@ package io.fabric8.crd.generator;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
-import io.fabric8.crd.generator.annotation.SchemaSwap;
 import io.fabric8.crd.generator.utils.Types;
 import io.fabric8.kubernetes.api.model.Duration;
 import io.fabric8.kubernetes.api.model.IntOrString;
@@ -60,6 +59,8 @@ public abstract class AbstractJsonSchema<T, B> {
   protected static final TypeDef DATE = TypeDef.forName(Date.class.getName());
   protected static final TypeRef DATE_REF = DATE.toReference();
 
+  private static final String VALUE = "value";
+
   private static final String INT_OR_STRING_MARKER = "int_or_string";
   private static final String STRING_MARKER = "string";
   private static final String INTEGER_MARKER = "integer";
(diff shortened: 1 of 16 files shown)


generator-annotations/src/main/java/io/fabric8/generator/annotation/Max.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+package io.fabric8.generator.annotation;
+
+import java.lang.annotation.*;
+
+@Target({ ElementType.ANNOTATION_TYPE, ElementType.FIELD, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Max {
+  double value();
+}
```
manusa, line 24: For the follow-ups, it might be interesting to add Javadoc to each annotation properly documenting its usage, and maybe linking the Kubernetes spec equivalent.

manusa (approved):

generator-annotations/src/main/java/io/fabric8/generator/annotation/Max.java
andreaTP: I'm failing to find a good "linkable" source with the specific field description, e.g.:
https://kubernetes.io/docs/reference/kubernetes-api/extend-resources/custom-resource-definition-v1/

Do you have any idea?

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
```diff
@@ -78,6 +79,11 @@
   public static final String ANNOTATION_JSON_IGNORE = "com.fasterxml.jackson.annotation.JsonIgnore";
   public static final String ANNOTATION_JSON_ANY_GETTER = "com.fasterxml.jackson.annotation.JsonAnyGetter";
   public static final String ANNOTATION_JSON_ANY_SETTER = "com.fasterxml.jackson.annotation.JsonAnySetter";
+  public static final String ANNOTATION_MIN = "io.fabric8.generator.annotation.Min";
```
metacosm, line 82: All annotations should be put in the same package, imo.

andreaTP: At this point we have annotations in 2 places:
 - `io.fabric8.generator.annotation` -> generic modifiers to represent CRD properties in Java code
 - `io.fabric8.crd.generator.annotation` -> specific annotations to tweak the specifically the `crd-generator`
 
 Do you see any other invariant @metacosm ?

generator-annotations/src/main/java/io/fabric8/generator/annotation/Max.java
manusa: No, I only found the go types sources when I was evaluating the field precision of the Max and Min configurations.

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
metacosm: As soon as the annotations are used in different contexts, it would be better for users to help discoverability, in particular, for the annotations to be in the same location even though they initially were developed for different purposes… Obviously, changing annotations location at this point wouldn't be backwards compatible so maybe just a matter of documentation? It might make sense to explain why they live in different packages so that users (and future maintainers, down the line 😸) don't think it's an arbitrary decision.

andreaTP: Added a brief JavaDoc on each annotation, let me know if you want other/more changes for this 🙂

generator-annotations/src/main/java/io/fabric8/generator/annotation/Max.java
andreaTP: added a generic reference to: `jsonschemaprops`  K8s API reference ... the best I have found in the wild ...

metacosm (approved):

andreaTP:
@manusa ready to merge?

manusa:
Releasing 6.1.1 ATM 
Once released and checked I'll start merging the pending PRs

manusa (approved):

andreaTP:
Thanks for the heads up @manusa !

manusa:
> WDYT? Should we move on with a separate java-generator issue?

I'm merging this one, so please, do so.

Issue #4350, opened 2022-08-24
@SchemaSwap can only be used once


xRodney:
### Describe the bug

In the [documentation](https://github.com/fabric8io/kubernetes-client/blob/master/doc/CRD-generator.md), it is written that it is "Same as SchemaFrom, but can be applied at any point in the class hierarchy"

I also found in [this PR](https://github.com/fabric8io/kubernetes-client/pull/3797): "The annotation serves to enable the use-case where you want to generate a valid CRD from a model that you don't own.". The PR references [keycloak operator](https://github.com/keycloak/keycloak/pull/9759) as an example of such model.

I believe I have a similar-enough use case:
* In my schema, I use a huge object tree (hundreds of fields in dozens of classes) which I do not own.
* Most of the classes are fine, only at a few places I need to tweak things.
* If I owned the classes, I would have just added a few @SchemaFrom's here and there.

I am led to believe that @SchemaSwap is the way how to achieve the same result without editing external code.
For each @SchemaFrom I would like to have on a field of some class I do not own, I would instead put one @SchemaSwap on the root level.

There are two problems with this:

1. @SchemaSwap is not repeatable. I am only entitled to one @SchemaSwap annotation per class. 
2. If the class and field referenced in @SchemaSwap is used multiple times in the class hierarchy, it is only replaced once.

### Fabric8 Kubernetes Client version

5.12.3

### Steps to reproduce

Generate CRD for this schema:

```
@Group("testresource.example.com")
@Version("v1")
@SchemaSwap(originalType = ExternalObject.class, fieldName = "shouldBeInt", targetType = Integer.class)
//@SchemaSwap(originalType = AnotherObject.class, fieldName = "shouldBeString", targetType = String.class)    // does not compile if uncommented
public class TestResource extends CustomResource<ExternalSpec, Void> implements Namespaced {
}

// I do not own these classes:

class ExternalSpec {
    ExternalObject first;
    ExternalObject second;
    AnotherObject third;
}

class ExternalObject {
    String shouldBeInt;
}

class AnotherObject {
    int shouldBeString;
}
```

It generates:

```
    schema:
      openAPIV3Schema:
        properties:
          spec:
            properties:
              first:
                properties:
                  shouldBeInt:
                    type: integer
                type: object
              second:
                properties:
                  shouldBeInt:
                    type: string
                type: object
              third:
                properties:
                  shouldBeString:
                    type: integer
                type: object
            type: object
          status:
            type: object
        type: object
[...]

PR #4354, opened 2022-08-26
fix: @SchemaSwap can only be used once

xRodney:
## Description

@SchemaSwap is now repeatable, so multiple modifications can be applied from the same resource root.
Fixed a bug when the swap was "used up" after just one usage, meaning that if the referenced type was used multiple times in object hierarchy, only the first one was swapped.

Fixes fabric8io/kubernetes-client#4350

## Type of change

## Checklist


crd-generator/api/src/test/java/io/fabric8/crd/generator/v1/JsonSchemaTest.java
```diff
@@ -178,15 +179,58 @@ void shouldExtractPropertiesSchemaFromExtractValueAnnotation() {
...
   @Test
   void shouldThrowIfSchemaSwapHasUnmatchedField() {
     TypeDef incorrectExtraction = Types.typeDefFrom(IncorrectExtraction.class);
-    assertThrows(IllegalArgumentException.class, () -> JsonSchema.from(incorrectExtraction));
+    IllegalArgumentException exception = assertThrows(IllegalArgumentException.class,
+        () -> JsonSchema.from(incorrectExtraction));
+    assertEquals(
+        "Unmatched SchemaSwaps: @SchemaSwap(originalType=io.fabric8.crd.example.extraction.ExtractionSpec, fieldName=\"FOO\", targetType=io"
+            + ".fabric8.crd.example.extraction.FooExtractor) on io.fabric8.crd.example.extraction.IncorrectExtraction",
+        exception.getMessage());
```
xRodney, line 223: The new implementation no longer differentiates between unmatched class and unmatched field. I my opinion the more detailed message compensates for it.

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
```diff
@@ -239,11 +256,9 @@ private T internalFromImpl(TypeDef definition, Set<String> visited, List<Interna
         continue;
       }
 
-      final PropertyFacade facade = new PropertyFacade(property, accessors, currentSchemaSwaps);
+      ClassRef potentialSchemaSwap = schemaSwaps.lookupAndMark(definition.toReference(), name).orElse(null);
```
andreaTP, line 259: If I correctly follow this change, here you cannot mark just yet the SchemaSwap as used, as it will mark the annotation to be "used" even in this condition:
 - Matched `Class`
 - The Class doesn't have the specified `Property` (e.g. a matching field)
 
 This operation should be performed only after `process` has been called on the `PropertyFacade` (and the `SchemaSwap` has been matched).

crd-generator/api/src/main/java/io/fabric8/crd/generator/InternalSchemaSwaps.java
```diff
@@ -0,0 +1,133 @@
...
+    @Override
+    public String toString() {
+      return new StringJoiner(", ", Key.class.getSimpleName() + "[", "]")
+          .add("originalType=" + originalType)
+          .add("fieldName='" + fieldName + "'")
+          .toString();
+    }
+  }
+
+  public static class Value {
```
andreaTP, line 92: This `Value` class spills out, either we give it a more meaningful name, or we contain it making it `private` and returning a more generic type from `getUnusedSwaps` (e.g. directly the resulting `String` to be used in the error message?)

Also,  given this implementation, `Value` can extend `Key`, I believe.

crd-generator/api/src/test/java/io/fabric8/crd/generator/v1/JsonSchemaTest.java
andreaTP: totally fine to change the message, but I believe that this change of behavior is caused by the fact that we are missing an invariant, see the comment above on matching the field.

crd-generator/api/src/test/java/io/fabric8/crd/generator/v1/JsonSchemaTest.java
```diff
@@ -178,15 +179,58 @@ void shouldExtractPropertiesSchemaFromExtractValueAnnotation() {
     assertNull(barProps.get("baz"));
   }
 
+  @Test
+  void shouldExtractPropertiesSchemaFromSchemaSwapAnnotations() {
```
andreaTP, line 183: what's the difference in between this test and the one extracted in `SchemaSwapSpec` ? Or is it a leftover?

andreaTP (changes requested):
@xRodney Thanks a lot for this PR!
I really like the overall changes that you did 🙂 , I added a few notes to be discussed.

A few additional requests, since you are fixing a bug can you add test cases for the relevant situations where the current implementation is failing (what you already described in the issue):
- `SchemaSwap` should be applied to all occurrences in the hierarchy
- `SchemaSwap` should be applied to deeply nested structures (e.g. 3 or more levels down)
- `SchemaSwap` should fail if the field name is not matched

crd-generator/api/src/test/java/io/fabric8/crd/generator/v1/JsonSchemaTest.java
xRodney: There are two very similar tests: this one in crd-generator/api project, and another in crd-generator/test. 
They both perform the same set of modifications on the same schema, but the difference is that one runs in reflection context and the other on schema generated by Apt. 
In an ideal world, that shouldn't matter, but the sundr library is not yet perfect with annotation arguments that themselves are annotations, and these are sometimes presented as AnnotationRef and sometimes directly as the target annotation (SchemaSwap). Therefore, the code in AbstractJsonSchema needs to compensate for this by using two code paths, which both need to be tested.

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
xRodney: schemaSwapt.lookupAndMark  accepts two arguments, the second being the property name. (Line 253: `String name = property.getName();`)
So I'm not sure I understand your concert?
Btw, I thought the case of "matched class, unmatched field" is being tested in JsonSchemaTest.shouldThrowIfSchemaSwapHasUnmatchedField. This test is still passing and I only extended it with assert on the error message. But the error is still thrown.

xRodney: Of is your concern that the `name` could be overriden by `@JsonProperty` & friends? IMHO it makes more sense for the `fieldName` parameter to reference the fields as is defined in Java, but that's just my opinion.

crd-generator/api/src/main/java/io/fabric8/crd/generator/InternalSchemaSwaps.java
xRodney: I renamed it to `InternalSchemaSwap`.
About extending `Key`, yeah, it's a possibility, but I would rather not do it, because `Key` is an immutable value class, designed to be used as, well, a HashMap key. That has implications on it's `equals` and `hashCode`.
`Value` (or now `InternalSchemaSwap`) is mutable, so either its `equals/hashCode` would have to only use a subset of fields or it could never be used as a map key - but being a subclass of `Key`, any `HashMap<Key, ...>` would allow that. So I chose to not deal with this at all and make it separate.

xRodney:
I have pushed one more test for deeply nested hierarchies, which also tests that the Swap is applied multiple times. 
As for the third case you mentioned, when name is not matched, I believe that case is already covered by a pre-existing test, but let's discuss that in that other thread.

crd-generator/api/src/main/java/io/fabric8/crd/generator/InternalSchemaSwaps.java
andreaTP: fair, but mark `Key` as `final` please.

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
andreaTP: > it makes more sense for the fieldName parameter to reference the fields as is defined in Java, but that's just my opinion.

This makes sense, but, using `name` might not be the name extracted by the `PropertyFacade` constructor:
https://github.com/fabric8io/kubernetes-client/blob/808084e0e348d5b377ac31fd89a979996867d93e/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java#L414-L426

in some cases, the `PropertyFacade` is extracting the field name from constructor/getter/setter 
and this implementation will not respect the extracted name as opposed to the underlying field name.
We can either refactor the logic and use the `PropertyFacade` name extraction logic everywhere or simply document this different behavior.

xRodney: I see, and the only way how that can happen is a `@JsonProperty` on the field itself or getter (get* or is*) or setter. 
I can only say that to me, as the user of the library, would never occur that `fieldName` could reference the name after transformations. 
Maybe it's because the word "field" is used, whose meaning in Java is well defined and is different from the term "property". (Field is a class variable, property is a getter/setter pair - which often is backed by a field). Also the annotation is `@JsonProperty`, not "JsonField".
Or maybe it's because the reason I am dealing with `@SchemaSwaps` in the first place is that I cannot place annotations (like `@JsonProperty`) on the field directly.
So with your permission, I will add Javadocs on `@SchemaSwap`, describing the currently implemented behaviour.

andreaTP: To me, this "limitation" is totally fine, please just make it explicit in this doc:
https://github.com/fabric8io/kubernetes-client/blob/master/doc/CRD-generator.md#iofabric8crdgeneratorannotationschemaswap

mentioning something along the lines:
> the name of the field is restricted to the original `fieldName` and should be backed by a matching concrete field of the matching class. Getters, setters, and constructors are not taken into consideration.

cc. @metacosm to check if he is on board with this decision.

andreaTP (approved):
Thanks again for this effort @xRodney !
LGTM, but let's wait for a little to see if @metacosm wanna chip in.

andreaTP:
@xRodney can you please fix the issues with the CI? `mvn spotless:apply -f crd-generator/pom.xml` should make it automatically.

xRodney:
Hey @andreaTP, thanks to you, too for your help.
I did the reformat and also at the last minute I noticed I forgot to delete the now unused `InternalSchemaSwap` inner class. And that led to realization that I can move the logic of `validateRemainingSchemaSwaps` into the new class `InternalSchemaSwaps`, so its inner class no longer leaks out and can be renamed back to `Value`.
And last but not least, I squashed everything into one commit.
So I'll wait for the CI to finish and for @metacosm's review and we're ready to take off.

xRodney:
Hi @andreaTP, it appears that you need to run the CI for me, with me being a first-time contributor? If so, could you please do it, so that we know if we are in the green?

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
metacosm: I think that should be fine but this should be documented as a breaking change, though…

metacosm (approved):

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
andreaTP: I agree that it "might" be breaking, but, at the same time, this PR is solving some fundamental issues with the implementation and I doubt anyone else is using it just yet.

andreaTP: Maybe we can add a line to the Changelog for reference, wdyt?

xRodney:
I have updated the changelog and added the breaking change as requested. In the latest iteration I have also added a few more tests about the case when targetType is void. And also rebased.
Please, run the CI one more time and if it passes, I think we are ready for merge.

andreaTP:
@xRodney we merged another PR slightly conflicting with the changes in this PR.
Would you mind rebasing? Please keep your JavaDoc style as it's much better 🙂 

Ping me and I will run the CI to validate and eventually move this forward 👍 thanks a lot for your effort!

xRodney:
Sure, it's done

andreaTP:
Unfortunately, the CI failure seems to be legit, @xRodney can you have a look?

xRodney:
Oops, missing import. Sorry about that, it's fixed now.

andreaTP (approved):
LGTM, thanks!

andreaTP:
thanks @xRodney !

PR #4406, opened 2022-09-08
[java gen] - Support generated annotations

### LINKED ISSUE #4384: [java-gen] - Generated Java model should include new specific annotations (min, max, etc.)

[Support for specific annotations](https://github.com/fabric8io/kubernetes-client/pull/4348) (e.g.: `@Max`, `@Min`, `@Pattern`, etc.), was added recently but the Java generator doesn't generate any of such annotations.

The Java generator should take the related keywords into account when they appear in CRDs (e.g.: `maximum`, `minimum`, `pattern`, etc.), and generate the corresponding annotations for them to appear in the Java model.

N/A

### Additional context

See:
* [Introduce specific annotations for the generators](https://github.com/fabric8io/kubernetes-client/pull/4348) and specifically https://github.com/fabric8io/kubernetes-client/pull/4348#issuecomment-1236555434
* [Add support for Min, Max, Pattern and Nullable on CRD generator fields](https://github.com/fabric8io/kubernetes-client/pull/4298)

fabiobrz:
## Description

The changes are related to those in https://github.com/fabric8io/kubernetes-client/pull/4348 
The generator will now generate the expected annotations, i.e. those defined in `io.fabric8.generator.annotation`, the same ones that a reverse CRD generation from the Java model take into account.

Fix #4384 

## Type of change

## Checklist

I can suash the commits as soon as this is ready for merge or even earlier in case we agree on it. 
FYI @andreaTP, CC  @manusa

### DIFF
--- a/java-generator/core/pom.xml
+++ b/java-generator/core/pom.xml
@@ -68,9 +68,8 @@
     </dependency>
 
     <dependency>
-      <groupId>javax.validation</groupId>
-      <artifactId>validation-api</artifactId>
-      <scope>provided</scope>
+      <groupId>io.fabric8</groupId>
+      <artifactId>generator-annotations</artifactId>
     </dependency>
 
     <!-- Testing -->
--- a/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
+++ b/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
@@ -16,16 +16,12 @@
 package io.fabric8.java.generator.nodes;
 
 import com.fasterxml.jackson.databind.JsonNode;
-import com.github.javaparser.ast.expr.Expression;
-import com.github.javaparser.ast.expr.NameExpr;
-import com.github.javaparser.ast.stmt.BlockStmt;
-import com.github.javaparser.ast.stmt.ReturnStmt;
 import io.fabric8.java.generator.Config;
 import io.fabric8.java.generator.exceptions.JavaGeneratorException;
 import io.fabric8.kubernetes.api.model.apiextensions.v1.JSONSchemaProps;
-import io.fabric8.kubernetes.client.utils.Serialization;
 
 import java.util.Locale;
+import java.util.Optional;
 import java.util.function.Function;
 
 import static io.fabric8.java.generator.nodes.Keywords.JAVA_KEYWORDS;
@@ -48,6 +44,22 @@ public abstract class AbstractJSONSchema2Pojo {
   protected final boolean isNullable;
   protected final JsonNode defaultValue;
 
+  protected Double maximum;
+  protected Double minimum;
+  protected String pattern;
+
+  public Double getMaximum() {
+    return maximum;
+  }
+
+  public Double getMinimum() {
+    return minimum;
+  }
+
+  public String getPattern() {
+    return pattern;
+  }
+
   public abstract String getType();
 
   public abstract GeneratorResult generateJava();
@@ -69,11 +81,17 @@ public abstract class AbstractJSONSchema2Pojo {
     return getType();
   }
 
-  protected AbstractJSONSchema2Pojo(Config config, String description, final boolean isNullable, JsonNode defaultValue) {
+  protected AbstractJSONSchema2Pojo(Config config, String description, final boolean isNullable, JsonNode defaultValue,
+      final ValidationProperties validationProperties) {
     this.config = config;
     this.description = description;
     this.isNullable = isNullable;
     this.defaultValue = defaultValue;
+    if (validationProperties != null) {
+      this.maximum = validationProperties.getMaximum();
+      this.minimum = validationProperties.getMinimum();
+      this.pattern = validationProperties.getPattern();
+    }
   }
 
   /** Takes a random string and manipulate it to be a valid Java identifier */
@@ -185,7 +203,17 @@ public abstract class AbstractJSONSchema2Pojo {
     final boolean isNullable = Boolean.TRUE.equals(prop.getNullable());
     switch (nt.getType()) {
       case PRIMITIVE:
-        return new JPrimitive(nt.getName(), config, prop.getDescription(), isNullable, prop.getDefault());
+        return new JPrimitive(
+            nt.getName(),
+            config,
+            prop.getDescription(),
+            isNullable,
+            prop.getDefault(),
+            ValidationProperties.Builder.getInstance()
+                .withMaximum(prop.getMaximum())
+                .withMinimum(prop.getMinimum())
+                .withPattern(prop.getPattern())
+                .build());
       case ARRAY:
         return new JArray(
             fromJsonSchema(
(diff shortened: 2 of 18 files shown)


java-generator/core/pom.xml
```diff
@@ -68,9 +68,8 @@
     </dependency>
 
     <dependency>
-      <groupId>javax.validation</groupId>
```
andreaTP, line 71: Nice! would you mind removing the dependency from the integration tests(and everwhere under the `java-generator` folder) and in the relevant docs?

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
```diff
@@ -69,11 +81,17 @@ protected String getClassType() {
     return getType();
   }
 
-  protected AbstractJSONSchema2Pojo(Config config, String description, final boolean isNullable, JsonNode defaultValue) {
+  protected AbstractJSONSchema2Pojo(Config config, String description, final boolean isNullable, JsonNode defaultValue,
+      final Optional<ValidationProperties> validationProperties) {
```
manusa, line 85: Please, never use Optionals as method parameters, use them only as return types.

The reasoning is very simple, consider I invoke the constructor with `super(theConfig, theDescription, theIsNullable, theDefaultalue, null)`, the NPE problem would still be there.

If you're willing to do this with an Optional, then the suggestion would be to keep the parameter as `ValidationProperties validationProperties`,
and then replace your constructor body with:
```java
Optional.ofNullable(values).ifPresent(values -> {
//...
```
Or better and simpler just use an if not null clause.

fabiobrz: +1, thanks @manusa - fixing.

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
```diff
@@ -229,6 +247,8 @@ public GeneratorResult generateJava() {
               new SingleMemberAnnotationExpr(
                   new Name("com.fasterxml.jackson.annotation.JsonSetter"),
                   new NameExpr("nulls = com.fasterxml.jackson.annotation.Nulls.SET")));
+          // let's also add Fabric8 specific `Nullable` annotation, which will be reflected on generated CRD
```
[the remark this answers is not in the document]
fabiobrz: Will do!

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
```diff
@@ -227,9 +255,74 @@ private static AbstractJSONSchema2Pojo fromJsonSchema(
...
+            prop.getDefault());
       default:
         throw new JavaGeneratorException("Unreachable " + nt.getType());
     }
   }
+
+  /**
+   * This stores the values for a generic JSON schema validation related properties, like `minimum`, `maximum`, etc.
+   */
+  public static class ValidationProperties {
```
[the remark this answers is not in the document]
fabiobrz: Yeah, makes sense. I followed the CRD `AbstractJsonSchema::SchemaPropsOptions` class, but I have nothing against moving it out.

java-generator/core/pom.xml
fabiobrz: Nice catch, thanks. Will do.

fabiobrz:
Hi @andreaTP and @manusa - latest changes pushed to address your review comments. Let me know if you spot anything else.

BTW - can I squash the commit now?

andreaTP (approved):
Thanks a lot for this contribution @fabiobrz !
LGTM!

fabiobrz:
> Thanks a lot for this contribution @fabiobrz ! LGTM!

Welcome @andreaTP - squashing my commits then!

fabiobrz:
Hi @manusa - the e2e tests failed after I squashed my commits, although it was green already before that.
I'll try to re-spin a run, but IMHO it should be an intermittent/random failure.

andreaTP:
cc. @manusa looking forward for your input here 🙏

manusa (approved):
LGTM, thx!

I think it would be worth adding a note on the the breaking changes list, since some users who might be using the Java generator, rely on the javax validation annotations that are no longer added.

This also raises the question: `as a developer relying on the java generator, how can I further transform the generated code to replace fabric8-fabric8 specific annotations with annotations that might be usedeful for my project (e.g. javax.validation)?`

andreaTP:
> worth adding a note on the the breaking changes list

I agree, just want to note that the Java-Generator is still in "preview", so breaking changes are fine there.
Maybe `6.2` is a good candidate to get the Java Generator out of the Preview?

Regarding your additional question, we should come up with something for sure ... 😓 I have a couple of half-baked ideas in mind but am not convinced yet, open to suggestions!

fabiobrz:
> I think it would be worth adding a note on the the breaking changes list, since some users who might be using the Java generator, rely on the javax validation annotations that are no longer added.

Hi @manusa - thanks for your feedback, will do.

> This also raises the question: as a developer relying on the java generator, how can I further transform the generated code to replace fabric8-fabric8 specific annotations with annotations that might be usedeful for my project (e.g. javax.validation)?

I've been discussing this shortly with @andreaTP as well last week.
Yesterday I was thinking about it and I came to the conclusion that maybe we'd want to create one issue for supporting the generation of validation constraints annotations per each framework type.

This would allow for them to be worked on asynchronously and for the support to different framework to be released in steps.
Maybe we could start with supporting the `javax.validation.*` as a PoC, in order to evaluate whether the proposal is actually fitting.

WDYT? CC @andreaTP

andreaTP:
> for supporting the generation of validation constraints annotations per each framework type

This is an interesting idea, but I'm afraid to add this complexity to the core `java-generator` project.

I'm thinking about adding the possibility to arbitrarily customize the generated code before emitting it, offering a Visitor or Traverse pattern that can perform generic post-processing.

On top of that mechanism, we can implement separate plugins for each of the features we want to support, and they can possibly live in separate repositories/projects.

Note: the described mechanism might be used as a foundation to implement optimizations, such as class deduplication and similar.

fabiobrz:
> > for supporting the generation of validation constraints annotations per each framework type
> 
> This is an interesting idea, but I'm afraid to add this complexity to the core `java-generator` project.
> 
> I'm thinking about adding the possibility to arbitrarily customize the generated code before emitting it, offering a Visitor or Traverse pattern that can perform generic post-processing.
> 
> On top of that mechanism, we can implement separate plugins for each of the features we want to support, and they can possibly live in separate repositories/projects.
> 
> Note: the described mechanism might be used as a foundation to implement optimizations, such as class deduplication and similar.

Thanks @andreaTP! 
+1 on loading the implementation dynamically, i.e. prepare the foundation for plugins. It makes sense to me and I actually thought of it as well, in relation to the complexity increase for the `core` module.

fabiobrz:
>  I think it would be worth adding a note on the the breaking changes list

Done, should be ready fr merge as soon as CI checks pass.

andreaTP:
Thanks!

PR #4409, opened 2022-09-08
Raw Handling

shawkins:
## Description
Here's a draft of consistently handling the raw type as object as described in #4289.  This is as far as things will generate before you hit some compile errors as the buildable logic won't add the relevant subtype methods like it would for HasMetadata.  Without changes to sundrio this is more breaking that I would have hoped for - ideally the subtypes would only include KubernetesResources.

I'm not sure how viable it would be (either at the pojo generation level or sundrio), but the next attempt would be to map to KuberentesResource but also add a generated method that allow the use of Map.

cc @manusa @iocanel 

## Type of change

 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [ ] Feature (non-breaking change which adds functionality)
 - [x] Breaking change (fix or feature that would cause existing functionality to change
 - [x] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [x] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [ ] I Added [CHANGELOG](https://github.com/fabric8io/kubernetes-client/blob/master/CHANGELOG.md) entry regarding this change
 - [ ] I have implemented unit tests to cover my changes
 - [ ] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/master/doc/CHEATSHEET.md) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [ ] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

PR #4414, opened 2022-09-13
using kubernetesresource as the raw type

shawkins:
## Description
This is an alternative to #4409 that uses KubernetesResource as the raw type, rather than object.

There are a couple of nuances.  

- for a little better backwards compatibility for models using Map<String, Object> it would be nice if RawExtension implemented Map<String, Object> - however that causes a problem with sundrio and results in build errors.  So the usage would need to be RawExtension.fromMap(...) https://github.com/fabric8io/kubernetes-client/compare/master...shawkins:kubernetes-client:raw_kuberesource?expand=1#diff-97d49d0a6ea3655ffb6a17863e03105b6a4822eade197fcf217ba5594d33d38dR73 - or new RawExtension(map), which would also be useful if RawExtension is added to the array of buildable resources.

- There are also sundrio build errors if all generated types remain KubernetesResources, so these changes see if the resource has a kind field to be considered a KubernetesResources.  If it does not, then we'll just use serializable instead.  That does not affect deserialization as those resources can't be deserialized by the KubernetesDeserializer.    However this change touches quite a few generated classes and could have other considerations around the usage of the KuberentesResource interface.

To minimize the changes / breakage I'll back out the changes to migrate from the Map<String, Object> to KuberentesResource and save those for a major release.

## Type of change

## Checklist


shawkins:
@manusa This commit shows a minimum amount of changes / breakage needed to non-generated files.  The commit to the generated will be quite large, so I'm leaving that out for now - I'm also not able to generate all the extensions.  For example knative fails with:

```
# knative.dev/pkg/changeset
vendor/knative.dev/pkg/changeset/commit.go:57:24: info.Settings undefined (type *debug.BuildInfo has no field or method Settings)
make: *** [Makefile:28: gobuild] Error 2
```

Not sure yet what is going on there.  After seeing which ones currently are actually using the Map<String, Object> for raw, I can further update the generate.go files.

To recap these changes are:
- Change the default mapping of raw from HasMetadata to KubernetesResource.  Unlike the Object mapping change, this preserves the builder methods like NamedExtensionFluent.withNewXXXExtension.  This corrects the Serialization logic when building a list of results, and allows for things that aren't HasMetadata to be used in raw locations.
- Introduces a kubernetes RawExtension, which is a replacement for the openshift RawExtension, that is used when deserializing a value that has no kind/apiVersion.  Not included yet in these changes is adding RawExtension to the Buildable array - that would allow buildable methods like withNewRawExtensionXXX and we may also want to add a Map based constructor as well for convenience.  It might be good for RawExtension to implement Map<String, Object>, but that seems to cause problems for sundrio.
- To satisfy sundrio objects that have no kind are no longer HasMetadata, but instead just serializable.  This does not affect what buildable methods are available, nor deserialization.  Not sure who else may be looking for this marker interface.

What they don't do:
- Change the existing raw mappings.  Map<String, Object> and the openshift RawExtension still exist.  All existing code should work except for a couple of narrow cases - like with KubernetesList / KubernetesBaseList typing, but we could workaround that more such as changing the constructors to use List<? extends KubernetesResource>.

What could be done differently:
- As Marc suggested all generated classes could extend RawExtension and the mapping could be changed from KubernetesResource to RawExtension.  However this is also a potentially breaking change - at the very least requiring users to regenerate or update custom model classes to the in hierarchy.

Migrating away from directly supporting map based on these changes would be a breaking change - the user would be expected to use RawExtension instead.  The biggest usability issue there is that the getter will become typed as KubernetesResource rather than Map.

shawkins:
Thanks @rohanKanojia it turned out to be the go version.  knative requires 1.18+.

Updated to pare back where KubernetesResource is used - many of the locations where the object model specifies raw can really only have typed values.  Since we don't differentiate between typed and HasMetadata, the most appropriate mapping is to still use HasMetadata.  This was the conclusion for BaseKubernetesList - it also applies to Template (you can create a Template with an untyped raw value, but you can't do anything with it).  

So to do this more thoroughly we'd need to evaluate on a case by case basis if the raw mapping actually accept untyped values, like NamedExtension.  Presumably all the places that were using Map and the previous RawExtension class, the answer is yes.  Unfortunately there's nothing to indicate this ahead of time or that would be easy to capture in the generation logic - it would seemingly take special case handling like was being done for the image raw extension.

This is obviously more involved that I had intended, so I'll probably take one more pass over it to see how to further minimize things.

manusa:
It seems there was an accidental commit (https://github.com/fabric8io/kubernetes-client/commit/7746090525604ba3e2af3470c4794b9ea4e5a2c6) of the vendor directory that you are reverting now.

It would be better if we separated that reversion to a separate commit or even a different PR. Then we can rebase this one so that the changes are clearer.

shawkins:
@andreaTP @manusa this introduces an AnyType class that is a base for IntOrString and RawExtension (which is still a KubernetesResource).  Alternatively we could leave IntOrString alone and just introduce AnyType, but also make it a KubernetesResource.

For now it is just an Object wrapper with some javadocs over the expected getValue types.  Are there additional helper methods that we would want?

andreaTP:
@shawkins thanks a lot for taking care of this, appreciated!

Regarding having `IntOrString` that extends `AnyType` I don't find in the spec any hint, I would not do it at first, but it's more of a gut feeling other than having any proof that it's the correct encoding, so not fussy on it.

Regarding helper functions I think that a `from`/`to` `JsonNode` would be a good start to cover a bunch of use cases.

shawkins:
> Regarding having IntOrString that extends AnyType I don't find in the spec any hint, I would not do it at first, but it's more of a gut feeling other than having any proof that it's the correct encoding, so not fussy on it.

It's related more to the Java implementation rather than any explicit schema hierarchy.  AnyType is simply and object wrapper - which is what IntOrString is as well.  IntOrString is just type restrictions layered on AnyType.

> Regarding helper functions I think that a from/to JsonNode would be a good start to cover a bunch of use cases.

Without additional restructuring I think those functions would need to go in same module as Serialization - so that we can reuse the same mapping logic.  

Converting from JsonNode should be unnecessary - any object, including JsonNode, is a valid value.  That can be called out in the JavaDocs.

For converting to JsonNode that would currently be Serialization.jsonMapper().valueToTree(object) - but we may not want to encourage that static usage as we'll ideally encapsulate that better.

kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/utils/SerializationTest.java
```diff
@@ -242,10 +244,10 @@ void unmarshalWithInvalidYamlShouldThrowException() {
   }
 
   @Test
-  @DisplayName("unmarshal, with invalid YAML resource, should throw exception")
-  void unmarshalWithInvalidResourceShouldThrowException() {
```
manusa, line 246: This reverts #4263 / #4206

Are we sure we want to proceed with this change? In which cases are we going to use our Serialization tooling to unmarshal an "invalid"/incomplete resource?

shawkins: > This reverts https://github.com/fabric8io/kubernetes-client/pull/4263 / https://github.com/fabric8io/kubernetes-client/issues/4206

The main qualm addressed with throwing an exception with 4263 was not returning an untyped GenericKubernetesResource, which is also HasMetadata.  Here we're returning a RawExtension - so there's no ambiguity.

manusa: :thinking: 

I feel that the following methods are still ambiguous:
```java
public static <T> T unmarshal(InputStream is) {
// ...
public static <T> T unmarshal(String str) {
```

We probably need to clean up this class (other PR/Issue).

I see value in a `Serialization.unmarshall` field that can return any instance of HasMetadata, list of HasMetadata, or maybe even KubernetesList, but only restricted for these types.

The current behavior for the referenced methods is sort of unpredictable.

shawkins: > I feel that the following methods are still ambiguous:

That's a different level of ambiguity.  What I'm mostly trying to do is remove special case handling.  If we have no type specified, then RawExtension makes just as much sense as throwing an exception - but allows us to remove the specialized check in Serialization.

> I see value in a Serialization.unmarshall field that can return any instance of HasMetadata, list of HasMetadata, or maybe even KubernetesList, but only restricted for these types.

That seems fine to add - Serialization.unmarshallHasMetadata or another signature with a hasmetadata flag.

Related to this - a yaml list of HasMetadata is currently problematic.  The root unmarshall method conversion from yaml only accounts for an object root - an array will fail with a class cast.

manusa (approved):

PR #4425, opened 2022-09-16
refactor: JsonSchema remove preserveSelfUnknownFields

manusa:
## Description
refactor: JsonSchema remove preserveSelfUnknownFields

relates to: #4398

## Type of change

## Checklist

### DIFF
--- a/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
+++ b/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
@@ -120,24 +120,24 @@ public abstract class AbstractJsonSchema<T, B> {
   }
 
   protected static class SchemaPropsOptions {
-    final Optional<Double> min;
-    final Optional<Double> max;
-    final Optional<String> pattern;
+    final Double min;
+    final Double max;
+    final String pattern;
     final boolean nullable;
     final boolean required;
 
     final boolean preserveUnknownFields;
 
     SchemaPropsOptions() {
-      min = Optional.empty();
-      max = Optional.empty();
-      pattern = Optional.empty();
+      min = null;
+      max = null;
+      pattern = null;
       nullable = false;
       required = false;
       preserveUnknownFields = false;
     }
 
-    public SchemaPropsOptions(Optional<Double> min, Optional<Double> max, Optional<String> pattern,
+    public SchemaPropsOptions(Double min, Double max, String pattern,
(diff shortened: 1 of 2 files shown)


crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
```diff
@@ -543,13 +522,7 @@ public Property process() {
...
         }
 
-        if (p.isPreserveUnknownFields()) {
-          preserveUnknownFields = true;
-        }
-
-        if (p.isPreserveSelfUnknownFields()) {
-          preserveSelfUnknownFields = true;
-        }
+        preserveUnknownFields = p.isPreserveUnknownFields();
```
andreaTP, line 525: If I remember correctly this method can be called multiple times on `object`s, so that the last property that affect the object may reset the flag, the gut feeling is to leave the `if`:

```java
if (p.isPreserveUnknownFields()) {
  preserveUnknownFields = p.isPreserveUnknownFields();
}
```

but haven't checked the code carefully.

andreaTP (commented):
Thanks a lot, @manusa for the cleanup!
One little note and LGTM.

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
manusa: I just removed the if-else clause because it seemed redundant, but what you say is true. I'll revert this change

andreaTP: Looking at it again, and the same comment might apply to the other properties as well (min, max, pattern). 👍

manusa: You're right. I changed them, please check again :pray:

andreaTP: This encoding works for me 👍 
I'm negatively surprised that the test suite is not able to catch a regression here 🙁

andreaTP (approved):
Given a green CI LGTM! Thanks again!

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
manusa: A next step would be to implement tests that verify the described expected behaviors (multipass processing, etc.)

PR #4429, opened 2022-09-19
[java-gen] Preserve unknown without type should translate to AnyType

### LINKED ISSUE #4426: how to deal with the crd field x-kubernetes-preserve-unknown-fields
### Describe the bug

I followed the instructions of [Generating Java from CRD](https://github.com/fabric8io/kubernetes-client/blob/master/doc/java-generation-from-CRD.md) and used Dapr's [Component CRD](https://github.com/dapr/dapr/blob/master/charts/dapr/crds/components.yaml)  to generate Java source code.

The CRD has a `value` field that uses `x-kubernetes-preserve-unknown-fields: true`, and the generated source code is as follows:
```
@com.fasterxml.jackson.annotation.JsonInclude(com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL)
@com.fasterxml.jackson.annotation.JsonPropertyOrder({})
@com.fasterxml.jackson.databind.annotation.JsonDeserialize(using = com.fasterxml.jackson.databind.JsonDeserializer.None.class)
public class Value implements io.fabric8.kubernetes.api.model.KubernetesResource {

    @com.fasterxml.jackson.annotation.JsonIgnore()
    private java.util.Map<String, Object> additionalProperties = new java.util.HashMap<>();

    @com.fasterxml.jackson.annotation.JsonAnyGetter()
    public java.util.Map<String, Object> getAdditionalProperties() {
        return additionalProperties;
    }

    @com.fasterxml.jackson.annotation.JsonAnySetter()
    public void setAdditionalProperties(java.util.Map<String, Object> additionalProperties) {
        this.additionalProperties = additionalProperties;
    }
}
```

But according to the component example given by Dapr, the value of the `value` field can be String, int, or other:
```
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: messagebus
spec:
  type: pubsub.mqtt
  version: v1
  metadata:
    - name: consumerID
      value: "{uuid}"
    - name: url
      value: "tcp://admin:public@localhost:1883"
    - name: qos
      value: 1
    - name: retain
      value: "false"
    - name: cleanSession
      value: "false"

```
So like **value: "tcp://admin:public@localhost:1883"**, how should I set it to Value Class?

6.1.1

### Steps to reproduce

...

### Expected behavior

...

### Runtime

Kubernetes (vanilla)

1.22

### Environment

Linux

_No response_

### Additional context

_No response_

andreaTP:
## Description
Resolves #4426 

The new encoding seems to be better aligned with the rules:
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 * Fix #4459: Fixed OSGi startup exceptions while using KubernetesClient/OpenShiftClient
 * Fix #4482: Fixing blocking behavior of okhttp log watch
 * Fix #4473: Fix regression in backoff interval introduced in #4365
-* Fix #4426: [java-generator] Encode a `JsonNode` instead of an Object if `x-kubernetes-preserve-unknown-fields` is present and the type is null.
+* Fix #4426: [java-generator] Encode an `AnyType` instead of an Object if `x-kubernetes-preserve-unknown-fields` is present and the type is null.
 
 #### Improvements
 * Fix #4348: Introduce specific annotations for the generators
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Fix #4459: Fixed OSGi startup exceptions while using KubernetesClient/OpenShiftClient
 * Fix #4482: Fixing blocking behavior of okhttp log watch
 * Fix #4473: Fix regression in backoff interval introduced in #4365
+* Fix #4426: [java-generator] Encode a `JsonNode` instead of an Object if `x-kubernetes-preserve-unknown-fields` is present and the type is null.
 
 #### Improvements
 * Fix #4348: Introduce specific annotations for the generators
--- a/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
+++ b/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
@@ -148,7 +148,7 @@ public abstract class AbstractJSONSchema2Pojo {
       return fromJsonSchema.apply(JPrimitiveNameAndType.INT_OR_STRING);
     } else if (type == null
         && Boolean.TRUE.equals(prop.getXKubernetesPreserveUnknownFields())) {
-      return fromJsonSchema.apply(JPrimitiveNameAndType.JSON_NODE);
+      return fromJsonSchema.apply(JPrimitiveNameAndType.ANY_TYPE);
     } else if (prop.getEnum() != null && prop.getEnum().size() > 0) {
       return fromJsonSchema.apply(new JEnumNameAndType(key));
     } else {
--- a/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
+++ b/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
@@ -148,7 +148,7 @@ public abstract class AbstractJSONSchema2Pojo {
       return fromJsonSchema.apply(JPrimitiveNameAndType.INT_OR_STRING);
     } else if (type == null
         && Boolean.TRUE.equals(prop.getXKubernetesPreserveUnknownFields())) {
-      return fromJsonSchema.apply(new JObjectNameAndType(key));
+      return fromJsonSchema.apply(JPrimitiveNameAndType.JSON_NODE);
     } else if (prop.getEnum() != null && prop.getEnum().size() > 0) {
       return fromJsonSchema.apply(new JEnumNameAndType(key));
     } else {
--- a/java-generator/it/src/it/preserve-unknown-objects/src/test/java/io/fabric8/it/certmanager/TestPreserveUnkonwnUsability.java
+++ b/java-generator/it/src/it/preserve-unknown-objects/src/test/java/io/fabric8/it/certmanager/TestPreserveUnkonwnUsability.java
@@ -1,0 +1,108 @@
+/**
+ * Copyright (C) 2015 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.fabric8.it.certmanager;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.IntNode;
+import com.fasterxml.jackson.databind.node.TextNode;
+import io.dapr.v1alpha1.Component;
+import io.dapr.v1alpha1.ComponentSpec;
+import io.dapr.v1alpha1.componentspec.Metadata;
+import io.fabric8.kubernetes.api.model.ObjectMeta;
+import io.fabric8.kubernetes.client.utils.Serialization;
+import org.junit.jupiter.api.Test;
+import io.fabric8.java.generator.testing.KubernetesResourceDiff;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.Files;
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class TestPreserveUnknownUsability {
+
+  @Test
+  void testDeserialization() {
+    // Arrange
+    Component sample =
+      Serialization.unmarshal(getClass().getResourceAsStream("/sample.yaml"), Component.class);
+
+    // Act
+    List<Metadata> metadataList = sample.getSpec().getMetadata();
+
+    // Assert
+    assertEquals(5, metadataList.size());
+  }
+
+  private Component createSampleComponent() throws Exception {
+    Component component = new Component();
+    ComponentSpec spec = new ComponentSpec();
+    spec.setType("pubsub.mqtt");
+    spec.setVersion("v1");
+
+    Metadata meta1 = new Metadata();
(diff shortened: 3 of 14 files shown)


andreaTP:
cc. @fabiobrz would love your review here 🙂

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
```diff
@@ -140,7 +140,7 @@ public static AbstractJSONSchema2Pojo fromJsonSchema(
       return fromJsonSchema.apply(JPrimitiveNameAndType.INT_OR_STRING);
     } else if (type == null
         && Boolean.TRUE.equals(prop.getXKubernetesPreserveUnknownFields())) {
-      return fromJsonSchema.apply(new JObjectNameAndType(key));
+      return fromJsonSchema.apply(JPrimitiveNameAndType.JSON_NODE);
```
fabiobrz, line 143: By doing this, we're using `JsonNode` abstraction as a part of the POJO, which might sound a little bit confusing. There's another use case in which we do this, partially, i.e. when encoding default values, but that is limited to the intermediate representation, rather than spanning to the actual final POJO. 
I understand this to be a better choice to be compliant with the spec, so basically I do agree. 
The only question left in my mind would actually be whether you've considered any other alternatives and what the rationale is for ending up with `JsonNode`.

java-generator/it/src/it/preserve-unknown-objects/src/test/java/io/fabric8/it/certmanager/TestPreserveUnkonwnUsability.java
```diff
@@ -0,0 +1,108 @@
...
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.Files;
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class TestPreserveUnknownUsability {
```
fabiobrz, line 37: This is an integration test, and I am fine with it, but shouldn't we add an approval test too? I see that the only CRD that would work in such case - apart the Dapr one in this PR - is the kamelet one. Maybe that would be the one to be used? WDYT?

java-generator/it/src/it/preserve-unknown-objects/src/test/java/io/fabric8/it/certmanager/TestPreserveUnkonwnUsability.java
```diff
@@ -0,0 +1,108 @@
...
+    spec.setType("pubsub.mqtt");
+    spec.setVersion("v1");
+
+    Metadata meta1 = new Metadata();
+    meta1.setName("consumerID");
+    meta1.setValue(new TextNode("{uuid}"));
+
+    Metadata meta2 = new Metadata();
+    meta2.setName("url");
+    meta2.setValue(new TextNode("tcp://admin:public@localhost:1883"));
```
fabiobrz, line 64: Just to add to my comment regarding the chosen implementation strategy, here - and in the other `set(...)` statements below - is where the mixing between the JSON definition and the Java model caught my eye.

fabiobrz (commented):
Hi @andreaTP, and thanks for this PR.
I dropped some comments, the main one being basically a request for clarification about the implementation strategy, and how to make it clear and visible for a better user experience.

Feel free to let me know what you think.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 #### Bugs
 * Fix #4369: Informers will retry with a backoff on list/watch failure as they did in 5.12 and prior.
 * Fix #4350: SchemaSwap annotation is now repeatable and is applied multiple times if classes are used more than once in the class hierarchy.
+* Fix #4426: [java-generator] Encode a `JsonNode` instead of an Object if `x-kubernetes-preserve-unknown-fields` is present and the type is null.
```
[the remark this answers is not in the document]
andreaTP: I aligned the handling of this situation in light of the better understanding of the spec:
https://github.com/fabric8io/kubernetes-client/pull/4429/commits/b9f5d4d19305f5967fca80aac9d015b3e50af188

The missing part was to remove the `type` in the case of `JsonNode` (it was `object`).

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
andreaTP: Great question @fabiobrz !

Let me try to answer thoroughly:
I have re-read the spec again and now I understand that in case of `type == null` and `x-kubernetes-preserve-unknown-fields` we should accept any value, including primitive types, arrays, maps and objects.

In Java this "can" be represented using an "object" and "JsonUnwrapped", but, IMHO this encoding is really awkward and has a number of downsides (e.g. it spills in the parent, depends on a specific Jackson functionality etc.)

Another alternative would be to map the field to `java.lang.Object` which gives us a "plain" Java construct without any guarantees.

The encoding with `JsonNode` gives us something that can be serialized and should conform to a valid Json which is all we can say about this field.
The downside is that we have little to no control and we leave everything in the hands of the final user, but, I can argue that there is almost nothing that can be done given the lack of guarantees.

java-generator/it/src/it/preserve-unknown-objects/src/test/java/io/fabric8/it/certmanager/TestPreserveUnkonwnUsability.java
andreaTP: You are 100% right, having to deal with `JsonNode` things is ugly and error-prone, but I do not have a better encoding that guarantees the necessary expressiveness and improves the Java integration.
Any idea, alternative, or suggestion would be highly appreciated!

CHANGELOG.md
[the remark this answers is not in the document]
fabiobrz: Thanks for addressing my comment, @andreaTP. I was actualy more concerned about tracking it by oggin a new issue, than addressing it here. 
For sure, the CRD generator would be aligned if we let the change go in, but I see the checks about test coverage and duplication aren't passing after it was added. 
Hence, the decision here seems to be about whether to keep this change, and make the CI checks pass, or to add it to a separate PR with its own tracker issue. 

This could also be the opportunity for discussing and defining a best practice for the upcoming CRD-gen and Java-gen fixes and enhancements.

WDYT?

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
fabiobrz: Thanks @andreaTP - I see what your reasoning is and I do agree on what you said about `JsonUnwrapped`.
For sure `Object` would offere less guarantees than `JsonNode`, although as said, it looks awkward as well how it mixes with the POJO representation, and this was my main concern.
We could let the change go in as-is, and agree on filing an issue for evaluating alternatives to improve things in the future, although it would break the APIs.

WDYT?

java-generator/it/src/it/preserve-unknown-objects/src/test/java/io/fabric8/it/certmanager/TestPreserveUnkonwnUsability.java
fabiobrz: +1, thanks, this can be resolved.

java-generator/it/src/it/preserve-unknown-objects/src/test/java/io/fabric8/it/certmanager/TestPreserveUnkonwnUsability.java
fabiobrz: At first, when I looked to the original poster issue, I immediately though about `Map`. But I must admit I haven't looked enough into the implications for this kind of solution.
ATM, is my opinion as well that the `JsonNode` based solution is the best tradeoff.

andreaTP: 👍

CHANGELOG.md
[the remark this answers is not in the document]
andreaTP: The SonarCloud checks should be taken "cum grano salis" and is not a mandatory check for merging 🙂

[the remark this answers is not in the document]
fabiobrz: Right then! :rocket:

fabiobrz:
Thanks @andreaTP, LGTM!

fabiobrz (approved):

andreaTP:
Thanks @shawkins for chipping in this discussion!
It's indeed related, we should discuss this better.

Quick note that in this PR we are fixing a usability issue where `Map<String, Object>`(and equivalent) doesn't allow to represent primitive types that are instead allowed.

rohanKanojia (approved):

andreaTP:
This PR is waiting for `AnyType` to be introduced in #4414 and changing the implementation to use it instead of `JsonNode`.

andreaTP:
Rebased to use the new `AnyType`, should be ready for review.

java-generator/it/src/it/preserve-unknown-objects/src/test/java/io/fabric8/it/certmanager/TestPreserveUnkonwnUsability.java
```diff
@@ -0,0 +1,112 @@
...
+      Serialization.unmarshal(getClass().getResourceAsStream("/sample.yaml"), Component.class);
+
+    // Act
+    List<Metadata> metadataList = sample.getSpec().getMetadata();
+
+    // Assert
+    assertEquals(5, metadataList.size());
+  }
+
+  private Component createSampleComponent() throws Exception {
```
andreaTP, line 53: @shawkins this test also validates your good work! Thanks!

shawkins: @andreaTP thanks for picking this up

andreaTP:
cc. @manusa

PR #4443, opened 2022-09-23
Add javax.Generated annotation to classes generated by the java-generator

### LINKED ISSUE #4351: Double generation from CRD to Java and back to CRD
### Describe the bug

I have the use case that I need to both generate CDR from Java, as well as generating Java from CDR. So I have both the annotation processor and maven plugin configured.

From what I can see, the maven plugin runs first and it generates Java code, and then the annotation processor kicks in, which totally makes sense because before compilation code needs to be generated.

The result is a roundtrip happens for the CDR and I get the generated one under `/META-INF/fabric8` along with the original one in a different place. This is not a real problem, but could be confusing.

6.0.0

#### Configure annotation processor

```xml
    <dependency>
      <groupId>io.fabric8</groupId>
      <artifactId>crd-generator-apt</artifactId>
      <scope>provided</scope>
    </dependency>
```

#### Configure maven plugin

```xml
      <plugin>
        <groupId>io.fabric8</groupId>
        <artifactId>java-generator-maven-plugin</artifactId>
        <executions>
          <execution>
            <goals>
              <goal>generate</goal>
            </goals>
          </execution>
        </executions>
        <configuration>
          <source>src/main/resources/kubernetes/crds</source>
        </configuration>
      </plugin>
```

#### Compile

`mvn compile`

### Expected behavior

Maybe there is a way to configure the annotation processor to skip generated code that I'm not aware of. The expected behaviour is not to generate CDR form generated code.

### Runtime

Kubernetes (vanilla)

1.24.2@latest

### Environment

macOS

```shell
NA
```

### Additional context

For a reason I have to use `5.12.3` annotation processor, but `6.0.0` maven plugin. Not sure whether that matters.

andreaTP:
## Description
Fix #4351

Emitting the annotation can be disabled using a hidden flag.
In the integration tests, this functionality is always disabled since it's not available in Java 8 (introduced in Java 9).

## Type of change

## Checklist


java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
```diff
@@ -39,6 +43,10 @@ public abstract class AbstractJSONSchema2Pojo {
   static final String OBJECT_CRD_TYPE = "object";
   static final String ARRAY_CRD_TYPE = "array";
 
+  public static final AnnotationExpr GENERATED_ANNOTATION = new SingleMemberAnnotationExpr(
+      new Name("javax.annotation.processing.Generated"),
+      new StringLiteralExpr("io.fabric8.java-generator"));
```
manusa, line 48: I think that the Generated annotation expects a fully qualified name of the Generator class (not sure what this might be in this context `CRGeneratorRunner.class.getName()` :shrug: )

andreaTP: You are very right ... 🤷 
Done!

fabiobrz (approved):
Hi @andreaTP - I have reviewed the changes in here, which LGTM, thanks!

honnix:
👍

PR #4466, opened 2022-09-30
feat: let users decide whether or not to close a watcher on exception

metacosm:
## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Watcher.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Watcher.java
@@ -43,6 +43,16 @@ public interface Watcher<T> {
    */
   void onClose(WatcherException cause);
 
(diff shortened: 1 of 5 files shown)


shawkins:
@metacosm This does not seem needed.  What problem is this addressing?  There are only three cases that close a watch:
- http gone, which has handling.  Any other explicit errors from kubernetes will trigger a retry, not a close.
- exhausting network retries, which the informer disables
- the event is unprocessable - this is now for specific reasons, invalid object, class cast, json processing exception.  There is no recovery action expected of a Watcher and certainly not by an Informer - the user would need to manipulate the cache, etc. to ensure consistency.  I don't think we should try to facilitate this - and it's not supported by the go client either.

metacosm:
Seems like the solution to only react after the fact is satisfactory enough at the moment, and, as pointed out by @shawkins, there are issues with not stopping the informer on unrecoverable errors, so closing this.

PR #4488, opened 2022-10-10
fix #4408: allowing for users to set the exception handling behavior

### LINKED ISSUE #4408: Informer initial start exception handling

The startup of an informer is treated differently than exception handling once it is running.  On startup any problem with the initial list or watch will be treated as a hard failure, while later on it will keep retrying.  See also #4365 

To conform with the go client we should let the informer simply keep retrying from the start.  If the user wants to know if it has ever started, there's the hasSynced method, or we can add another CompletableFuture similar to stopped.

We could apply this change to Informers stated by runnableInformer without much of a risk of compatibility issues.  All other paths though the user is probably expecting an informer that has already synced to be returned.

_No response_

### Additional context

_No response_

shawkins:
## Description
Fix #4408 

cc @csviri 

I couldn't come up with a good way to deal with this all under the covers, so it seemed best to introduce an explicit handler with a signature that's sufficient to cover our default behavior.  This makes the final solution much closer to some of the earlier proposals.  For the method:

```
boolean retryAfterException(boolean isStarted, Throwable t);
```
our current behavior is:

```
(b, t) -> b && !(t instanceof WatcherException)
```
That is we'll retry once started and only if it's not a WatcherException.  

The expectation here is that if you use the runnableInformer method, you can change this behavior by calling something like exceptionHandler((b, t) -> true)

In reviewing the go implementation, I see the following differences:

- Their WatchErrorHandler is doc'd as only informational - however the reflector is passed in, so you can easily kill it if you wanted.
- They will always retry, so the WatchErrorHandler method does not return anything.
- The equivalent of a watch onClose exception is not passed to their WatchErrorHandler.  I think their assumption is that you'll immediately be trying a fresh list / watch and then you'll experience a problem with one of those calls if something is still wrong. 

I've added a helper method to ExceptionHandler to check for deserialization issues - but that is not currently used in any built-in way.   If you use something like exceptionHandler((b, t) -> true), but there is a jsonprocessing exception with the list call or with a watch event the result would be that you'll just keep retrying - a fresh list operation, then watch.  If you don't want to make an assumption that things will self heal (a missing conversion webhook is installed later), then ExceptionHandler.isDeserializationException could be used - exceptionHandler((b, t) -> !ExceptionHandler.isDeserializationException(t))

With a major release we can then determine if the default behavior should become more generally to just keep retrying.

Along with this is a small breaking change to make the return type for start and stopped to be a CompletionStage rather than a CompletableFuture - I realized that since we don't want the user completing and don't support canceling, we should be using the narrower type instead.  Since the stopped method was just introduced we may be able to change that one without too much concern (that would also need backported).  If anyone has any qualms with start changing as well, I'll revert that.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/impl/SharedInformerFactoryImpl.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/impl/SharedInformerFactoryImpl.java
@@ -53,7 +53,7 @@ public class SharedInformerFactoryImpl implements SharedInformerFactory {
   private String name;
   private String namespace;
 
-  private KubernetesClient client;
+  private final KubernetesClient client;
 
   public SharedInformerFactoryImpl(KubernetesClient client) {
     this.client = client;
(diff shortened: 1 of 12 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/impl/SharedInformerFactoryImpl.java
```diff
@@ -106,11 +107,11 @@ public synchronized <T> SharedIndexInformer<T> getExistingSharedIndexInformer(Cl
 
   @Override
   public synchronized Future<Void> startAllRegisteredInformers() {
-    List<CompletableFuture<Void>> startInformerTasks = new ArrayList<>();
+    List<CompletionStage<Void>> startInformerTasks = new ArrayList<>();
```
manusa, line 110: This doesn't seem to be compliant with line 128.
I'm going to add a subsequent commit to ensure this won't cause any issue (there shouldn't be any issue because the underlying impl will be CompletableFuture, but the code is wrong nonetheless).

manusa (approved):
LGTM, thx!

PR #4490, opened 2022-10-11
improving how httpclients can be configured 

### LINKED ISSUE #4471: KubernetesClientBuilder enhancements for common HttpClient customizations
> First of all, I should not have to create Kubernetes client first so that I can get http client builder to replace interceptor and then get http client factory and finally create Kubernetes clients from builder using that http client factory. This is pretty messy builder pattern implementation from API point of view.

That's obviously not intended.  There are really two issues here.  The first is that there's no current way exposed for getting the default HttpClient Factory.  A method for that can easily be extracted from HttpClientUtils.createHttpClient.  The second is that the factory logic assumes customization via sub-classing (see the XXXClientFactory.additionalConfig methods).  There wasn't a mechanism envisioned for modifying across any factory - such as modifying the interceptors in a common way.

> Why not pass HttpClientBuilder or HttpClientFactoryBuilder to KubernetesClientBuilder?

Since we added HttpClient.getFactory, HttpClient.Builder is viable as a possible argument to KubernetesClientBuilder.  To fully decompose it's also possible to change HttpClient.Factory.createHttpClient(Config config) to be HttpClient.Factory.processConfig(Config config, HttpClient.Builder).

That looks approximately like:

```java
HttpClient.Factory factory = HttpClient.Factory.new();
HttpClient.Builder builder = factory.newBuilder();
// make modifications that could be overridden by processing the config
factory.processConfig(config, builder);
// make modifications that override the config
builder.addOrReplaceInterceptor ...
// then supply the builder to the KubernetesClientBuilder
```

However it's also possible to just associate a Consumer<HttpClient.Builder> with the KubernetesClientBuilder:

```java
new KubernetesClientBuilder()
  .withConfig(config)
  .withAdditionalHttpClientConfiguration(b -> b.addOrReplaceInterceptor... )
  .build();
```

That eliminates a few of the steps.

_Originally posted by @shawkins in https://github.com/fabric8io/kubernetes-client/issues/2271#issuecomment-1263038923_

### LINKED ISSUE #4472: withRequestConfig support needs refined

The implementation is currently based upon creating entirely new client instances.  The old implementation was based upon the assumption that of threadlocal access to config properties is no longer valid with sync clients.

Separate the requestconfig into values that are truly applicable on a per request basis and those that require a new config / client.  It should in theory be possible to control things like retry behavior on a per request basis.

_No response_

### Additional context

_No response_

shawkins:
## Description
The first commit is to fix #4472 - refined withRequestConfig so that a new client is not created.  This removes the need / ambiguity or exposing the factory via the client - which is needed for #4471.  It also makes withRequestConfig much lighter weight as it no longer creates an entirely new client.  On the downside it requires a lot of wiring and a new Inteceptor method so that it can pick up the modified, rather than the default Config.  The 6.0 changes allowed the maxConcurrentRequests and maxConcurrentRequestsPerHost to actually work - but as far as I can tell those did not work in 5.x.  I vote for removing those methods from the RequestConfig entirely.  I'll update the changes with docs based upon whatever we agree upon.

The second commit fixes #4471 - it adds a KubernetesClientBuilder.withHttpClientBuilderConsumer that can be used to configure the Builder that is created.  We still need the KubernetesClientBuilder to take a Factory as TokenRefreshInterceptor expects to use the newBuilder() method to create a "fresh" client instance - that logic could be refactored of course, but that's too much for right now.

## Type of change

## Checklist


manusa:
> but as far as I can tell those did not work in 5.x. I vote for removing those methods from the RequestConfig entirely. I'll update the changes with docs based upon whatever we agree upon.

This seems fine, we can deprecate them and remove them in a subsequent release

manusa:
I think I broke the feature while rebasing :( sorry :(
I hope you still have your local copy @shawkins

shawkins:
Sure, let me see.

shawkins:
> This seems fine, we can deprecate them and remove them in a subsequent release

The pr currently has them removed - the rationale is that since they are probably used through the builder, a user unfortunately won't know they are deprecated unless they read the release notes.

manusa:
> > This seems fine, we can deprecate them and remove them in a subsequent release
> 
> The pr currently has them removed - the rationale is that since they are probably used through the builder, a user unfortunately won't know they are deprecated unless they read the release notes.

that works too

manusa (approved):
LGTM, thx

Issue #4487, opened 2022-10-10
Schema for multimaps is wrong


xRodney:
### Describe the bug

Hello again,

after #4413 will have been merged, crd generator will no longer fail on multimaps, which are present in keycloak-operator crds.

Unfortunately, they will still be generated wrong.

### Fabric8 Kubernetes Client version

SNAPSHOT

### Steps to reproduce

(same as in #4357)

```
@Group("map.fabric8.io")
@Version("v1alpha1")
public class ContainingMaps extends  CustomResource<ContainingMapsSpec, Void> {}

public class ContainingMapsSpec {
  private MyMultiMap<String, String> test3;
}

public class MyMultiMap<K, V> extends HashMap<K, List<V>> {}
```

The field `test3` is treated as if it was `Map<String, String>`:

```
      openAPIV3Schema:
        properties:
          status:
            type: object
          spec:
            properties:
              test3:
                additionalProperties:
                  type: string
                type: object

```

### Expected behavior

The field `test3` should be treated as if it was `Map<String, List<String>>`:

```
 openAPIV3Schema:
        properties:
          status:
            type: object
          spec:
            properties:
              test3:
                additionalProperties:
                  items:
                    type: string
                  type: array
                type: object



```

### Runtime

other (please specify in additional context)

### Kubernetes API Server version

other (please specify in additional context)

### Environment

other (please specify in additional context)

### Additional context

Compared to #4413, this is rather low severity, as it can easily be worked around using `@SchemaFrom` and/or `@SchemaSwap`.

In my assessment, this should be improved in sundrio and I have already filed an issue there (sundrio/sundrio#341) and a PR (sundrio/sundrio#342). If I'm right and my contribution is merged, here we should only upgrade sundrio and maybe add a test and a changelog entry.

PR #4501, opened 2022-10-14
fix: Stack overflow on multi maps (and other non-trivial generic clas…

xRodney:
## Description
…ses), schema for multimaps is wrong

Logic in sundrio that is responsible for replacing generic type arguments on methods and properties with their concrete instantiations from subclasses contained a bug, in which it looped infinitely if the same type argument name was used at multiple classes in the hierarchy. A common occurence were multimaps.

A very similar code was also present in crd-generator, also suffering from the same bug.

Sundrio has been updated to account for this situation, and updated utilities from sundrio have been introduced to crd-generator, which replace the previous implementation of the same algorithm.

A test has been added to validate that multimaps (like `class MultiMap<K,V> implements Map<K,List<V>>`) are generated correctly.

Fixes fabric8io/kubernetes-client#4357
Fixes fabric8io/kubernetes-client#4487
Supersedes closes #4504 

## Type of change

## Checklist

### DIFF
--- a/crd-generator/test/src/test/java/io/fabric8/crd/generator/maps/ContainingMapsCRDTest.java
+++ b/crd-generator/test/src/test/java/io/fabric8/crd/generator/maps/ContainingMapsCRDTest.java
@@ -1,0 +1,64 @@
+/**
+ * Copyright (C) 2015 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.fabric8.crd.generator.maps;
+
+import io.fabric8.kubernetes.api.model.apiextensions.v1.CustomResourceDefinition;
+import io.fabric8.kubernetes.api.model.apiextensions.v1.CustomResourceDefinitionVersion;
+import io.fabric8.kubernetes.api.model.apiextensions.v1.JSONSchemaProps;
+import io.fabric8.kubernetes.client.utils.Serialization;
+import org.junit.jupiter.api.Test;
+
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+class ContainingMapsCRDTest {
+
+  @Test
+  void testCrd() {
+    CustomResourceDefinition d = Serialization.unmarshal(getClass().getClassLoader()
+        .getResourceAsStream("META-INF/fabric8/containingmaps.map.fabric8.io-v1.yml"),
+        CustomResourceDefinition.class);
+    assertNotNull(d);
+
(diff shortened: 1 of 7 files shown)


xRodney:
It seems that sundrio 0.93.1 is not yet fully released. I will make it Ready for review once it is.

xRodney:
Sundrio has been released (thanks @iocanel), this is now ready for review.
(That one failing E2E test is a false alarm, in my opinion)

andreaTP (commented):
Overall this looks good, thanks a lot @xRodney !

I added one question and if you can find more tests to cover possible edge cases it would be nice (e.g. different number of generic type parameters etc. etc.).

crd-generator/api/src/test/java/io/fabric8/crd/example/map/ContainingMapsSpec.java
```diff
@@ -32,4 +33,8 @@ public Map<String, Map<String, List<Boolean>>> getTest2() {
     return test2;
   }
 
+  private MyMultiMap<String, String> test3;
+
+  class MyMultiMap<K, V> extends HashMap<K, List<V>> {
```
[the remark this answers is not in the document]
xRodney: [The new implementation](https://github.com/sundrio/sundrio/blob/main/model/utils/src/main/java/io/sundr/model/functions/TypeCast.java) investigates the whole inheritance tree to see what type arguments are used on `Map`. 

In your example, the parameters on any reference are ignored and the actual type used is always `Map<Integer, Boolean>`, so that's what is rendered. 
(Except ` WARN Property 'test3' with 'io.fabric8.crd.example.map.ContainingMapsSpec.MyMultiMap<java.lang.String,java.lang.String>' key type is mapped to 'string' because of CRD schemas limitations` :D)

```

  private MyMultiMap<String, String> test3;

  class MyMultiMap<K, V> extends HashMap<Integer, Boolean> {
  }
```

becomes:
```
test3:
  additionalProperties:
    type: boolean
  type: object
```

I will add test covering this and other cases as suggested.

xRodney:
I have added more tests to `crd-generator-api`. I also copied most of the cases to `crd-generator-test`, because I was worried about possible differences between Sundrio Reflection and APT contexts. But all seems to be good.

Btw, there is already quite an extensive test suite in sundrio [here](https://github.com/sundrio/sundrio/blob/main/model/utils/src/test/java/io/sundr/model/functions/TypeCastTest.java) and [here](https://github.com/sundrio/sundrio/blob/main/model/utils/src/test/java/io/sundr/model/utils/CollectionsTest.java)

crd-generator/test/src/test/java/io/fabric8/crd/generator/maps/ContainingMapsCRDTest.java
```diff
@@ -0,0 +1,64 @@
...
+import io.fabric8.kubernetes.api.model.apiextensions.v1.JSONSchemaProps;
+import io.fabric8.kubernetes.client.utils.Serialization;
+import org.junit.jupiter.api.Test;
+
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+class ContainingMapsCRDTest {
```
manusa, line 29: Most of these tests could be refactored to use single assertions with AssertJ, but it's not big deal

manusa: OK, this is probably a port of `CRDGeneratorTest`. That one needs some refactoring too :sweat_smile:

manusa (approved):
LGTM (codewise)

andreaTP (approved):
Lgtm! Thanks a lot for this fix @xRodney ! Much appreciated!

PR #4512, opened 2022-10-20
Fix StackOverflow on cyclic references involving collections.

### LINKED ISSUE #4510: Stack overflow on recursive types (again)
### Describe the bug

Hi,

Unfortunately, even after the recent fixes for multimaps, keycloak-operator CRD still does not compile, it still fails with StackOverflowError.
But the stack trace is different this time, and is not connected with multimaps.

This can be reproduced on current master (to be 6.2)

SNAPSHOT

### Steps to reproduce

I have been able to reproduce the issue in a unit test:

```
@Group("sample.fabric8.io")
@Version("v1alpha1")
public class Cyclic extends CustomResource<CyclicSpec, CyclicStatus> implements Namespaced {
}

public class CyclicSpec {
  private List<Ref> ref;
}

public class Ref {
  private List<Ref> ref;
}
```

You will probably recognize this as a variation of an existing test. But the existing test has both fields as of simple `Ref` type, not `List<Ref>`. It turns out, both `List`s are necessary for the Stack Overflow to occur.

### Expected behavior

Definitely no StackOverflow. 
If the field is otherwise unannotated, I expect to get an error telling me about cyclic references.
But I should be able to use `@JsonIgnore`, `@SchemaFrom` or `@SchemaSwap` to exclude the field from generation or replace it with another type.

### Runtime

### Environment

_No response_

### Additional context

The stack overflow happens in class `ClassDependenciesVisitor`:

```
// process all references to see if they need to be added or not
    type.getReferences().forEach(c -> {
      final String fqn = c.getFullyQualifiedName();
      if (ignore(fqn)) {
        return;
      }

      // check if we're dealing with a collection type to extract parameter types if existing
      if (fqn.startsWith("java.util") && (Collections.isCollection(c) || Collections.IS_MAP.apply(c))) {
        c.getArguments().forEach(this::processTypeRef);    // <<<<<<<<<<<<<<<<<<<< HERE
      } else {
        // otherwise, process all non-JDK classes that we haven't already processed
        if (!ignore(fqn) && classesForCR.add(fqn)) {
          // deal with generic arguments if present
          c.getArguments().forEach(this::processTypeRef);
        }
      }
    });
```

This visitor, it appears, is only used for tests / debugging purposes. The problematic portion can be simplified (or just commented out) and after necessary `@SchemaSwap`s, the keycloak CRD finally generates.

Unfortunately, as it stands now, the end users cannot work this around using `@SchemaSwap`s or similar, because the SO happens in a code path that ignores these annotations.

xRodney:
## Description

The logic in ClassDependenciesVisitor was prone to SO when there were cyclic references in the schema. This should fix it, and the tests are still passing, so hopefully it is safe. 

Fixes #4510

~TBH, I do not fully understand the purpose of ClassDependenciesVisitor, I realize it it used in tests and maybe when someone would want to include the generator in their own code? I would be grateful for some pointers.~

It appears that it exists mostly to facilitate conditional hot reload when used with quarkus, as implemented here https://github.com/quarkiverse/quarkus-operator-sdk/blob/main/core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/OperatorSDKProcessor.java#L187

This still works with my changes, see my comment below.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Bugs
 * Fix #4159: ensure the token refresh obeys how the Config was created
 * Fix #4491: added a more explicit shutdown exception for okhttp
+* Fix #4510: Fix StackOverflow on cyclic references involving collections.
 * Fix #4534: Java Generator CLI default handling of skipGeneratedAnnotations
 * Fix #4535: The shell command string will now have single quotes sanitized
 * Fix #4543: (Java Generator) additionalProperties JsonAny setter method generated as setAdditionalProperty
--- a/crd-generator/api/src/main/java/io/fabric8/crd/generator/visitor/ClassDependenciesVisitor.java
+++ b/crd-generator/api/src/main/java/io/fabric8/crd/generator/visitor/ClassDependenciesVisitor.java
@@ -33,7 +33,7 @@ public class ClassDependenciesVisitor extends TypedVisitor<TypeDefBuilder> {
   private static final Map<String, Set<String>> traversedClasses = new HashMap<>();
   private static final Map<String, Set<String>> crdNameToCrClass = new HashMap<>();
   private final Set<String> classesForCR;
-  private final Set<String> processed = new HashSet<>();
+  private final Set<ClassRef> processed = new HashSet<>();
 
   public ClassDependenciesVisitor(String crClassName, String crdName) {
     // need to record all classes associated with the different versions of the CR (not the CRD spec)
@@ -49,7 +49,7 @@ public class ClassDependenciesVisitor extends TypedVisitor<TypeDefBuilder> {
     // note that we cannot simply check the traversed class set to know if a class has been processed because classes
     // are usually added to the traversed set before they're looked at in depth
     final String className = type.getFullyQualifiedName();
-    if (ignore(className) || processed.contains(className)) {
+    if (ignore(className)) {
       return;
     }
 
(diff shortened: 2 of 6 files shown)


andreaTP:
Hey @xRodney ! Thanks for this follow-up!

Arrived at this point do you think it's worth adding an integration test using the `RealmRepresentation` from the published jar `keycloak-models`?

Regarding your question about `ClassDependenciesVisitor` @metacosm should be the best person to answer.

andreaTP:
Thanks, a lot @xRodney ! At least, this guarantees us that we are at the end of the tunnel 🙂 super glad you picked it up!

Let @manusa and @rohanKanojia decide whether they think adding this hard dependency to test is a good idea or not.
Meanwhile, this guarantees us some kind of confidence this is going to work 🥳

xRodney:
Hi, @manusa and @rohanKanojia, do you have any opinion on this?
If you feel this needs a deeper discussion, I would be happy to extract the Keycloak tests to a separate PR and just fix the bug here. It's up to you.

xRodney:
Okay, so regarding my original uncertainty regarding the purpose of `ClassDependenciesVisitor`:
Iit appears that it exists mostly to facilitate conditional hot reload when used with quarkus, as implemented here https://github.com/quarkiverse/quarkus-operator-sdk/blob/main/core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/OperatorSDKProcessor.java#L187 
Originally, I was testing it as a raw java operator, but with quarkus it really shines (when it does not SO, I mean :D )
So I booted up quarkus-operator-sdk and replaced dependencies with my changes from this PR, and hot reload seems to work as expected. So I think we should be fine.

About Keycloak dependency:
Since there seems to be not much interest, I propose to exclude it to a separate PR. I really wish that the bugfix makes it to the 6.3 release, and I feel that the (lack of) discussion about a test would unnecessarily delay the merge.

manusa (approved):
code-wise looks good.
:pray: Andrea/Chris check if feature-wise it's OK.

andreaTP (approved):
LGTM, thanks for the separate PR on the Keycloak CR 👍

PR #4521, opened 2022-10-22
JobExample doesn't use deprecated methods

Tariqs:
I have updated the job creation code according to the version 6 and above.

Related to https://stackoverflow.com/questions/74148689

## Type of change

## Checklist

### DIFF
--- a/kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/JobExample.java
+++ b/kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/JobExample.java
@@ -32,51 +32,53 @@ import java.util.concurrent.TimeUnit;
  * Creates a simple run to complete job that computes π to 2000 places and prints it out.
  */
 public class JobExample {
-    private static final Logger logger = LoggerFactory.getLogger(JobExample.class);
+  private static final Logger logger = LoggerFactory.getLogger(JobExample.class);
 
-    public static void main(String[] args) {
-        final ConfigBuilder configBuilder = new ConfigBuilder();
-        if (args.length > 0) {
-          configBuilder.withMasterUrl(args[0]);
-        }
-        try (KubernetesClient client = new KubernetesClientBuilder().withConfig(configBuilder.build()).build()) {
-          final String namespace = "default";
-          final Job job = new JobBuilder()
-            .withApiVersion("batch/v1")
-            .withNewMetadata()
-            .withName("pi")
-            .withLabels(Collections.singletonMap("label1", "maximum-length-of-63-characters"))
-            .withAnnotations(Collections.singletonMap("annotation1", "some-very-long-annotation"))
-            .endMetadata()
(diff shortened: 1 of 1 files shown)


kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/JobExample.java
```diff
@@ -62,12 +62,16 @@ public static void main(String[] args) {
             .endSpec()
             .build();
 
-          logger.info("Creating job pi.");
+          logger.info("Creating job pi.");                             
           client.batch().v1().jobs().inNamespace(namespace).createOrReplace(job);
 
+          //If you are using version kubernetes-client version 6 or above, use the following code to create job
+          //client.batch().v1().jobs().inNamespace("default").resource(job).create();
+
```
manusa, line 70: Please replace the affected code with valid code instead of adding it as a comment.
The examples should be aligned with whatever version they're published along with (i.e. these examples should be aligned with v6)

manusa (changes requested):
Please, check the comments, and also make sure to respect the code-base formatting.

kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/JobExample.java
shawkins: @Tariqs as long as the job object has the appropriate apiVersion set, then you can use:

client.resource(job).inNamespace("default").create()

Tariqs: @shawkins What that apiVersion? Please specify

shawkins: @Tariqs as long as the resource has the appropriate apiVersion and kind (it will by default), you can just use client.resource(object) to perform any of the common create/update/delete operations.  

You only need to call the typed api - batch().v1().jobs() - to call type specific operations, or to get a typed list.

In this example there are other Job specific operations, like getLog, being performed so it would make sense to use:

```
ScalableResource<job> jobResource = client.batch().v1().jobs().inNamespace("default").resource(job);
jobResource.createOrReplace();
...
String jobLog = jobResource.getLog();
```

Tariqs: @shawkins When I tried to use it, IntelliJ showed me as deprecated method. I did call like you have mentioned `batch().v1().jobs() `.

Tariqs: @shawkins Please refer to this post where I have explained the deprecation issue https://stackoverflow.com/questions/74148689/kubernetes-job-creation-method-showing-as-deprecated-in-intellij-i-am-using-fab/74153050#74153050

manusa (approved):

PR #4559, opened 2022-11-04
Remove ignored_fields from JObject

### LINKED ISSUE #4546: "description" is a protected keyword and is ignored on generation, can we somehow allow this for the content
### Component java-generator ignores "description" field definition in content

I don't understand how to use CustomRessourceDefinitions containing "description" as field in the content. Due to the fact that it is part of the IGNORED_FIELDS in JObject.

I get that "description" is a protected keyword somehow to actually also describe an object in open API and that is therefore put to javadoccomments, but what if the Object itself has an field named description.

When I try generating an object with
description:
   type: string

It will not appear in the generated code.

I actually don't have an idea how to implement it, but I would like to have the possibility to use description also as content, not only as interface explaination.

Currently I don't use the generated code, but generated it once and added the missing fields. This should not be the prefered way.

### Additional context

Maybe there is some hidden proerty to the generator to use it differently, but I did not find any code in the project that would let me think I can use "description" as ones object field.

andreaTP:
## Description
Fix #4546

## Type of change

## Checklist

### DIFF
--- a/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
+++ b/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
@@ -37,15 +37,6 @@ import java.util.stream.Collectors;
 
 public class JObject extends AbstractJSONSchema2Pojo implements JObjectExtraAnnotations {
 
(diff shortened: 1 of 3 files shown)


andreaTP:
cc @mteubner

manusa (approved):

PR #4561, opened 2022-11-04
[java-generator] Fix additionalProperties ser/deser

### LINKED ISSUE #4543: Additional Attributes are generated to be ignored, because of wrong setting of JsonAnySetter annotation
### Describe the bug

Component : java-generator

The generation of additionalProperties out of a CRD is not creating the access methods as expected. If there is an object in the CRD defined with configuration:
                            type: object
                            x-kubernetes-preserve-unknown-fields: true

It will generate a java class for that configuration, but the mapper will not fill the Map. Therefore the object will not contain any data.

6.2.0

### Steps to reproduce

1. Create a CRD having type object and x-kubernetes-preserve-unknown-fields: true
2. generate the java code
3. read a CR with additionalAttributes like 
         - configuration:
           description: "Map does not contain this content"
           enabled: true
           name: "Erronious content"
4. Use a mapper to read that yaml into you generated Java Object

All the addtionalAttributes are not present in the Map

### Expected behavior

Every unknown field should be accessible with in that additionalAttributes map via the "key"; i.E. "description".

---
Proposal, instead of generating:

    @com.fasterxml.jackson.annotation.JsonAnyGetter()
    public java.util.Map<String, Object> getAdditionalProperties() {
        return additionalProperties;
    }
    @com.fasterxml.jackson.annotation.JsonAnySetter()
    public void setAdditionalProperties(java.util.Map<String, Object> additionalProperties) {
        this.additionalProperties = additionalProperties;
    }

---
we may need to generate 

    @com.fasterxml.jackson.annotation.JsonAnyGetter()
    public java.util.Map<String, Object> any() {
        return additionalProperties;
    }

    @com.fasterxml.jackson.annotation.JsonAnySetter()
    public void setAdditionalProperties(String name, Object value) {
        additionalProperties.put(name, value);
    }

    public Object getAdditionalProperties(String name) {
        return additionalProperties.get(name);
    }

---

It definetly works for String objects as unknown entities. I don't know what happends with complex objects.

### Runtime

Kubernetes (vanilla), other (please specify in additional context)

1.25.3@latest

### Environment

Windows

_No response_

### Additional context

_No response_

andreaTP:
## Description
Fix #4543 

## Type of change

## Checklist

### DIFF
--- a/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
+++ b/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
@@ -22,8 +22,10 @@ import com.github.javaparser.ast.NodeList;
 import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
 import com.github.javaparser.ast.body.EnumDeclaration;
 import com.github.javaparser.ast.body.FieldDeclaration;
+import com.github.javaparser.ast.body.MethodDeclaration;
 import com.github.javaparser.ast.body.VariableDeclarator;
 import com.github.javaparser.ast.expr.*;
+import com.github.javaparser.ast.stmt.BlockStmt;
 import com.github.javaparser.ast.type.ClassOrInterfaceType;
 import com.github.javaparser.utils.StringEscapeUtils;
 import io.fabric8.java.generator.Config;
@@ -280,6 +282,13 @@ public class JObject extends AbstractJSONSchema2Pojo implements JObjectExtraAnno
 
       objField.createGetter().addAnnotation("com.fasterxml.jackson.annotation.JsonAnyGetter");
       objField.createSetter().addAnnotation("com.fasterxml.jackson.annotation.JsonAnySetter");
+
+      MethodDeclaration additionalSetter = clz.addMethod("setAdditionalProperties", Modifier.Keyword.PUBLIC);
+      additionalSetter.addAnnotation("com.fasterxml.jackson.annotation.JsonAnySetter");
+      additionalSetter.addParameter("String", "key");
+      additionalSetter.addParameter("Object", "value");
+      additionalSetter
+          .setBody(new BlockStmt().addStatement(new NameExpr("this." + Keywords.ADDITIONAL_PROPERTIES + ".put(key, value);")));
     }
 
     buffer.add(new GeneratorResult.ClassResult(this.className, cu));
--- a/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
+++ b/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
@@ -283,7 +283,7 @@ public class JObject extends AbstractJSONSchema2Pojo implements JObjectExtraAnno
       objField.createGetter().addAnnotation("com.fasterxml.jackson.annotation.JsonAnyGetter");
       objField.createSetter().addAnnotation("com.fasterxml.jackson.annotation.JsonAnySetter");
 
-      MethodDeclaration additionalSetter = clz.addMethod("setAdditionalProperties", Modifier.Keyword.PUBLIC);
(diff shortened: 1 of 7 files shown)


andreaTP:
cc @mteubner

java-generator/it/src/it/cert-manager/expected/Auth.expected
```diff
@@ -82,4 +82,9 @@ public class Auth implements io.fabric8.kubernetes.api.model.KubernetesResource
     public void setAdditionalProperties(java.util.Map<String, Object> additionalProperties) {
         this.additionalProperties = additionalProperties;
     }
+
+    @com.fasterxml.jackson.annotation.JsonAnySetter()
+    public void setAdditionalProperties(String key, Object value) {
```
shawkins, line 87: Should this be setAdditionalProperty singular?

andreaTP: Surprisingly not, the integration test fails if I make it singular.

mteubner: So probably some jackson magic :)

andreaTP: yup, at least now is properly tested 🙂

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
```diff
@@ -280,6 +282,13 @@ public GeneratorResult generateJava() {
 
       objField.createGetter().addAnnotation("com.fasterxml.jackson.annotation.JsonAnyGetter");
       objField.createSetter().addAnnotation("com.fasterxml.jackson.annotation.JsonAnySetter");
+
+      MethodDeclaration additionalSetter = clz.addMethod("setAdditionalProperties", Modifier.Keyword.PUBLIC);
```
shawkins, line 286: @andreaTP I mean this would need to change to singular as well.

andreaTP: @shawkins that's what I changed to generate the corresponding output everywhere else.

shawkins, line 286:
```suggestion
      MethodDeclaration additionalSetter = clz.addMethod("setAdditionalProperty", Modifier.Keyword.PUBLIC);
```

Sorry, I may not be being clear.  I would expect both the test and this logic to say setAdditionalProperty singular.  That lines up with the model generation logic in the main project - Pod.setAdditionalProperty.

andreaTP: Right, that's what I did, changing both breaks the functionality I'm fixing here as demonstrated by the additional test included in this PR.

If you are still in doubt I propose to have a quick sync on it to clarify somewhere next week ;-)

shawkins: Sorry, it would probably be best to compare to what's already in use.  The core model classes are generated with:

```
    @JsonIgnore
    private Map<String, Object> additionalProperties = new HashMap<String, Object>();

   @JsonAnyGetter
    public Map<String, Object> getAdditionalProperties() {
        return this.additionalProperties;
    }

    @JsonAnySetter
    public void setAdditionalProperty(String name, Object value) {
        this.additionalProperties.put(name, value);
    }
```

due to a limitation in the jsonschema2pojo, 

```
@JsonIgnore
void setAdditionalProperties(Map<String, Object> properties)
```

 being generated by the class annotations:

```
@Setter
@Accessors(prefix = {
    "_",
    ""
})
```

As far as I know that handles all of the issues we have seen with serdes and builder logic.

Could you try aligning to that?

andreaTP: @shawkins sorry for the long forth and back, the integration test failure was due to a different reason ... fixed now 👍

shawkins (approved):
LGTM

manusa (approved):

Issue #3246, opened 2021-06-16
ensure delete returns only when the delete operation is fully finished


victornoel:
Hi,

If I'm calling delete on a job for example, it takes time for the operation to really finish, but the delete call returns early, even if I used the FOREGROUND DeletionPropagation (not sure it's really linked to the matter ^^).

It would make sense to add an option so that the call only returns when everything is deleted as expected. A bit like the `--wait` parameter of kubectl works in that context (see https://github.com/kubernetes/kubernetes/issues/42594).

victornoel:
Related to #1474

victornoel:
Potential workaround:
```java
client.batch().v1().jobs()
    .inNamespace(namespace)
    .withName(name)
    .waitUntilCondition(Objects::isNull, 10, TimeUnit.SECONDS);
```

manusa:
There's a different behavior between what the REST API does when performing a deletion with the foreground propagation policy and what `kubectl` does.

When you perform a `kubectl delete ... --cascade='foreground`, the command waits until every dependent resource is deleted to complete the execution.

However, when you perform the same operation using the REST API (with `foreground` PropagationPolicy), the requests returns as soon as the resource has been modified and the `deletionTimestamp` has been added to the metadata. I'm not sure if the client should take care of performing this wait/blocking for you.

Your suggestion is exactly what you need to do if you really need to provide blocking until the resource was effectively deleted.

victornoel:
@manusa I understand. It makes sense indeed, thank you for the explanation, I will continue with my solution then!

PR #4578, opened 2022-11-15
fix #4516: adding a blocking delete operation

### LINKED ISSUE #4516: Replacement for deleteExisting behavior.

deletingExisting was removed to help streamline the api interfaces (it wasn't universally available for Resource) and because the implementation wasn't based upon exact deletion tracking - which is now possible that delete can return the deletion list.

In kubectl apply and replace have a force option.  That is relevant to our create, replace, and createOrReplace operations.  It would also be applicable to some of the patch methods, which allow for our current server side apply behavior.

One option is to provide a higher level blocking delete operation #3246 - similar to the logic in kubectl we would track by uid is possible what resources were deleted to make sure that intervening creates don't cause indefinite waits.  

That could look like  Deletable.forceDelete returning a CompletionStage<List<StatusDetails>> - that would give the user flexibility in how long to wait.  The downside is that it's not fluent.

Another option is to add a boolean force parameter to the given affected methods - resource.replace(true) - that has the downside of affecting quite a few methods.

Finally to provide the force option via a method - resource.force().create() - we'd need another interface used at both ListVisitFromServerWritable and WritableOperation that adds a force method which would need a new interface to allow for CreateOrReplaceable or patch methods to then be called.  It's that last part about patch methods that make this a little less appetizing of a solution - until we have a higher level method for server side apply, I don't think this one is a great option.

_No response_

### Additional context

_No response_

shawkins:
## Description
Fixes #4516 with the ability to make delete a blocking operation.

The context changes are due to elevating timeout to the base context.  To avoid multiple inheritance issues with Timeoutable it was converted to non-generic.

This option turns out to be a lot simpler than adding the force option to serverSideApply and replace - as all of the delete options (timeout, grace period, propagation) are needed for those contexts as well.  And we just introduced a NonDeletingOperation interface which further complicates what could be common api point for specifying delete options.

The blocking delete is similar to what exists in the kubectl client when --wait is used as a parameter (delete, apply, replace).  The returned statusdetails are used to determine the set of uids that you're tracking.  Once none of those exist, you'll consider things deleted.  A corner case is the possibility of incomplete statusdetails - that was noted in an old issue, but I'm not sure that it is still relevant and kubectl doesn't consider that case either.

This simplifies the workaround to just two calls:
op.withTimeout(10, TimeUnit.SECONDS).delete();
op.create();

This is also applicable to #3246 and similar issues.

## Type of change

## Checklist


shawkins:
Updated with a changelog and rebased to master.

One change from the previous commit is that the blocking method was generalized so that it would work for list contexts to be applied in parallel, rather than serially.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/RollingOperationContext.java
```diff
@@ -1,112 +0,0 @@
...
-    return new RollingOperationContext(podOperationContext, rolling, rollingTimeout, rollingTimeUnit);
-  }
-
-  public RollingOperationContext withRollingTimeout(long rollingTimeout) {
-    return new RollingOperationContext(podOperationContext, rolling, rollingTimeout, rollingTimeUnit);
-  }
-
-  public RollingOperationContext withRollingTimeUnit(TimeUnit rollingTimeUnit) {
-    return new RollingOperationContext(podOperationContext, rolling, rollingTimeout, rollingTimeUnit);
-  }
```
manusa, line 65: how are these options being set now?

manusa: nevermind, just saw the new lombok builder usage at RollableScalableResourceOperation

manusa (approved):

PR #4579, opened 2022-11-15
Removing internal usage of registerCustomKind

shawkins:
## Description
This follows up on a next step of #3972 - remove the internal usage of registerCustomKind.  The static nature of that registration was problematic both for us and for users as the behavior would seem inconsistent.  

This change highlighted that several packages are missing from the KubernetesDeserializer - https://github.com/fabric8io/kubernetes-client/commit/8db8124ef3ac69b980b9ba32dceac0cb1c0ff459 - that commit won't be needed against master if #4511 is committed first, but may be considered for back-porting.

The strategy here is to use some additional jackson logic to convey an element type with list parsing.  list is the only built-in path where we may not have a type already associated with the processing - the other calls, such as a get, will pass the expected type directly; have handling to convert, such as with watch events; or can just use the generic.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/OperationSupport.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/OperationSupport.java
@@ -48,6 +48,7 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InterruptedIOException;
+import java.lang.reflect.Type;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URL;
@@ -566,7 +567,12 @@ public class OperationSupport {
    */
   private <T> T handleResponse(HttpRequest.Builder requestBuilder, Class<T> type, Map<String, String> parameters)
       throws IOException {
-    return waitForResult(handleResponse(httpClient, requestBuilder, type, parameters));
+    return waitForResult(handleResponse(httpClient, requestBuilder, new TypeReference<T>() {
+      @Override
+      public Type getType() {
+        return type;
+      }
+    }, parameters));
   }
 
   /**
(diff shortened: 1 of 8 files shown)


shawkins:
> This change highlighted that several packages are missing from the KubernetesDeserializer - https://github.com/fabric8io/kubernetes-client/commit/8db8124ef3ac69b980b9ba32dceac0cb1c0ff459 - that commit won't be needed against master if https://github.com/fabric8io/kubernetes-client/pull/4511 is committed first, but may be considered for back-porting.

Separated to #4594

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/OperationSupport.java
```diff
@@ -579,7 +585,8 @@ private <T> T handleResponse(HttpRequest.Builder requestBuilder, Class<T> type,
    *
    * @return Returns a de-serialized object as api server response of provided type.
    */
-  protected <T> CompletableFuture<T> handleResponse(HttpClient client, HttpRequest.Builder requestBuilder, Class<T> type,
+  protected <T> CompletableFuture<T> handleResponse(HttpClient client, HttpRequest.Builder requestBuilder,
```
manusa, line 588: Wouldn't it be easier to provide an overloaded method instead of duplicating the TypeReference inner-class instantiation in most of the calling methods?

shawkins: I don't like how many overloaded methods there are already in OperationSupport, so it didn't seem worth it as it would just combine 2 calls.

manusa: sounds good

manusa (approved):

PR #4615, opened 2022-11-28
Making raw/generic buildable

### LINKED ISSUE #4540: Builder methods expecting HasMetadata/KubernetesResource are not correct
When the builder logic generates the handling for a method like IngressControllerSpec.unsupportedProperties it creates a method impl for withUnsupportedProperties that does an instanceOf check for the known buildable types.  If you use something like GenericKubernetesResource, or RawExtension that is not an explicitly known type - it is silently ignored.  At the very least an exception should be thrown.

For usability at a minimum we need to add GenericKubernetesResource and RawExtension as buildable everywhere - with either a method added or a pointer from the exception when an unknown type is used about converting to these types - either wrapping with RawExtension, or using Serialization.jsonMapper().convertValue(resource, GenericKubernetesResource.class).

cc @iocanel @manusa

shawkins:
## Description

Fix #4540

This addresses #4540 as much as we can on the fabric8 side, it will allow the use of generic or raw extension in builders methods that expect a kubernetesresource.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Fix #4535: The shell command string will now have single quotes sanitized
 * Fix #4543: (Java Generator) additionalProperties JsonAny setter method generated as setAdditionalProperty
 * Fix #4547: preventing timing issues with leader election cancel
+* Fix #4540: treating GenericKubernetesResource and RawExtension as buildable
 * Fix #4569: fixing jdk httpclient regression with 0 timeouts
 
 #### Improvements
--- a/extensions/camel-k/generator-v1/Makefile
+++ b/extensions/camel-k/generator-v1/Makefile
@@ -20,7 +20,7 @@ all: build
 
 build: gobuild
 	pushd ../model-v1 && \
-	mvn clean install -o && \
+	mvn -Pgenerate clean install -DskipTests -o && \
 	popd
 
 gobuild:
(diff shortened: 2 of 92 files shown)


manusa:
Change looks good, but I think we're missing adding the profile to the Makefiles or somewhere else. Consider that the [gobuild.yml](https://github.com/fabric8io/kubernetes-client/blob/f34f3b540594118e21d8942cb84e39b423e628d0/.github/workflows/gobuild.yml) workflow ensures that whenever there are changes in the models, the generated files are persisted too.

Maybe also document about this new Maven profile in https://github.com/fabric8io/kubernetes-client/blob/f5524268e410560cd44eb86c7bb8ee44327f7c39/kubernetes-model-generator/README.md.

shawkins:
> Maybe also document about this new Maven profile

The profile is not new, it's the one that's already used for the kubernetes-model-generator.  I had assumed the extension make files would have just been copied from there, but alas they were not - thus the missing profile.  I'll update the make files.

manusa:
> The profile is not new,

image

Apparently I created those :rofl: 

> I had assumed the extension make files would have just been copied from there, but alas they were not

No, they were not. There's been multiple actions to bring them together in the past, but the process is still ongoing. This seems to be a further step in completing the task. :relieved:

shawkins:
@manusa should be good now

manusa (approved):

Issue #4604, opened 2022-11-26
Can't access ClusterServiceVersions on non OpenShift cluster


jeffmaury:
### Describe the bug

In order to list operators installed on a cluster, we should use the `operators.coreos.com/v1alpha1:ClusterServiceVersion`.

This resource is available through `OpenShiftOperatorHubAPIGroupDSL`. But this one is available through the OpenShiftClient.

But this resource is available on raw Kubernetes cluster once you installed the Operator Lifecycle Manager.

So today it is possible only if user uses `OpenShiftOperatorHubAPIGroupClient` passing the Kubernetes client as a parameter which I'm not sure is a public supported API

### Fabric8 Kubernetes Client version

5.12.4

### Steps to reproduce

See https://github.com/redhat-developer/intellij-openshift-connector/pull/484/files#diff-e332bfa50ae6e995739ad7ae9869734c6299f0d4286bf72d9fc3876207fe89fbR485

### Expected behavior

 Should be possible to adapt `KubernetesClient` to `OpenShiftOperatorHubAPIGroupDSL`

### Runtime

Kubernetes (vanilla)

### Kubernetes API Server version

1.23

### Environment

Windows

shawkins:
@jeffmaury did you try the adapt method on 5.12?  If you received an exception about openshift support, then please try fabric8 6+ as the adapt method no longer perform additional support checks - as long as the target implementation is available, you'll get that client.  There is now a separate supports method.

jeffmaury:
Adapt did not work with 5.12, we can't migrate yet to 6+ as it's not backward compatible but it's on our roadmap

rohanKanojia:
as a workaround you can try accessing ClusterServiceVersion using [GenericKubernetesResource API](https://github.com/fabric8io/kubernetes-client/blob/master/doc/CHEATSHEET.md#resource-typeless-api)

jeffmaury:
I need a typed access to this resource so I found another workaround

manusa:
> I need a typed access to this resource so I found another workaround

What was the other workaround?

---

Not sure what your workaround was. Here is another suggestion:

If the model is in a separate module, add the dependency to that module and then register the model in the KubernetesDeserializer.

jeffmaury:
> > I need a typed access to this resource so I found another workaround
> 
> What was the other workaround?
> 
> Not sure what your workaround was. Here is another suggestion:
> 
[...]

PR #4631, opened 2022-12-02
fix (extensions) : isSupported doesn't check all of the applicable API Groups (#4447)

### LINKED ISSUE #4447: KnativeClient.isSupported doesn't check all of the applicable API Groups
### Describe the bug

Initially reported: https://github.com/quarkusio/quarkus/issues/28171
With fix: https://github.com/quarkusio/quarkus/pull/28172

The current implementation of `isSupported()` for `KnativeClient` is:
```java
  public boolean isSupported() {
    return hasApiGroup("knative.dev", true);
  }
```

This is not taking into consideration all of the API groups our models support:
* serving.knative.dev
* flows.knative.dev
* networking.internal.knative.dev
* etc.

6.0.0

### Steps to reproduce

- In a cluster with Knative CRDs, invoke the `KnativeClient.isSupported` method.
- `false` is returned instead of `true`

### Expected behavior

`true` should be returned instead

### Suggested fix

The [check](https://github.com/fabric8io/kubernetes-client/blob/0308e299313a12c72b9313730a397e894d7278b0/extensions/knative/client/src/main/java/io/fabric8/knative/client/DefaultKnativeClient.java#L242-L246) should not be exact:
```java
  public boolean isSupported() {
    return hasApiGroup("knative.dev", false);
  }
```

rohanKanojia:
## Description
Fix #4447

+ ~Extension client interfaces should extend SupportTestingClient rather than implementations~
+ Disable exact apiGroup match in isSupported method call in extension clients. This effects Istio, Knative and Tekton extensions that use more than one apiGroups

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Bugs
 * Fix #4590: only using a builder if there are visitors
 * Fix #4159: ensure the token refresh obeys how the Config was created
+* Fix #4447: `isSupported` doesn't check all of the applicable API Groups
 * Fix #4491: added a more explicit shutdown exception for okhttp
 * Fix #4510: Fix StackOverflow on cyclic references involving collections.
 * Fix #4534: Java Generator CLI default handling of skipGeneratedAnnotations
--- a/extensions/camel-k/client/pom.xml
+++ b/extensions/camel-k/client/pom.xml
@@ -84,7 +84,17 @@
     </dependency>
     <dependency>
       <groupId>org.junit.jupiter</groupId>
-      <artifactId>junit-jupiter-migrationsupport</artifactId>
+      <artifactId>junit-jupiter-params</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/extensions/camel-k/client/src/main/java/io/fabric8/camelk/client/DefaultCamelKClient.java
+++ b/extensions/camel-k/client/src/main/java/io/fabric8/camelk/client/DefaultCamelKClient.java
@@ -62,7 +62,7 @@ public class DefaultCamelKClient extends ExtensionRootClientAdapter<DefaultCamel
 
   @Override
   public boolean isSupported() {
-    return hasApiGroup("camel.apache.org", true);
+    return hasApiGroup("camel.apache.org", false);
   }
 
 }
--- a/extensions/camel-k/client/src/test/java/io/fabric8/camelk/client/CamelKClientAdaptTest.java
+++ b/extensions/camel-k/client/src/test/java/io/fabric8/camelk/client/CamelKClientAdaptTest.java
@@ -1,0 +1,82 @@
+/**
+ * Copyright (C) 2015 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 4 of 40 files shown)


shawkins:
> Extension client interfaces should extend SupportTestingClient rather than implementations

@rohanKanojia @manusa the intent moving forward was to only expose the explicit support testing for the openshift client - https://github.com/fabric8io/kubernetes-client/blob/db396dc89e9290b8f87ad59ddbf359de3d3d50db/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/extension/SupportTestingClient.java#L22

For the extensions, this is supposed to be an implicit mechanism that provided the same functionality as how their adapter logic used to call ExtensionAdapterSupport.isAdaptable.  I was thinking in future releases we'd simply remove these checks.  Since the extensions can be multi-group / multi-version clients, it's best for the user to check for explicit support of a resource.

manusa:
I'm not sure what happened here.
The scope of the issue (at least when I created it) was just to disable the exact flag from the `isSupported` method implementation.

I also feel kind of uncomfortable with the usage of an interface in production code that is specific for testing purposes.

Steven, could you please give a few bullet points on how we should move forward with the interface and its usage (more specific tasks).

shawkins:
> Steven, could you please give a few bullet points on how we should move forward with the interface and its usage (more specific tasks).

We mostly wanted Client.isAdaptable to go away - it has been deprecated.

The client impls that implemented SupportTestingClient did so to support checking their support via Client.isAdaptable.  The client interfaces that extend SupportTestingClient, which I think is just the OpenShiftClient, are supposed to have the isSupported as part of their public contract.

Rather than doing:

```
if (client.isAdaptable(OpenShiftClient.class)) {
  OpenShiftClient openShiftClient = client.adapt(OpenShiftClient.class); 
  ...
}
```
you are supposed to do:

```
OpenShiftClient openShiftClient = client.adapt(OpenShiftClient.class); 
if (openShiftClient.isSupported()) { 
  ...
}
```
There are several rationales:
 - we've made the adapt operation always work - it's free of support checks - so isAdaptable is no longer semantically meaningful
 - we need to think about why we're exposing XXXClient.isSupported - as mentioned for multi-group / multi-version clients it's not appropriately expressive.

rohanKanojia:
I think these extensions also use isAdaptable method:

https://github.com/fabric8io/kubernetes-client/blob/f6902ec848ee74ead29037cbf11adf61abfeff5a/extensions/camel-k/tests/src/test/java/io/fabric8/camelk/test/AdaptTest.java#L33

I added interface because i noticed isAdaptable is deprecated and I could not find a way to do equivalent using new adapt method. What should be expected behavior of adapt method if underlying cluster doesn't support extension's apiGroup. Should it return null if it doesn't support?

manusa:
> I added interface because i noticed isAdaptable is deprecated and I could not find a way to do equivalent using new adapt method. What should be expected behavior of adapt method if underlying cluster doesn't support extension's apiGroup. Should it return null if it doesn't support?

adapt doesn't care about the cluster having support for a given API group, isSupported should be used instead.

rohanKanojia:
isSupported method is not available in current extension interfaces (KnativeClient, TektonClient etc.) .

manusa:
> isSupported method is not available in current extension interfaces (KnativeClient, TektonClient etc.) .

SupportTestingClient.isSupported is only used in Client.isAdaptable BaseClient implementation which is deprecated, thus the interface is deprecated too.

For 4447 the only requirement is to make the implementation on the default extension client implementations not exact.

In a nutshell, the following statement is not OK:

> - Extension client interfaces should extend SupportTestingClient rather than implementations

We can later consider getting rid altogether of the interface and whatever implementation remains, but I would rather do that in the scope of a different issue where we can actually discuss the implications.

shawkins:
> isSupported method is not available in current extension interfaces (KnativeClient, TektonClient etc.) .

Yes, and I think we want to keep it that way.  

The openshift client support check is more unique - it's primarily checking the url, then looking for any openshift group.  There's a case to be made that such a broad check is not needed there either - or at the very least needs more documentation.  As reported in #4604 you can just as easily have a regular kubernetes cluster return true from this check by having any of the openshift crds/operators installed - but it's certainly not actually a general openshift cluster.

So rather than doing:

```
XXXClient myClient = client.adapt(XXXClient.class); 
if (XXXClient.isSupported()) { 
  ...
}
```

you are supposed to check down to a resource level:

```
if (client.supports(MyResource.class)) {
  ...
}
```

We can of course also entertain adding client.resource(class).supports().

manusa (approved):

shawkins:
@manusa in the last meeting we talked about whether it would make more sense here just to go ahead and remove the deprecated isAdaptable and all of these checks from the extensions.  Would you rather that be a follow-on task instead?

manusa:
Yes. This should be tackled separately. I just needed to release this one so the temp fix can be removed from Quarkus.

PR #4642, opened 2022-12-05
fix #4598: support for free-form objects in crd generator

### LINKED ISSUE #4598: Support for Free-Form Objects in crd generator

I'm trying to create a spec property as a [free-form object](https://swagger.io/docs/specification/data-models/dictionaries/) from my POJO but it seems that it's not possible.

I wanted to add a `map/object` property that accepts any kind of value (string, int, objects..).
I've tried in these ways
```
Map<String, ?> config; 
Map<String, Object> config; 
```
Both generate this schema:

```
config:
  additionalProperties:
    type: object         
  type: object
```

My goal is to generate instead

```
config:
  additionalProperties: {}
  type: object
```
or 
```
config:
  additionalProperties: true
  type: object
```

Be able to generate a spec like:

```
config:
  additionalProperties: true
  type: object
```

_No response_

### Additional context

Using latest version

nicoloboschi:
## Description
Fix #4598 

Maps with Object as value type (e.g. Map<String, Object>) will now produce a spec like 

```
config:
   additionalProperties: true
   type: object
```

this is a breaking change somehow, the same source code now generated a different crd.
However before the generated code was more restrictive.

Source code:
```
Map<String, Object> config;
```

now generates:
```
config:
   additionalProperties: true
   type: object
```

before:
```
config:
  additionalProperties:
    type: object         
  type: object
```
which is the same as above except values of only type of objects were accepted.
To restore the same generated code the user will have to modify the source code to:

```
Map config;
```

## Type of change

 - [] Bug fix (non-breaking change which fixes an issue)
 - [x ] Feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### 6.3-SNAPSHOT
 
 #### Bugs
+* Fix #4598: support for free-form objects in crd generato
 * Fix #4590: only using a builder if there are visitors
 * Fix #4159: ensure the token refresh obeys how the Config was created
 * Fix #4491: added a more explicit shutdown exception for okhttp
--- a/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
+++ b/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
@@ -637,6 +637,12 @@ public abstract class AbstractJsonSchema<T, B> {
       }
 
       final TypeRef valueType = TypeAs.UNWRAP_MAP_VALUE_OF.apply(typeRef);
+      if (valueType instanceof ClassRef) {
+        ClassRef classRef = (ClassRef) valueType;
+        if (Object.class.getName().equals(classRef.getFullyQualifiedName())) {
+          return mapLikeProperty(true);
+        }
+      }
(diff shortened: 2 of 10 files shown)


andreaTP:
Hi @nicoloboschi ,

Thanks a lot for this PR!

I'm struggling to find documentation on how this encoding correctly applies to the underlying data structure, do you have any link to share?

So far I found:
 - [link](https://github.com/swagger-api/swagger-codegen/issues/11619) where `addtionalProperties: true` is used to extend an actual definition, but, in our case there are no other fields in the definition
 - [link](https://github.com/OAI/OpenAPI-Specification/issues/668) where I understand that `additionalProperties: True` is an empty schema and it will accept primitive and object types (which is mapped using `AnyType`)
 
In both cases, I see arguments that make me consider the current implementation more closely respecting the semantics than the proposed one, but I might be missing something, would you mind elaborating?

cc. @shawkins @metacosm

nicoloboschi:
@andreaTP I was following this spec: https://swagger.io/docs/specification/data-models/dictionaries/ (Free form objects) but I've just realized that this is not supported on kubernetes CRD validator. 

I would like to not use ´io.fabric8.kubernetes.api.model.AnyType config;´ since I want to keep the Map in order auto-parse the config. 

I'll close my pull

PR #4644, opened 2022-12-05
improve: generate crds in parallel and optimize code

nicoloboschi:
## Description
In my project I have 10 CRDs classes. 
I've taken a CPU flamegraph and I figured out the usage of java streams api adds a relevant overhead while visiting specs/status classes via reflection. 

Also the visitors can access the crd in parallel since the base typeref is stateless.

Changes included:
- Enable concurrent crd visitors using nCPU threads.
- Moved some hot path code from using streams to old plain java for-loop

FYI I did a similar fix in sundrio: https://github.com/sundrio/sundrio/pull/352

The net result of this patch is that generation time moved from ~30 seconds to to ~11 seconds.  

## Type of change

## Checklist

### DIFF
--- a/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractCustomResourceHandler.java
+++ b/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractCustomResourceHandler.java
@@ -18,15 +18,22 @@ package io.fabric8.crd.generator;
 import io.fabric8.crd.generator.decorator.Decorator;
 import io.fabric8.crd.generator.visitor.*;
 import io.fabric8.kubernetes.client.utils.Utils;
+import io.sundr.builder.Visitor;
 import io.sundr.model.AnnotationRef;
 import io.sundr.model.Property;
 import io.sundr.model.TypeDef;
 import io.sundr.model.TypeDefBuilder;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.stream.Collectors;
 
 /**
@@ -36,9 +43,11 @@ import java.util.stream.Collectors;
 public abstract class AbstractCustomResourceHandler {
 
   protected final Resources resources;
+  private final boolean parallel;
 
-  protected AbstractCustomResourceHandler(Resources resources) {
+  protected AbstractCustomResourceHandler(Resources resources, boolean parallel) {
     this.resources = resources;
+    this.parallel = parallel;
   }
 
   public void handle(CustomResourceInfo config) {
@@ -54,20 +63,18 @@ public abstract class AbstractCustomResourceHandler {
 
     ClassDependenciesVisitor traversedClassesVisitor = new ClassDependenciesVisitor(config.crClassName(), name);
 
-    TypeDefBuilder builder = new TypeDefBuilder(def);
+    List<Visitor<TypeDefBuilder>> visitors = new ArrayList<>();
     if (config.specClassName().isPresent()) {
-      builder.accept(specReplicasPathDetector);
+      visitors.add(specReplicasPathDetector);
     }
-
     if (config.statusClassName().isPresent()) {
-      builder.accept(statusReplicasPathDetector);
+      visitors.add(statusReplicasPathDetector);
     }
+    visitors.add(labelSelectorPathDetector);
+    visitors.add(additionalPrinterColumnDetector);
+    visitors.add(traversedClassesVisitor);
 
-    def = builder
-        .accept(labelSelectorPathDetector)
-        .accept(additionalPrinterColumnDetector)
-        .accept(traversedClassesVisitor)
-        .build();
+    visitTypeDef(def, visitors);
 
     addDecorators(config, def, specReplicasPathDetector.getPath(),
         statusReplicasPathDetector.getPath(), labelSelectorPathDetector.getPath());
@@ -89,6 +96,54 @@ public abstract class AbstractCustomResourceHandler {
       resources.decorate(
           getPrinterColumnDecorator(name, version, path, type, column, description, format));
     });
+  }
+
+  private TypeDef visitTypeDef(TypeDef def, List<Visitor<TypeDefBuilder>> visitors) {
+    if (visitors.isEmpty()) {
+      return def;
+    }
+    if (parallel) {
+      return visitTypeDefInParallel(def, visitors);
+    } else {
+      return visitTypeDefSequentially(def, visitors);
+    }
+  }
+
+  private TypeDef visitTypeDefSequentially(TypeDef def, List<Visitor<TypeDefBuilder>> visitors) {
+    TypeDefBuilder builder = new TypeDefBuilder(def);
+    for (Visitor<TypeDefBuilder> visitor : visitors) {
+      builder.accept(visitor);
+    }
+    return builder.build();
+  }
+
+  private TypeDef visitTypeDefInParallel(TypeDef def, List<Visitor<TypeDefBuilder>> visitors) {
+    final ExecutorService executorService = Executors.newFixedThreadPool(
+        Math.min(visitors.size(), Runtime.getRuntime().availableProcessors()));
+    try {
+      List<CompletableFuture<Void>> futures = new ArrayList<>();
+      for (Visitor<TypeDefBuilder> visitor : visitors) {
+        futures.add(CompletableFuture.runAsync(() -> {
+          // in this case we're not building a new typedef,
+          // instead we just need to traverse the object graph.
+          TypeDefBuilder builder = new TypeDefBuilder(def);
+          builder.accept(visitor);
+        }, executorService));
+      }
+      try {
+        CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).get();
+      } catch (InterruptedException interruptedException) {
+        Thread.currentThread().interrupt();
+      } catch (ExecutionException ex) {
+        if (ex.getCause() instanceof RuntimeException) {
+          throw (RuntimeException) ex.getCause();
+        }
+        throw new RuntimeException(ex.getCause());
+      }
+    } finally {
+      executorService.shutdown();
+    }
+    return def;
   }
 
   /**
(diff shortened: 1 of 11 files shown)


andreaTP:
Hi @nicoloboschi , thanks for this PR! This sounds super interesting indeed!
But, since I have already experienced issues with concurrency + sundrio + crd-gen, I need to summon @metacosm and @iocanel to review this change 🙏

andreaTP:
@nicoloboschi are you still actively working on this? if that's the case, would you mind converting to Draft?

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractCustomResourceHandler.java
```diff
@@ -54,21 +60,53 @@ public void handle(CustomResourceInfo config) {
 
     ClassDependenciesVisitor traversedClassesVisitor = new ClassDependenciesVisitor(config.crClassName(), name);
 
-    TypeDefBuilder builder = new TypeDefBuilder(def);
-    if (config.specClassName().isPresent()) {
-      builder.accept(specReplicasPathDetector);
-    }
+    final ExecutorService executorService = Executors.newFixedThreadPool(
+      Math.min(5, Runtime.getRuntime().availableProcessors()));
```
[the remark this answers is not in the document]
nicoloboschi: They are not going to compete for any resource because they are all independent tasks (that's what I've seen but I don't know this code very deeply). 

I haven't measured the cache hit ratio but the results show the parallel execution is drastically faster.

iocanel (approved):
Overall, looks good.
Added a comment with two suggestion, which I consider nice ot have.

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractCustomResourceHandler.java
```diff
@@ -54,21 +60,51 @@ public void handle(CustomResourceInfo config) {
...
-    if (config.specClassName().isPresent()) {
-      builder.accept(specReplicasPathDetector);
-    }
+    final ExecutorService executorService = Executors.newFixedThreadPool(
+        Math.min(5, Runtime.getRuntime().availableProcessors()));
+    try {
+      List<CompletableFuture<Void>> futures = new ArrayList<>();
+      if (config.specClassName().isPresent()) {
+        futures.add(CompletableFuture.runAsync(() -> {
+          TypeDefBuilder builder = new TypeDefBuilder(def);
```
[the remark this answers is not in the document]
nicoloboschi: >However, it shouldn't make a difference as in this particular case we are not interested in building something (we completely ignore that), but instead we just need to traverse the object graph.

yes, exactly, we just do a LOT of recursive calls and the only output is stored in the visitor object itself (`getPath()`)

I've already tried with the var-args solution but it doesn't help.

Sharing my dummy benchmark numbers
- using quarkus-operator-sdk mojo to generate crd
- jdk 17
- Intel mac 6 core
- 8 CRD in the project with ~200 props in total, spec and status classes in each one
- 3 rounds per solution
Results:
- current version: ~40s
- concurrency of 5: ~18s
- var args:  ~45s

nicoloboschi:
@iocanel any chance we can merge this? 

this is my latest comment about performances comparison


- using quarkus-operator-sdk mojo to generate crd
- jdk 17
- Intel mac 6 core
- 8 CRD in the project with ~200 props in total, spec and status classes in each one
- 3 rounds per solution

Results:
- current version: ~40s
- concurrency of 5: ~18s
- var args: ~45s

nicoloboschi:
@iocanel @andreaTP any chance to see this pull moving forward? 🤞

manusa:
Hi @nicoloboschi 
I still need to review this and analyze if it might cause trouble in certain environments. AFAIK running the generation in several parallel processes might lead to inconsistent results.
One of the options is to make this a configuration option.

nicoloboschi:
> Hi @nicoloboschi I still need to review this and analyze if it might cause trouble in certain environments. AFAIK running the generation in several parallel processes might lead to inconsistent results. One of the options is to make this a configuration option.

@manusa I can totally add a flag. I haven't find any configuration present at the moment. Do you have any suggestion on how to make it configurable? 

For example, I could add a system property but I'm not sure how it is supposed to be set if someone is using the `crd-generator-apt` module

andreaTP:
> Do you have any suggestion on how to make it configurable?

Currently, the crd generator is purely using annotations and has no additional configuration:
https://github.com/fabric8io/kubernetes-client/tree/009974fe5fe634cbd6d1210780b99ff82e251f2d/crd-generator/api/src/main/java/io/fabric8/crd/generator/annotation

nicoloboschi:
- I've made it configurable and disabled by default.
- I added an option in `CRDGenerator` for enabling the parallel execution. 
- For the apt processor I've added an option `io.fabric8.crd.generator.parallel`. 
- I added the apt processor option example in the test pom.xml
```
<compilerArgs>
            <arg>-Aio.fabric8.crd.generator.parallel=false</arg>
</compilerArgs>
```
- I've added CRDGenerator tests with parallel enabled
- I tested locally the `crd-generator/tests` suite with parallel enabled but I've left it disabled since it's better to test the default configuration in this case

@andreaTP let me know if I missed something

andreaTP:
I just skimmed through on the phone, and, making parallel execution optional makes a lot of sense.
I think that the Maven setting is a bit "hidden" and you probably want to mention it in the docs.

I might be wrong, but it looks like there are a few tiny bits that are not source compatible (which is appropriate to merge before 6.4) and @metacosm would need to have a look and follow up in the Quarkus Operator SDK.

nicoloboschi:
I created the related pr in for the quarkus sdk: https://github.com/quarkiverse/quarkus-operator-sdk/pull/488 (it's only a draft since we need this to get released)

crd-generator/test/pom.xml
```diff
@@ -90,6 +90,14 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <compilerArgs>
+            <arg>-Aio.fabric8.crd.generator.parallel=false</arg>
```
andreaTP, line 97: this should be redundant as it's already the default, isn't it?

andreaTP:
Thanks for the follow-ups @nicoloboschi , LGTM.

andreaTP (approved):

crd-generator/test/pom.xml
nicoloboschi: yes it is. I wanted to make it explicit and leave it there as example since it's not trivial to add processor options

andreaTP: 👍

metacosm (approved):
LGTM  but I haven't checked downstream projects yet.

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractCustomResourceHandler.java
```diff
@@ -91,6 +98,54 @@ public void handle(CustomResourceInfo config) {
...
+    for (Visitor<TypeDefBuilder> visitor : visitors) {
+      builder.accept(visitor);
+    }
+    return builder.build();
+  }
+
+  private TypeDef visitTypeDefInParallel(TypeDef def, List<Visitor<TypeDefBuilder>> visitors) {
+    if (visitors.isEmpty()) {
+      return def;
+    }
```
gastaldi, line 120: Perhaps move this `if` to the `visitTypeDef` method since it is useful for `visitTypeDefSequentially` too?

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractCustomResourceHandler.java
```diff
@@ -91,6 +98,54 @@ public void handle(CustomResourceInfo config) {
...
+      for (Visitor<TypeDefBuilder> visitor : visitors) {
+        futures.add(CompletableFuture.runAsync(() -> {
+          // in this case we're not building a new typedef,
+          // instead we just need to traverse the object graph.
+          TypeDefBuilder builder = new TypeDefBuilder(def);
+          builder.accept(visitor);
+        }, executorService));
+      }
+      try {
+        CompletableFuture.allOf(futures.toArray(new CompletableFuture[] {})).get();
```
gastaldi, line 134:
```suggestion
        CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).get();
```

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractCustomResourceHandler.java
nicoloboschi: done

nicoloboschi:
@gastaldi I've addressed your comments, please review it again

doc/CRD-generator.md
```diff
@@ -442,3 +442,32 @@ The CRD generator will add the additional `labels`:
 | `io.fabric8.crd.generator.annotation.Labels`                 | Additional `labels` in `metadata`                                                     |
 
 A field of type `com.fasterxml.jackson.databind.JsonNode` is encoded as an empty object with `x-kubernetes-preserve-unknown-fields: true` defined.
+
+## Experimental
+
+### Generate CRDs in parallel
+Starting from 6.4, it's possible to speed up the CRDs generation by using parallel computation.
```
gastaldi, line 449: 6.4.0 is already released, since this change doesn't break the API it could be introduced in a patch 6.4.x release

crd-generator/api/src/main/java/io/fabric8/crd/generator/CRDGenerator.java
```diff
@@ -68,6 +69,11 @@ public CRDGenerator withOutput(CRDOutput<? extends OutputStream> output) {
     return this;
   }
 
+  public CRDGenerator withParallelGenerationEnabled(boolean parallel) {
```
gastaldi, line 72: Just an idea: what if the ExecutorService is passed instead? That would give a better control over the execution.

gastaldi (approved):
LGTM. I'd appreciate if you could please squash the commits. Good job!

crd-generator/api/src/main/java/io/fabric8/crd/generator/CRDGenerator.java
nicoloboschi: In general I think this makes sense. 
In this specific case, given the crd generation is a short-running task, I don't believe it's worth to manage the thread pool at higher level

nicoloboschi:
Fixed the doc, rebased to the current master and squashed @gastaldi

manusa (approved):
Awesome work, thx!

Issue #4574, opened 2022-11-10
Client.resource(resource).get() does not return latest resource


gyfora:
### Describe the bug

I have been trying to use the following code snippet to get the latest version of a given custom resource:
```
var latest = client.resource(resource).get();
```

However it seems that this does not actually return the latest version but simply the resource I passed to the call...

On the contrary when I do:
```
var latest = client.resources(MyCR.class, MyCRList.class)
                                    .inNamespace(ns)
                                    .withName(name)
                                    .get();
```  
it works and gives the latest resource.

This seems to be an inconsitency in the api.                                    

### Fabric8 Kubernetes Client version

6.2.0

### Steps to reproduce

Compare the output of:
```
var latest = client.resource(resource).get();
```

vs

```
var latest = client.resources(MyCR.class, MyCRList.class)
                                    .inNamespace(ns)
                                    .withName(name)
                                    .get();
```

### Expected behavior

They should both return the latest version of the resource. But only the second works

### Runtime

minikube

### Kubernetes API Server version

1.25.3@latest

### Environment

macOS

shawkins:
> This seems to be an inconsitency in the api.

You need to use fromServer() to force the fetch for a get when using resource(foo).  When the docs refer to the "current item" that does include the one passed into the resource method.

gyfora:
Maybe I just don't understand the design here but I find it extremely odd that a call to a client with a get returns the exact object that you passed to it.

I am sure many people will intuitively misunderstand this

shawkins:
> Maybe I just don't understand the design here but I find it extremely odd that a call to a client with a get returns the exact object that you passed to it.

I'm not sure what the rationale was for having this naming / functioning.  Feel free to capture something if want to propose different signatures.

> I am sure many people will intuitively misunderstand this

Sure, feel free to open a pr for any missing javadocs.

shawkins:
> Do you think it makes sense to change the default behaviour to help users? I think if they just want to get the same resource back they would not call the client.

This behavior was already how the client functioned for the existing resource and resourceList methods, so it carried forward once the resource method was generalized - as such existing usage wouldn't change.  I don't think it was spelled out in the release notes that you need to call fromServer when switching to the resource api from the other style - that was an over site.

As for changing the default behavior of what a get means that can be expanded to @manusa @rohanKanojia and others.

manusa:
> As for changing the default behavior of what a get means that can be expanded to...

[...]

PR #4651, opened 2022-12-08
Fix #3973: removing the need for parameter processing in Serialization

### LINKED ISSUE #3973: Further consolidate and expand resource methods

shawkins:
## Description
Moves #3972 further along by deprecating Serialization methods accepting parameters.  Also support for Parameterizable has been removed - it should have been rarely used.  The code to handle template parameters in now localized to processLocally. 

## Checklist


shawkins:
@manusa @rohanKanojia sorry I expanded this pr.  The changes are:
1. Deprecate Serialization taking responsibility for parameter replacement.
2. Consolidate all of the template parameter replacement logic to the TemplateOperationsImpl - it had been in three places.  To do this we are getting rid of the Parameterizable interface.  It was only needed in scenarios where there was a non-string parameter, which we now handle.  We also had overloaded the load logic in several places related to templates - however that causes a duplication in logic and makes the behavior harder to document / understand.  To replace the behavior of templates().load, there is now a templates().from method - which is appropriately documented.  I understand that all of this would be a breaking change, but since it only affects templates it may be acceptable for 6.x.  If not we can hold off until 7 and put deprecations in now.  
3. With a lot of the logic cleared out, I started to make more progress on refining away the need for openshift-client jar.  I can't continue on that path until #4574 is included as more conflicts will emerge.

Should I continue with this pr with the serialization refinements, or just wait for #4574 and this are merged before picking that up again?

shawkins:
Closing in favor of starting with #4683 

Will refine the second pr out of what's here / what's on #4662

PR #4678, opened 2022-12-14
fix #4663: changes to help prevent the client from waiting indefinitely

### LINKED ISSUE #4663: Seemingly indefinite wait at a get operation
A get operation failed to complete, far exceeding the default request timeout - which is not directly enforced by OperationSupport.waitForResult.  There's an assumption that the httpclient layer will enforce the timeout.  However if okhttp has exhausted it's max requests, it will not enforce the timeout while the request is enqueued.  We have seen this as an issue with websockets, which for some reason count against the request total - so watches / informers will eventually exhaust the max requests.

Two changes are possible here - some enforcement of the request / read timeout in waitForResult, and take responsibility in the okhttp httpclient logic for incrementing the max requests for long running requests (websockets, even http watches if possible, but that seems more difficult to detect).

shawkins:
## Description
Fix #4663

Addresses #4663 by enforcing the request timeout at a higher level and setting connection / request limits to unlimited.

@manusa in looking at the current state what I see is that after 64 websockets okhttp will stop accepting additional websockets or regular requests.  It appears that at least with http1 against the mockserver, you can start as many regular requests you want - waiting to consume the body does not hold the connection, it is immediately released back to the pool.

For jetty 64 concurrent regular requests is all it supports - https://www.eclipse.org/lists/jetty-dev/msg02788.html#:~:text=So%20if%20you%20send%20requests,next%20one%20will%20be%20rejected. maxConnectionsPerDestination.  This number does not affect websockets.  The only way this would be exhausted currently is for the user to have 64 long running requests going at the same time - http watches, log watches, etc. 

For jdk there doesn't seem to be built-in limitations.

You had mentioned possibly creating some abstract test across the http clients for this.  There were a couple of drawbacks to my initial attempts, so I haven't included that in this pr.

Based upon what we talked about for this meeting, we're ok with this breakage of the config properties and the potential for thread allocation issues with okttp - but at least those should be accompanied by a stacktrace.

@vietj we touched on this a little in our discussion about connection or request limits in vertx.  Because we may start a number of long running operations simultaneously (non-websockets could still be things like http watches or following a log), we'd like to simply to just turn off any limit by default.  If there are some system or other upper end limits that we should be aware of, we'd want to address later by ensuring that user is aware the system has become exhausted - the current behavior with okhttp makes it completely non-obvious why the fabric8 client seems to be doing nothing.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 #### _**Note**_: Breaking changes
 * Fix #4574: fromServer has been deprecated - it no longer needs to be called.  All get() operations will fetch the resource(s) from the api server.  If you need the context item that was passed in from a resource, load, or resourceList methods, use the item or items method.
 * Fix #4633: client.run().withRunConfig was deprecated.  Use withNewRunConfig instead.
+* Fix #4663: Config.maxConcurrentRequests and Config.maxConcurrentRequestsPerHost will no longer be used.  Instead they will default to unlimited for all clients.  Due to the ability of the fabric8 client to start long running requests (either websocket or regular http) and how this is treated by the underlying clients you can easily exhaust these values and enter a state where the client is unresponsive without any additional information on what is occurring.
 
 ### 6.3.1 (2022-12-15)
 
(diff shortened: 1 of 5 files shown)


oscerd (approved):

manusa:
Hi @shawkins 

I think we started working on similar stuff at the same time :)

I'm currently implementing the tests to ensure that at least a given amount (to be polished based on findings such as the ones you exposed + others) of concurrent connections can be achieved by **every* HttpClient implementation.

As you say, we might not be able to guarantee unlimited requests for all clients. If this is the case, then we should probably revert to providing the configuration option and throw an exception or log a warning in case an incompatible with the selected HttpClient implementation is provided.

Anyway, please let me finish the tests and then you can rebase these changes on top.

shawkins:
> I think we started working on similar stuff at the same time :)

My initial thought with this pr was to push for inclusion in 6.3.1, but then decided there is an easy workaround for okhttp and the jetty limitation should really never be hit, so that didn't seem necessary.

I realized the potential for overlap, so I didn't try to refine my tests beyond what would initially prove out these changes.

manusa (approved):

PR #4687, opened 2022-12-16
Test/simultaneous HttpClient connections

manusa:
## Description

Relates to #4663 
Relates to #4678

Hi @shawkins 

I'm still working on this, but maybe we want to merge it so you can rebase #4678 on top and make sure that the fixes are working.

For Jetty and HTTP/1.x I manged to tune it so at least 1024 concurrent connections are feasible.

I did try to add the test implementation for JDK, but it was not passing. As said, if you rebase your changes on top of this after we merge it maybe you can figure out why.

I'm still working on the tests, and maybe some other features. Again, if you think this can help with #4678, we can go ahead and merge it and I'll provide follow-up PRs.

## Type of change

## Checklist


shawkins:
> For Jetty and HTTP/1.x I manged to tune it so at least 1024 concurrent connections are feasible.

Yes changing the setMaxConnectionsPerDestination on the sharedHttpClient should address that.  

Adjusting the queue size seems like it would only be needed for circumstances of creating substantially more connections in a tight loop correct.  I think you're forcing it to be needed by the test introducing a wait in the websocket handling - that will only happen in very specific circumstances in our code.  Only if it's a ExecWatchInputStream that is not being read will the flow control introduce waits - it seems ok for the user to start seeing exceptions in this case.

> I did try to add the test implementation for JDK, but it was not passing. 

The test is probably not making the correct assumption.  It doesn't matter whether you are using 1 or 1024 connections, the jdk client will leave them open / cached.  I think there is a system property to control that.  It should matter more that processing is complete.  However what I had seen with okhttp / http1 with the mockserver is that processing would complete regardless with a higher number of open requests at a time because the connections were eligible for reuse even before the bodies were completely read by our layer.  I think your test confirms the same thing - even without any adjustments to the okhttp requests, the http1Connections test will pass.

As for what change goes in first, I don't have a preference.

oscerd (approved):

manusa:
> I think you're forcing it to be needed by the test introducing a wait in the websocket handling - that will only happen in very specific circumstances in our code

The test is in an early iteration stage. I introduced that blocking call there to ensure that the upgraded HTTP requests remained active. However, I'm quite sure I should be able to remove this and still provide an adequate assertion -WIP-.

> Adjusting the queue size seems like it would only be needed for circumstances of creating substantially more connections in a tight loop correct. 

Obviously the current state of the test required this. However, by inspecting the code I'm not sure that the default capped executor (200 threads) will also enforce some sort of connection limit, since the executor is used at the client connector level. If this is not the case and the executor is only used for message processing (hopefully I'll be able to determine this with the refactored test), I will remove the related changes.

> ...with a higher number of open requests at a time because the connections were eligible for reuse even before the bodies were completely read by our layer.

That's why I'm forcing a chunked response. My assumption here is that follow-up chunks are not requested until the response has begun processing. This behavior was tested with Jetty, and the test will fail in case the max requests are capped.

I haven't tried yet with OkHttp. If I'm not able to make the test fail, then we'll need to think in another option to keep the socket open.

manusa:
Hi @shawkins 

I'm still working on these, but they are on a better shape now. You might want to check another look in case you spot something else.

> Adjusting the queue size seems like it would only be needed for circumstances of creating substantially more connections in a tight loop correct

After refactoring the WebSocket test and removing the forced block in the WebSocket processing, it's no longer necessary to increase the queue. I'm still not sure why this queue is set at the connector level though.

> ...with a higher number of open requests at a time because the connections were eligible for reuse even before the bodies were completely read by our layer.

You were right, OkHttp was passing the test even when the response was not being processed yet.

I refactored the test so that the response headers are only sent after all of the request connections are established and waiting. This one seems to make the OkHttp implementation fail. Haven't checked yet if increasing the max number of connections makes the test pass, but I hope so.

I'm not sure if the HTTP/2 tests are still necessary or with these is enough.

Please share your thoughts.

httpclient-jetty/src/main/java/io/fabric8/kubernetes/client/jetty/JettyHttpClientBuilder.java
```diff
@@ -65,6 +68,11 @@ public JettyHttpClient build() {
       sharedWebSocketClient.setConnectTimeout(connectTimeout.toMillis());
     }
     sharedHttpClient.setFollowRedirects(followRedirects);
+    sharedHttpClient.setMaxConnectionsPerDestination(MAX_CONNECTIONS);
+    sharedWebSocketClient.getHttpClient().setMaxConnectionsPerDestination(MAX_CONNECTIONS);
```
shawkins, line 72: In looking at the websocket logic, the http connection pool appears only to be used for the establishment of the upgrade request - once the websocket is established it does not count against that pool.  I think your tests should pass without setting this, but of course setting it higher shouldn't be a problem either.

kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractSimultaneousConnectionsTest.java
```diff
@@ -0,0 +1,214 @@
...
+        client.consumeBytes(request, (value, asyncBody) -> asyncBody.consume())
+            .whenComplete((r, e) -> asyncResponses.add(r));
+      }
+      assertThat(activeConnections.await(20, TimeUnit.SECONDS)).isTrue();
+      assertThat(asyncResponses).extracting(HttpResponse::code).containsOnly(204);
+    } finally {
+      httpServer.stop(0);
+      httpExecutor.shutdownNow();
+    }
+    assertThat(serverSocketFactory.activeConnections()).isZero();
```
shawkins, line 126: Isn't this assertion problematic for JDK?

manusa: Haven't implemented the JDK tests yet. If it fails, maybe further adjusting the predicate might work. I'll check

manusa: OK, this one should be removed, thought it was part of the WS test. This is legacy from the previous test implementation.

httpclient-jetty/src/main/java/io/fabric8/kubernetes/client/jetty/JettyHttpClientBuilder.java
manusa: Consider the wording for the test where the `simultaneous` word was chosen instead of `concurrent`. 
Despite being a very unlikely scenario, what I'm trying to ensure is that regardless of concurrency, or simultaneity, if a client opens `n` connections at the same time and later on they all upgrade simultaneously, the implementation should work.

```
----connect------upgrade---->
----connect------upgrade---->
vs.
-----connect-----upgrade---------->
----------------connect---upgrade->
-connect----upgrade--------------->
```

manusa: So an additional test should be provided to assert this situation

manusa:
Added the tests for the Jdk HttpClient implementation. They are passing locally.

manusa:
Hi @shawkins 

This should be ready now. Please take a look whenever you can.

I finally merged #4678 first and then rebased this one on top.

Unless I missed something, the OkHttp fix was applied in the wrong place and the tests weren't passing (I guess they would have passed on a derived client instance though (:thinking:).
This PR also moves the dispatcher's connection limit to  the `OkHttpClientBuilderImpl` class where the rest of global settings are set.

I also included an additional test for WebSockets were we actually test that there can be 1024 simultaneous HTTP connections before upgrading (in relation to https://github.com/fabric8io/kubernetes-client/pull/4687#discussion_r1052403903).

manusa:
OK, there's still the issue with Windows :weary:

manusa:
After spending a few hours trying to figure out where the bottleneck is on Windows I've decided not to waste more time and disable the tests in this platform.

I'm not sure where the cap is coming from, but the number simultaneous connections varies from execution to execution but never goes beyond 500, and never grows despite having different timeouts (i.e. the issue is not that Windows is slower executing the test).

shawkins:
> Unless I missed something, the OkHttp fix was applied in the wrong place and the tests weren't passing (I guess they would have passed on a derived client instance though (thinking).

It's not about whether it's derived, it's about whether the builder was created using a Config.  The code to manipulate the dispatcher from the Config was in that spot initially, so it stayed there after my change.  It's of course fine to move it further down as it no longer depends on the Config and will be applicable to any HttpClient that is created (although we only create the config-less ones for token requests, and those aren't shared clients).

> After spending a few hours trying to figure out where the bottleneck is on Windows I've decided not to waste more time and disable the tests in this platform.

Is it a thread allocation issue?  For OkHttp that will be several thousand threads.

shawkins (approved):
LGTM

manusa:
> Is it a thread allocation issue? For OkHttp that will be several thousand threads.

It's non-deterministic and fails for all client implementations.
It's also strange because it fails with both web-servers (OkHttp Mock, + native HttpServer). At some point the number of concurrent requests reach a limit and no more threads _seam_ to be allocated.

AFAIK Windows should support an unlimited amount of threads or run out of memory, but none of this is happening.

PR #4694, opened 2022-12-19
[java-generator] Overridable package name

### LINKED ISSUE #4621: [java-generator] Make the package name overridable.

Most of the time we are switching codebases to use the `java-generator` the big diff is in the import.
e.g.:
https://github.com/bf2fc6cc711aee1a0c2a/e2e-test-suite/pull/417/files

Making the package name configurable/overridable externally is going to improve a lot in this regard, the only challenge would be when multiple CRDs are provided, we need an easy way to provide an intuitive `Map<String, String>` to perform the substitution.

_No response_

### Additional context

_No response_

andreaTP:
## Description
Fix #4621

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Fix #4637: all pod operations that require a ready / succeeded pod may use withReadyWaitTimeout, which supersedes withLogWaitTimeout.
 * Fix #4633: provided inline access to all RunConfig builder methods via run().withNewRunConfig()
 * Fix #4670: the initial informer listing will use a resourceVersion of 0 to utilize the watch cache if possible.  This means that the initial cache state when the informer is returned, or the start future is completed, may not be as fresh as the previous behavior which forced the latest version.  It will of course become more consistent as the watch will already have been established.
+* Fix #4694: [java-generator] Option to override the package name of the generated code.
 
 #### Dependency Upgrade
 
--- a/java-generator/cli/src/main/java/io/fabric8/java/generator/cli/GenerateJavaSources.java
+++ b/java-generator/cli/src/main/java/io/fabric8/java/generator/cli/GenerateJavaSources.java
@@ -23,6 +23,7 @@ import picocli.CommandLine.Command;
 import picocli.CommandLine.Option;
 
 import java.io.File;
+import java.util.Map;
 
 @Command(name = "java-gen", mixinStandardHelpOptions = true, helpCommand = true, versionProvider = KubernetesClientVersionProvider.class)
 public class GenerateJavaSources implements Runnable {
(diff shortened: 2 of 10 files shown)


rohanKanojia (approved):

rohanKanojia:
@andreaTP : Do you think we should add documentation for this new flag in [doc/java-generation-from-CRD.md](https://github.com/fabric8io/kubernetes-client/blob/master/doc/java-generation-from-CRD.md)

andreaTP:
@rohanKanojia good call!
We should re-generate the documentation from CLI and Maven: https://github.com/fabric8io/kubernetes-client/blob/master/doc/java-generation-from-CRD.md#usage
Tracked here: 4707 since I have a few improvements in the pipeline and I would like to do it once at the end to avoid continuous git conflicts.

CHANGELOG.md
```diff
@@ -12,6 +12,11 @@
 
 #### _**Note**_: Breaking changes
 
+### 6.4.0
+
+#### New features
+* Fix #4694: [java-generator] Option to override the package name of the generated code.
+
```
manusa, line 19: This should be moved to the SNAPSHOT section, I guess that since we released 6.3.1 while you were working on this you didn't notice.

manusa (commented):
Sorry for the belated review.

Besides the next _boring_ section, I think it would be good to update the docs too (however, you created another more generic issue for that).

While I like the solution, I'm not completely convinced about its intuitiveness.

As I see it, we're offering the possibility to users to map the final computed, conventional **packages** to a different name. However, this is under the assumption that the user is aware of what the final package name will be according to our internal convention (and this is were I'm probably making wrong assumptions).

So basically we're telling the user to configure the final step in the following flow:
`gorup -> convention.package -> custom.package`

I insist that maybe users would be more comfortable with what you proposed here and this is just to raise a point and a discussion. Wouldn't it be more intuitive for users to map the group/name to a package name:
```xml
<broker.amq.io>com.example.crd.amq</broker.amq.io>
```

Although now that I'm seeing this, my proposal has the problem of how to include the version (maybe nested, but this would be ugly).

Anyway, I'd just like to know your thoughts on this before probably merging it as it is. Otherwise, it's pretty cool ;) :raised_hands:

andreaTP:
@manusa you are right that the current implementation is not great in terms of usability, but I cannot find a better implementation of it 🙁 

The reasoning to be applied here is the following:
- integrate the generator into your project
- you are not satisfied with one or more of the package names
- you get the current and apply the override in the configuration

Matching on the plain `group` would have been slightly better, but you have then the problem with the `version` and everything becomes messy again.

Possible improvements over this might be to log the package names and provide the user with some information in the terminal ... but, again this is not going to be great ...

manusa (approved):

manusa (approved):

Issue #4701, opened 2022-12-20
Share openshift tokens across openshift clients


shawkins:
If the adapt call is made multiple times, this will create different OpenShiftOAuthInterceptors - each will maintain it's own token.

_Originally posted by @shawkins in https://github.com/fabric8io/kubernetes-client/issues/4673#issuecomment-1351503927_

shawkins:
This affects usage where the user needs an openshift client, but initially is provided with just the base kuberenetes client.  This should not be typical.  It's more likely that the instance is already an openshift client, which is also the workaround to this issue - make sure the root client that is passed around is the openshift client.

PR #4702, opened 2022-12-20
fix #4698: refining authentication mechanisms to be clearer

### LINKED ISSUE #4698: Remove unnecessary Basic Auth in Openshift requests.

Related with #4673
When execute a request with Openshift client, the token requested after a failure (403) is stored in an AtomicReference for future requests in OpenShiftOAuthInterceptor. 

But in the `before` method in the interceptor, there is a condition:
```
if (Utils.isNotNullOrEmpty(token)
        && (headers.headers(AUTHORIZATION).isEmpty() || Utils.isNullOrEmpty(headers.headers(AUTHORIZATION).get(0)))) {
      setAuthHeader(builder, token);
    }
```
that will never be true because the HttpClientUtils.HeaderInterceptor add a Basic Authorization header if your client has username and password configured; that it is necessary for the authentication requests performed by the OpenShiftOAuthInterceptor.
So, the result is the token stored is never reused, and every request to openshift performs the authentication.

A way to avoid the Basic header for Openshift.

If I try to walkaround like this:

```
....
new KubernetesClientBuilder().withConfig(config)
        .withHttpClientBuilderConsumer(builder -> {
          builder.addOrReplaceInterceptor("HEADER", new NoBasicHeaderInterceptor(config));
        })
        .withHttpClientFactory(factory)
        .withTaskExecutor(executorService).build().adapt(OpenShiftClient.class);

....

  private static final class NoBasicHeaderInterceptor implements Interceptor {
    private final Config config;

    private NoBasicHeaderInterceptor(final Config config) {
      this.config = config;
    }

    @Override
    public Interceptor withConfig(final Config config) {
      return new NoBasicHeaderInterceptor(config);
    }

    @Override
    public void before(final BasicBuilder builder, final HttpHeaders headers) {
      if (Utils.isNotNullOrEmpty(this.config.getOauthToken())) {
        builder.header("Authorization", "Bearer " + this.config.getOauthToken());
      }
      if ((this.config.getCustomHeaders() != null) && !this.config.getCustomHeaders().isEmpty()) {
        for (final Map.Entry<String, String> entry : this.config.getCustomHeaders().entrySet()) {
          builder.header(entry.getKey(), entry.getValue());
        }
      }
      if ((this.config.getUserAgent() != null) && !this.config.getUserAgent().isEmpty()) {
        builder.setHeader("User-Agent", this.config.getUserAgent());
      }
    }
  }
```

then, the first Authentication request to Openshift fails with 401 (the one to `.well-known/oauth-authorization-server` endpoint). Curiously, the second one, to the url returned by the previous request, it adds manually the Basic header.

A possible solution could be avoid Basic header filter for openshift, and add it manually for the first request too.

_No response_

### Additional context

_No response_

shawkins:
## Description

Fix #4698

Addresses #4698: this makes auth mechanisms more distinct.  It gives full control over setting the auth header to the token interceptors (which could be renamed).  

The proposed behavior:

For kubernetes: if username and password is supplied, then only basic auth will be used.  If not, then the token logic will be used.

For openshift: there is some confusing overlap between the usage of username / password and setting the token / token provider.  It makes sense to me to keep these distinct.  So if a username and password are provided, then we'll use that generate a token - after the first request(s) fail, once there has been a successful refresh the token will be reused.  If a username / password are not set, then we'll use the token from the config, which may also be provided by a token provider.

All of this would be added to appropriate documentation if it seems good.

There is additional logic to guard against sharing the token with requests where the user has used withRequestConfig to override the username / password.  @manusa seemed open to another issue that would deprecate withRequestConfig (either completely or specific fields).

## Type of change

## Checklist


shawkins:
@manusa and others.  Upon further review the openshift logic probably does need to stay somewhat intertwined.

So with this change the openshift logic will try to use the token (which could also be from a tokenprovider).  If that fails, it will then see if a username and password are supplied in the config passed into the interceptor.  If they are supplied, then we'll authenticate to get a fresh token - if we get a one, it will be persisted in the kubeconfig.

If a username and password aren't supplied we'll try to refresh the config (just like the base kubernetes logic) and use the updated token (we don't bother detecting if it hasn't actually changed).

Just like the kuberentes token refresh we set the token on the current config when there's been a possible change - this is an alternative to the atomicreference approach - I've modified here so that it's a volatile field for better multi-threaded behavior.

This also refactored the kuberentes token refresh such that the config passed into the interceptor is used to determine key details of the kubeconfig - what file it came from, and what the current context is.  Without that, we were relying on the static / default logic.

What's not to like:
- While the kubernetes logic clearly separates how username and password are used, the openshift logic does not.
- Being able to override the username / password in the requestconfig complicates things, seems like we shouldn't allow that with #4708.  In particular I'm a little fuzzy about what it should mean to openshift when change the username, but don't change the current context.
- Setting the token on the current Config seems a little hackish.  It does partially address #4701.  However we still aren't officially "sharing" tokens across different openshift clients if they are configured to use a username / password because there's no refreshing of the config in that case.  A workaround is that if you pass an OpenShiftConfig to the KuberentesClientBuilder that will be used by all of the OpenShiftClients created by adapt calls.  Alternatives could include refining the OAuthTokenProvider for our internal usage as well.

shawkins:
oc behavior:

https://github.com/openshift/oc/blob/a4cfdc79461db44d482b315e4e894f32d6c68ae2/pkg/cli/login/loginoptions.go#L197

- if we have a token try that first.
- if not and there's no password, then match the username against the kubeconfig (we are already doing this the first time we load the config)
- otherwise use the password to obtain a token

After the config is updated, then it is saved here https://github.com/openshift/oc/blob/a4cfdc79461db44d482b315e4e894f32d6c68ae2/pkg/cli/login/login.go#L179

We seem to be reasonable mimicking this behavior.

The only additional corner case I can think of is that the existing code was somehow trying to account for the case of using the OpenShiftClient against a non-openshift server that was configured to use basic auth, in which case it was trying to do the basic auth first.  However this does not seem to be supported by oc, so it should be acceptable to not do this.

Also oc does support a few other auth schemes - https://github.com/openshift/oc/blob/33f8aed30246e726258c51be27c21c60a2b005f4/pkg/helpers/tokencmd/request_token.go#L92 - but that is beyond the scope of this issue.

shawkins:
Test runs not withstanding this should be ready...

manusa (approved):
LGTM, thx :raised_hands: 

The analysis here is :top: https://github.com/fabric8io/kubernetes-client/pull/4702#issuecomment-1377574025

PR #4712, opened 2022-12-26
feat (kubernetes-client) : Add DSL support for creating ConfigMap from file (#4184)

### LINKED ISSUE #4184: How Can I Create ConfigMap From File
Hi, I want to create a configmap from file, is there any api implements the action `kubectl create configmap [name] --from-flie=`

rohanKanojia:
## Description

Fix #4184

+ Added ConfigMapResource for modifying default ConfigMap operations
+ Added FromFileCreatable DSL method to provide file
+ Move KubernetesResourceUtilTest to `io.fabric8.kubernetes.client.utils`

## Type of change

 - [X]  Feature (non-breaking change which adds functionality)

## Checklist


oscerd (approved):

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/dsl/FromFileCreatable.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.fabric8.kubernetes.client.dsl;
+
+public interface FromFileCreatable<T> {
```
shawkins, line 18: Do you anticipate using this across types?  If not, it may be best to merge with ConfigMapResource or wherever this functionality lands.  Templates have a similar bit of functionality, but it's based upon load as it does not support directories.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/dsl/ConfigMapResource.java
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.fabric8.kubernetes.client.dsl;
+
+import io.fabric8.kubernetes.api.model.ConfigMap;
+
+public interface ConfigMapResource extends Resource<ConfigMap>,
```
shawkins, line 20: As a method on ConfigMapResource you can do something like configMaps().resource(item).fromFile, or configMaps().load(stream).fromFile.  This could be put at a higher level, like configMaps().fromFile(name, path), to eliminate a bit of overlap between the context item and what's coming from the file.

shawkins (commented):
Looks good, just a couple of thoughts about the dsl method.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/dsl/FromFileCreatable.java
rohanKanojia: I thought this might be useful in case we decide to support DSL method for creating Secrets from file in future. There is a `--from-file` option in `kubectl` for Secrets as well. Do you think this makes sense? Otherwise I can merge this with ConfigMapResource.

shawkins: > Do you think this makes sense?

That make sense, just make sure to remove the reference to ConfigMap.  Based upon kubectl this could even be a vararg of dirOrFilePath.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/dsl/ConfigMapResource.java
rohanKanojia: I moved it a level up by making `client.configMaps()` return `FromFileCreatableOperation` but now it's not able to use  client like this `client.configMaps().inNamespace("foo").fromFile(..)`

Do you think we should consider this scenario? There is also an option to add an overloaded `fromFile` method that receives `ObjectMeta` instead of `name`

shawkins: > but now it's not able to use client like this client.configMaps().inNamespace("foo").fromFile(..)

To support that either MixedOperation would need to be generic to allow for a refined NonNamespaceOperation type, or there's NamespaceableResource - that allow a Resource to have a namespace specified when the root operation does not have one.

> There is also an option to add an overloaded fromFile method that receives ObjectMeta instead of name

Makes sense, that would make this more generalized than what kubectl supports.

manusa: Considering both semantics and consistency, I'd expect that `client.resource` and `client.configMaps().fromFile` did exactly the same. 

So IMHO (and just with the context from reading this thread) I'd expect;
`client.configMaps().fromFile($theFile).inNamespace("default").operation()`

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/dsl/FromFileCreatable.java
```diff
@@ -0,0 +1,27 @@
...
+  /**
+   * Create new resource from a directory or file contents.
+   *
+   * @param name resource name
+   * @param dirOrFilePath a file or directory path
+   * @return Resource instance for operations to do with this resource
+   */
+  T fromFile(String name, String... dirOrFilePath);
```
manusa, line 26: I don't really like the fact that we are using Strings to specify files or paths.
What are the advantages of this vs. `T fromFile(Path path)` or `T fromFile(File file)`?

manusa (changes requested):
Looks good, the only change I'd request would be to make private any of the utility methods that are not used elsewhere (KubernetesResourceUtil).

rohanKanojia:
I've exposed these methods in KubernetesResourceUtil publicly because I want to later use them in Eclipse JKube in place of these methods in [`org.eclipse.jkube.kit.enricher.api.util.KubernetesResourceUtil`](https://github.com/eclipse/jkube/blob/master/jkube-kit/enricher/api/src/main/java/org/eclipse/jkube/kit/enricher/api/util/KubernetesResourceUtil.java#L632-L705):

- KubernetesResourceUtil.createConfigMapEntry
- KubernetesResourceUtil.isFileWithBinaryContent
- KubernetesResourceUtil.addNewEntriesFromDirectoryToExistingConfigMap
- KubernetesResourceUtil.addNewEntryToExistingConfigMap

manusa:
> I've exposed these methods in KubernetesResourceUtil publicly because I want to later use them in Eclipse JKube in place of these methods in [`org.eclipse.jkube.kit.enricher.api.util.KubernetesResourceUtil`](https://github.com/eclipse/jkube/blob/master/jkube-kit/enricher/api/src/main/java/org/eclipse/jkube/kit/enricher/api/util/KubernetesResourceUtil.java#L632-L705):
> 
> * KubernetesResourceUtil.createConfigMapEntry
> * KubernetesResourceUtil.isFileWithBinaryContent
> * KubernetesResourceUtil.addNewEntriesFromDirectoryToExistingConfigMap
> * KubernetesResourceUtil.addNewEntryToExistingConfigMap

What for? shouldn't JKube just use the client's ConfigMap from file method for whatever it's trying to achieve?

rohanKanojia:
In JKube, we have functionality (ConfigMap fragment's `jkube.eclipse.org/cm/` annotation) that requires adding entries from a file, directory, or annotations to the existing ConfigMap.

manusa:
Let's discuss this on a call. We cannot expose stuff in the client just because JKube might need the internals. I haven't checked our current implementation in JKube, but I'm sure this can be easily walked around with the new client functionality.

rohanKanojia:
@manusa : I've marked internal methods as `private`. I think we should be able to work around this in JKube by creating temporary ConfigMaps and merging them with main ConfigMap

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/dsl/FromFileCreatable.java
shawkins: Comparing to kubectl, the strings can actually be key file pairs - key=some.file

manusa:
> > @manusa : I've marked internal methods as private. I think we should be able to work around this in JKube by creating temporary ConfigMaps and merging them with main ConfigMap
> 
> If it's too cumbersome, I'm fine with doing something similar to what Rohan had initially - add a method to ConfigMapResource for adding an entry - configMaps().resource(x).addEntry(name, key, path).replace()

Maybe we can consider this as a future improvement. It certainly looks useful for the DSL interface.

manusa (approved):

manusa (changes requested):
The tests use path separators and aren't suitable for windows

manusa (approved):

manusa (changes requested):
After further checking the commits and of the the fix you provided yesterday I saw you actually didn't address this comment https://github.com/fabric8io/kubernetes-client/pull/4712#pullrequestreview-1240043221

Let's at least discuss it before we proceed.

rohanKanojia:
I changed the vararg type from String to Path in [FromFileCreatable](https://github.com/rohanKanojia/kubernetes-client/blob/pr/issue4184/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/dsl/FromFileCreatable.java#L28) . Isn't this what you were referring to?  Or you also meant removing `name` parameter?

manusa:
> I changed the vararg type from String to Path in [FromFileCreatable](https://github.com/rohanKanojia/kubernetes-client/blob/pr/issue4184/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/dsl/FromFileCreatable.java#L28) . Isn't this what you were referring to? Or you also meant removing `name` parameter?

In my [first comment](https://github.com/fabric8io/kubernetes-client/pull/4712#pullrequestreview-1240043221) I expected a change to either `fromFile(Path path)` or `fromFile(File file)`.

The current implementation makes it even worst, providing a variable Path argument makes it extremely confusing for users (i.e. what is happening to those paths, are they concatenated? are they all considered? what if the first path is relative and the second is absolute?)

rohanKanojia:
Vararg parameter was suggested by @shawkins in order to align it better with `kubectl`:
```
    --from-file=[]:
	Key file can be specified using its file path, in which case file
	basename will be used as configmap key, or optionally with a key and
	file path, in which case the given key will be used.  Specifying a
	directory will iterate each named file in the directory whose basename
	is a valid configmap key.
```

manusa:
> Vararg parameter was suggested by @shawkins in order to align it better with `kubectl`:
> 
> [#4712 (comment)](https://github.com/fabric8io/kubernetes-client/pull/4712#discussion_r1057915172)
> 
> ```
>     --from-file=[]:
> 	Key file can be specified using its file path, in which case file
> 	basename will be used as configmap key, or optionally with a key and
> 	file path, in which case the given key will be used.  Specifying a
> 	directory will iterate each named file in the directory whose basename
> 	is a valid configmap key.
> ```

As I said, it's better if we discuss this. The current implementation in no way aligns to what kubectl does when specified with key-file.

rohanKanojia:
At the moment `fromFile(String name, Path... dirOrFilePath)` corresponds to ConfigMap's name and list of files/directories that would get added as entries in ConfigMap's data value

manusa:
> At the moment `fromFile(String name, Path... dirOrFilePath)` corresponds to **ConfigMap's name** and list of files/directories that would get added as entries in ConfigMap's data value

So to align with kubectl what you want is `configMaps().fromFile($key, $file).fromFile($key, $file2).fromFile($key2, $file3).fromFile($fileKeyFromName);`.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesResourceUtil.java
```diff
@@ -457,6 +470,92 @@ public static Secret createDockerRegistrySecret(String dockerServer, String user
...
 
+  /**
+   * Create new ConfigMap from files/directories
+   *
+   * @param name name of Configmap to create
+   * @param dirOrFilePaths a var-arg for directory of file paths.
+   * @return ConfigMap with data as key-value pair of file names and their contents
+   * @throws IOException in case of failure while reading file
+   */
+  public static ConfigMap createNewConfigMapFromDirOrFiles(final String name, final Path... dirOrFilePaths)
```
sunix, line 481: maybe rename it to `createConfigMapFromDirOrFiles`

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesResourceUtil.java
```diff
@@ -457,6 +470,92 @@ public static Secret createDockerRegistrySecret(String dockerServer, String user
...
+          .onMalformedInput(CodingErrorAction.REPORT)
+          .onUnmappableCharacter(CodingErrorAction.REPORT)
+          .decode(ByteBuffer.wrap(bytes));
+      return false;
+    } catch (CharacterCodingException e) {
+      return true;
+    }
+  }
+
+  private static void addNewEntriesFromDirectoryToExistingConfigMap(ConfigMapBuilder configMapBuilder, final Path path)
```
sunix, line 518: Maybe rename it to `addEntriesFromDirectoryToExistingConfigMap`

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesResourceUtil.java
```diff
@@ -457,6 +470,92 @@ public static Secret createDockerRegistrySecret(String dockerServer, String user
...
+        try {
+          addNewEntryToExistingConfigMap(configMapBuilder, createConfigMapEntry(file.getFileName().toString(), file), file);
+        } catch (IOException e) {
+          throw new IllegalArgumentException(e);
+        }
+      });
+    }
+  }
+
+  private static void addConfigMapEntryFromFile(ConfigMapBuilder configMapBuilder, final String name, final String key,
```
sunix, line 531: maybe rename it to `addEntryFromFileToExistingConfigMap`

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesResourceUtil.java
```diff
@@ -457,6 +470,92 @@ public static Secret createDockerRegistrySecret(String dockerServer, String user
...
+      final Path file) throws IOException {
+    configMapBuilder.withNewMetadata().withName(name).endMetadata();
+    String entryKey = Optional.ofNullable(key).orElse(file.toFile().getName());
+    Map.Entry<String, String> configMapEntry = createConfigMapEntry(entryKey, file);
+    addNewEntryToExistingConfigMap(configMapBuilder, configMapEntry, file);
+  }
+
+  private static void addNewEntryToExistingConfigMap(ConfigMapBuilder configMapBuilder, Map.Entry<String, String> entry,
```
sunix, line 539: addEntryToExistingConfigMap

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesResourceUtil.java
```diff
@@ -457,6 +470,92 @@ public static Secret createDockerRegistrySecret(String dockerServer, String user
...
+      final Path file)
+      throws IOException {
+    if (isFileWithBinaryContent(file)) {
+      configMapBuilder.addToBinaryData(entry.getKey(), entry.getValue());
+    } else {
+      configMapBuilder.addToData(entry.getKey(), entry.getValue());
+    }
+  }
+
+  private static void addNewEntriesFromDirOrFileToExistingConfigMap(ConfigMapBuilder configMapBuilder, final String name,
```
sunix, line 549: `addEntriesFromDirOrFileToExistingConfigMap`

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesResourceUtil.java
sunix: naming: should we omit `Existing` ? would be `addEntryConfigMap`

sunix (commented):
Few minor naming suggestions:
- Simplification
  - `addNewEntry*` to `addEntry*`
  - `*ToExistingConfigMap` should we omit `Existing` ?
- Aligning naming through all the method:
    - you should choose `ConfigMapEntry...` or just `Entry` but not one in some method and the other in another method.
     - Same for `ToExistingConfigMap`: sometime used in the name, sometime not ?

rohanKanojia:
I have updated `fromFile` to be a chained method instead of varargs. I've also changed method names as suggested by Sun

shawkins:
@rohanKanojia @manusa it may be simpler from the start to just allow fromFile to modify the existing resource -  https://github.com/fabric8io/kubernetes-client/compare/master...shawkins:rohanKanojia-pr/issue4184?expand=1

sunix (approved):

shawkins:
@rohanKanojia any thoughts on https://github.com/fabric8io/kubernetes-client/compare/master...shawkins:rohanKanojia-pr/issue4184?expand=1 - since the logic here is already trying to account for modifying the current / context item with the usage of fromFile, that could just be exposed.

Here you can do: 

configMaps().fromFile(...).withName("x").replace() - this will create a new configmap x, add a file entry, then perform the replace (or any other resource operation)

With that other commit we have:

configMaps().withName("x").fromFile(...).replace() - that will create or edit x if it exists to add the file entry, then perform the replace (or any other resource operation)

The advantages are:
- gives the ability to directly manipulate an existing configmap
- requires fewer interface changes as it's based upon a resource subclass, rather than an operation customization

The disadvantages are:
- the method name isn't as semantically clear, you have to read what fromFile does to understand that it can create a new context item based upon the current if it exists either locally or on the server.  So this likely would need a different method and interface name.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/dsl/NameableNamespaceableResource.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.fabric8.kubernetes.client.dsl;
+
+public interface NameableNamespaceableResource<T> extends Resource<T> {
+  Resource<T> withName(String name);
```
shawkins, line 19: This could be NamespaceableResource so that withName and inNamespace could be called in either order.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/core/v1/ConfigMapOperationsImpl.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+    return new ConfigMapOperationsImpl(context.withName(name).withItem(item));
+  }
+
+  private ConfigMap createNewOrMergeWithExistingConfigMap(String key, Path dirOrFilePath) {
+    try {
+      ConfigMap item = getItem();
+      ConfigMap tempConfigMap = createNewConfigMapWithEntry(key, dirOrFilePath);
+
+      ConfigMap mergedConfigMap = mergeConfigMapData(item, tempConfigMap);
```
shawkins, line 75: Couldn't this be done with just the builder logic? https://github.com/fabric8io/kubernetes-client/compare/master...shawkins:rohanKanojia-pr/issue4184?expand=1#diff-7ff2d3334a0db4e97ce2cd61970f03169cb05b1ee85885b1fcddd79230bf9196R59 

Or is the merging logic needed for another purpose?

rohanKanojia:
I had talked to @manusa about this, we agreed let's get a basic version working merged first then we can iterate over making improvements.

shawkins:
> I had talked to @manusa about this, we agreed let's get a basic version working merged first then we can iterate over making improvements.

It's more of an alternative api, rather than an improvement - the version you have basically provides the same functionality under the covers.  Modifying the api after a release will unfortunately require more effort as it will be a breaking change.

rohanKanojia:
Okay, Shall I cherry-pick your commit onto my branch? Or do you want to take over this PR?

shawkins:
> Okay, Shall I cherry-pick your commit onto my branch? Or do you want to take over this PR?

Not necessarily either - if the desire is to keep fromFile as a root operation for creating a configmap from scratch, then we'll just move forward with what you have, and I did leave a couple of review comments just in case.

If fromFile is to be more general, then we can consider what's in the other commit.  If the implicit behavior is too confusing, it could be explicit with:

configMaps().withName("x").fromFile(...)... - will only work if "x" exists

configMaps().withName("x").newItem().fromFile(...)... - manipulates an empty configmap

rohanKanojia:
Let's discuss this on the Triage call next Wednesday to decide which approach we should pick.

rohanKanojia:
@shawkins : +1 from my side for 2 and 3.

rohanKanojia:
Superceded by #5173

PR #4732, opened 2023-01-04
fix #4729: completablefuture cancel will close the underlying resource

### LINKED ISSUE #4729: Ensure websockets and http responses are closed if the future is canceled
Related to #4726 and #4720 there was an assumption made based upon the jdk implementation that canceling the returned future from buildAsync or consumeBytes would close the the underlying resource if it occurred before the result were delivered.  However our intervening logic means that we're dealing with a downstream future that is canceled - the upstream will still complete normally and could leak the connection.  This will only happen in cases where the user manually closes something like a watch or a port forward before it initially completes.

shawkins:
## Description

Fix #4729 

@manusa  in reviewing the other recent issues along these lines, I think this is another corner-ish case.  It addresses the possibility of a resource leak due to an ill-defined expectation of cancelling the CompletableFutures returned by the http client layer.  If the cancel happens before the future is completed, then the resource may not get closed.

The fix here is to bake that responsibility into the standardhttpclient.  A different way of addressing this is to instead have HttpClient return a CompletionStage - which makes it clear that you are instead supposed to "cancel" via a thenAccept or whenComplete handler that closes things out.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/StandardHttpClient.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/StandardHttpClient.java
@@ -18,7 +18,9 @@ package io.fabric8.kubernetes.client.http;
 
 import io.fabric8.kubernetes.client.http.AsyncBody.Consumer;
 import io.fabric8.kubernetes.client.http.WebSocket.Listener;
+import io.fabric8.kubernetes.client.utils.Utils;
 
+import java.io.Closeable;
 import java.nio.ByteBuffer;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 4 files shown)


shawkins:
Based upon our call we agreed this doesn't generally happen - only in a couple of places do we return something that would have a cancelable future that is not already completed.  Also for sendAsync it would only happen after the initial request timeout.

But it does seem like a good idea to make these changes as complete as possible as to not necessitate closure via reference tracking, so I'll update the sendAsync logic as well.

shawkins:
@manusa should be ready now.  Upstream cancellation support has been added for all of the futures returned at a top level - which matches some of our usage pattern.  It's important to consider that this isn't necessarily a great practice - you must cancel the return future.  Any new downstream futures (thenApply... etc.) will not cause the cancellation of the upstream.  CompletionStage is probably the better api choice in the long run.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/StandardHttpClient.java
```diff
@@ -72,7 +84,23 @@ public CompletableFuture<HttpResponse<AsyncBody>> consumeBytes(HttpRequest reque
       });
     }
 
-    return cf;
+    return withUpstreamCancelation(cf, AsyncBody::cancel);
+  }
+
+  static <V> CompletableFuture<HttpResponse<V>> withUpstreamCancelation(CompletableFuture<HttpResponse<V>> cf,
```
rohanKanojia, line 90:
```suggestion
  static <V> CompletableFuture<HttpResponse<V>> withUpstreamCancellation(CompletableFuture<HttpResponse<V>> cf,
```

rohanKanojia (approved):

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/StandardHttpClient.java
shawkins: This looks to be a US vs British spelling difference https://www.grammarly.com/blog/canceled-vs-cancelled/ - since Java also uses the double L, I'll change it here also.

manusa (approved):

PR #4736, opened 2023-01-05
fix #4733 refinements to the connections test

### LINKED ISSUE #4733: deflake concurrent connection tests
On two recent runs I'm seeing issues with the 20 second wait https://github.com/fabric8io/kubernetes-client/actions/runs/3839873556/jobs/6538227145#step:4:38275 in the websocket test (one with jdk the other with okhttp).  Running locally, I've reproduced that and sometimes failed on one of the other tests as well.

One issue could be a problem with the HttpServer - if you remove the waits before returning the response, then jetty seems to consistently fail at 200 requests.  JDK seems to fail at around 400 requests.  Checking the websocket futures it's clear in this scenario that we're flooding the server to the point of failure: 

```Caused by: org.eclipse.jetty.websocket.api.exceptions.UpgradeException: org.eclipse.jetty.websocket.core.exception.UpgradeException: Failed to upgrade to websocket: Unexpected HTTP Response Status Code: 500 Internal Server Error
	at org.eclipse.jetty.websocket.common.JettyWebSocketFrameHandler.convertCause(JettyWebSocketFrameHandler.java:532)
	... 27 more
Caused by: org.eclipse.jetty.websocket.core.exception.UpgradeException: Failed to upgrade to websocket: Unexpected HTTP Response Status Code: 500 Internal Server Error
	... 21 more
java.util.concurrent.ExecutionException: io.fabric8.kubernetes.client.http.WebSocketHandshakeException
``` 

The test logic is not aware of that possibility as the websocket future is not being checked.

The strategy of introducing a wait before returning the headers also holds a thread per request.  The tight loop creation of requests for jdk in particular is prone to creating an additional thousand threads because the tasks are executed to quickly against a cached thread pool for thread reuse.  This means there can be 2 or more thousand threads running at the same time (I believe okhttp will use 2 threads per websocket as well), which could be an issue on older / smaller vms (you'll need several gigs of stack space).  My local system doesn't seem to have a thread allocation failure until around 32000 threads, so I'm not sure how much this is an issue with my local testing.  I did notice that with the header waits left in I could get more successful test completions with a wait longer than 20 seconds - sometimes up to 60 seconds - but wasn't able to quickly track down what was contributing to that additional time.

cc @manusa

shawkins:
## Description
A possible resolution for #4733 - websocket connection attempts are made tolerant to server-side errors.  At least jdk and jetty seem prone to flooding the server and getting a 500 response before the context is ever created.  It also allows context calls to delegate the response delay to a separate thread pool - this greatly reduces the threads created / held for those tests.  It also relaxes the latch wait to make it more tolerant to longer running behavior we sometimes see on github.  We could reduce thread usage more by slowing down the connection creation, or using a different executor for jdk.

Since the http1WebSocketConnections makes use of the OkHttp server it will still have quite a few threads in play - 2000 on the server side and up to 1000 on the client side (for OkHttp).

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractSimultaneousConnectionsTest.java
+++ b/kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractSimultaneousConnectionsTest.java
@@ -16,6 +16,7 @@
 package io.fabric8.kubernetes.client.http;
 
 import com.sun.net.httpserver.HttpServer;
+import io.fabric8.kubernetes.client.utils.Utils;
 import okhttp3.Protocol;
 import okhttp3.Response;
 import okhttp3.WebSocketListener;
@@ -51,6 +52,7 @@ import java.util.stream.IntStream;
 import javax.net.ServerSocketFactory;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.fail;
 
 public abstract class AbstractSimultaneousConnectionsTest {
 
@@ -62,6 +64,7 @@ public abstract class AbstractSimultaneousConnectionsTest {
   private RegisteredServerSocketFactory serverSocketFactory;
   private MockWebServer mockWebServer;
   private ExecutorService httpExecutor;
+  private ExecutorService responseService;
   private HttpServer httpServer;
 
   private HttpClient.Builder clientBuilder;
@@ -72,6 +75,7 @@ public abstract class AbstractSimultaneousConnectionsTest {
     mockWebServer = new MockWebServer();
     mockWebServer.setServerSocketFactory(serverSocketFactory);
     httpExecutor = Executors.newCachedThreadPool();
+    responseService = Executors.newFixedThreadPool(4);
     httpServer = HttpServer.create(new InetSocketAddress(0), 0);
     httpServer.setExecutor(httpExecutor);
     httpServer.start();
@@ -84,6 +88,7 @@ public abstract class AbstractSimultaneousConnectionsTest {
     mockWebServer.shutdown();
     httpServer.stop(0);
     httpExecutor.shutdownNow();
+    responseService.shutdownNow();
   }
 
   protected abstract HttpClient.Factory getHttpClientFactory();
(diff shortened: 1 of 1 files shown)


manusa:
> It also allows context calls to delegate the response delay to a separate thread pool - this greatly reduces the threads created / held for those tests.

IMHO the `http1Connections` test should not be altered. This one is checking how the client would behave in the harshest of conditions when a user opens hundreds of simultaneous connections to the cluster (e.g. creating informers for multiple resources in a cluster with hundreds of namespaces -jenkins ci-). This is also assuming that this test is not causing trouble as it is in the CI (I haven't seen it, neither it's reported in the issue, but maybe it is causing trouble).

> websocket connection attempts are made tolerant to server-side errors.

This suite is obviously causing trouble with websockets and resource management at the CI level or maybe even for developer local systems. As [I mentioned in the issue](https://github.com/fabric8io/kubernetes-client/issues/4733#issuecomment-1371868152), I'd prefer to reduce the number of connections (maybe 512), instead of tampering with the test.

Also the fact that `http1WebSocketConnectionsBeforeUpgrade` fails randomly but `http1Connections` might be revealing problems that we may be able to fix in our implementation (also based on the assumption that `http1Connections` is not failing).

I'm also not very fond of the retry approach. I'm more inclined for the throttling (with maybe some jitter) proposed in #4748 to avoid the server flooding.

As a side note, one of the future changes I'd like to see is the removal of the OkHttp dependency for our mock server. In the end we might end up implementing a our own HTTP server.

shawkins:
> IMHO the http1Connections test should not be altered. 

I don't follow your reasoning here. This is not altering the test as much as it is altering the server side behavior - it is unnecessarily holding threads.

> I'm also not very fond of the retry approach. I'm more inclined for the throttling (with maybe some jitter) proposed in https://github.com/fabric8io/kubernetes-client/pull/4748 to avoid the server flooding.

I'm mostly fine with either, but it brings into question how do you know how much throttling will be needed.  Part of the issue is that we are simply ignoring exceptions for a failed connection - we have no idea if that is due to flooding or not and if we still see failures at 1 ms, we'll have to try 2, then 4 etc.  With the retry approach as coded, we will hit the desired concurrency within the desired window without further guesswork.

manusa:
Let's discuss later today, I think it'll be quicker discussing this in a call.

shawkins:
We'll move forward with #4752

PR #4740, opened 2023-01-06
Fix #4739: Mock server doesn't handle optimistic concurrency control

### LINKED ISSUE #4739: Mock server doesn't handle optimistic concurrency control
### Describe the bug

When using `@EnableKubernetesMockClient` the behaviour of `replace()` with a stale resource version does not reflect the behaviour of a real kubernetes apiserver. A real server will respond with an HTTP 409 (conflict) status code and associated message, but the mock honours the request and returns HTTP 200.

SNAPSHOT

### Steps to reproduce

The following code should throw a `KubernetesClientException`, but does not

```java
// Create a resource
final ConfigMap original = client.resources(ConfigMap.class).resource(
        new ConfigMapBuilder().withNewMetadata().withName("name").endMetadata()
          .addToData("key", "value").addToData("change", "me").build())
      .create();

// Replace it, changing the metadata.resourceVersion
final ConfigMap replacementV1 = new ConfigMapBuilder(original).addToData("key", "value2").build();
client.resources(ConfigMap.class).resource(replacementV1).replace();

// Replace it again, but using the original's resource version
final ConfigMap replacementV2 = new ConfigMapBuilder(original).addToData("key", "value3").build();
client.resources(ConfigMap.class).resource(replacementV2).lockResourceVersion(original.getMetadata().getResourceVersion()).replace();
```

### Expected behavior

The final client call should throw `KubernetesClientException` with a 409 status code and a message indicating that the resource has been modified. The full message is something like 

> Failure executing: PUT at: https://192.168.39.15:8443/api/v1/namespaces/default/configmaps/name. 
> Message: Operation cannot be fulfilled on configmaps "name": the object has been modified; 
> please apply your changes to the latest version and try again. 

### Runtime

minikube

1.25.3@latest

### Environment

Linux

_No response_

### Additional context

_No response_

tombentley:
## Description

This PR fixes #4739.

A couple of minor points worthy of discussion:

1. I added the check to the path for `PATCH` as well as `PUT` requests, but I haven't added a unit test for this case.
2. When constructing the error message Kubernetes refers to the resource type using the singular name of the resource (e.g. `configmap`), but I don't have ready access to that in the `KubernetesCrudDispatcherHandler`, so I've used the `metadata.kind` instead (e.g. `ConfigMap`). Therefore there is a slight difference in the `KubernetesClientException` error message.

## Type of change

## Checklist

### DIFF
--- a/junit/kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/crud/PatchHandler.java
+++ b/junit/kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/crud/PatchHandler.java
@@ -63,6 +63,7 @@ public class PatchHandler implements KubernetesCrudDispatcherHandler {
       fullPatch = persistence.merge(currentResource, requestBody);
     }
     validatePath(query, fullPatch);
+    validateResourceVersion(currentResource, fullPatch);
 
     final JsonNode updatedResource;
     if (isStatusPath(path)) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Bugs
 
 #### Improvements
+* Fix #4739: honor optimistic concurrency control semantics in the mock server for `PUT` and `PATCH` requests.
 
 #### Dependency Upgrade
 
--- a/junit/kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/crud/KubernetesCrudDispatcherHandler.java
+++ b/junit/kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/crud/KubernetesCrudDispatcherHandler.java
@@ -24,10 +24,13 @@ import io.fabric8.mockwebserver.crud.AttributeSet;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.RecordedRequest;
 
+import java.net.HttpURLConnection;
+
 import static io.fabric8.kubernetes.client.http.StandardHttpHeaders.CONTENT_TYPE;
 import static io.fabric8.kubernetes.client.server.mock.KubernetesAttributesExtractor.NAME;
 import static io.fabric8.kubernetes.client.server.mock.KubernetesAttributesExtractor.NAMESPACE;
 import static io.fabric8.kubernetes.client.server.mock.crud.KubernetesCrudPersistence.METADATA;
+import static io.fabric8.kubernetes.client.server.mock.crud.KubernetesCrudPersistence.RESOURCE_VERSION;
 import static java.net.HttpURLConnection.HTTP_BAD_REQUEST;
 
 @FunctionalInterface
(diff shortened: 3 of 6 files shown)


tombentley:
The final three checkboxes in the checklist don't really apply to this PR:

* Some checks failed, but not due to any changes in the code I touched (AFAICT). e.g. I notice that the e2e tests seem to be failing for master.
* The PR doesn't actually change any code that interacts with Kubernetes or OpenShift

I am therefore marking this as ready for review.

shawkins (approved):
LGTM

shawkins:
#4751 should address the test failure.  @tombentley can you run mvn spotless:apply to clean up the style check?

tombentley:
@shawkins done.

rohanKanojia (approved):

junit/kubernetes-server-mock/src/main/java/io/fabric8/kubernetes/client/server/mock/crud/PatchHandler.java
```diff
@@ -63,6 +63,7 @@ public MockResponse handle(String path, String contentType, String requestBody)
       fullPatch = persistence.merge(currentResource, requestBody);
     }
     validatePath(query, fullPatch);
+    validateResourceVersion(currentResource, fullPatch);
```
manusa, line 66: We're validating the resource version in all patch scenarios which AFAIU is not correct

shawkins: This validation would fail only if the user includes a resourceVersion in the patch.  I believe this is correct.

manusa (approved):

manusa:
I added the missing tests for the PATCH operation. Also ensuring/documenting that the resource version is only checked for JSON Merge patching modes and ignored otherwise. 
Merging as soon as CI is green.

Thx for the contribution :raised_hands:

shawkins:
> Also ensuring/documenting that the resource version is only checked for JSON Merge patching modes and ignored otherwise.

Just to make sure we're on the same page, with a real api server it is checked on the server for all patch types.  However for our object based json patch we're explicitly removing the resourceVersion on the client side.

manusa:
> Just to make sure we're on the same page, with a real api server it is checked on the server for all patch types. However for our object based json patch we're explicitly removing the resourceVersion on the client side.

Unless I made a mistake on my side, `differentResourceVersionOk` was passing on a real cluster. I'm going to retry. Unless I made a mistake JSON patch was allowed even if a different resource version was passed.

manusa:
Checked again, that test passes on OpenShift DevSandbox (4.12 | 1.25.4)

manusa:
> > Unless I made a mistake JSON patch was allowed even if a different resource version was passed.
> 
> If you try to do something like
> 
> kubectl patch deployment patch-demo --type='json' -p='[{"op": "replace", "path": "/metadata/resourceVersion", "value":"1"}]'
> 
> it should fail with a conflict.

It's clear it should fail, but try the test with a real cluster (adding  `client = new KubernetesClientBuilder().build() to the first line of the mentioned test case should do it). At some point we're removing the replace operation from the list of operations we send to the server.

shawkins:
> And that's why it's important for me to test>document these things instead of relying on what seems to be logic (I only trust code) :)

Sorry if I'm not being very clear.  The test that was added reinforces potentially suspect client behavior and does not test the mock server handling.  To test the mock server, which validates all patch types, you can instead use the patch method that takes the patch as a string.

> My suggestion is that we follow up with an issue to see if this behavior needs to be changed or accept our current.

Yes the json patch behavior and the replace / lock operation don't align well with kubectl, it would be good to review that.

manusa:
OK, I think I'm not clear.

I do understand that the scenario is wrong to verify the behavior `server-side`.

However, this is the way it behaves for our client ATM. We don't have more tests around that ensure this.

I will add a clarification comment in the test, and yet another test with the explicit replace operation for the resource version that should fail.

manusa:
Hi @shawkins 
Please check the last commit https://github.com/fabric8io/kubernetes-client/pull/4740/commits/a5a35392f27fa6f381c62c017cc81f8df81d7dc6

We can either move forward with this (dirty TODO message), or further move that test case elsewhere.

shawkins:
> Please check the last commit https://github.com/fabric8io/kubernetes-client/commit/6a486a7395f7e352ce726f6b9ce8197a35c66461

Yes that looks good.

PR #4748, opened 2023-01-09
Tune the client to handle up to 1024 concurrent connections

vietj:


httpclient-vertx/src/main/java/io/fabric8/kubernetes/client/vertx/VertxHttpClientFactory.java
```diff
@@ -76,6 +76,9 @@ public VertxHttpClient build() {
 
       WebClientOptions options = new WebClientOptions();
 
+      options.setMaxPoolSize(1024);
```
shawkins, line 79: Are there any downsides to leaving this unlimited?  We didn't assume any limit for the other implementations.

kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractSimultaneousConnectionsTest.java
```diff
@@ -114,6 +114,10 @@ public void http1Connections() throws Exception {
           .build();
       for (int it = 0; it < MAX_HTTP_1_CONNECTIONS; it++) {
         asyncResponses.add(client.consumeBytes(request, (value, asyncBody) -> asyncBody.consume()));
+        // Throttle connection creation to keep cool with the server otherwise it might reset connections
```
shawkins, line 117: @manusa @vietj there's also #4736 - in that I didn't want to make any assumption about rate limiting the creation, so instead I added logic to keep retrying if there are failures.  

OkHttp must be doing something different internally as it didn't seem to trigger any ill-effects on the server-side - but jdk, jetty, and vertx seem to cause flooding that isn't handled at least for websockets.  The original exceptions for vertx in these tests though seemed to be stackoverflows - is there something more severe going on once vertx starts seeing reset connections?

I'm not opposed to slowing down the creation in these tests - it will help with the jdk async thread pool usage as well - just want to make sure there's not other considerations in play.

httpclient-vertx/src/main/java/io/fabric8/kubernetes/client/vertx/VertxHttpClientFactory.java
vietj: The downside is a potential resource exhaustion on the client, but also on the origin server. It is up to you to decide how you want to do that.

Note that the Jetty server has a max value but does not enforce it by default.

kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractSimultaneousConnectionsTest.java
vietj: That does not slow a lot the tests according to my tests, the tests are still very fast, the HTTP/1 test takes 250ms on my machine to pass.

Jetty might throttle internally the creation of connection.

shawkins: > That does not slow a lot the tests according to my tests, the tests are still very fast, the HTTP/1 test takes 250ms on my machine to pass.

I'm less concerned with the runtime of the test and how any issue with 500 from the server is handled.

Without these changes, here's what's in the log from a run of the tests https://gist.github.com/shawkins/4f174fb0c25d30f5524f42a9ea5d3cba - but I'm not sure if that's directly due to the test run, shutdown, etc.

httpclient-vertx/src/main/java/io/fabric8/kubernetes/client/vertx/VertxHttpClientFactory.java
shawkins: > The downside is a potential resource exhaustion on the client, but also on the origin server. It is up to you to decide how you want to do that.

We'll just assume unlimited, then we can later limits should anyone experience a problem - but we'll probably want to put that logic in the StandardHttpClient such that it's common to all http clients and make it as proactive as possible to inform  the user exactly what is happening.

> Note that the Jetty server has a max value but does not enforce it by default.

These tests are using com.sun.net.httpserver.HttpServer and the okhttp3.mockwebserver.MockWebServer at least up to 1024 connections they don't seem to require any additional configuration either.

vietj: the issue might arise when multiple clients are targeting the same server with the same level of concurrency, but I don't know about your use cases.

kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractSimultaneousConnectionsTest.java
vietj: interesting it looks like a bug in Vertx that happens in that case, thanks I'll report it and we'll get a fix for that

vietj: I think it happens because we have many connection on the same pool that cannot connect subsequently

httpclient-vertx/src/main/java/io/fabric8/kubernetes/client/vertx/VertxHttpClientFactory.java
manusa: I'm not sure is the responsibility of the (Kubernetes) client to protect a Kubernetes server->cluster from being flooded.

I understand this might be concerning for any standard use-case where the target server is a regular HTTP server. However, in this case, the API server should be strong enough to handle or scale when a given number of concurrent connections are performed (Come on, it's Kubernetes after all ;))

The decision to remove the hard limit on the client-side was to achieve a consistent behavior across all `HttpClient` implementation with the minimum effort.

kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractSimultaneousConnectionsTest.java
manusa: Checking #4736

shawkins:
@manusa @vietj to move forward here we'll remove the waits that were added in the tests - this will be handled by #4752 

Then we'll set the limits to max int.

vietj:
shall I close this PR @shawkins then ?

manusa:
I just rebased the branch with the changes from #4752 

I'll create a new PR to the branch with the change to + `setMaxPoolSize` which should replace this one.

PR #4752, opened 2023-01-11
test: refine simultaneous connections test

manusa:
## Description

**Partial** alternative to #4736

- Introduced delayed HttpHandler, delays response delivery until a certain amount of requests/exchanges have been accumulated.
- Introduced artificial back-pressure using a client-server shared CyclicBarrier. New requests are only sent when the server receives and is able to process the previous request.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractSimultaneousConnectionsTest.java
+++ b/kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractSimultaneousConnectionsTest.java
@@ -48,8 +48,9 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.CyclicBarrier;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.TimeoutException;
 import java.util.stream.IntStream;
 
 import javax.net.ServerSocketFactory;
@@ -104,9 +105,9 @@ public abstract class AbstractSimultaneousConnectionsTest {
   @DisplayName("Should be able to make 2048 simultaneous HTTP/1.x connections before processing the response")
   @DisabledOnOs(OS.WINDOWS)
   public void http1Connections() throws Exception {
-    final CyclicBarrier barrier = new CyclicBarrier(2);
-    httpServer.createContext("/http",
-        new DelayedResponseHandler(MAX_HTTP_1_CONNECTIONS, barrier, exchange -> exchange.sendResponseHeaders(204, -1)));
+    final DelayedResponseHandler handler = new DelayedResponseHandler(MAX_HTTP_1_CONNECTIONS,
+        exchange -> exchange.sendResponseHeaders(204, -1));
+    httpServer.createContext("/http", handler);
     try (final HttpClient client = clientBuilder.build()) {
       final Collection<CompletableFuture<HttpResponse<AsyncBody>>> asyncResponses = ConcurrentHashMap.newKeySet();
       final HttpRequest request = client.newHttpRequestBuilder()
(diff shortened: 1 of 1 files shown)


oscerd (approved):

kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractSimultaneousConnectionsTest.java
```diff
@@ -156,41 +151,90 @@ public void http1WebSocketConnections() throws Exception {
...
+          }
+        }
+      });
+    }
+
+    @Override
+    public void handle(HttpExchange exchange) throws IOException {
+      connectionCount.incrementAndGet();
+      try {
+        barrier.await(1, TimeUnit.SECONDS);
```
shawkins, line 226: I'm fine with using a mechanism like this rather than retries - I was not sure how intentional the flooding was, so I preserved that initial behavior just in case that was something you were trying to test.

You can also consider using a Semaphore instead of CyclicBarrier as there's no need to wait here, just increase the permits for the loop.

kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractSimultaneousConnectionsTest.java
```diff
@@ -156,41 +151,90 @@ public void http1WebSocketConnections() throws Exception {
...
+    private final AtomicInteger connectionCount;
+    private final CompletableFuture<Void> done;
+
+    private DelayedResponseHandler(int requestCount, CyclicBarrier barrier, HttpHandler handler) {
+      this.requestCount = requestCount;
+      this.barrier = barrier;
+      exchanges = ConcurrentHashMap.newKeySet();
+      connectionCount = new AtomicInteger(0);
+      done = new CompletableFuture<>();
+      done.thenRunAsync(() -> {
```
shawkins, line 211: In local testing a saw two types of errors on the client side - 500s from before the request is ever processed by the server, and errors after createContext has been called, but before the headers are sent.  That is it appeared that we could flood the output buffering of the HttpServer.  I also didn't see at first glance if this was a non-blocking operation, so for these reasons used a bounded theadpool, rather than relying for this rather than relying on the common pool.

kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractSimultaneousConnectionsTest.java
```diff
@@ -94,29 +101,22 @@ private void withHttp1() throws IOException {
...
       final HttpRequest request = client.newHttpRequestBuilder()
           .uri(String.format("http://localhost:%s/http", httpServer.getAddress().getPort()))
           .build();
       for (int it = 0; it < MAX_HTTP_1_CONNECTIONS; it++) {
         asyncResponses.add(client.consumeBytes(request, (value, asyncBody) -> asyncBody.consume()));
+        barrier.await(1, TimeUnit.SECONDS);
       }
-      assertThat(activeConnections.await(20, TimeUnit.SECONDS)).isTrue();
-      CompletableFuture.allOf(asyncResponses.toArray(new CompletableFuture[0])).get(10, TimeUnit.SECONDS);
+      CompletableFuture.allOf(asyncResponses.toArray(new CompletableFuture[0])).get(20, TimeUnit.SECONDS);
```
shawkins, line 119: The 20 second value seemed problematic in some of my testing with the non-okhttp clients, thus the change to how the other pr was awaiting this and the similar timeouts.  The improved thread management made things better, but you may want to consider increasing this for github runs just to aliviate the possibility of spurious failures.

kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractSimultaneousConnectionsTest.java
manusa: The await is just a safety mechanism to make sure that the tests end as soon as possible in case there are any other non-contemplated errors that may hang the its execution. This way the test will halt from any of the sides in case one of them becomes unresponsive.
Maybe a semaphore is better suited for this, but don't know exactly how to implement the same safety features.

manusa: Switching to semaphore was not creating enough back-pressure for the JDK implementation which failed randomly (about 25% of times in my local system), especially for `http1WebSocketConnectionsBeforeUpgrade`.
I've switched back to Cyclic Barrier. The slight block in the handler seems to do the job (ran same test >500 times with 0 failures locally -even with RejectedExecutions which caused test failures with Semaphore-)

shawkins: So even with a semaphore the re-try logic from the other pr was still necessary.  Long story short - the HttpServer can't handle very much - hopefully this will be reliable enough for now in github runs, but it does make a good case for moving the retry logic into our standardhttpclient to be more agnostic to this situation.

manusa: Yes, perfectly summarized.

Issue #4735, opened 2023-01-05
OpenShift s2i binary builds seem broken in 6.3


manusa:
### Describe the bug

While upgrading Eclipse JKube to use Kubernetes Client 6.3.1 https://github.com/eclipse/jkube/pull/1988, we've detected that binary builds are not completed.

It seems that the binary upload is not working as expected and the HTTP request never completes.

The call:
```java
client.buildConfigs()
  .inNamespace(namespace)
  .withName(buildName)
  .instantiateBinary()
  .fromFile(binaryTarFile);
```

will hang forever.

There's a very simple unit test to verify this logic:
https://github.com/fabric8io/kubernetes-client/blob/1e944fc144d514e5a7ef875e0552be039fed75cf/kubernetes-tests/src/test/java/io/fabric8/openshift/client/server/mock/BuildConfigTest.java#L208

But there's no real e2e test verifying this scenario.

### Fabric8 Kubernetes Client version

6.3.1

### Steps to reproduce

Create an OpenShift binary build:
```java
client.buildConfigs()
  .inNamespace(namespace)
  .withName(buildName)
  .instantiateBinary()
  .fromFile(binaryTarFile);
```

### Expected behavior

The binary build should complete successfully

### Runtime

OpenShift

### Kubernetes API Server version

other (please specify in additional context)

### Environment

Linux

### Additional context

Point of failure:
https://github.com/fabric8io/kubernetes-client/blob/1d3e263cc6216f4a3adad6b677539186179f41c2/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/build/BuildConfigOperationsImpl.java#L267-L271

https://github.com/fabric8io/kubernetes-client/blob/1d3e263cc6216f4a3adad6b677539186179f41c2/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/OperationSupport.java#L515

shawkins:
More than likely the underlying issue is not related to this particular call - it's just a post of string value that is no different than a typical create call.  This in part was to be addressed by #4663 - but it would be great if your scenario is reproducible to get a confirmation of what is occurring.

manusa:
> What client is being used in this scenario?

OkHttp / OpenShiftClient

[...]

PR #4766, opened 2023-01-13
fix: StandardHttpClient sends Expect 100-continue header value

manusa:
## Description

Relates to #4735 (might fix it)

fix: StandardHttpClient sends Expect 100-continue header value

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,8 @@
 #### Bugs
 * Fix #4249 #4726: prevent the over-logging of errors after the websocket has been closed
 * Fix #4650: allowing for comments at the end of certificate files
-
 * Fix #4668: use acme.cert-manager.io ApiGroup for Orders and Challenges
+* Fix #4735: StandardHttpClient sends Expect 100-continue header value
 
 #### Improvements
 * Fix #4637: all pod operations that require a ready / succeeded pod may use withReadyWaitTimeout, which supersedes withLogWaitTimeout.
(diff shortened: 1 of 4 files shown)


manusa:
@shawkins FYI, the JDK client sends the "wrong" header value

oscerd (approved):

shawkins:
Indeed https://github.com/openjdk/jdk/blob/master/src/java.net.http/share/classes/jdk/internal/net/http/Exchange.java#L536

And you confirmed that the api server does require the case sensitive value?  Maybe @maxandersen can push through another jdk fix.

manusa:
> Indeed https://github.com/openjdk/jdk/blob/master/src/java.net.http/share/classes/jdk/internal/net/http/Exchange.java#L536
> 
> And you confirmed that the api server does require the case sensitive value? Maybe @maxandersen can push through another jdk fix.

No, sorry, I haven't confirmed yet that changing the header value fixes it.

manusa:
:+1: Confirmed, JKube's `oc:build` works again once the header is sent with the **correct** value `100-continue`.

manusa:
I'm merging this one, we should probably discuss what to do with the JDK implementation.

shawkins:
@manusa reviewing this more, the burden of the fix should be in the underlying go library: https://github.com/golang/go/issues/57824 

If for whatever reason that is not possible, we can push for a jdk change.

manusa:
This is more aligned with what I remember about the header value. However, I think both sides would benefit from the fix.

PR #4796, opened 2023-01-26
fix #4794: clarifying event processing wrt to a manual informer stop

### LINKED ISSUE #4794: Add a convenient way to make sure a `SharedIndexInformer` is done watching

Right now, the only way to make sure a `SharedIndexInformer` is done watching is by polling:
```java
  boolean isWatching();
```

`SharedIndexInformer` exposes a way to get notified when a `stop` has been requested:
```java
  CompletionStage<Void> stopped();
```
... but, as of version 6.4.0, the watch is likely to be still running _after_ completion:
```java
  public void stop() {
    stopFuture.complete(null);
[...]
    stopWatcher();
  }
```

Either:
1. make above `stop` blocking
2. or call `stopFuture.complete(null)` from within:
```java
  private synchronized void stopWatcher() {
    Optional.ofNullable(watchFuture).ifPresent(theFuture -> {
      watchFuture = null;
      theFuture.cancel(true);
      theFuture.whenComplete((w, t) -> {
        if (w != null) {
          stopWatch(w);
        }
      });
    });
  }
```
3. or expose something like:
```java
  CompletionStage<Void> doneWatching();
```

```java
            while (informer.isWatching()) {
                LOGGER.warn("Informer still watching after shutdown...");
                // pause thread...
            }
```

### Additional context

_No response_

shawkins:
## Description

Fix #4794 

Addresses #4794 ensures event processing will have stopped when the stopped future is completed.  And will also complete the future after isWatching returns false.  However since we have been setting the watching flag proactively to false (and even at the AbstractWatchManager level we call onClose immediately with the Watch.close call), this does not mean that the websocket / http connection is fully gone. 

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/impl/cache/Reflector.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/impl/cache/Reflector.java
@@ -72,13 +72,13 @@ public class Reflector<T extends HasMetadata, L extends KubernetesResourceList<T
   }
 
   public void stop() {
-    stopFuture.complete(null);
     startFuture.completeExceptionally(new KubernetesClientException("informer manually stopped before starting"));
     Future<?> future = reconnectFuture;
     if (future != null) {
       future.cancel(true);
     }
     stopWatcher();
+    stopFuture.complete(null);
   }
 
   private synchronized void stopWatcher() {
@@ -220,6 +220,9 @@ public class Reflector<T extends HasMetadata, L extends KubernetesResourceList<T
 
     @Override
     public void eventReceived(Action action, T resource) {
+      if (stopFuture.isDone()) {
+        return;
+      }
       if (action == null) {
         throw new KubernetesClientException("Unrecognized event for " + Reflector.this);
       }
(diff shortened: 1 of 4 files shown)


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/impl/cache/Reflector.java
```diff
@@ -220,6 +220,9 @@ class ReflectorWatcher implements Watcher<T> {
 
     @Override
     public void eventReceived(Action action, T resource) {
+      if (stopFuture.isDone()) {
+        return;
+      }
```
manusa, line 225: I'm fine with this change, but not sure if it might be breaking/unexpected for some users.

Just worried that some users might report that events are missing.

_I don't think K8s emits resource-related events to confirm/ack the Watch has been closed, so shouldn't be a problem_

shawkins: I don't think that would be an issue.  Those are events that are processed after stop has been called.

rdesgroppes:
Thanks, @shawkins!

PR #4825, opened 2023-02-02
fix #4788 moving retry logic into the standard layer

### LINKED ISSUE #4788: Move exponential backoff handling to standard client layer
Our high level enforcement of exponential back off is causing several code paths to go without this support.  It would be best to move this logic to a single lower level place.

shawkins:
## Description

Fix #4788 

This makes the retry logic common for all requests - `sendAsync`, `consumeBytes`, and websockets.  There wasn't a great way of making the client executor available due to when the httpclient is created vs when the executor is obtained - further refactoring could probably make this work, but for now since the httpclient calls are non-blocking I'm just running the delayed task in the scheduling thread.

@manusa see if you want to refine the max connections test based upon this.

This does blur the line between the request and watch retries.  Low level issues can be handled by the request retry, while higher level ones, such as repeated error status responses would still need to be handled by the watch retry.

## Type of change

## Checklist


oscerd (approved):

manusa:
> @manusa see if you want to refine the max connections test based upon this.

I think the test is fine as it is. In the end we didn't add the retry logic to the test. The test "improves" the embedded HttpServer reliability (also for OkHttp MockServer). All in all, I think that having more solid Mock servers will allow us to identify and isolate implementation problems in the future.

manusa (approved):

Issue #4802, opened 2023-01-27
config.refresh() erases token specified when building initial config


Vlatombe:
### Describe the bug

Relates to https://github.com/fabric8io/kubernetes-client/issues/4519

This is essentially the same problem but using a fixed token instead of a kubeconfig file when building the initial configuration.

### Fabric8 Kubernetes Client version

6.3.1

### Steps to reproduce

Instantiate the client using

```
KubernetesClient client = new KubernetesClientBuilder().withConfig(new ConfigBuilder().withMasterUrl("https://url-to-my-kubernetes-master").withCaCertData("my-ca-cert-data").withOauthToken("my-token").build()).build();
```

and also have a file at `/var/run/secrets/kubernetes.io/serviceaccount/token` with some different token.

After 1 minute, calls start failing with `Unauthorized` errors.

### Expected behavior

Calls don't fail after 1 minute

### Runtime

Kubernetes (vanilla)

### Kubernetes API Server version

1.23

### Environment

Amazon

shawkins:
Yes this is some problematic behavior of the refresh logic that was noticed in #4702 as well.  As you note this will happen only if you specify the token and have another token that can be picked up by the refresh logic - we're not currently tracking if the token and other fields were set from the auto configure or after the fact by the user.

As a workaround I think you can prevent this from happening if you call config.setAutoConfigure(false) on the config you pass to the builder.

This does not affect the openshift logic with #4702 as it will only try to refresh on a authentication failure, however the base token refresh interceptor will attempt to proactively refresh - which is causing this issue.  There is currently no way to adjust that 1 minute interval.

Vlatombe:
I found `new ConfigBuilder(Config.empty());` to be effective to prevent even the initial autoconfiguration.

shawkins:
To recap - we will currently bypass refresh if autoconfigure is false and there is no associated kubeconfig file.  If there isn't a kubeconfig file that is found which provides a competing token, you can either start with an empty / non-auto configured builder, or use withAutoConfigure

new ConfigBuilder().withMasterUrl("https://url-to-my-kubernetes-master").withCaCertData("my-ca-cert-data").withOauthToken("my-token").withAutoConfigure(false).build()

@manusa @rohanKanojia @Vlatombe The broader question is over expected behavior - if you call withOauthToken, or withOauthTokenProvider, should we track that so that the config is not proactively refreshed?  If the token does expire, then should we try the more typical refresh mechanisms?

[...]

PR #4837, opened 2023-02-06
fix #4708 refining what is possible via the requestconfig

### LINKED ISSUE #4708: Refine what is possible via withRequestConfig
The RequestConfig object exposes quite a few timeouts, authentication properties, and other things that we are allowing to be set on a per request basis if one starts with NamespacedKubernetesClient.withRequestConfig.

There are several issues with this:
- in the code it's effectively part of the Config, which is otherwise immutable.  That is confusing to deal with and requires extra logic to pass around to the interceptors.
- passing alternative authentication is confusing to the logic in the interceptors.  Although that may be improved by #4702.

It would be good to re-assess what about this feature is actually needed, and if it should just be part of the regular dsl - that is instead of client.adapt(NamespacedKubernetesClient.class).withRequestConfig(new RequestConfigBuilder().withImpersonateUsername("name").build()).(c -> ...)

should we just offer

client.withImpersonateUsername("name")...

With methods for whatever seems still useful.

shawkins:
## Description

Fixes #4708

There are several goals here.  The first is to ensure that only config values that can be changed without creating a new httpclient can be set on the RequestConfig - that means connectionTimeout cannot be there.  The next is that allowing for authentication related overrides complicates the logic and any documentation we'd want to create around the refresh interceptors, so it's not a good idea to allow that to be set on a per request basis.  It also seems unnecessary to override the oauthtokenprovider as the user already has control over how the token should be provided.

The next goal was to minimize our changes to the Config (the full application of those changes is often unclear and/or may require a new httpclient) - after this the config is nearly immutable and not copied - we're no longer creating copies to override the RequestConfig, nor for the namespace handling.  This means that places in the client which were referencing the config to get RequestConfig values were changed to reference a contextual RequestConfig.  The only logic that is still modifying the config is the token refresh - and this will likely get refined with #4802.  

To clean up how Interceptors reference the config, there is not an explicit parameter that finds tags associated with the request - this is not actually on the HttpRequest as we create those in other places and it's simpler not to have to find the tags that were applicable at that point and only deal with this concept as part of the Interceptors.  Due to this and the change to using the standard logic we can pass the full request, not just the headers into the interceptor methods.  After #4825 we'll be able to also add a generalized after method for regular requests - this should give us the full set of methods needed to introduce generalized logging interception.

This did not attempt to introduce new entry logic for override the requestconfig - it is still through the withRequestConfig method.  If we do provide other DSL entry points for such modifications they should use the same internal method for constructing a new Client / derived HttpClient that will have the appropriate overrides.

This does have conflicts with #4825 - but they will be straight-forward to resolve after one of them is committed.

## Type of change

## Checklist


manusa:
Tried to quickly merge master but it needs more work :sweat:

shawkins:
> Tried to quickly merge master but it needs more work

That's likely - there have been some overlapping changes of late.  After this goes in and #4877, then the extra fields will be removed as well.

manusa:
Shall we merge #4877 first? (seems cleaner at the moment)

shawkins:
> Shall we merge https://github.com/fabric8io/kubernetes-client/pull/4877 first? (seems cleaner at the moment)

That's good with me, I'll clean this one up after that.

shawkins:
@manusa should be good now

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/Interceptor.java
```diff
@@ -16,41 +16,23 @@
...
-   *
-   * @param config
-   * @return
-   */
-  default Interceptor withConfig(Config config) {
-    return this;
-  }
+  public interface RequestTags {
+
+    <T> T getTag(Class<T> type);
```
manusa, line 25: In the future we should document what a Request tag is and what can it be used for

manusa (approved):

PR #4839, opened 2023-02-06
Fix/pod it

shawkins:
## Description
Possibly some additional changes for #4838

We aren't checking the upload return value, which can mask that it has failed.  Also there is a status being constructed that is not used.

The logs for the recent failed runs do contain:

2023-02-06T17:58:45.1252271Z [OkHttp WebSocket https://192.168.49.2:8443/...] WARN io.fabric8.kubernetes.client.dsl.internal.ExecWebSocketListener - Received [java.net.SocketException], with message:[Connection or outbound has been closed] after ExecWebSocketListener is closed, Ignoring.

But those can only occur after onClose or onError has already been called.  It shouldn't be happening after onError as the exitCode future would be completed with an exception.  If it is occurring after onClose, but the exitCode is not completed, that will currently cause the upload method timeout to be exceeded (there's no log the timeout was exceeded), so now it will complete the exitcode with an exception.  It's not clear from the log that this is actually occurring as it didn't look like the timeout was exceeded.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/ExecWebSocketListener.java
@@ -260,24 +260,15 @@ public class ExecWebSocketListener implements ExecWatch, AutoCloseable, WebSocke
 
   @Override
   public void onError(WebSocket webSocket, Throwable t) {
-
-    //If we already called onClosed() or onFailed() before, we need to abort.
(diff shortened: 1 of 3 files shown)


manusa:
I'll cherry pick the production code fix into my branch. I think I did some additional changed after you cherry-picked my original commit.

manusa:
There's still something going on. I'm not able to reproduce locally despite running the test continuously. My gut feeling is that it has to do with an asynchronous processing of the packets, but as far as I can see, the related code looks OK.

PR #4860, opened 2023-02-09
fix #4675: adding an informer timeout handled directly in our client

### LINKED ISSUE #4675: Set a timeout client side in informer watches
> I am wondering why Fabric8 is not specifying default timeout for watch requests?

Its never been requested as a feature.  The understanding is that api server will enforce its default timeout (30 minutes I believe), with some jitter of its own.

_Originally posted by @shawkins in https://github.com/fabric8io/kubernetes-client/issues/4670#issuecomment-1348622780_

shawkins:
## Description

Fix #4675

This seems like a little less work than updating the mock and other tests with fixed expectations.  Here we'll trigger the shutdown of the websocket after 5 - 10 minutes (the same as the go client).

Another thought is to eliminate the random strategy specifically for the mock client, but that also seems hackish.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
@@ -39,6 +39,7 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -48,6 +49,13 @@ import java.util.function.Supplier;
 import static java.net.HttpURLConnection.HTTP_GONE;
 
 public abstract class AbstractWatchManager<T extends HasMetadata> implements Watch {
+
+  public static class WatchRequestState {
+
+    private final AtomicBoolean reconnected = new AtomicBoolean();
+    private final AtomicBoolean closed = new AtomicBoolean();
+
+  }
 
   private static final Logger logger = LoggerFactory.getLogger(AbstractWatchManager.class);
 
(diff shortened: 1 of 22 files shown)


manusa:
In this PR we're killing the Watch client-wise as opposed to sending a parameter with a timeout. This is because our MockServer doesn't support parameters with variable values.

We'll probably want to follow up **in the future** with the improvement of the MockServer to allow for variable path parameters and then sending the timeout parameter in the URL for the Watch request including the jitter.

shawkins:
Addressed the conflict and also added a comment / refinement about the message processing behavior - what matters is how things are handled at the watchmanager.  We do not, and should not, try and guard against further message processing in the reflector.

shawkins:
After reviewing again decided this is best handled by an existing method on the http / websocket watch handlers - closeRequest is already supposed to trigger a graceful restart.  Doing this via the exception mechanism is not as good as it forces a re-list as well.

shawkins:
One, hopefully last, refinement which also is now setting the url timeout as well as a fallback.  This also highlights the scope of the hard-coded expectation issue.  There are close to 40 places where this appears and several of those are shared among multiple tests.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
```diff
@@ -87,9 +97,20 @@
...
+  protected abstract void start(URL url, Map<String, String> headers, WatchRequestState state);
+
+  /**
+   * Attempt to gracefully close the current request.
+   * <p>
+   * If forceClosed has not been set, then it's expected that the watch will
+   * attempt to reconnect
+   */
+  public void closeRequest() {
+    Optional.ofNullable(latestRequestState).ifPresent(s -> s.closed.set(true));
```
sunix, line 109:
```suggestion
    Optional.ofNullable(latestRequestState).ifPresent(state -> state.closed.set(true));
```

sunix (approved):

manusa (approved):

PR #4867, opened 2023-02-10
[java-generator] performance improvementes

### LINKED ISSUE #4865: Improve performance of the java-generator
Suggestion from @mariofusco as he used async profiler against the `java-generator`, thanks for the amazing work 🙏 !

He noticed that one possible optimization would be to reduce the amount of work that JavaParser should do when parsing plain Strings e.g. [here](https://github.com/fabric8io/kubernetes-client/blob/7ff51aa97548ec4f3d1da4803ef7b585b57fe53f/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java#L159) :

```java
clz.addImplementedType("io.fabric8.kubernetes.api.model.KubernetesResource");
```

can be more efficient narrowing down the scope of the JavaParser string parsing:

```java
clz.addImplementedType( toClassOrInterfaceType( "io.fabric8.kubernetes.api.model.KubernetesResource" ) );
```

mariofusco:
Fix #4865

This PR:
- replaces some wasteful hidden invocations to `StaticJavaParser.parseClassOrInterfaceType` with explicit javaparser AST nodes constructors
- splits the code generation and writing phases
- performs the writings in independent async tasks

### DIFF
--- a/java-generator/core/src/main/java/io/fabric8/java/generator/FileJavaGenerator.java
+++ b/java-generator/core/src/main/java/io/fabric8/java/generator/FileJavaGenerator.java
@@ -116,7 +116,7 @@ public class FileJavaGenerator implements JavaGenerator {
               }
               return Stream.empty();
             })
-      .collect(Collectors.toList());
+        .collect(Collectors.toList());
   }
 
   private void writeCRCompilationUnits(File basePath, List<WritableCRCompilationUnit> wCUs) {
--- a/java-generator/core/src/main/java/io/fabric8/java/generator/FileJavaGenerator.java
+++ b/java-generator/core/src/main/java/io/fabric8/java/generator/FileJavaGenerator.java
@@ -32,6 +32,11 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static io.fabric8.java.generator.CRGeneratorRunner.groupToPackage;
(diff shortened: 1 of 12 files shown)


oscerd (approved):

java-generator/core/src/main/java/io/fabric8/java/generator/FileJavaGenerator.java
```diff
@@ -83,32 +85,50 @@ private void runOnSingleSource(File source, File basePath) {
...
+            })
+      .collect(Collectors.toList());
+  }
+
+  private void writeCRCompilationUnits(File basePath, List<WritableCRCompilationUnit> wCUs) {
+    CompletableFuture<Void>[] futures = new CompletableFuture[wCUs.size()];
+
+    int i = 0;
+    for (WritableCRCompilationUnit w : wCUs) {
+      futures[i++] = CompletableFuture.runAsync(() -> w.writeAllJavaClasses(basePath));
```
andreaTP, line 127: Have you attempted to use a different Thread Pool than the `ForkJoinPool.commonPool` since it's used also by the underlying `parallelStreams`?

java-generator/core/src/main/java/io/fabric8/java/generator/FileJavaGenerator.java
```diff
@@ -83,32 +85,50 @@ private void runOnSingleSource(File source, File basePath) {
...
+              }
+              return Stream.empty();
+            })
+      .collect(Collectors.toList());
+  }
+
+  private void writeCRCompilationUnits(File basePath, List<WritableCRCompilationUnit> wCUs) {
+    CompletableFuture<Void>[] futures = new CompletableFuture[wCUs.size()];
+
+    int i = 0;
```
andreaTP, line 125: [nitpick] can you, please, move `wCUs.size()` to a val (e.g. `total`) and reuse it as a boundary in the following `for` loops?

java-generator/core/src/main/java/io/fabric8/java/generator/FileJavaGenerator.java
```diff
@@ -83,32 +85,50 @@ private void runOnSingleSource(File source, File basePath) {
...
+    for (WritableCRCompilationUnit w : wCUs) {
```
andreaTP, line 126:
```suggestion
    for (int i = 0; i < total; i++) {
```

java-generator/core/src/main/java/io/fabric8/java/generator/FileJavaGenerator.java
```diff
@@ -83,32 +85,50 @@ private void runOnSingleSource(File source, File basePath) {
...
+      futures[i++] = CompletableFuture.runAsync(() -> w.writeAllJavaClasses(basePath));
+    }
+
+    for (int j = 0; j < futures.length; j++) {
```
andreaTP, line 130:
```suggestion
    for (int j = 0; j < total; j++) {
```

andreaTP (commented):
Amazing job @mariofusco 🎉 🚀 🚀 🚀 
Thanks a ton!!!

A minor style nitpick and it's good to go.
Can you share the perf difference before/after this change?

mariofusco:
> Can you share the perf difference before/after this change?

In reality I can no longer run my main because I don't have anymore the yaml that I was using to test it. Is it possible to push it on main or somewhere else? That said I'm afraid that the code generation triggered by that file doesn't last long enough to see any reliably measurable difference. Maybe we could find a bigger test case and put it in a JMH benchmark?

java-generator/core/src/main/java/io/fabric8/java/generator/FileJavaGenerator.java
mariofusco: At that point the code generation performed in the parallel stream is finished and that pool should be completely available to write files. Also consider that creating a brand new thread pool could be very likely much more expansive than writing the file, that's why imo it is a better idea to reuse an already existing pool. Anyway if we had the JMH benchmark that I'm proposing in the other comment we could properly measure this.

java-generator/core/src/main/java/io/fabric8/java/generator/FileJavaGenerator.java
mariofusco: Done.

java-generator/core/src/main/java/io/fabric8/java/generator/FileJavaGenerator.java
mariofusco: Done.

java-generator/core/src/main/java/io/fabric8/java/generator/FileJavaGenerator.java
mariofusco: Done.

java-generator/core/src/main/java/io/fabric8/java/generator/FileJavaGenerator.java
andreaTP: You can find the yaml I was using here: https://github.com/kserve/kserve/releases/download/v0.10.0/kserve.yaml
feel free to add it to the `CompilationTest` along with this PR.

Feel free to add a JMH benchmark if you think that it's useful, but I'm convinced that this PR is perfectly enough for a first iteration.
In case performance issues becomes a problem we can revisit the implementation later on :+1:

Thanks again!

mariofusco:
I added a JMH benchmark to demonstrate the performance improvement of this PR and got the following results.

Before this PR:
```
Benchmark                                 Mode  Cnt  Score   Error  Units
SourceCodeGenerationBenchmark.benchmark  thrpt   10  1.451 ± 0.036  ops/s
```

After
```
Benchmark                                 Mode  Cnt  Score   Error  Units
SourceCodeGenerationBenchmark.benchmark  thrpt   10  2.845 ± 0.201  ops/s
```

andreaTP (approved):
Amazing job @mariofusco , good to go! 🎉

manusa (approved):
LGTM, thx!

Awesome job :raised_hands:

PR #4887, opened 2023-02-15
fix #4885: addressing a potential hang with jdk streams

### LINKED ISSUE #4885: JDK client hanging on PodIT
Some of the e2e tests with the jdk client are hanging during the run of PodIT - it's not clear which test is the issue.  It should be one of the ones that deals with a stream read or other action that has an indefinite timeout.

shawkins:
## Description

Fix #4885

Addresses PodIT hanging failure with jdk by checking for complete prior to waiting - the issue is that the request for more with jdk will be executed in the calling thread which already holds the lock and can modify the complete flag prior to the wait.

There are also some failures on the JobIT log - but I have not been able to reproduce that.  There are a couple of changes here to help with that logic in general - one is better localize when more is requested.  That should be done inline with the processing call, rather than in a completion handler because it looks like they can be executed in parallel.  The other is that I don't see now anywhere in the jdk client that guarantees the supplied buffers won't change after be passed to consume.  Since we already handled that case with jetty we should do the same here.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Fix #4818: [java-generator] Escape `*/` in generated JavaDocs
 * Fix #4823: (java-generator) handle special characters in field names
 * Fix #4723: [java-generator] Fix a race in the use of JavaParser hitting large CRDs
+* Fix #4885: addresses a potential hang in the jdk client with exec stream reading
 
 #### Improvements
 * Fix #3805: DeletionTimestamp and Finalizer support in Mock server.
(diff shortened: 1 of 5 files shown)


oscerd (approved):

shawkins:
> There are also some failures on the JobIT log - but I have not been able to reproduce that.

I think I see how that is possible now - the initial speculative change was wrong.  We need to handle the done in the serialexecutor as well to ensure that the buffers that have already been sent are processed first - otherwise we're risking dropping some of that.  This seems to be more prevalent with jdk / jetty than okhttp likely due to the threading model of how consume is handled - it's an async task for okhttp.

manusa (approved):
LGTM, thx!

shawkins:
@manusa things look better with the latest e2e run.  There is a familiar looking podit failure with vertx, but on 1.25 - so that shouldn't be using containerd correct? https://github.com/fabric8io/kubernetes-client/actions/runs/4200862573/jobs/7287316314#step:5:1022

I'll see if that can be reproduced.

manusa:
> @manusa things look better with the latest e2e run. There is a familiar looking podit failure with vertx, but on 1.25 - so that shouldn't be using containerd correct? https://github.com/fabric8io/kubernetes-client/actions/runs/4200862573/jobs/7287316314#step:5:1022
> 
> I'll see if that can be reproduced.

Nope, that one is using Minikube on baremetal with Docker.

shawkins:
> Nope, that one is using Minikube on baremetal with Docker.

It did spawn #4891

PR #4896, opened 2023-02-20
fix #4861: various changes to refine resourceVersion and replace

### LINKED ISSUE #4861: Revisit resourceVersion and locking behavior

There are several paradigms that the fabric8 client uses wrt to resourceVersions - which are not all entirely consistent.

- kubectl does not have the same concept of replacement - only if the user leaves the item resourceVersion unset does it make a single attempt to get the freshest resourceVersion and make the put call.
- kubectl does not have the same concept of locking, which is basically only necessary for the fabric8 client due to offering a replace operation that by default will try to use a fresher resourceVersion.  Our notion of locking doesn't extend to other operations such as patching.
- json patch (edit method) always removes the resourceVersion - as mentioned above given the limited scope of locking, there's no way to then lock an edit.
- the other patches are handled in the same way as kubectl - that is if the resourceVersion is included it will be considered for optimistic locking.

We should try to make things as consistent as possible, and/or follow the behavior of kubectl.  That could mean either expanding / formalizing the notion of locking, or removing it - with documentation on what leaving the resourceVersion populated on an item does.

_No response_

### Additional context

_No response_

shawkins:
## Description

Closes #4861 

Introduces several breaking changes to implement what's described in #4861.

Potential concerns:
- there's no direct replacement for createOrReplace. Users are directed instead to use one of several alternatives.
- there's likely a behavioral change for people using patch methods that default to or specify the json patch type, which may require them to manually set their resourceVersions to null.  We could consider instead introducing an "unlock" method in the dsl that does this for the user - resource.unlock().{patch, update, serverSideApply} - this is generally preferable to lock as you are opting into overwriting concurrent changes.
- it will be a little confusing coming from kubectl - our update will be the same as their replace and our replace will be deprecated.

cc @manusa  @csviri

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,11 @@
 * Fix #4659: The SupportTestingClient interface has been deprecated.  Please use one of the supports methods or getApiGroup to determine what is available on the api server.
 * Fix #4825: removed or deprecated/moved methods that are unrelated to the rolling timeout from ImageEditReplacePatchable.  Deprecated rollout methods for timeout and edit - future versions will not support
 * Fix #4826: removed RequestConfig upload connection and rolling timeouts.  Both were no longer used with no plans to re-introduce their usage.
+* Fix #4861: several breaking changes related to resourceVersion handling and the replace operation:
+  - replace is deprecated, you should use update instead.  If you set the resourceVersion to null it will not be optimistically locked
+  - json patch methods using an item for the diff generation such as edit or patch will no longer omit the resourceVersion in the patch.  If you want the patch to be unlocked, then set the resourceVersion to null on the item to be patched.
+  - createOrReplace is deprecated, you should use server side apply instead.
+  - internal logic to mimic an apply that modify an item prior to a json patch is deprecated - you should instead build the item to be patched off of base version, such as with the edit method.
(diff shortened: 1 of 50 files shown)


csviri (approved):
LGTM

shawkins:
@manusa it definitely looks like the github minikube execution makes conflicts much more prevalent.  

The long term goal should be that edit is locking, this matches the kubectl behavior: https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/edit/edit.go#L49

I'm open to:
- making edit specifically unlocked, but leaving patch locked.  And document that in later releases this behavior will change
- updating test case either to remove the resourceVersion as part of the edit, or adding an unlock dsl method as mentioned above.

The best I can can up with for an in-between behavior would be generation aware - that is upon a conflict check to see if the generation has changed.  If it has not, then update the resourceVersion and try again.  The kubectl behavior on conflict is to suggest to the user they can force a put.

manusa:
I'm fine with keeping the locking behavior.

Tomorrow I'll update the PR fixing the tests and adding a more explicit breaking change note.

manusa (approved):

PR #4936, opened 2023-03-01
fix: list found httpclient implementations and output selected one

### LINKED ISSUE #4935: Improve logging when multiple HTTP client factories are found

A warning is currently issued if multiple HTTP client factories are found on the classpath. However, the message is not as helpful as it could be, in particular when it comes to the fact that an implementation is automatically selected.

The message should list the found factories so that it's easier for users to remove extra dependencies since it might not be obvious, especially when the client is consumed transitively. Moreover, the message should specify which implementation was selected. 

_No response_

### Additional context

_No response_

metacosm:
## Description

Fixes #4935

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/HttpClientUtils.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/HttpClientUtils.java
@@ -31,13 +31,14 @@ import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.util.Base64;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.ServiceLoader;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -73,7 +74,6 @@ public class HttpClientUtils {
   private static final Pattern IPV4_PATTERN = Pattern.compile(
       "(http://|https://)?(?<ipAddressOrSubnet>(([01]?\\d\\d?|2[0-4]\\d|25[0-5])\\.){3}([01]?\\d\\d?|2[0-4]\\d|25[0-5])(\\/([1-2]\\d|3[0-2]|\\d))?)(\\D+|$)");
   private static final Pattern INVALID_HOST_PATTERN = Pattern.compile("[^\\da-zA-Z.\\-/:]+");
-  private static final AtomicBoolean MULTIPLE_HTTP_CLIENT_WARNING_LOGGED = new AtomicBoolean();
 
   private HttpClientUtils() {
   }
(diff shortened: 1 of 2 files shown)


oscerd (approved):

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/HttpClientUtils.java
```diff
@@ -145,22 +145,27 @@ public static HttpClient.Factory getHttpClientFactory() {
             "No httpclient implementations found on the context classloader, please ensure your classpath includes an implementation jar");
       }
     }
+    LOGGER.info("Using httpclient {} factory", factory.getClass().getName());
     return factory;
   }
 
   private static HttpClient.Factory getFactory(ServiceLoader<HttpClient.Factory> loader) {
     HttpClient.Factory factory = null;
+    List<String> detectedFactories = new ArrayList<>(7);
```
shawkins, line 154: 7?  That's aspirational.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/HttpClientUtils.java
```diff
@@ -73,7 +74,6 @@ public void before(BasicBuilder builder, HttpRequest request, RequestTags tags)
   private static final Pattern IPV4_PATTERN = Pattern.compile(
       "(http://|https://)?(?<ipAddressOrSubnet>(([01]?\\d\\d?|2[0-4]\\d|25[0-5])\\.){3}([01]?\\d\\d?|2[0-4]\\d|25[0-5])(\\/([1-2]\\d|3[0-2]|\\d))?)(\\D+|$)");
   private static final Pattern INVALID_HOST_PATTERN = Pattern.compile("[^\\da-zA-Z.\\-/:]+");
-  private static final AtomicBoolean MULTIPLE_HTTP_CLIENT_WARNING_LOGGED = new AtomicBoolean();
```
shawkins, line 76: This could get a little chatty in some extreme case - the user has multiple http clients in the classpath, relies upon the default logic to create one of the client types, but specifies the other one programmatically.

metacosm: Shouldn't they all get picked up by the ServiceLoader at once, though?

metacosm: Or is it a matter that the method might be called several times, thus leading to the warning being output several times? If that's the case, this actually raises the question of whether or not we should record the chosen factory "once and for all" to avoid going through the resolution process every time a new client is created? Another potential issue is whether or not we want to support creating clients with implicitly different http clients?

shawkins: > Or is it a matter that the method might be called several times, thus leading to the warning being output several times?

Yes, every client KubernetesClient creation that does not specify a HttpClient.Factory makes this call.

> If that's the case, this actually raises the question of whether or not we should record the chosen factory "once and for all" to avoid going through the resolution process every time a new client is created? 

Possibly.  Some people would cache the ServiceLoader, it's iterator will provide new instances should they be found later.  However I would not want this to be static as the current HttpClientUtils is.  There are several other usages of ServiceLoader in fabric8 that are created for each use.  The only one that is not I believe is the KubernetesDeserializer because the resulting mapping is held statically.

> Another potential issue is whether or not we want to support creating clients with implicitly different http clients?

That seems fine if your circumstances allow for that - you'd have to be doing something odd like dynamically installing new services or changing the priority of an existing one for the selection to change.

metacosm: I have to admit that's a little concerning because I was under the (maybe misguided) impression that creating a client was a rather fast operation. If we do http client resolution each time now, i.e. class path scanning, it might not be as much. If installing new services dynamically is not considered a core use case, it should be fine to always use the same resolved instance all the time (after the initial lookup).

shawkins: > (maybe misguided) impression that creating a client was a rather fast operation

It should be especially in comparison to how long it should be used - any remote round trip will end up being orders of magnitude greater than the time to create the client.

>  If we do http client resolution each time now, i.e. class path scanning, it might not be as much

The ServiceToURLProvider has the same approach and you could make the same case about places that make repeated calls to loadClass.

> If installing new services dynamically is not considered a core use case, it should be fine to always use the same resolved instance all the time (after the initial lookup)

It's mostly a question of where to put it - if you want to refactor so that it's held on the KuberenetesClientBuilder in some fashion that would be fine, but I would avoid doing anything statically.

shawkins (approved):
Taking a step back, this should be fine for almost all usage.  There will not be repeated info / warn logs for our internal creation of httpclients (TokenRefreshInterceptor) because we pass the factory to it.  Also a user wouldn't see additional logs if they keep re-using the same KuberenetesClientBuilder to build clients - as the factory is held on the builder.

rohanKanojia (approved):

manusa (approved):

Issue #4911, opened 2023-02-23
Timeout consolidation


shawkins:
There's a recognition on several issues that there are too many timeouts that are not all well defined.

On https://github.com/eclipse/jkube/issues/2024 it was agreed that writeTimeout should be removed as primary concern.

The other timeouts:
1. Config/RequestConfig requestTimeout - applied as the HttpClient readTimeout and enforced at OperationSupport.waitForResult
1. Config/RequestConfig scaleTimeout - used for scale operations when the wait boolean is specified.  Wth the rolling logic deprecated #4826 the internal call to scaling will be removed, so the only place will be from the user scale call.
1. Config/RequestConfig websocketTimeout - the readTimeout specifically for websocket requests - used two different ways.  Watches set it as the readTimeout on the client, while exec waits on the initial future for that amount of time.
1. readyWaitTimeout / logWaitTimeout - many pod operations check if the pod is in a good state prior to performing the operation
1. delete timeout - uses a common context timeout value, turns the delete into a blocking operation - applied for each delete when operating over a collection
1. openshift buildconfig timeout - also uses the common context timeout value - applies to how long it takes to upload the build and get the response back 
1. Config/RequestConfig connectionTimeout - applied as the connection timeout

Let's discuss how to collapse / combine these.

PR #4949, opened 2023-03-03
Timeout consolidation

shawkins:
## Description
Addresses #4911 

- removes writeTimeout
- connectionTimeout remains on the HttpFactory, but the readTimeout is moved to the HttpRequest
- implements regular request timeout for vertx
- Adds a connectionTimeout to WebSockets to clarify how it is applied.
- removes the usage of RequestConfig.websocketTimeout, instead we'll use requestTimeout

Api Changes:
- RequestConfig.scaleTimeout will be deprecated
- the delete and scale timeout handling are effectively the same, so that logic was consolidated - withTimeout(...).delete(), or withTimeout(...).scale(count) - mean wait up to that much for the operation to complete.  If not specified, the operation returns "immediately" when the initial response comes back from the server.  Alternatives to this would include specifying the wait directly in the call - scale(count, time, timeunit)

Remaining timeouts:
1. Config/RequestConfig connectionTimeout - applied as the connection timeout
1. Config/RequestConfig requestTimeout - for blocking operations applied as the request readTimeout and enforced at OperationSupport.waitForResult.  
1. readyWaitTimeout / logWaitTimeout - many pod operations check if the pod is in a good state prior to performing the operation
1. delete / scale timeout - uses a common context timeout value, turns the delete / scale into a blocking operation - applied for each delete when operating over a collection
1. openshift buildconfig timeout - also uses the common context timeout value - applies to how long it takes to upload the build and get the response back
1. waitUntilReady / waitUntilCondition - a timeout is supplied as part of the call.

Issues:
- The meaning / usage of a 0 timeout is not consistent.  In most places in the api it's used in a "java" way meaning don't wait.  However the httpclients will generally interpret that as wait indefinitely as will the buildconfig logic.  So withTimeout(0).delete and buildConfig.withTimeout(0)...fromFile mean different things.
- using a common withTimeout method will quickly become problematic with what was done here to move it close to the delete and scale methods.  Instead it would have to move much higher, probably on Resource - resource.withTimeout(...).create().  At that point it will greatly overlap with the requestTimeout.  You could even try to take it as far as resource.withTimeout(...).waitUntilReady() 
- there's no RequestConfig entry for the log/pod ready timeout, it's just hard coded to be 5 seconds.  This is behaviorally different than the other timeouts because it has soft enforcement - wait at most this long, but try the operation anyway.

cc @manusa @rohanKanojia let me know your thoughts so I can take this further

## Type of change

## Checklist


shawkins:
Cleared up the conflicts.  @manusa @rohanKanojia do you have any feedback on the issues listed?  In particular the expended usage of a common timeout method and its location in the dsl.

shawkins:
@manusa a couple of small refinements here wrt the fix that was just checked in, you only need to call tag when the RequestConfig changes.  Otherwise these changes clarify the requestTimeout as a request level property.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/LogWatchCallback.java
```diff
@@ -74,14 +69,10 @@ private void cleanUp() {
 
   public LogWatchCallback callAndWait(HttpClient client, URL url) {
     HttpRequest request = client.newHttpRequestBuilder().url(url).build();
-    HttpClient clone = client.newBuilder()
-        .tag(Optional.ofNullable(context.getRequestConfig()).map(RequestConfigBuilder::new).orElse(new RequestConfigBuilder())
-            .withRequestTimeout(0).build())
-        .readTimeout(0, TimeUnit.MILLISECONDS).build();
```
manusa, line 80: shouldn't we ensure that the context timeout is `0`?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatchHTTPManager.java
```diff
@@ -46,18 +44,12 @@ public WatchHTTPManager(final HttpClient client,
       final ListOptions listOptions, final Watcher<T> watcher, final int reconnectInterval,
       final int reconnectLimit)
       throws MalformedURLException {
-    super(
-        watcher, baseOperation, listOptions, reconnectLimit, reconnectInterval,
-        () -> client.newBuilder()
-            .tag(new RequestConfigBuilder(baseOperation.getRequestConfig()).withRequestTimeout(0).build())
-            .readTimeout(0, TimeUnit.MILLISECONDS)
-            .forStreaming()
-            .build());
```
manusa, line 55: Same as before, shouldn't we ensure a request config read timeout of `0`?

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/core/v1/PodOperationsImpl.java
```diff
@@ -202,35 +200,28 @@ public Loggable withLogWaitTimeout(Integer logWaitTimeout) {
   @Override
   public PortForward portForward(int port, ReadableByteChannel in, WritableByteChannel out) {
     try {
-      return new PortForwarderWebsocket(httpClient, this.context.getExecutor()).forward(getResourceUrl(), port, in, out);
+      return new PortForwarderWebsocket(httpClient, this.context.getExecutor(), getRequestConfig().getRequestTimeout())
```
manusa, line 203: Same as others, shouldn't we enforce a `0` timeout value?

shawkins: No, the application of the timeout is on a per request basis.  The "default" is 0.

shawkins: Let's take a step back.  The previous enforcement of the request timeout was something that was applied to every request - however it wasn't really applicable to every request, and thus in places we overrode to 0.

What I'm proposing here is that the "default" request timeout from an api perspective is 0 - that's how jetty, jdk, and vertx already operate.  Then only for the operations where the config / context request timeout is applicable, we'll apply it or the related websocket connect timeout.  The reqeustConfig does not need to be further consulted - it was already used to set appropriate timeout.

manusa: OK, yes, this aligns with what I inferred from the changes. However, aren't the users still capable of overriding the request config for a specific context?

```
client.adapt(NamespacedKubernetesClient.class).withRequestConfig(new RequestConfigBuilder//...
new KubernetesClientBuilder().withConfig(new ConfigBuilder().withRequestTimeout(//...
```

Considering this (maybe I'm still wrong), a user might want to configure some generic timeouts for their HTTP requests, without knowing that they might be actually breaking something else.

shawkins: > However, aren't the users still capable of overriding the request config for a specific context?

Yes, and that requestTimeout would then be applied to applicable requests.

> Considering this (maybe I'm still wrong), a user might want to configure some generic timeouts for their HTTP requests, without knowing that they might be actually breaking something else.

It might be good to clarify that part of the issue is that we have a semantic problem.  The user property is called requestTimeout, while the api level one is a "readTimeout" - for okhttp this means if you don't get anything back from a response stream in that time, then you timeout.  For the other implementations it should be applied for the entire request / response cycle.  

However neither of these is greatly applicable to things like log streaming, which we were overriding at the api level to 0.  That api level setting to 0 would trump anything that user did via the RequestConfig - how this always worked before was that the overriden RequestConfig is logically applied first to create the derived client, then we manipulate that new client instance.  

Am I understanding you correctly that you want new timeout functionality where by a user could use the RequestConfig to override the timeout behavior of following logs (or other operations where we had been explicitly setting it to 0)?

manusa: > Am I understanding you correctly that you want new timeout functionality where by a user could use the RequestConfig to override the timeout behavior of following logs (or other operations where we had been explicitly setting it to 0)?

Nope, just the opposite, I want to make sure that the user can't override _any_ timeout config in any way that would allow to trip the OpenShift Builds, log requests, watchers, and so on.

I'll probably add some behavioral tests tomorrow on top of this PR. This way I'll probably get a better overview too. There are still things that I can't figure out until I try them out properly. And there also seem to be some remains -or not- (for example the BuildConfigOperationsImpl constructor still has the workaround I added to overcome one of the initial timeout bugs)

shawkins: > for example the BuildConfigOperationsImpl constructor still has the workaround I added to overcome one of the initial timeout bugs

Yes, that should be removed.

shawkins: > Nope, just the opposite, I want to make sure that the user can't override any timeout config in any way that would allow to trip the OpenShift Builds, log requests, watchers, and so on.

Added some changes that clarify the nature of readTimeout as the full request timeout.

The expected behavior:
* the requestTimeout is applied to the HttpRequest via our calling logic, and is not considered separately by the StandardHttpClient.  Whatever is on the HttpRequest is what will be used.
* The Config/RequestConfig.requestTimeout is used for most simple requests via OperationSupport.withRequestTimeout.  That can be overriden pragmatically by NamespacedKubernetesClient.withRequestConfig - which is the only place that OperationContext.withRequestConfig is called from. It is also the only path that calls BaseClient.newClient that uses tag to associate a new RequestConfig with the derived HttpClient.  
* BuildConfigOperationsImpl will use the timeout associated with the context (defaults to 0), the requestConfig does not come into play.
* All of the places that were explicitly using a 0 readTimeout no longer need to do that - that is now the default.

manusa:
Hi @shawkins 

I think the PR is good to go.
I added an extra commit with some tests. Please check RequestTimeoutTest and BuildConfigOperationsTimeoutTest. They should ensure that the new timeouts behave either as per previous behavior or as per new convention/specification.
Maybe I missed some test, if so, please indicate which other parts should be tested and we can add them in a subsequent PR.

kubernetes-client/src/test/java/io/fabric8/kubernetes/client/behavior/RequestTimeoutTest.java
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  @DisplayName("Standard Kubernetes Object operations, with overridden config, have overridden timeout")
+  @Test
+  void standardOperationsHaveOverriddenTimeoutInConfig() {
+    // Given
+    factory.getInstances().iterator().next()
+        .expect("/api/v1/namespaces/.+/pods.+", new TestHttpResponse<AsyncBody>().withCode(200).withBody(new TestAsyncBody()));
+    // When
+    client.getConfiguration().getRequestConfig().setRequestTimeout(1337000);
```
shawkins, line 91: Directly modifying the RequestConfig held by the Config may not work in perpetuity - ideally we're moving toward having the Config and stuff under it as immutable.  At the very least this shouldn't be considered a supported path.

manusa: Shouldn't be supported but it's possible now.
I'll add a note to clarify that this is not something intended to be used by users.

shawkins:
> Please check RequestTimeoutTest and BuildConfigOperationsTimeoutTest. 

Looks good.  Thanks @manusa

Issue #4704, opened 2022-12-21
Coordinate token refresh


shawkins:
There is no synchronization between concurrent requests - each will attempt to authorize.  The regular TokenRefreshInterceptor has the same behavior.  This will be more nuanced to address as ideally we don't want to block the other threads.

_Originally posted by @shawkins in https://github.com/fabric8io/kubernetes-client/issues/4673#issuecomment-1351503927_

Issue #2112, opened 2020-04-05
Support for the expiration of the token


egor-ryashin:
Right now the client begins to fail once the expiration time comes. For example if the client is created in the environment with the following `.kube/config`:
```
user:
  auth-provider:
    config:
      access-token: ...
      cmd-args: config config-helper --format=json
      cmd-path: /.../google-cloud-sdk/bin/gcloud
      expiry: "2020-04-04T21:52:48Z"
      expiry-key: '{.credential.token_expiry}'
      token-key: '{.credential.access_token}'
    name: gcp
```
after the time 2020-04-04T21:52:48Z comes the client throws exception:
```
Caused by: io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: ... Message: Unauthorized! Token may have expired! Please log-in again. Unauthorized.
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.requestFailure(OperationSupport.java:510) ~[kubernetes-client-4.6.1.jar:na]
```
And the application requires the token refresh (by running `kubectl` command) and creation of a new a client instance (which reads the new token).

rohanKanojia:
This seems somehow related to #2111

fabriziofortino:
This is quite an annoying issue. To work this around, I have created a utility function that makes a `kubectl version` call on each context I need to interact with:

```kotlin
private fun tryK8sTokenRefresh(contexts: List<String?>) = runBlocking(Dispatchers.IO) {
        val kubeConfig = KubeConfigUtils.parseConfig(File(Config.getKubeconfigFilename()))
        kubeConfig
            .contexts
            .filter { nc -> contexts.isEmpty() || contexts.contains(nc.name) }
            .toList().pmap {
                // Make a request on each cluster to refresh the access tokens
                val versionCmd = "kubectl version --context=${it.name}".runCommand()
                if (versionCmd.first != 0) {
                    verbose(
                        "An error occurred connecting to $it. Check your k8s configuration.\n${versionCmd.second}",
                        err = true
                    )
                    true
                } else {
                    verbose("Refreshed token for ${it.name} -> $versionCmd")
                    false
                }
            }
    }
```

subnetmarco:
Any update on this issue?

manusa:
Is this still happening? I recall we added automatic token refresh mechanism for a few scenarios.

If it is, is anyone with a GCP cluster willing to contribute a fix?

hossman:
[...]

PR #4951, opened 2023-03-06
fix (kubernetes-client-api) :Add new autoOAuthToken field in Config to differentiate between user and autoconfigured tokens (#4802)

### LINKED ISSUE #4802: config.refresh() erases token specified when building initial config

rohanKanojia:
## Description
Fix #4802 

+ Add a new string field autoOAuthToken in Config, this field would be
      used by Config internally to store token obtained during
      autoconfiguration process.
+ Remove setOAuthToken references from token interceptors and Config, it
      would only be called by user (during `new ConfigBuilder().withOAuthToken("...")` call)
      whenever custom token is required.
+ getOAuthToken would give most priority to token from provider, then
      actual value of `oauthToken` and finally `autoOAuthToken` value. All
      interceptors would keep using getOAuthToken to get resolved OAuth
      token value

Updated TokenRefreshInterceptor / OpenShiftOAuthInterceptor to match these expectations:

_Orignally posted by @shawkins_:

- [X] if it's a user token(`oauthToken` is set), don't do any refresh logic at all.  
- [X] if we don't have an oauth token provider, then every minute we could do a proactive "soft" refresh the Config (like the current refresh method) - that may do some file reads or even some exec/system calls, so it's best if it's not happening inline with the http requests and if applicable and the token is nearing expiration do the oidc call.  
- [X] If there is an auth failure and there's no user token or token provider we should start with the soft refresh
- [X] if a token is now there and came from the execconfig or something other than the kubeconfig (service account or env), then use it.
- [X] If the token still isn't there or came from the kubeconfig and the kubeconfig provides the appropriate AuthProvider, then we should do the oidc refresh

## Type of change

## Checklist


oscerd (approved):

openshift-client/src/main/java/io/fabric8/openshift/client/internal/OpenShiftOAuthInterceptor.java
```diff
@@ -94,6 +94,9 @@ public void before(BasicBuilder builder, HttpRequest httpRequest, RequestTags ta
 
   @Override
   public CompletableFuture<Boolean> afterFailure(Builder builder, HttpResponse<?> response, RequestTags tags) {
+    if (config.isTokenNonRefreshable()) {
+      return CompletableFuture.completedFuture(true);
```
shawkins, line 98: Returning true here and in the other interceptor means to try the request again, but since no modifications have been made on the builder, it would just re-issue the existing request.

Including the token provider in isTokenNonRefreshable could be confusing as it may refresh.  That is:

if token source == user, then do nothing
if token source == provider, then attempt to get the token again and try once more.

shawkins (changes requested):
This looks good.  It's the easiest way to do this with the existing code structure.

- the user OauthTokenProvider will take precedence.
- the user oauthToken will be used next  
- if neither of those were used, then every minute there will be a soft refresh or an open id refresh is the token is near expiration.  The code was already doing this, but didn't check for expiration, so we probably can resolve the main thrust of #2112.  The other issues mentioned on there likely remain.
- on unauthorized failure there will be a last chance refresh

Improvements that can be done later:
- Don't update the Config - move the autoOathToken and related refresh logic off of the Config and onto to the refresh interceptors.  But this requires a lot of changing of how we "refresh" the Config.
- don't re-attempt the request if the token hasn't actually changed from the initial request
- coordinate the refresh #4704 - and perform the time based soft refresh in a non IO thread.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/Config.java
```diff
@@ -971,12 +979,18 @@ public static String getKeyAlgorithm(String clientKeyFile, String clientKeyData)
     return null;
   }
 
+  public String getUserConfiguredOauthToken() {
```
shawkins, line 982: This and the getAutoOAuthToken may need jsonignores

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/TokenRefreshInterceptor.java
```diff
@@ -84,14 +84,22 @@ public CompletableFuture<Boolean> afterFailure(BasicBuilder headerBuilder, HttpR
   }
 
   private CompletableFuture<Boolean> refreshToken(BasicBuilder headerBuilder) {
+    if (config.getOauthTokenProvider() != null) {
+      String tokenFromProvider = config.getOauthTokenProvider().getToken();
+      if (tokenFromProvider != null && !tokenFromProvider.isEmpty()) {
+        return CompletableFuture.completedFuture(overrideNewAccessTokenToConfig(tokenFromProvider, headerBuilder, config));
+      } else if (config.getUserConfiguredOauthToken() != null && !config.getUserConfiguredOauthToken().isEmpty()) {
```
shawkins, line 91: this needs to be in a separate block after the oathtokenprovider check

shawkins:
@rohanKanojia @manusa opened a pr against this pr that updates to latest (as of yesterday), and takes the changes a little further https://github.com/rohanKanojia/kubernetes-client/pull/176/commits/d4fba9d42e2956d00e9b94e8cf2866c269e54fc6 - which further clarifies the behavior and tries to localize it better into the TokenRefreshInterceptor.  The only thing that may be odd is that I removed some checks for empty string because they were not being applied consistently - they can of course be added back.

shawkins:
LGTM, should be ready after resolving conflicts.

shawkins (approved):

manusa (approved):

Issue #4923, opened 2023-02-27
Failures with PodIT.copyFile


shawkins:
There are several failures with copyFile in https://github.com/fabric8io/kubernetes-client/actions/runs/4268450566/jobs/7430955144 observed against vertx and jdk.

This was after the change that removed the blocking requirement for onClose, but that doesn't appear to be the issue as there's a failure with jdk - it still requires a request prior to handling onClose.  The likely explanation is that there is an extra request call being issued such that there is demand to not only process the out data message, but the following exit code message.  Since the data write is async and the exit code message handling is not, this would lead to a race condition.  While we can handle the exit code message via the SerialExecutor, it seems worthwhile to track down where the requests may be getting off.

shawkins:
I am unable to reproduce this.  Adding artificial waits before and during the async write and trying all of the clients does not trigger this - only manually adding an errant request call seems to create the problem.

In running the readDirMissingException against vertx the I did see some reproducing behavior - but it's quite odd.  Several times with the vertx client I've seen onClose __before__ data messages.  That would cause both the readDirMissingException and the copyFile errors as we'll no more messages are expected and close / cleanup any output.  If the message ordering issue is kube problem, then this would explain the jdk failure as well.

cc @vietj @manusa is it possible that we're making an invalid assumption wrt when it's possible to see a websocket close?

shawkins:
@vietj @manusa spent a little more time with and the upload issue.  There are broadly 3 problems:

kubernetes seems to drop stdIn https://github.com/kubernetes/kubernetes/issues/112834
kuberentes seems to drop stdOut https://github.com/kubernetes/kubernetes/issues/60140

The vert.x issues with download / stdOut are definitely a problem with our VertxWebSocket handling - the pausing only applies to data frames, not close.  So there is a race condition that allows close to be emitted before the data frames if our reading of the stdOut lags. We have a couple of options to address that.  One is to remove WebSocket back pressure from vertx - the streams will naturally buffer the results - but if left unread could be a memory issue.  Another would be an upstream vertx change to add close to the pending rather than handling it immediately.

However the JDK client retains a strict ordering - but we still have seen failures https://github.com/fabric8io/kubernetes-client/actions/runs/4268450566/jobs/7430955144#step:5:448

[...]

PR #4959, opened 2023-03-09
fix #4910 / #4923 addressing inconsistent behavior with exec streams

### LINKED ISSUE #4910: Issues with upload / failures with PodIT
With kubernetes 1.26 we are still seeing errors with pod upload:

```
Error:  Failures: 
Error:    PodIT.uploadBinaryFile:318 array lengths differ, expected: <16385> but was: <8192>
Error:    PodIT.uploadFile:298 expected: <I'm uploaded> but was: <>
```

https://github.com/fabric8io/kubernetes-client/actions/runs/4252695615/jobs/7396899761#step:6:579

With additional trace logging failed test runs showed the same data being sent to the server either way - as measured down at the level at which okhttp writes to the ssl socket.

There doesn't seem to be a good explanation for this other than we are more prone to seeing it on the containerd minikube runtime.  

The issue does seem to match one reported upstream https://github.com/kubernetes/kubernetes/issues/112834

We may want to add a formal warning about this behavior so that users are aware of potential data loss.

shawkins:
## Description
Only 1 real issue was fully tracked down with #4910 / #4923 - vertx closeHandler is immediate.  That needs changed to the endHandler, which is notified in order.  Reviewing the vertx logic does also highlight that we'll have a problem if the server ever does send a ping - but that should not currently be the case.  The only "ping" sent currently is the 0 byte message.

For the common failures some additional debugging has been added to the failure message to narrow in on where the problem lies.

I had thought there was a concurrency issue with PodOperationsImpl.readTo, but that shouldn't be the case - the stream was still being closed in the calling method's try.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@
 * Fix #4891: address vertx not completely reading exec streams
 * Fix #4899: BuildConfigs.instantiateBinary().fromFile() does not time out
 * Fix #4908: using the response headers in the vertx response
+* Fix #4910: addressing inconsistent behavior with pod exec operations
+* Fix #4923: addressing inconsistent behavior with pod exec operations
 * Fix #4931: using coarse grain locking for all mock server operations
 * Fix #4947: typo in HttpClient.Factory scoring system logic
 * Fix #4928: allows non-okhttp clients to handle invalid status
--- a/httpclient-vertx/src/main/java/io/fabric8/kubernetes/client/vertx/VertxWebSocket.java
+++ b/httpclient-vertx/src/main/java/io/fabric8/kubernetes/client/vertx/VertxWebSocket.java
@@ -44,7 +44,14 @@ class VertxWebSocket implements WebSocket {
       ws.pause();
       listener.onMessage(this, msg);
     });
-    ws.closeHandler(v -> listener.onClose(this, ws.closeStatusCode(), ws.closeReason()));
+    // if the server sends a ping, we're in trouble with our fetch strategy as there is
+    // no ping handler to increase the demand - this should not be an immediate issue as
+    // the api server does not seem to be sending pings
+
+    // if for whatever reason we send a ping, pong counts against the demand, so we need more
+    ws.pongHandler(b -> ws.fetch(1));
+    // use end, not close, because close is processed immediately vs. end is in frame order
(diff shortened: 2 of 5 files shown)


shawkins:
https://github.com/fabric8io/kubernetes-client/actions/runs/4376515845/jobs/7659072908#step:6:450 confirms a fresh instance where some of the stdIn upload is dropped.  Options here include:
- adding a time delay between when the last data is sent and close
- attempt to check inline with the operation or afterwards.  File upload and directory upload have divergent logic that makes this difficult.  We may want to consolidate to just tar and a two step process - send the tar.gz, then in a separate exec check the length or checksum, then extract.

https://github.com/fabric8io/kubernetes-client/actions/runs/4268450566/jobs/7430955144#step:5:448 since there is no upload could be an issue with stdOut download - but I could never quite confirm that locally.  Initially it seemed like an issue with the stream close, but that turned out to not be the case.  I'd like to get another reproduction of that locally or via the github runs before drawing more conclusions about how to proceed.

shawkins:
@manusa the first commit addresses the vertx close problem and some misc. cleanups.  The second commit is a speculative change help address the upload problem - basically throwing a few more messages at the server and a little wait to help ensure that what was already sent is processed.  This was done because I couldn't think of a good way an inline, or a post check that wouldn't require even more code.

manusa:
So from the final code changes I see we're addressing 3 things here:
- Vert.x ping-pong handling: request more frames in case a pong is received (added a handler instead of disabling client-sent pings)
- Vert.x replacing `closeHandler` with `endHandler` which waits for the previous frames to be processed as opposed to closing the stream immediately
- Pod (Download) Operations: moved the exitCode blocking retrieval to the try-with-resources block. OutputStream is clearly closed after the server process has completed.

There still remains a few issues to solve:
- Delaying the close operation server-side for tar extract operation (might complete before the buffers are flushed) -Upload-
- Upload/Download checksums

_please feel free to edit this comment to make it more accurate_

manusa:
I also think that we might want to add a little complexity to the upload operation (maybe also consolidate into a single implementation regardless of file or directory -like you proposed-). Perform 2 commands, one for the upload, one for the checksum. The upload command is only completed (local close processing is delayed) after the second command is successful. The checksum command is executed after the upload completes (websocket remains open) and is retried `n` times until success. I understand that this should cover the issue, but maybe I'm misinterpreting something.

manusa (approved):

PR #4968, opened 2023-03-13
fix #4910: speculative changes for upload

### LINKED ISSUE #4910: Issues with upload / failures with PodIT

shawkins:
## Description
Trying to fully address the upload failures.

- they seem to happen more with the single file upload rather than with the directory upload, but that could simply be because we have more tests doing single file uploads.  Comparing to kubectl, they do everything through tar, so I tried to do that here.  It will limit the files to 8GB.  However commons compress tar does not work well with streams of unknown size.  There is an ugly reflection based workaround, which may be too hackish.
- I couldn't get completion detection working very well with just a single exec if we use gzip.  You have to send a very large post amount of data to force the tar -z to give up on the stdIn without an explicit close.  Here again, like kubectl I've just dropped using gzip - https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/cp/cp.go
- Assuming the missing data may still happen with tar, at the end of the tar a hidden file that signals we're properly done.  A background process will see that and we'll pick that up and echo to stdOut.  The other approach to this is to copy the file in one exec, then perform a size / termination check, expansion, and removal of the temp file as a second exec.  We'd be required to use that approach to retain compression.

@manusa and others - any thoughts on the first and last points?

## Type of change

## Checklist


oscerd (approved):

shawkins:
We can workaround 1 by shading the class ourselves if/until an upstream fix is supplied.  The only other problem is that commons-compress is currently an optional dependency, it will be required then for all file uploads.

I'll decouple the fix of EOF detection until after we hit more of the upload issues - if for some reason they are more problematic with piping cat to file then we may not need to do anything else.

shawkins:
@manusa @rohanKanojia if no one objects to the approach here, I'll go ahead and fix up the unit tests to work with the new upload verification logic.

shawkins:
Based upon the failures in https://github.com/fabric8io/kubernetes-client/actions/runs/4503720351/jobs/7927685644?pr=4968 and the lack of responses back from the server, we can infer that while we had written all of the bytes, the final bytes had not been written on the server side.  So the flushing of stdin is not predictable and we cannot simply wait on it.  We either would need to send more data through, then truncate or we can simply keep retrying.  It seems to make more sense to do the latter.

kubernetes-itests/src/test/java/io/fabric8/kubernetes/PodIT.java
```diff
@@ -289,16 +291,36 @@ void uploadFile(String uploadPath) throws IOException {
...
     try (InputStream checkIs = podResource.file(uploadPath).read();
         BufferedReader br = new BufferedReader(new InputStreamReader(checkIs, StandardCharsets.UTF_8))) {
       String result = br.lines().collect(Collectors.joining(System.lineSeparator()));
       assertEquals("I'm uploaded", result, () -> checkFile(podResource, null, uploadPath));
     }
   }
 
+  void retryUpload(BooleanSupplier operation) {
+    Awaitility.await().atMost(60, TimeUnit.SECONDS).until(operation::getAsBoolean);
+  }
```
manusa, line 305: I'm not really convinced about this, might eventually hide something else.

shawkins: Certainly as we are only dealing with the boolean return so this does not qualify exactly the nature of the failed upload.  However most other problems are likely to be deterministic, in which case we'll exceed the timeout.  Only other somewhat rare non-deterministic problems would be missed.

PR #4976, opened 2023-03-16
fix #4975: using common scale logic for all resources

### LINKED ISSUE #4975: Ability to Scale CustomResource

I am looking for an alternative of scaling custom resource like with following command:
```
kubectl scale activemqartemis/artemis-broker -n activemq-artemis-operator --replicas=0 # --replicas=10
```

Currently I see that only standard kubernetes resources are supported like deployment, statefulset, replicaset etc.
The thing is, that if I scaledown SS or RS our "deployment" uses, after `oc scale ss` operation finishes, out ClusterOperator reverts it back to previous state. (Which expects to call oc scale CR to update it's real value).

Ability to call something like 
`myGeneratedCRClient().inNamespace(namespace).resource(myResource).scale(int);`

_No response_

### Additional context

_No response_

shawkins:
## Description
Addresses #4975 by moving scale operations to Resource.

Moved the main scale operation from a spec change to the subresource.  And modified the wait logic to use the subresource as well - which means we need to switch to polling instead.  This will conflict with #4949.

The v1beta1 I was unsure of - it uses beta1.scale instead.  So I just moved the legacy logic over there.  Some additional testing might clarify if the common logic will be okay there, or we'll just roll off it eventually...

JobResource was already Scalable, but it doesn't actually support the scale subresouce, so it needs to retain its current scale count methods.  If you call scale() on it or any other resource that does not support the scale subresource, you'll get a 404.  Or we can update all of them to a better message if desired.

It's possible to save a Scale GET when calling scale count, but that requires a bit more logic as the return type from a scale call is a Scale, not the resource type.

This does highlight the limitation of our dryRun logic - that's currently applied only to WriteableOperation, but that of course can include Scale, so there's currently no way to specify a dryRun scale operation.

## Type of change

## Checklist


rohanKanojia (approved):

rohanKanojia:
@shawkins : Is it possible to add an integration test to verify `scale` is working for Custom Resources specifically?

shawkins:
> @shawkins : Is it possible to add an integration test to verify `scale` is working for Custom Resources specifically?

Added a fully generic test.

manusa (approved):

PR #4981, opened 2023-03-20
Fix: Vert.x HttpClient stackoverflow

manusa:
Authored by @vietj

Relates to https://github.com/quarkusio/quarkus/issues/31872

The implementation of the InputStream to ReadStream adapter can sometimes do a stack overflow when the InputStream buffer extraction on a worker thread is faster than the event-loop. We should guard against this recursion and trampoline when the reads starts to pile (8).


shawkins (approved):
LGTM

geoand:
I love this simple trampoline implementation :)

PR #4989, opened 2023-03-22
Upgrade Tekton Model to v0.47.0

dziemba:
## Description
- Upgrade Tekton Pipelines to v0.47.0
- Upgrade Tekton Triggers to v0.23.0
- Add the new `v1` Tekton Pipelines API Group
- Merge generator code into single package per module (to avoid duplicating boilerplate + version drifts between API groups) and remove obsolete Golang setup (e.g. vendoring)
- Rename `ArrayOrString` to `ParamValue` (has been renamed upstream)
- Fix various tests, delete obsolete ones

## Type of change

## Checklist


rohanKanojia:
@dziemba : I think we should keep old generators v1alpha1 and v1beta1 for backward compatibility.

dziemba:
First off: This whole PR turned out a bit larger than I anticipated. If you like me to split up some things or revert some of the (opinionated) changes, I'm happy to do that :) 

> I think we should keep old generators v1alpha1 and v1beta1 for backward compatibility.

Do you mean keeping them pinned to the older Tekton versions? I thought about that, and backwards compatibility would be super nice to have. The problem I see with that approach though is how Tekton creates new APIs. They start in alpha, then get promoted to beta and then to stable apparently. 

Screenshot 2023-03-23 at 10 27 22

Looking at the changes for v1alpha1, we are losing most of the old objects, which breaks backwards compatibility. But we also get two new objects, which we wouldn't be able to use if we're pinning v1alpha1 to an older Tekton version.

I think ideally we would need to have multiple packages for different Tekton versions, that each contain all API group models for that given version. That will quickly become a lot of code and maintenance effort though. Especially since Tekton is still on v0.x.

How do you suggest we proceed here? I'm happy to revert the changes to v1alpha1 back to how it was, but I'm a bit worried that won't be very future-proof.

rohanKanojia:
@shawkins @manusa : Could you please provide some guidance on what should be the appropriate strategy here? Shall we keep old apiGroups to older versions or switch to the latest versions.

dziemba:
@rohanKanojia @shawkins @manusa Any update on which direction to go in? I'm happy to change this PR as you see fit :)

dziemba:
@vdemeester @afrittoli @bobcatfish ([Tekton topical owners for API Versioning](https://github.com/tektoncd/pipeline/blob/main/topical-ownership.md)) do you have any suggestions on what the right approach here is?

vdemeester:
I am not sure I fully followed but tekton follow the same practices as k8s. API versions go from alpha to stable. Some alpha types can be removed as, as state in most api practice, alpha is.. breakable, from one release to the other, breaking changes on field or removal of a type can happen. 

I think a library should focus more on the api it supports than tekton version themselves. A gist from the [doc](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md) 
- a `v1` type today should work the same in one year (~10/12 releases) as everyting has to be done in a backward compatible way
- a `v1beta1` type today should work the same in 9 month — after that "some" fields could be removed if properly marked as deprecated
- a `v1alpha1` type can break tomorrow

As of today (0.46), most types are either in `v1` (the main ones, like `Pipeline`, `Task`, …) and `v1beta1` (`CustomRun`, …).

extensions/tekton/examples/src/main/java/io/fabric8/tekton/api/examples/TaskCreate.java
```diff
@@ -1,51 +0,0 @@
...
- */
-package io.fabric8.tekton.api.examples;
-
-import io.fabric8.tekton.client.DefaultTektonClient;
-import io.fabric8.tekton.client.TektonClient;
-import io.fabric8.tekton.pipeline.v1beta1.Task;
-import io.fabric8.tekton.pipeline.v1beta1.TaskBuilder;
-import io.fabric8.tekton.pipeline.v1beta1.TaskList;
-
-public class TaskCreate {
```
rohanKanojia, line 24: I can still see v1beta1 Task in generated resources. Why do we need to delete this example?

extensions/tekton/examples/src/main/java/io/fabric8/tekton/api/examples/v1beta1/TaskRunCreate.java
```diff
@@ -1,60 +0,0 @@
...
-          .endResources()
-          .endSpec()
-          .build();
-
-      // Create TaskRun
-      taskRun = tektonClient.v1beta1().taskRuns().inNamespace(namespace).create(taskRun);
-      System.out.println("Created: " + taskRun.getMetadata().getName());
-
-      // List TaskRun
-      TaskRunList taskRunList = tektonClient.v1beta1().taskRuns().inNamespace(namespace).list();
```
rohanKanojia, line 56: I can still see v1beta1 TaskRun in generated resources. Why do we need to delete this example?

extensions/tekton/tests/src/test/java/io/fabric8/tekton/pipeline/v1beta1/TaskTest.java
```diff
@@ -32,7 +32,7 @@ void testTaskFromJSON() throws IOException {
     ObjectMapper mapper = new ObjectMapper();
     Task task = mapper.readValue(is, Task.class);
     assertEquals("Task", task.getKind());
-    assertEquals("tekton.dev/v1beta1", task.getApiVersion());
+    assertEquals("tekton.dev/v1", task.getApiVersion());
```
rohanKanojia, line 35: This doesn't look okay to me. Could you please create a separate test for v1 Task?

extensions/tekton/generator/go.mod
```diff
@@ -0,0 +1,82 @@
+module github.com/fabric8io/kubernetes-client/extensions/tekton/generator
+
+require (
+	github.com/fabric8io/kubernetes-client/generator v0.0.0
+	github.com/tektoncd/pipeline v0.46.0
```
rohanKanojia, line 5: I see that [0.47.0](https://github.com/tektoncd/pipeline/tree/v0.47.0) also got released a few days ago. Shall we bump to latest version?

extensions/tekton/generator/cmd/generate/generate-v1beta1.go
```diff
@@ -30,6 +30,7 @@ func main() {
 	// no other types need to be defined as they are auto discovered
 	crdLists := map[reflect.Type]schemagen.CrdScope{
 		// v1beta1
```
rohanKanojia, line 32: I also see a new resource `ResolutionRequest` in `resolution.tekton.dev/v1beta1` apiGroup. Do you think we should add it?

extensions/tekton/tests/src/test/java/io/fabric8/tekton/pipeline/v1beta1/TaskTest.java
dziemba: Ah yes, this is still in the `v1beta1` namespace, that's no good. Should I duplicate this test for v1 and leave the v1beta1 one as-is?

extensions/tekton/examples/src/main/java/io/fabric8/tekton/api/examples/TaskCreate.java
dziemba: I tried to run these and they didn't work. I thought it makes sense to clean up broken examples. I can have another look at this if you want.

dziemba:
Thanks for the review, I'll address the feedback soon

extensions/tekton/generator/cmd/generate/generate-v1beta1.go
dziemba: I'll have a look at that

extensions/tekton/examples/src/main/java/io/fabric8/tekton/api/examples/v1beta1/TaskRunCreate.java
dziemba: I restored this file

extensions/tekton/examples/src/main/java/io/fabric8/tekton/api/examples/TaskCreate.java
dziemba: I restored this file

extensions/tekton/tests/src/test/java/io/fabric8/tekton/pipeline/v1beta1/TaskTest.java
dziemba: I restored this as-is and make a new test for the `v1` one

dziemba:
@rohanKanojia I addressed all feedback now, please have another look :)

manusa:
Hi @dziemba,

Sorry for the late feedback. This morning we noticed that some of the structs are missing.

More specifically, the Interceptor in the triggers module (https://pkg.go.dev/github.com/tektoncd/triggers@v0.23.1/pkg/apis/triggers/v1alpha1#Interceptor).

Could you please add that one too?

dziemba:
@manusa Sure, I've now added the `Interceptor` types.

manusa:
> @manusa Sure, I've now added the `Interceptor` types.
> 
> Update: I forgot to add the new type to the API DSL, I just pushed a new commit that includes these changes as well.

Hi @dziemba, sorry for the late updates. We've been trying to release a stable 6.6 version.

If you covered everything we mentioned, then it should be good to go. Don't worry about changelog conflicts or similar. If you checked the `Allow edits and access to secrets by maintainers` I should be able to take care of that.

rohanKanojia:
This is also affected by https://github.com/fabric8io/kubernetes-client/issues/5113 . We should refactor trigger generators to remove the split package warning.

rohanKanojia (approved):

metacosm:
Should we update triggers to 0.24 that got released last week?

manusa:
CHANGELOG was not correctly updated, I'll fix after merging to avoid re-running all pipelines.

Improvement entries need to be removed

manusa (approved):

manusa:
LGTM, thx @dziemba!

PR #5007, opened 2023-03-29
[java-generator] - Handling potentially duplicated fields which could be generated by deprecated schema properties

### LINKED ISSUE #4793: [java-generator] In some circumstances the generated code contains duplicated fields.
### Describe the bug

The Java Generator is emitting fields without checking that a field with a colliding name has been already generated.

SNAPSHOT

### Steps to reproduce

- `jbang io.fabric8:java-generator-cli:6.4.0 -u https://raw.githubusercontent.com/projectcalico/calico/f1a1611acb98d9187f48bbbe2227301aa69f0499/manifests/canal.yaml -t ./`
- check the file `org/projectcalico/crd/v1/IPPoolSpec.java`
- there are two fields `private Boolean natOutgoing;` one referencing the deprecated field `nat-outgoing` and one `natOutgoing`

### Expected behavior

The code emitted by the java-generator should compile.

There are a bunch of strategies in case a colliding field has been found:

- prepending an `_` (or something)
- adding a trailing number
- checking if there is a "deprecation" note in the description and skip(or apply a different mangling) to one of the field
- something in between the previous

### Runtime

Kubernetes (vanilla)

1.25.3@latest

### Environment

Linux, macOS

_No response_

### Additional context

found validating the issue raised by @sfc-gh-valiu

fabiobrz:
## Description

A naive approach for fixing duplicate occurrences of fields that could be generated from different schema properties.

Fixes #4793 

CC @andreaTP 

## Type of change

## Checklist


andreaTP:
Nice to see you around @fabiobrz ! Thanks a lot for getting to this, love the approach!
I'll ask a few questions in-line 🙂

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
```diff
@@ -94,10 +95,41 @@ public JObject(
       for (Map.Entry<String, JSONSchemaProps> field : fields.entrySet()) {
         String nextPrefix = (config.getPrefixStrategy() == Config.Prefix.ALWAYS) ? classPrefix : "";
         String nextSuffix = (config.getSuffixStrategy() == Config.Suffix.ALWAYS) ? classSuffix : "";
+        String fieldKey = field.getKey();
+        // let's handle duplicates
+        final Map<String, Map<String, JSONSchemaProps>> duplicates = new HashMap<>();
+        // counting the current field duplicates
+        final List<Map.Entry<String, JSONSchemaProps>> fieldDuplicates = fields.entrySet().stream()
```
andreaTP, line 102: Despite the fact that I really like the readability of this solution it involves, at the moment, a pretty expensive computation.
Have you attempted to perform the field name check/sanitization once per `JObject` before the loop iterating on the fields (e.g. before line 95)?

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
```diff
@@ -94,10 +95,41 @@ public JObject(
...
+            .filter(f -> sanitizeString(field.getKey()).equals(sanitizeString(f.getKey())))
+            .collect(Collectors.toList());
+        final Long duplicatesCount = fieldDuplicates.stream().count();
+        final Boolean duplicatesExist = duplicatesCount > 1;
+        if (duplicatesExist) {
+          // we want to throw an exception on some duplicates missing requirements. At the moment, the only one we
+          // enforce is for a (duplicatesCount - 1) number of duplicates to be marked as deprecated
+          final Long deprecatedDuplicates = fieldDuplicates.stream()
+              .filter(d -> d.getValue().getDescription().startsWith(DEPRECATED_FIELD_MARKER)).count();
```
andreaTP, line 111: I would keep this a little more relaxed like:

```
getDescription().toLowercase().contains("deprecated")
```
wdyt?

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
```diff
@@ -94,10 +95,41 @@ public JObject(
...
+          if (deprecatedDuplicates != (duplicatesCount - 1)) {
```
andreaTP, line 112: I'm perfectly fine with this approach, but I would love to quickly discuss this, have you considered alternative approaches?
E.g. append a count number on duplicates instead of failing

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
```diff
@@ -94,10 +95,41 @@ public JObject(
...
+                    duplicatesCount, deprecatedDuplicates));
+          }
+          final String fieldName = sanitizeString(field.getKey());
+          Map<String, JSONSchemaProps> existingDuplicates = duplicates.get(fieldName);
+          if (existingDuplicates == null) {
+            existingDuplicates = new HashMap<>();
+            duplicates.put(fieldName, existingDuplicates);
+          }
+          if (field.getValue().getDescription().startsWith(DEPRECATED_FIELD_MARKER)) {
+            fieldKey += String.format("-deprecated%d", existingDuplicates.keySet().size());
```
andreaTP, line 124: uh, noticing it now, I would take one of the two:
- fail on > 1 duplicate
- append the counter to duplicates but omit the counter when there is a single element

andreaTP, line 124: as an addition, we can also annotate the field with 'Deprecated'

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
fabiobrz: I am not sure, to be honest. Any description containing such string would mark the field as deprecated, even when it is not. I'd be more strict here and keep it as is. That being said, it would need a correspondent approach in crd-generator.
WDYT?

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
fabiobrz: Can you share a pointer to some spec draft? I only found http://json-schema.org/draft/2020-12/json-schema-validation.html#name-deprecated (see https://github.com/json-schema-org/json-schema-spec/pull/737/files) which makes me think we could rely on this rather than on the description?

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
fabiobrz: The approach I considered allows for more than one deprecated duplicated field, allowing for exactly one non-deprecated duplicated field.
But if you think this use case is not reasonable, we can definitely switch to a simpler check. I don't have a strong opinion, feel free to let me know what you prefer.

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
fabiobrz: See https://github.com/fabric8io/kubernetes-client/pull/5007#discussion_r1151953888
+1 on removing the counter when there's just one duplicate.

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
andreaTP: Anything we do here is already well beyond the compatibility with the crd-gen ...
I would not be worried as we are handling edge cases.
I think we can agree on a compromise:
- keep `startsWith` but make sure we trim first the string
- lower the requirement using `toLowerCase` and match on `deprecated`

I do believe this is a reasonable compromise :-)

fabiobrz: +1, thanks, fixing!

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
andreaTP: Personally, I'm fine with this but I will probably prefer allowing non-deprecated duplicates as well.
I don't have a super strong opinion here either, I would love to have more feedback on this cc. @manusa @metacosm @shawkins

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
andreaTP: I've never seen the `deprecated` field used in the CRDs sadly.
Here I was asking to use JavaParser to add the `@Deprecated` annotation on the Java field.

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
```diff
@@ -243,6 +275,10 @@ public GeneratorResult generateJava() {
                         + ")"));
           }
         }
+
+        if (k.matches(".*-deprecated(\\d$)?$")) {
```
andreaTP, line 279: Can we avoid this regexp continuously being computed on the hot path?
E.g. you can set a boolean here: https://github.com/fabric8io/kubernetes-client/pull/5007/files#diff-c55bf304c5b566fc02eed93f3960ca65807f07dcaa34dc364a43f953463df066R107

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
```diff
@@ -314,4 +350,8 @@ static ClassOrInterfaceType toClassOrInterfaceType(String className) {
     return withoutDollars.indexOf('<') >= 0 ? StaticJavaParser.parseClassOrInterfaceType(withoutDollars)
         : new ClassOrInterfaceType(null, withoutDollars);
   }
+
+  private static String sanitizeFieldNameFromDefinition(Map.Entry<String, JSONSchemaProps> fieldDefinition) {
```
andreaTP, line 354: [nitpick] we are using this method exactly once, I would not do this refactoring ...

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
```diff
@@ -75,11 +76,42 @@ public JObject(
...
       for (Map.Entry<String, JSONSchemaProps> field : fields.entrySet()) {
+        String fieldKey = field.getKey();
+        // lookup the duplicated field properties map
+        final Map.Entry<String, Map<String, JSONSchemaProps>> fieldDuplicatesDefinition = groupedFieldDefinitions.entrySet()
+            .stream()
+            .filter(d -> d.getValue().containsKey(field.getKey()))
+            .findFirst().get();
+        if (fieldDuplicatesDefinition.getValue().entrySet().size() > 1) {
+          Set<Map.Entry<String, JSONSchemaProps>> fieldDuplicates = fieldDuplicatesDefinition.getValue().entrySet();
+          final Long duplicatesCount = fieldDuplicates.stream().count();
```
andreaTP, line 96: length?

andreaTP (commented):
Added a few comments, and let's wait to see if someone has a stronger opinion on the logic, amazing work! Thanks a lot @fabiobrz always a pleasure to see you around!

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
fabiobrz: `generateJava()` uses `this.fields` as the iteration base, hence IIUC you mean to add a boolean to `AbstractJsonSchema2Pojo`, right?

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
fabiobrz: You're definitely right, thanks, fixing!

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
andreaTP: I mean that we don't need to base this check on the name of the field, but we can set a boolean when mangling the name directly in first place.

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
andreaTP: Just avoid to convert to a stream :-)

fabiobrz: Indeed, makes sense :slightly_smiling_face:

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
fabiobrz: Right, I am not sure I did what you expected for me to do :smile: ... but I did remove the regex from the hot path and stored the deprecated keys when mangling them. WDYT?

fabiobrz:
> Added a few comments, and let's wait to see if someone has a stronger opinion on the logic, amazing work! Thanks a lot @fabiobrz always a pleasure to see you around!

Thanks @andreaTP for your kind words, hugely appreciated by the way. :slightly_smiling_face: 
I am definitely glad to contribute to the project, as you know!

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
fabiobrz: I eventually decided to push a change that reflects what you initially asked for @andreaTP. 
Having more than one duplicated field seems really to be a rare (and I'd say bad) use case, which shouldn't be supported.
Additionally, this allows for us to enforce a more strict requirement about the reason for having such duplicated fields, i.e. one of them has to be deprecated. 

The implementation looks better this way to me, let me know what you think.

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
```diff
@@ -75,11 +77,53 @@ public JObject(
...
+
       for (Map.Entry<String, JSONSchemaProps> field : fields.entrySet()) {
+        String fieldKey = field.getKey();
+        // lookup the duplicated field properties map
+        final Map.Entry<String, Map<String, JSONSchemaProps>> fieldDuplicatesDefinition = groupedFieldDefinitions.entrySet()
+            .stream()
+            .filter(d -> d.getValue().containsKey(field.getKey()))
+            .findFirst().get();
+        final int duplicatesCount = fieldDuplicatesDefinition.getValue().entrySet().size();
+        if (duplicatesCount > 1) {
```
andreaTP, line 95: `fieldDuplicatesDefinition` is generated from a `findFirst()` call, if I'm not missing anything this condition is not going to be triggered.
Can you add a `GeneratorTest` to cover this aspect, please 🙏 ?

andreaTP (changes requested):
Added one more comment, thanks for bearing with me!

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
andreaTP: I'm ok with starting with the logic proposed here!

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
andreaTP: it's a good first iteration, agreed 👍

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
fabiobrz: > we should be able to do a direct lookup:
> `groupedFieldDefinitions.get(AbstractJSONSchema2Pojo.sanitizeString(f.getKey()));`

Indeed, I went the long way, when instead it was just there at hand, thanks. Fixing!

> The note on testing still applies if that's not too much work ...
Sure. I'll add a couple of tests for the requirements we're enforcing: no more than 1 duplicated field && the duplicated field must be deprecated.

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
```diff
@@ -75,11 +77,50 @@ public JObject(
...
+          .collect(Collectors.groupingBy(
+              (f) -> AbstractJSONSchema2Pojo.sanitizeString(f.getKey()),
+              Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)));
+
       for (Map.Entry<String, JSONSchemaProps> field : fields.entrySet()) {
+        String fieldKey = field.getKey();
+        // lookup the duplicated field properties map
+        final Map<String, JSONSchemaProps> fieldDuplicatesDefinition = groupedFieldDefinitions
+            .get(AbstractJSONSchema2Pojo.sanitizeString(field.getKey()));
+        final int duplicatesCount = fieldDuplicatesDefinition.entrySet().size();
```
andreaTP, line 92:
```suggestion
        final int duplicatesCount = fieldDuplicatesDefinition.size();
```

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
```diff
@@ -75,11 +77,50 @@ public JObject(
...
+          // for exactly 1 field duplicate to exist
+          if (duplicatesCount > 2) {
+            throw new JavaGeneratorException(
+                String.format("The %s field has %d duplicates, which is not a supported configuration",
+                    field.getKey(),
+                    duplicatesCount - 1));
+          }
+          // another requirement that we enforce is that if one field duplicate exists, then it's because it has
+          // been marked as deprecated
+          final long deprecatedDuplicates = fieldDuplicatesDefinition.entrySet().stream()
```
andreaTP, line 105: I think we can use `anyMatch` instead of counting exactly the number of duplicates (now it should be a fixed number).

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
fabiobrz: Correct, thanks, fixing!

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
fabiobrz: I'd use `noneMatch`:

```
final Boolean noDeprecatedDuplicatesExist = fieldDuplicatesDefinition.entrySet().stream()
              .noneMatch(d -> d.getValue().getDescription().trim().toLowerCase().startsWith(DEPRECATED_FIELD_MARKER));
```

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
fabiobrz: This is fixed, resolving.

andreaTP (approved):
The last comments are just nitpicking, thanks for the great work @fabiobrz !
This is already good to go as soon as the CI is green and conflicts resolved 👍 
LGTM

manusa (approved):

PR #5010, opened 2023-03-29
fix #5009: ensuring underlying serialization is used

### LINKED ISSUE #5009: Invalid serialization of wrapped polymorphic types
https://github.com/quarkiverse/quarkus-operator-sdk/issues/543 is due to the UntypedFieldTypeModule inhibiting the polymophic handling.

shawkins:
## Description
Fixes #5009 by converting the logic in BeanPropertyWriterDelegate to use it's super methods.  This is like the logic in Jdk8OptionalBeanPropertyWriter, which is also installed via a BeanSerializerModifier.  @manusa if you want to future proof this then a check can be added that the delegate class is exactly BeanSerializerModifier, and not a subclass.  The corner case being that another module is contributing a conflicting BeanSerializerModifier - note that the Jdk8Module does not concern itself with that.

Alternatively the pattern with filtered properties appears to be that direct references to the delegates are used.  However the api does not make it clear what is required to override for this to work as expected, so there seems to be a greater maintenance burden.

To limit potential future issues the scope of when the XXXDelegates was limited to only when an any value method is found - that should help in custom scenarios as internally we'll generate additionalProperties on all kubernetes resources.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Fix #4963: Openshift Client return 403 when use websocket
 * Fix #4985: triggering the immediate cleanup of the okhttp idle task
 * fix #5002: Jetty response completion accounts for header processing
+* Fix #5009: addressing issue with serialization of wrapped polymophic types
 
 #### Improvements
 * Fix #4434: Update CronJobIT to use `batch/v1` CronJob instead
--- a/kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/utils/SerializationWrappedPolymorphicTest.java
+++ b/kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/utils/SerializationWrappedPolymorphicTest.java
@@ -1,0 +1,119 @@
+/**
+ * Copyright (C) 2015 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.fabric8.kubernetes.client.utils;
+
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.fasterxml.jackson.annotation.JsonTypeName;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import io.fabric8.kubernetes.api.model.KubernetesResource;
+import io.fabric8.kubernetes.api.model.Namespaced;
+import io.fabric8.kubernetes.client.CustomResource;
+import io.fabric8.kubernetes.model.annotation.Group;
+import io.fabric8.kubernetes.model.annotation.Version;
+import lombok.Data;
+import org.junit.jupiter.api.Test;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class SerializationWrappedPolymorphicTest {
+
+  @Group("example.com")
+  @Version("v1alpha1")
+  static class TestCR extends CustomResource<TestCR.TestCRSpec, Void> implements Namespaced {
+
+    @JsonDeserialize(using = JsonDeserializer.None.class)
+    static class TestCRSpec implements KubernetesResource {
+
+      @JsonIgnore
+      private Map<String, Object> additionalProperties = new HashMap<String, Object>();
+
+      private List<Child> children = new ArrayList<>();
+
+      public List<Child> getChildren() {
+        return children;
+      }
+
+      public void setChildren(List<Child> children) {
+        this.children = children;
+      }
+
+      @JsonAnyGetter
+      public Map<String, Object> getAdditionalProperties() {
+        return this.additionalProperties;
+      }
+
+      @JsonAnySetter
+      public void setAdditionalProperty(String name, Object value) {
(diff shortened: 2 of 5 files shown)


shawkins:
A couple of additional thoughts:
- the jackson apis are ill-suited for this task.  Either you are dependent on trying to guess how much delegation is needed, or you can only have one modification via a subclass of a BeanPropertyWriter per objectmapper.  Even with this pr we have more than likely need methods like unwrappingWriter implemented to preserve our customized behavior in additional circumstances.  
- At worst if the BeanPropertyWriterDelegate is not used the object written will be invalid - but the user may clean things up by setting the appropriate field to null/default if omitted, or removing the additionalProperty.
   - So we may want to consider removing this writer behavior altogether in favor of making the user explicitly resolve the ambiguity.
- At least for fabric8 core this whole untyped problem was introduced via template handling.  If we instead move template logic to strictly use generic or raw, we could remove, or make optional, the untyped handling.  That would require the user to do relevant parameter substitution before converting to the strongly typed form.  This becomes somewhat cleaner / easier if we move forward with the changes that make the serialization logic non-static.

@manusa @metacosm @rohanKanojia what are your thoughts?

manusa:
> A couple of additional thoughts:
> 
>     * the jackson apis are ill-suited for this task.  Either you are dependent on trying to guess how much delegation is needed, or you can only have one modification via a subclass of a BeanPropertyWriter per objectmapper.  Even with this pr we have more than likely need methods like unwrappingWriter implemented to preserve our customized behavior in additional circumstances.
> 
>     * At worst if the BeanPropertyWriterDelegate is not used the object written will be invalid - but the user may clean things up by setting the appropriate field to null/default if omitted, or removing the additionalProperty.
>       
>       * So we may want to consider removing this writer behavior altogether in favor of making the user explicitly resolve the ambiguity.
> 
>     * At least for fabric8 core this whole untyped problem was introduced via template handling.  If we instead move template logic to strictly use generic or raw, we could remove, or make optional, the untyped handling.  That would require the user to do relevant parameter substitution before converting to the strongly typed form.  This becomes somewhat cleaner / easier if we move forward with the changes that make the serialization logic non-static.
> 
> 
> @manusa @metacosm @rohanKanojia what are your thoughts?

This has been the first issue reported with our customized writer/reader wrappers in a very long period. I don't really think that this is that problematic.

We have a strong usage of this in JKube where we do modifications to typed resources that might contain values which are not handled by the Java types but that might later be replaced.

This is the same for Templates. Moving that handling to a Generic Resource works internally (this is partially what was done as a workaround before we introduced this fix). However, I'm sure that many users do modifications of the Templates and do expect a Template type.

In addition, internally yes, moving Template Handling to `GenericResource` works for most cases (read, write, create).

However, and to be more explicit about what won't work and should be expected to:
```java
client.templates().withName("template").edit(t -> new TemplateBuilder(t)/* ... */);
```

There are many other Jackson modules around that provide similar behavior, so I can't really think of a better and more suitable approach that works for all scenarios. IMHO the strict delegate pattern (prior to this PR changes) was good enough. An edge case has now been identified and it's easy to reproduce, adding a test and overriding the rest of methods to make the test pass and complete the unfinished delegate pattern should be just fine.

manusa:
As an additional item.
Once serialization is non-static, everything becomes much easier because we can add methods to customize how the Client deals with serialization. I really think it's better to fix the bug and then concentrate the effort in providing per-client-instance serialization capabilities.

shawkins:
> This has been the first issue reported with our customized writer/reader wrappers in a very long period. I don't really think that this is that problematic.

It's hard to guess how problematic it will be.  I agree that for the built-in objects/serialization that is generated by the project logic what we had was good enough.  It appears that the likely corner cases will come from custom resources where the full range of jackson could be used.  Here it was a problem with wrapping, it's also clear that the unwrapped case isn't currently handled.

> client.templates().withName("template").edit(t -> new TemplateBuilder(t)/* ... */);

Yes that would still work as generic.  The usability issue with strictly raw or generic is the conversion needed when manipulating the objects - you may want to convert to your expected type when calling getObjects, and when setting you'll need to convert or wrap because there's common base other than KubernetesResource.  We don't currently provide good helper methods for that.   

The alternative to strict generic would be to allow the parsing of KubernetesResource to gracefully handle an untyped situation in a different way - that is convert inline to a generic at the nearest KubernetesResource parent.  However I'm not really sure that's easily doable, so you'd likely have to always parse KubernetesResource as generic, then see if it could be converted to the typed response.  If that fails, then leave it as the generic.

> IMHO the strict delegate pattern (prior to this PR changes) was good enough. An edge case has now been identified and it's easy to reproduce, adding a test and overriding the rest of methods to make the test pass and complete the unfinished delegate pattern should be just fine.

That's fine, feel free to change the pr.  I've offered a solution and stated what the possible weak point is, but you can certainly change it as you see fit.

But I'd also give consideration to dropping this write manipulation altogether.

> I really think it's better to fix the bug and then concentrate the effort in providing per-client-instance serialization capabilities.

I think I'll have to keep peeling smaller changes out of that large pr.  As a single bit of work it touches too much stuff and still is incomplete in several places.

manusa:
I really think we should stick with this PR that seems to work.
Once GH and Maven are stable I think I'll add a subsequent PR to make the processing lazier at the module level and that's it.

> I think I'll have to keep peeling smaller changes out of that large pr. As a single bit of work it touches too much stuff and still is incomplete in several places.

This is where I think our efforts should be aiming. A non-static Serializer would make everything so much easier.

manusa (approved):

manusa:
I think the Istio implementation broke

manusa:
I'll provide a quick fix and release SNAPSHOTS to unblock the OperatorSDK dependent team. We can then update this one.

shawkins:
@manusa the istio tests are failing because this is hitting the problematic case - the BeanPropertyWriter is already a subclass UnwrappingBeanPropertyWriter, so overriding without preserving the delegate throws it away.  It seems like jackson should have hit this before - the combination of UnwrappingBeanPropertyWriter and Jdk8OptionalBeanPropertyWriter for example would cause the same problem.  Adding the check of the class type will prevent this problem, but also inhibit the unmatched logic at object contexts.  That may be okay as that can be expressly null, but it doesn't seem worth having a separate case.

I'll capture an upstream issue for the project to have / provide a real base instance that wraps an existing reader / writer - they should use it internally for things like UnwrappingBeanPropertyWriter, Jdk8OptionalBeanPropertyWriter, etc.

This pr can be closed.

manusa:
This case is the perfect example of what I was worried when we discussed about the delegate-super topic.
I'm not sure why some the upstream Writers follow different pattern implementations, not sure which was first and why this diverged over time.

shawkins:
> This case is the perfect example of what I was worried when we discussed about the delegate-super topic.

I'm honestly worried about both cases - there's no clarity in their api about what is needed.

manusa:
> I'm honestly worried about both cases - there's no clarity in their api about what is needed.

I would need to check the Git history, but I assume that over time things diverged from the original reasoning.

But I'm sure that the composition of several of this writers most likely and eventually ends up in "tragedy".

shawkins:
https://github.com/FasterXML/jackson-databind/issues/3862 is the upstream issue for this.

PR #5025, opened 2023-04-03
feat: additional callback methods for Interceptor interface

manusa:
## Description

Relates to:
- https://github.com/eclipse/jkube/issues/1950
- https://github.com/eclipse/jkube/issues/2000

Added two more methods to the Interceptor interface. The main intention is to provide enough functionality to implement an `HttpLoggingInterceptor` compatible with every `HttpClient` implementation.

The provided infrastructure will allow to implement (in a subsequent PR) an `HttpLoggingInterceptor` to replace the OkHttp-specific intereceptor which is also flawed.

@shawkins thoughts? As far as I can see, these changes will enable the `HttpLoggingInterceptor` that can be activated based on the configured logging-level. If this one's OK, I'll send the interceptor implementation after merging.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/Interceptor.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/Interceptor.java
@@ -16,11 +16,13 @@
 
 package io.fabric8.kubernetes.client.http;
 
+import java.nio.ByteBuffer;
+import java.util.List;
 import java.util.concurrent.CompletableFuture;
 
 public interface Interceptor {
 
-  public interface RequestTags {
+  interface RequestTags {
 
     <T> T getTag(Class<T> type);
 
@@ -33,6 +35,31 @@ public interface Interceptor {
    * @param request the current request
    */
   default void before(BasicBuilder builder, HttpRequest request, RequestTags tags) {
+  }
+
+  /**
+   * Called after a non-WebSocket HTTP response is received. The body might or might not be already consumed.
+   * <p>
+   * Should be used to analyze response codes and headers, original response shouldn't be altered.
+   *
+   * @param response the response received from the server.
+   */
+  default void after(HttpResponse<?> response) {
+
+  }
+
+  // In case we want to encapsulate to spy the responses from the server
+
+  /**
+   * Called before a request to allow the encapsulation of the provided consumer.
+   * <p>
+   *
+   * @param consumer the original consumer.
+   * @param request the HTTP request.
+   * @return the consumer to use.
+   */
+  default AsyncBody.Consumer<List<ByteBuffer>> consumer(AsyncBody.Consumer<List<ByteBuffer>> consumer, HttpRequest request) {
+    return consumer;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


oscerd (approved):

shawkins:
> thoughts? 

I'm presuming that you'll want this interceptor to be last in the chain, and that we'll log just based upon the after method - using the request from the response.  This is a probably close enough to the logging users will want to see.  

From our perspective correlating to the incoming user level request prior to the before modifications may be interesting, but has been made hard to do because no call spans both and we don't have unique request identifiers.  This also won't necessarily correlate intermediate requests by the interceptors - failed backwards compatibility and OpenShiftOAuthInterceptor token refreshes will be logged as separate requests.

> As far as I can see, these changes will enable the HttpLoggingInterceptor that can be activated based on the configured logging-level. If this one's OK, I'll send the interceptor implementation after merging.

Are you also looking to log the requests by the TokenRefreshInteceptor?

manusa:
> Are you also looking to log the requests by the TokenRefreshInteceptor?
> OpenShiftOAuthInterceptor, and so on

I assume that users want the complete log of the HTTP conversation regardless of its origin (client request, interceptor, etc). So I guess we do want to log these requests too.

TBH I didn't actually consider this scenario, but as far as composition looks, those requests should be logged too.

> I'm presuming that you'll want this interceptor to be last in the chain, and that we'll log just based upon the after method - using the request from the response

I'm not sure if we actually want this interceptor to be applied first instead. Anyway, the interceptor should be internal and added by us in a similar way as we are currently adding the OkHttp interceptor. So ensuring it's the first or last in the chain should be easy.

The interceptor should log same as OkHttp (I haven't checked but I assume):
- Request URI
- Request Headers
- Request Body
- Response Status
- Response Headers
- Response Body

shawkins:
> TBH I didn't actually consider this scenario, but as far as composition looks, those requests should be logged too.

The TokenRefreshInteceptor creates a new client instance, which is not based upon the Configuration - it uses the plain HttpClient.Factory.newBuilder() method.

> and the response body by the encapsulated consumer

I'm unsure how easily this will be applied.  Proxying the consumer will give us access to the bytes, but not the broader cycle - without additional changes you'd have to log the bytes as they are received because there is no separate notification to the consumer that we're done.

shawkins:
> The TokenRefreshInteceptor problem could be tackled by applying the LoggingInterceptor at the HttpClient layer.

Sure, it just becomes less customizable that way.

> Did you have something in mind?

There's no single great solution.  We'd have to look at content type and length to have some understanding of what's in the response - otherwise we can't easily tell a difference at this layer between binary and test or log streaming (or other large / long running download), and just a more typical response.  We'll probably have to buffer that up to a given size, then log in chunks if it looks like it's text.  Yes something like onBodyDone could work:

```
         @Override
          public AsyncBody.Consumer<List<ByteBuffer>> consumer(AsyncBody.Consumer<List<ByteBuffer>> consumer,
              HttpRequest request) {
            AtomicBoolean added = new AtomicBoolean();
            // some other state for buffering / length

            BiConsumer<Void, Throwable> onBodyDone = (v, t) -> {
              // some logging
            };

            return (value, asyncBody) -> {
              if (!added.compareAndSet(false, true)) {
               asyncBody.done().whenComplete(onBodyDone);
              }
              // buffer the bytes if text, keep a count if binary
              // log once the buffer is full
              consumer.consume(value, asyncBody);
            };
          }
```

manusa:
OK, so shall I merge this and we can discuss the logging approach once I provide the PoC?

shawkins:
> OK, so shall I merge this and we can discuss the logging approach once I provide the PoC?

Sure, that's fine.

PR #5037, opened 2023-04-06
feat: generic HttpLoggingInterceptor

manusa:
## Description

Relates to (should fix):
- https://github.com/eclipse/jkube/issues/1950
- https://github.com/eclipse/jkube/issues/2000

**PoC** to replace the OkHttp-specific `HttpLoggingInterceptor` with a Fabric8-specific implementation that works across all of our `HttpClient`s.

The `HttpLoggingInterceptor` implementation is following a deferred approach. This means that the messages are only logged once the server responds and the response is completely processed by the client. This allows to log the request (and its body), the response, and the response body in a neat and grouped way. Other alternatives can be considered.

The `Interceptor` interface `after` method had to be further modified (on top of changes already provided in #5025), to allow to uniquely identify the original requests. HttpClients freely copy the original request to propagate it around which prevents keeping a consistent identifier (HttpRequests are copied either to propagate them, or to start new ones)

Additional changes had to be completed in the WebSocket handling in order to allow to fully intercept and log the HTTP upgrade requests and responses. We should look into completely removing the `WebSocketResponse` wrapper class in the future.

Some questions/topics
- Do we prefer the neat and sorted deferred logging, or shall we consider immediate logging instead?
- Log format is similar to that provided by cURL, but can likely be improved
- Does this look good in general? should we keep moving in this direction?

I'll continue by implementing tests and addressing any comments once I get some feedback about this.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/HttpLoggingInterceptor.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/HttpLoggingInterceptor.java
@@ -25,6 +25,8 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
 
 public class HttpLoggingInterceptor implements Interceptor {
 
@@ -67,9 +69,13 @@ public class HttpLoggingInterceptor implements Interceptor {
   private static final class DeferredLoggingConsumer
       implements AsyncBody.Consumer<List<ByteBuffer>> {
 
+    private static final long MAX_BODY_SIZE = 2097152L; // 2MiB
+
     private final HttpLogger httpLogger;
     private final HttpRequest originalRequest;
     private final AsyncBody.Consumer<List<ByteBuffer>> originalConsumer;
+    private final AtomicBoolean logResponseBody;
+    private final AtomicLong responseBodySize;
     private final Queue<ByteBuffer> responseBody;
 
     public DeferredLoggingConsumer(HttpLogger httpLogger, HttpRequest originalRequest,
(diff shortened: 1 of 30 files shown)


kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/HttpLoggingInterceptor.java
```diff
@@ -0,0 +1,206 @@
...
+      this.cleanUp = cleanUp;
+      originalResponse = new AtomicReference<>();
+      asyncBodyDoneProcessed = new AtomicBoolean(false);
+      responseBody = new ConcurrentLinkedQueue<>();
+    }
+
+    @Override
+    public void consume(List<ByteBuffer> value, AsyncBody asyncBody) throws Exception {
+      // TODO: Should try to detect if the body is text or not? (HttpLoggingInterceptor.isPlainText)
+      value.stream().map(BufferUtil::copy).forEach(responseBody::add); // Potential leak
```
shawkins, line 117: Yes doing something like a long running log watch, an http watch, etc. would indefinitely build up here.  This will need to be bounded in some way - time, size, or both.

manusa:
Hi @shawkins,
I'd like to see if we can fit this into 6.6 to overcome some of the downstream issues with regards to the logging interceptor.
I'll try to address the issue with the body logging.
Regarding the rest of open questions, any opinions?
Do you think we can go forward with this approach and already replace the OkHttp logging interceptor? Or is there anything critical that should be added prior to removing the other interceptor and introducing this one?

shawkins:
@manusa a concern with the current pr is that it will catch exactly 1 response in the chain - depending upon the relative position of the logging interceptor.  To catch all intermediate requests, for example backwards compatibility modifications, we need more notification of after so that each can be logged.  One possible way of doing this is to expand the after calls (I realize this is different than how before is handled) - https://github.com/fabric8io/kubernetes-client/commit/4004b7b1515bb12310c836406e2e084834ea7cbb such a change would also help cut down on the state held by the logging interceptor

shawkins (approved):
LGTM

PR #5047, opened 2023-04-12
fix #5036: addressing the handling of non-connection errors

### LINKED ISSUE #5036: Allow configuring Max Frame Size in Vert.x HTTP Client

In some cases, it is necessary to transmit larger amount of data via the K8s Client. The particular use case for this is an Operator with a large CR. E.g. in the Keycloak Operator, we are seeing a lot of error messages like:
```
[io.ver.cor.net.imp.ConnectionBase] (vert.x-eventloop-thread-1) Max frame length of 65536 has been exceeded.
```
This is not happening with e.g. OKHttp Client. However, Quarkus and the Operator SDK is moving to Vert.x HTTP Client, so using Vert.x will be inevitable.

Max Frame Size needs to be configurable.

_No response_

### Additional context

_No response_

shawkins:
## Description
Addresses the additional concerns raised in #5036

This still needs to increase or allow setting the relevant limits in vertx and jetty.

cc @vietj 

## Type of change

## Checklist

### DIFF
--- a/httpclient-jetty/src/main/java/io/fabric8/kubernetes/client/jetty/JettyHttpClientBuilder.java
+++ b/httpclient-jetty/src/main/java/io/fabric8/kubernetes/client/jetty/JettyHttpClientBuilder.java
@@ -39,6 +39,8 @@ public class JettyHttpClientBuilder
     extends StandardHttpClientBuilder<JettyHttpClient, JettyHttpClientFactory, JettyHttpClientBuilder> {
 
   private static final int MAX_CONNECTIONS = Integer.MAX_VALUE;
+  // the default for etcd seems to be 3 MB, but we'll default to unlimited to have the same behavior across clients
+  private static final int MAX_WS_MESSAGE_SIZE = Integer.MAX_VALUE;
 
   public JettyHttpClientBuilder(JettyHttpClientFactory clientFactory) {
     super(clientFactory);
@@ -62,6 +64,10 @@ public class JettyHttpClientBuilder
     }
     HttpClient sharedHttpClient = new HttpClient(newTransport(sslContextFactory, preferHttp11));
     WebSocketClient sharedWebSocketClient = new WebSocketClient(new HttpClient(newTransport(sslContextFactory, preferHttp11)));
+    sharedWebSocketClient.setMaxBinaryMessageSize(MAX_WS_MESSAGE_SIZE);
+    // the api-server does not seem to fragment messages, so the frames can be very large
+    sharedWebSocketClient.setMaxFrameSize(MAX_WS_MESSAGE_SIZE);
+    sharedWebSocketClient.setMaxTextMessageSize(MAX_WS_MESSAGE_SIZE);
     sharedWebSocketClient.setIdleTimeout(Duration.ZERO);
     if (connectTimeout != null) {
       sharedHttpClient.setConnectTimeout(connectTimeout.toMillis());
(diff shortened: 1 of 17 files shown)


oscerd (approved):

shawkins:
Updated the max frame and message values for jetty and vertx using 2 MB - which roughly comes from accommodating the max size of a configmap https://kubernetes.io/docs/concepts/configuration/configmap/#motivation 

The other option is to just set the values to unlimited - which seems to be what okhttp and jdk effectively do.

vmuzikar:
@shawkins Thank you for providing the fix!

What are the implications of having unlimited frame size at the client side?

IMHO it'd make sense to align with other HTTP Clients (OkHttp) and set it to unlimited.

shawkins:
> What are the implications of having unlimited frame size at the client side?

I don't believe there is much risk, it just provides a more friendly error than an OOM should an unexpectedly large resources be retrieved.  Given that OkHttp was used for years without explicit limits and we didn't have any issues that I'm aware of, I'm okay with unlimited - but without additional input from @vietj and @manusa I thought I'd start off with at least what could be a "principled" bound.

httpclient-jetty/src/main/java/io/fabric8/kubernetes/client/jetty/JettyHttpClientBuilder.java
```diff
@@ -39,6 +39,8 @@
     extends StandardHttpClientBuilder<JettyHttpClient, JettyHttpClientFactory, JettyHttpClientBuilder> {
 
   private static final int MAX_CONNECTIONS = Integer.MAX_VALUE;
+  // the max data in a config map is 1 MiB, so pad a little beyond that
+  private static final int MAX_WS_MESSAGE_SIZE = 1 << 21;
```
sunix, line 43: 1 MiB ? or 2 ?
> Updated the max frame and message values for jetty and vertx using 2 MB - which roughly comes from accommodating the max size of a configmap https://kubernetes.io/docs/concepts/configuration/configmap/#motivation

shawkins: With some further searching using etcd as a keyword it looks like kubernetes sets the limit on the etcd server to 3 MB per request - https://github.com/kubernetes/kubernetes/issues/105863   
https://stackoverflow.com/questions/73549236/argo-and-kubernetes-request-entity-to-large-limit-is-3145728
- rather than the etcd default of 1.5 MB.  So the 1 MB limit for a configmap is some other limit that is being enforced.  After discussion I'll update this pr to be unlimited and note in the resolution the concern about protecting the client memory from attack, which in this case would imply the api server has been compromised in some way.

shawkins:
@manusa updated to umlimited, and logged the issue of configuration as a separate concern #5062

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/WatcherWebSocketListener.java
```diff
@@ -42,9 +43,13 @@ public void onOpen(final WebSocket webSocket) {
...
   @Override
-  public void onError(WebSocket webSocket, Throwable t) {
-    logger.debug("WebSocket error received", t);
-    manager.scheduleReconnect(state);
+  public void onError(WebSocket webSocket, Throwable t, boolean connectionError) {
+    if (connectionError) {
+      logger.debug("WebSocket error received", t);
+      manager.scheduleReconnect(state);
+    } else {
+      manager.close(new WatcherException("Could not process websocket message", t));
```
shawkins, line 51: The downside to the approach is that we have not fully captured what the recoverable error types are, and we'll need to update the handling further.  The alternative here is to log at an error level and still try to reconnect - we'll still potentially need to refine the handling further to avoid looping on something that is non-recoverable.

sunix (approved):

metacosm (approved):

manusa (approved):

PR #5090, opened 2023-05-02
[java-generator] - Gradle plugin

### LINKED ISSUE #4829: Gradle Plugin Task for CRD Generation

At the Moment we are using this 

```
tasks.register('createCrds') {
    doLast {
        def config = new io.fabric8.java.generator.Config()
        def runner = new io.fabric8.java.generator.FileJavaGenerator(config, new File("${projectDir}/chart/templates/Test.yaml"))
        runner.run(
            new File("${buildDir}/generated/sources/crds/main/java")
        )
    }
}
```

Like [fabric8-maven-plugin](https://github.com/fabric8io/fabric8-maven-plugin)
We want also a Plugin for Gradle to generate CRD and not using this above.

_No response_

### Additional context

This Code above leeds issue when fabric8 is updating this methodes we are using

fabiobrz:
## Description
Adding a basic implementation for a Java Generator Gradle plugin

FYI @andreaTP 

Fix #4829 

## Type of change

## Checklist


java-generator/it/src/it/plugin/gradle/simple/build.gradle
```diff
@@ -0,0 +1,31 @@
...
+// Apply a specific Java toolchain to ease working on different environments.
+java {
+    toolchain {
+        languageVersion = JavaLanguageVersion.of(11)
+    }
+}
+
+javaGen {
+    source = file('artemis-crd.yml')
+    target = file('tmp/generated/')
```
andreaTP, line 29: Question, is the `target` folder already included in the "Source folders" of Gradle?

java-generator/it/src/test/java/io/fabric8/java/generator/gradle/plugin/ITGradleRunnerExtension.java
```diff
@@ -0,0 +1,72 @@
...
+ * Simplified/updated version of
+ * https://github.com/eclipse/jkube/blob/master/gradle-plugin/it/src/test/java/org/eclipse/jkube/gradle/plugin/tests/ITGradleRunnerExtension.java
+ */
+public class ITGradleRunnerExtension implements BeforeEachCallback, AfterEachCallback {
+  private GradleRunner gradleRunner;
+
+  @Override
+  public void beforeEach(ExtensionContext context) throws Exception {
+    gradleRunner = GradleRunner.create()
+        .withGradleDistribution(new URI("https://services.gradle.org/distributions/gradle-8.1-bin.zip"))
```
andreaTP, line 39: would it be possible to bind the gradle version to a Maven property so that it will be updated by dependabot?

java-generator/gradle-plugin/src/main/java/io/fabric8/java/generator/gradle/plugin/JavaGeneratorPluginExtension.java
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+   * The Download target folder for CRDs downloaded from remote URLs
+   *
+   */
+  public abstract DirectoryProperty getDownloadTarget();
+
+  public File getDownloadTargetOrDefault() {
+    return this.getDownloadTarget().getAsFile()
+        .getOrElse(this.gradleProject.getLayout().getProjectDirectory().dir("tmp").dir("input").getAsFile());
```
andreaTP, line 64: is it "gradle idiomatic" to create a `tmp` folder at the root of the project?
I'm not sure I have seen this pattern before ...

java-generator/it/src/it/plugin/gradle/simple/expected/ActiveMQArtemisSpec.expected
```diff
@@ -0,0 +1,169 @@
+package io.amq.broker.v1beta1;
```
andreaTP, line 1: instead of checking the content of the generated files would it be possible to verify in a test that the generated sources have been compiled and are available on the classpath?

java-generator/gradle-plugin/src/main/java/io/fabric8/java/generator/gradle/plugin/JavaGeneratorPluginExtension.java
```diff
@@ -0,0 +1,128 @@
...
+  public File getTargetOrDefault() {
+    return this.getTarget().getAsFile()
+        .getOrElse(this.gradleProject.getLayout().getProjectDirectory().dir("tmp").dir("output").getAsFile());
+  }
+
+  /**
+   * Generate uppercase Enums
+   *
+   */
+  public abstract Property<Boolean> getEnumUppercase();
```
andreaTP, line 82: Instead of flattening the `Config` down here, I think it's more convenient (especially for maintenance) to re-use a private instance of Config: https://github.com/fabric8io/kubernetes-client/blob/master/java-generator/core/src/main/java/io/fabric8/java/generator/Config.java
so that we can maintain defaults etc. etc. in a single place.

andreaTP (commented):
Outstanding work @fabiobrz , thanks a ton!
Even an integration test! ❤️ 

I added a few questions/notes, especially due to my inexperience with Gradle plugins.
I think that @manusa input would be helpful here 🙂

java-generator/gradle-plugin/pom.xml
```diff
@@ -0,0 +1,122 @@
...
+      <artifactId>java-generator-core</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.codehaus.groovy</groupId>
+      <artifactId>groovy-all</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.gradle</groupId>
+      <artifactId>gradle-base-services</artifactId>
```
andreaTP, line 54: Forgot to mention, but I think that most of these dependencies should be marked as `provided`.

java-generator/it/src/it/plugin/gradle/simple/build.gradle
fabiobrz: I am not sure to get the gist here: is your questio about the `target` property key or rather about its value? The reason I am asking: do we want to changethe default _value_ or the test one to something like `build/generated/` or do we want to chenge the _key_ to something in the Gradle semantics? 
If the latter is the case we should ask ourselves whether we agree on having configuration key names to differe from the ones of the Maven plugin.
WDYT?

java-generator/it/src/test/java/io/fabric8/java/generator/gradle/plugin/ITGradleRunnerExtension.java
fabiobrz: Yep. I think this is doable by adding a `*.properties` file that binds a property value to one defined by the Maven project that's executing the tests.
BTW - I must say this would diverge from the version of the Gradle artifacts we're using, which is behind ATM (see https://github.com/fabiobrz/kubernetes-client/blob/java-gen.gradle-plugin/java-generator/pom.xml#L43) and which doesn't seemm to be actively released lately, see e.g. `gradle-core` [1].
WDYT?

[1]
https://mvnrepository.com/artifact/org.gradle/gradle-core?repo=gradle-libs-releases-local

java-generator/gradle-plugin/src/main/java/io/fabric8/java/generator/gradle/plugin/JavaGeneratorPluginExtension.java
fabiobrz: No, you're right, thanks for pointing this out. We could set this default differenlty. I'll have a look.

java-generator/gradle-plugin/pom.xml
fabiobrz: Yep, I noticed that is the case in Jkube. I'll have a look again and fix it, thanks!

java-generator/it/src/it/plugin/gradle/simple/build.gradle
andreaTP: Sorry for the confusion, I'm referring to the "default value".

java-generator/it/src/test/java/io/fabric8/java/generator/gradle/plugin/ITGradleRunnerExtension.java
andreaTP: You should be able to skip the `properties` file and directly generate from Maaven a Java source file containing the version, we usually use something like this: https://stackoverflow.com/a/36628755

Regarding the gradle versions issues, I'm not really familiar with them ... maybe @manusa ?

java-generator/gradle-plugin/src/main/java/io/fabric8/java/generator/gradle/plugin/JavaGeneratorPluginExtension.java
andreaTP: thanks 👍

java-generator/it/src/it/plugin/gradle/simple/build.gradle
fabiobrz: This should be fixed now

java-generator/gradle-plugin/src/main/java/io/fabric8/java/generator/gradle/plugin/JavaGeneratorPluginExtension.java
fabiobrz: This should be fixed now.

java-generator/it/src/it/plugin/gradle/simple/expected/ActiveMQArtemisSpec.expected
fabiobrz: This should be fixed now.

java-generator/it/src/test/java/io/fabric8/java/generator/gradle/plugin/ITGradleRunnerExtension.java
fabiobrz: Should be fixed now. Currently the Gradle binaries version is set by default to `8.1`. Can be overridden via `-Dgradle-binaries.version=...`

java-generator/gradle-plugin/src/main/java/io/fabric8/java/generator/gradle/plugin/JavaGeneratorPluginExtension.java
```diff
@@ -33,10 +31,20 @@ public abstract class JavaGeneratorPluginExtension {
   public static final String NAME = "javaGen";
   private Project gradleProject;
 
+  private Config javaGeneratorConfig = new Config(
```
andreaTP, line 34:
```suggestion
  private Config javaGeneratorConfig = Config.builder().build();
```

java-generator/it/src/test/java/io/fabric8/java/generator/gradle/plugin/SimpleIT.java
```diff
@@ -0,0 +1,119 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.*;
+
+class SimpleIT {
+
+  @RegisterExtension
+  private final ITGradleRunnerExtension gradleRunner = new ITGradleRunnerExtension();
+
+  static final List<File> TESTED_CRDS = Stream.of(
```
andreaTP, line 38: In other tests we are using the `junit` `TempDir`, e.g.:
https://github.com/fabric8io/kubernetes-client/blob/fe579f0eab6ece572cee8a184f0614864a8c95a6/java-generator/core/src/test/java/io/fabric8/java/generator/CompilationTest.java#L45-L46

would it be possible to apply the same pattern?

andreaTP: applies at line 45, sorry for the confusion.

java-generator/gradle-plugin/src/main/java/io/fabric8/java/generator/gradle/plugin/JavaGeneratorPluginExtension.java
```diff
@@ -0,0 +1,156 @@
...
+  /**
+   * The Download target folder for CRDs downloaded from remote URLs
+   *
+   */
+  public abstract DirectoryProperty getDownloadTarget();
+
+  public File getDownloadTargetOrDefault() {
+    return this.getDownloadTarget().getAsFile()
+        .getOrElse(this.gradleProject.getLayout().getProjectDirectory()
+            .dir("src")
```
andreaTP, line 76: Usually, we download the CRD to a "temp" or "target" dir instead of a "managed source folder", something like:
`build/crds` wdyt?

java-generator/it/src/it/plugin/gradle/simple/build.gradle
```diff
@@ -0,0 +1,49 @@
...
+// Apply a specific Java toolchain to ease working on different environments.
+java {
+    toolchain {
+        languageVersion = JavaLanguageVersion.of(11)
+    }
+}
+
+javaGen {
+    urls = [
+      "https://raw.githubusercontent.com/artemiscloud/activemq-artemis-operator/v1.0.11/deploy/crds/broker_activemqartemis_crd.yaml",
```
andreaTP, line 43: So far, we have attempted to keep the build and test "local", would it be possible to commit one CRD and test against it?

java-generator/it/src/main/java-templates/io/fabric8/java/generator/testing/GradleProperties.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.fabric8.java.generator.testing;
+
+public final class GradleProperties {
```
andreaTP, line 18: 👍

andreaTP (changes requested):
A last round of minor comments and is good to go! Thanks again @fabiobrz !

java-generator/it/src/test/java/io/fabric8/java/generator/gradle/plugin/ITGradleRunnerExtension.java
andreaTP: perfect 👍

java-generator/it/src/test/java/io/fabric8/java/generator/gradle/plugin/SimpleIT.java
fabiobrz: I am wondering whether this is the right use case for `@TempDir`. Maybe I am missing some pieces but according to the docs its lifecycle would be managed by JUnit. Which clashes with the fact that those directories are instead managed by the Gradle build. `resources` is an exception, but I'll use it to store the CRD locally.

java-generator/gradle-plugin/src/main/java/io/fabric8/java/generator/gradle/plugin/JavaGeneratorPluginExtension.java
fabiobrz: Fixed, thanks.

java-generator/gradle-plugin/src/main/java/io/fabric8/java/generator/gradle/plugin/JavaGeneratorPluginExtension.java
fabiobrz: `build` would be in the Gradle lifecycle managment scope. If we want to use a default here - then either we live with a Java project structure and let the current structure be or we use something different, e.g.: `tmp/crds`, although this would mean for us to ger back to my initial implementation, basically.
TBH, we're generating Java sources, which to my eyes makes the current implementation legit, rather than the proposed one. I can be proved wrong, thoug, just let me know what you think.

java-generator/it/src/test/java/io/fabric8/java/generator/gradle/plugin/SimpleIT.java
andreaTP: Oh, if those are managed by Gradle it's fine, missed it 👍

java-generator/gradle-plugin/src/main/java/io/fabric8/java/generator/gradle/plugin/JavaGeneratorPluginExtension.java
fabiobrz: Yep, done!

java-generator/gradle-plugin/src/main/java/io/fabric8/java/generator/gradle/plugin/JavaGeneratorPluginExtension.java
fabiobrz: Now I got what you mean, i.e. putput the downloaded files to the build folder, so calling `clean` would... clean :slightly_smiling_face: 
Done, thanks!

java-generator/it/src/it/plugin/gradle/simple/build.gradle
fabiobrz: Yep, done!

fabiobrz:
Hi @andreaTP - I added one unit test, at least for the plugin class itself, and some documentation. I hope I addressed all the concerns you pointed out. Marking the PR as ready.

Feel free to let me know in case you have additional thoughts.

andreaTP (approved):
Amazing work LGTM, thanks a lot @fabiobrz !

manusa (approved):
LGTM, thx!

manusa (approved):
LGTM, thx!

PR #5138, opened 2023-05-16
fix #5135: adding an edit method to each model object

### LINKED ISSUE #5135: Possible Sundrio improvements
https://github.com/sundrio/sundrio/pull/368 will cut down on the model jar sizes by around 20% and will also decrease the runtime footprint. 

Other possible opportunities to reduce the size of the generated models:
- suppress the generation of nested builders in places where we're using HasMetadata to represent raw.  The available types are usually incomplete as the module and buildable references don't contain all of the types (for example Deployment is missing for Template objects), and the content could actually be free form.
- 6 methods are generated which contain logic for determining the type / builder per nested collection field.  It seems like this could be consolidated to a single method for determining the builder - and/or as discussed with @metacosm we could make all of our model objects Editable and omit the instanceof check for types and let the general builderOf logic catch it.
- removeMatchingFromXXX are not necessary as we're already dealing with a VisitableBuilder

@manusa @metacosm @iocanel - do any of these seem worthwhile?

shawkins:
## Description

In scope of #5135

This will enable users / us to do something like pod.edit().build() as a clone operation.

I've also opened https://github.com/sundrio/sundrio/pull/376 so that we can align to the common BaseFluent.builderOf method for obtaining a builder.  Or in the interim we may need to keep our similar logic and just update it some more and make it static.  From there we'll use that in ResourceHandlerImpl and in Serialization (for cloning).

## Type of change

## Checklist


shawkins:
I'll see if I can resolve then NPE this upstream in sundrio before moving forward with this hack based approach.

shawkins:
> I'll see if I can resolve then NPE this upstream in sundrio before moving forward with this hack based approach.

Opened https://github.com/sundrio/sundrio/issues/396 but could not see how to immediately resolve it.  We'll probably need to use this workaround if we want to make the pojos editable any time soon.

iocanel:
Some thoughts:

####  On `BaseFluent.builderOf` 
AFAIR this uses `Class.forName` and its something I would really like to avoid. Ideally, this should move down one level into the concrete fluent (where it can be re-implemented without reflection).

#### kubernetes model builder package
One way of solving the chicken and egg thing with jsonschema2pojo would be to generate the package in separate jar, that jsonschema2pojo can depend on. The rest of the model can depend on that jar and not regenerate the model.

A completely different approach would be to add the method using a `lombok` style byte code manipulation (however this would have to be done on sundrio).

#### edit() vs toBuilder()
It doesn't really matter which one will be used, but ...
It would be really nice if whatever method is used is expressed using an interface. It would make things easier when parsing files, and apply visitors etc.

shawkins:
> AFAIR this uses Class.forName and its something I would really like to avoid. Ideally, this should move down one level into the concrete fluent (where it can be re-implemented without reflection).

The first check in builderOf is for if the item is Editable.  

> One way of solving the chicken and egg thing with jsonschema2pojo would be to generate the package in separate jar, that jsonschema2pojo can depend on. The rest of the model can depend on that jar and not regenerate the model.

That is in this pr - a dummy class is used to trigger the creation of the fabric8 builder package in a different class.

> It would be really nice if whatever method is used is expressed using an interface. It would make things easier when parsing files, and apply visitors etc.

If you are open to it upstream, then toBuilder could be added as a default method.

From a end user perspective it's still a little awkward as an interface (we talked about this a little on a sundrio issue) - as there's currently no additional bounds, you'll need to at least cast to VisitableBuilder.

shawkins:
@metacosm @manusa marking this as ready.  A follow-up will be to add the toBuilder method to the Editable interface.

metacosm (approved):
Looks OK to me.

manusa (approved):

Issue #4624, opened 2022-11-30
How to use only http in watch?


liziwl:
I have some problem using websocket, however using pure http is working well. 

Are there any config to set watcher using only http protocol?

rohanKanojia:
KubernetesClient seems to use http watch if it fails to start websocket initially. What behavior are you observing?

liziwl:
I use official client works fine.

```
io.fabric8.kubernetes.client.KubernetesClientException: An error has occurred.

	at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:103)
	at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:97)
	at io.fabric8.kubernetes.client.dsl.internal.WatchConnectionManager.lambda$run$2(WatchConnectionManager.java:133)
	at java.util.concurrent.CompletableFuture.uniHandle(CompletableFuture.java:836)
	at java.util.concurrent.CompletableFuture$UniHandle.tryFire(CompletableFuture.java:811)
	at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:488)
	at java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:1990)
	at io.fabric8.kubernetes.client.okhttp.OkHttpWebSocketImpl$BuilderImpl$1.onFailure(OkHttpWebSocketImpl.java:71)
	at okhttp3.internal.ws.RealWebSocket.failWebSocket(RealWebSocket.kt:592)
	at okhttp3.internal.ws.RealWebSocket$connect$1.onFailure(RealWebSocket.kt:202)
	at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:525)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:750)
	Suppressed: java.lang.Throwable: waiting here
		at io.fabric8.kubernetes.client.utils.Utils.waitUntilReady(Utils.java:168)
		at io.fabric8.kubernetes.client.utils.Utils.waitUntilReadyOrFail(Utils.java:179)
		at io.fabric8.kubernetes.client.dsl.internal.WatchConnectionManager.waitUntilReady(WatchConnectionManager.java:96)
[...]

PR #5153, opened 2023-05-19
fix #5152: expanding the error detection

shawkins:
## Description
As seen on #5152 some jdk client implementations do call onError with connection related errors (my local one would call onClose instead).  The proposed change is to further differentiate what are protocol errors, so that only select exceptions are treated as terminal.  Note that logic like this would be applicable in the http scenario as well, but no changes were made there.

Another possible issue I noticed in doing these changes is the behavior in WatchConnectionManager.start - if a websocket has already been established, then a subsequent connection attempt can short-circuit on failure before scheduling another attempt if a 200 or 503 occurs.  That short-circuit should only be for the first attempt, where we fall-back to an http watch instead.  It seems like there is probably some improvement to be had with the http case - also #4624 was never addressed.

However comparing to the go client, it will only retry a watch when the connection is refused or if there have been too many requests: https://sourcegraph.com/github.com/kubernetes/client-go@2a5f18df73b70cb85c26a3785b06162f3d513cf5/-/blob/tools/cache/reflector.go?L418 - so it seems like they would have a similar issue if 200 or 503 where returned.

## Type of change

## Checklist


shawkins:
@manusa @scholzj this refines the changes in #5047 to address https://github.com/fabric8io/kubernetes-client/pull/5047#discussion_r1171443013 which was observed in #5152.  We'll specifically look for a protocolexception as the only case to assume we should proactively terminate - with the expectation that the httpclient will have normalized to that.  For everything else there will be some logging based upon the inferred severity and number of times we've seen that termination since we've made progress with the watch.

The calls to scheduleReconnect have been consolidated to ensure consistent handling of exceptions - the http watch exceptions have not been fully normalized, but at worst this will result in additional info logs or looping on unresolvable protocol errors.

scholzj:
Thanks @shawkins

shawkins:
@manusa at least for the jdk client this causes pretty bad behavior with watches / informers, so another 6.6 may be in order.  For the other clients it shouldn't be as impactful.

shawkins:
@manusa @scholzj After updating to master this was updated to harden the websocket implementation contracts - returning a boolean for send and sendClose is a little problematic.  A non-terminal error writing would mean that any additional writes would be invalid.  To ensure the expected behavior and let us know if something is wrong, callbacks will log and even terminate the websocket if possible.  Similar logic / logging was added / made more robust for sendClose - with a delayed termination if a remote close has not been received.  

There is currently no method exposed for the vertx websocket to force a termination, so if anyone hits one of the error logs, we'll need to have something added or that specific error fixed.

scholzj (approved):
Thanks a lot for the effort you put into this @shawkins!

manusa (approved):

rohanKanojia (approved):

Issue #4606, opened 2022-11-27
Patch with empty collections


shawkins:
### Is your enhancement related to a problem? Please describe

For non-json patches - json merge, strategic merge - if an object has an empty collection intended to remove items, that will simply be omitted from the serialization and no change will be made.

Server side apply because of the tracking of the managed fields and different semantics, will remove a collection as expected.

### Describe the solution you'd like

The existing model logic and patch methods make this difficult to deal with.  The fields are annotated to omit if empty.  The builder logic will always set a default empty list or map.

To make this "work" the implementation logic for the non-json patches would also have to use a base object to determine what is being modified, then the serialization logic adapted to account for explicitly empty collections - this would likely involve a conversion to the map representation and explicitly setting the empty collections, they would then be emitted with the serialization.

One option is for edit / accept to have a PatchContext parameter.  Those operations have a clear base.  Unfortunately with all of the other methods because the json merge / strategic merge could refer to a partial realization of the object it's not clear what the base is.

Another option is to just recommend that users move on to using server side apply.

manusa:
The problem with most solutions is that they are not very clear and user-friendly. We'll also offering yet another configuration option to the already flexible patch operation.
The proposed solution also requires having an original object, which in the end would imply some of the replace/update requirements which might not make it ideal for patches.

IMO we should wait to see if this is something that cannot be workedaround and requested by the community. I'm not sure if @csviri still needs this ¿?

csviri:
@manusa no I don't need it anymore.

PR #5222, opened 2023-06-08
fix #5214: explicitly omitting nulls by default

shawkins:
## Description

Resolves the discussion on #5214 - we'll treat the 6.6 / 6.7 behavior as a regression and switch to explicitly omitting nulls by default.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### 6.8-SNAPSHOT
 
 #### Bugs
+* Fix #5214: null values are omitted by default, which means custom resources by in large won't need JsonIncludes annotations.  The only time that is required is when null is to be preserved via @JsonInclude(value = Include.ALWAYS) on the relevant field.
 * Fix #5218: No export for `io.fabric8.tekton.triggers.internal.knative.pkg.apis.duck.v1beta1` in tekton v1beta1 triggers model
 * Fix #5224: Ensuring jetty sets the User-Agent header
 * Fix #4225: Enum fields written in generated crd yaml
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesSerialization.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesSerialization.java
@@ -16,6 +16,8 @@
 
 package io.fabric8.kubernetes.client.utils;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.DeserializationConfig;
@@ -89,6 +91,8 @@ public class KubernetesSerialization {
   protected void configureMapper(ObjectMapper mapper) {
     mapper.registerModules(new JavaTimeModule(), unmatchedFieldTypeModule);
     mapper.disable(DeserializationFeature.FAIL_ON_INVALID_SUBTYPE);
+    // omit null fields, but keep null map values
+    mapper.setDefaultPropertyInclusion(JsonInclude.Value.construct(Include.NON_NULL, Include.ALWAYS));
     HandlerInstantiator instanciator = mapper.getDeserializationConfig().getHandlerInstantiator();
     mapper.setConfig(mapper.getDeserializationConfig().with(new HandlerInstantiator() {
 
--- a/doc/CHEATSHEET.md
+++ b/doc/CHEATSHEET.md
@@ -1859,6 +1859,10 @@ import io.fabric8.kubernetes.model.annotation.Version;
 public class CronTab extends CustomResource<CronTabSpec, CronTabStatus> implements Namespaced {
 }
 ```
+
+**Note:** Null values in your custom resource will be omitted by default by the client or when directly using KubernetesSerialization, or the static Serialization methods.  If you have a situation where a null value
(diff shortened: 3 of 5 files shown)


manusa:
Unrelated to this change (I think), but worth noting and related to the general functionality, especially for patch scenarios.

Our Kubernetes Resource objects are currently initialized with empty arrays (I think we did add this behavior a while ago). This poses a problem when using builder to generate objects that will be sent to the API server in a JSON_MERGE (`application/merge-patch+json`) PATCH request. If the remote object has an entry in the array, since the sent patch JSON contains the field with an empty array, the remote field will be overwritten and cleared.

shawkins:
> Unrelated to this change (I think), but worth noting and related to the general functionality, especially for patch scenarios.
> 
> Our Kubernetes Resource objects are currently initialized with empty arrays (I think we did add this behavior a while ago). This poses a problem when using builder to generate objects that will be sent to the API server in a JSON_MERGE (`application/merge-patch+json`) PATCH request. If the remote object has an entry in the array, since the sent patch JSON contains the field with an empty array, the remote field will be overwritten and cleared.

@manusa for the built-in resources where we initialize empty collections there is also the annotation: @JsonInclude(JsonInclude.Include.NON_EMPTY) - so the expectation moving forward is that neither a null nor an empty collection will be included in the serialization by default.  So I don't think the situation you are describing currently occurs with the built-in resources.  However the opposite problem exists - this was discussed on #4606 but wasn't addressed.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### 6.8-SNAPSHOT
```
andreaTP, line 3: Given the discussion, this fix should probably be included in a 6.7 patch release.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesSerialization.java
```diff
@@ -89,6 +91,8 @@ public KubernetesSerialization(ObjectMapper mapper, boolean searchClassloaders)
   protected void configureMapper(ObjectMapper mapper) {
     mapper.registerModules(new JavaTimeModule(), unmatchedFieldTypeModule);
     mapper.disable(DeserializationFeature.FAIL_ON_INVALID_SUBTYPE);
+    // omit null fields, but keep null map values
+    mapper.setDefaultPropertyInclusion(JsonInclude.Value.construct(Include.NON_NULL, Include.ALWAYS));
```
andreaTP, line 95: Note for the future, is it possible to let the user customize those options completely?

andreaTP (approved):
Super late at this party, but thanks a lot for the changes @shawkins !

I agree this is the right direction, a couple of minor notes.

manusa:
> So I don't think the situation you are describing currently occurs with the built-in resources

It does occur (for sure at least on 6.6, but I guess this is the same on master and earlier versions too) for `DeploymentConfig`, for example.

I think this has to do with the `javaOmitEmpty` JSON Schema configuration that's computed using the `getPropertyDescriptor` function:
 
https://github.com/fabric8io/kubernetes-client/blob/60230428cb90b4e632913e85925536f9537b1280/kubernetes-model-generator/pkg/schemagen/generate.go#L313

And the fact that builders are initialized with empty collections.

The following link contains a concrete example of why this is a problem:
https://github.com/manusa/yakd/blob/6d942632ee3ad5e738ba31c0b2fe45bac99f3c99/src/main/java/com/marcnuri/yakd/quickstarts/dashboard/deploymentconfigs/DeploymentConfigService.java#L90-L97

Basically when sending the patch, the JSON sent to the server contains something like:
```
{"spec": {"replicas": 1, "triggers": []}} 
```

Since this is a JSON_MERGE patch, the `triggers` are wiped out from the server resource.

shawkins:
> It does occur (for sure at least on 6.6, but I guess this is the same on master and earlier versions too) for DeploymentConfig, for example.

So yes that would be a bug with the omitEmpty logic.  I recall specifically adding a workaround for all maps in Fabric8DefaultRule but wasn't aware we had an inconsistency with some lists (others lists of either simple or complex types like objectmeta.finalizers or ownerreferences do end up annotated appropriately).  I'd be fine with another override in Fabric8DefaultRule to act as a catch all.

> And the fact that builders are initialized with empty collections.

Builders make quite a few assumptions about null / empty - the initialization, no null map values, and no null complex type collection values, and probably more...  At least the builder behavior to default to an empty collection was part of the motivation of the change to initialize all of pojos with empty collections.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesSerialization.java
shawkins: Only in an advanced scenario usage scenario.  The configureMapper method is protected and may be overriden if someone wants a create a custom KubernetesSerialization subclass.  However I wanted to work through the quarkus integration and especially a plan to minimize the number of classes referenced by the kubernetesdeserializer - I'm thinking we need provide a way to lazily load - before open things up even more.

andreaTP: Great idea!

shawkins:
@manusa the rationale for the inconsistency in the deploymentconfig spec triggers is that they are not marked as omitempty upstream: https://github.com/openshift/api/blob/master/apps/v1/types.go#L63

As you point out our builders are opionated about treating an empty and null collection the same - and we're already initializing the trigger list to empty, so I would be fine to just force the annotation.  That can either be done in the generator or via the default rule.  Do you want that as part of this pr?

PR #5231, opened 2023-06-11
fix #5215: refining docs and serverSideApply support

### LINKED ISSUE #5215: Usage of serverSideApply as a replacement for createOrReplace
Relates to #5176

Doing a conversion of createOrReplace to serverSideApply with Keycloak makes me think that adding an unlock method, which we have talked about in the past, is a good idea.  That would make the full replacement:

resource.unlock().forceConflicts().serverSideApply()

A more nuanced issue is that kubernetes will sometimes update the managedFields time even if there are no other changes in the resource, causing a new revision that would not be created via just an update.  Logic that is sensitive to new revisions, such as operators, would need filtering to address this. 

@csviri - is this something that has already been seen with the operator sdk?

I can't find any information on how to avoid this, in fact it looks like a kubernetes bug based upon the documentation https://github.com/kubernetes/kubernetes/blob/6f116dc4d5ce996aac4a9ad48a16e72b059f9439/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1253 

cc @manusa

shawkins:
## Description
Addresses the first option from https://github.com/fabric8io/kubernetes-client/issues/5215#issuecomment-1579372933

@manusa @csviri let's discuss what javadoc updates should accompany this, and what if any additional logic changes.

I'm assuming for now that we won't be adding an unlock function - the application would be for situations like:
- resource.unlock().edit/accept - but you have access to the object in the lamda, so you can modify it yourself
- some signatures of resource.patch - in particular resource.unlock().patch() - where you must otherwise be congnisant of not modifying the resourceVersion coming from an informer
- resource.unlock().update() - where you must otherwise be congnisant of not modifying the resourceVersion coming from an informer

Or for other operations what we wish to be locked by default - as noted in the faq there are currently some that fall into that category, but that may not be intentional.

## Type of change

## Checklist


csviri (approved):
LGTM

shawkins:
Updated to include the faq link in the javadocs.

manusa (approved):

doc/FAQ.md
```diff
@@ -119,3 +119,46 @@ we should provide it like this:
...
+
+- Delete is not locked by default, you may use the applicable lockResourceVersion method if you want the delete to apply only to a specific resourceVersion
+
+- Specialized mutative operations are generally unlocked - this includes scale, rolling operations (resume, pause, restart), and others.  The rationale is that you want the narrow operation to succeed even if there has been an intervening change.  If you encounter a situation where you require these operations be locked, please raise an issue so that we can see about making the locking function applicable.
+
+- A handful of additional operations, such as undo, updateImage, and others are currently locked by default.  This may not be intentional - under the covers this is apply a json patch; older versions of json patching that created the resource diff were unlocked by default.  If you encounter an exception due to a concurrent modification performing an operation that seems like it should ignore that possibility by default please raise an issue.
```
metacosm, line 148: Late to the party but this paragraph doesn't really parse for me. The first sentence says that these operations are locked yet the subsequent sentences seem to imply that this is because we use json patch apply, which was unlocked by default. I think we should clarify this.

doc/FAQ.md
```diff
@@ -119,3 +119,46 @@ we should provide it like this:
 ```java
 NO_PROXY: localhost,127.0.0.1,.google.com,.github.com
 ```
+
+### Optimistic Locking Behavior
+
+Unfortunately it's a little complicated as it depends on what operation you are doing - we'll work towards ensuring the Javadocs are as informative as possible.  Here is quick overview:
+
+- Basic mutative operations such as update and all variations of patch (patch, edit, accept, serverSideApply) that operate on a given item - are all locked to the resourceVersion on that item.  If you don't want this behavior then set the resourceVersion to null:
```
metacosm, line 127: Maybe a sentence or a link to a reference on the locking behavior in Kubernetes would help?

doc/FAQ.md
```diff
@@ -119,3 +119,46 @@ we should provide it like this:
...
+
+createOrReplace was introduced as an alternative to the kubectl apply operation.  Over the years there were quite a few issues highlighting where the behavior was different, and there were only limited workarounds and improvements offered.  Given the additional complexity of matching the kubectl client side apply behavior, that was never offered as an option.  Now that there is first class support for serverSideApply it can be use, but it comes with a couple of caveats:
+
+- you will want to use forceConficts - resource.forceConflicts().serverSideApply() - this is especially true when acting as a controller updating a resource that [manages](https://kubernetes.io/docs/reference/using-api/server-side-apply/#using-server-side-apply-in-a-controller)
+
+- for some resource types serverSideApply may cause vacuous revisions - see https://github.com/kubernetes/kubernetes/issues/118519 - if you are being informed of modifications on those resources you must either filter those out, don't perform the serverSideApply, or use the [java-operator-sdk](https://github.com/java-operator-sdk/java-operator-sdk) that should handle that possibility already.
```
metacosm, line 158: ❤️

doc/FAQ.md
shawkins: Old behavior - they were implicitly unlocked
New behavior - they are locked, but that may not be intentional

doc/FAQ.md
shawkins: Sure, there will be another pr to more throughly cover server side apply, that can be part of it.

PR #5246, opened 2023-06-15
fix #5223: using a separate annotation to control cycle expansion

shawkins:
## Description
Here's something similar to the previous pr, but with a completely separate annotation / logic - if that helps make the feature seem more palatable.  The replacement is happening at a higher-level (class rather than property), which seems to simplify the mechanics. cc @metacosm 

Now that the config swaps have been removed keycloak usage of schemaswap looks like:
```
@SchemaSwap(originalType = GroupRepresentation.class, fieldName = "subGroups", targetType = org.keycloak.representations.overrides.NoSubGroupsGroupRepresentationList.class)
@SchemaSwap(originalType = RealmRepresentation.class, fieldName = "components", targetType = org.keycloak.representations.overrides.ComponentExportRepresentationMap.class)
@SchemaSwap(originalType = ComponentExportRepresentation.class, fieldName = "subComponents", targetType = org.keycloak.representations.overrides.NoSubcomponentsComponentExportRepresentationMap.class)
@SchemaSwap(originalType = ScopeRepresentation.class, fieldName = "policies")
@SchemaSwap(originalType = ScopeRepresentation.class, fieldName = "resources")
public class KeycloakRealmImport extends CustomResource<KeycloakRealmImportSpec, KeycloakRealmImportStatus> implements Namespaced {

}
```
Updating to use this pr looks like:
```
@SchemaExpand(originalType = ScopeRepresentation.class, terminalType = AnyType.class, depth = 3)
@SchemaExpand(originalType = GroupRepresentation.class, terminalType = AnyType.class, depth = 3)
@SchemaExpand(originalType = ComponentExportRepresentation.class, terminalType = AnyType.class, depth = 3)
public class KeycloakRealmImport extends CustomResource<KeycloakRealmImportSpec, KeycloakRealmImportStatus> implements Namespaced {

}
```
One up side is that since SchemaSwap is no longer used, there's a potential to deprecate it - as that seems undesirable to @andreaTP to maintain in its current form.

Obviously the crd size, even for a depth of 3, is greatly increased.  I don't think a depth of 10 is even possible with the default kubernetes resource size limit.

One thing that I noticed that if you use void as the replacement there are a couple of behaviors depending on where the class is used.  If it's a field or a collection, then it will be omitted.  If it's a map value however, it will instead become object - it might be better to omit the map instead.

## Type of change

## Checklist

### DIFF
--- a/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
+++ b/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
@@ -17,6 +17,7 @@ package io.fabric8.crd.generator;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import io.fabric8.crd.generator.annotation.SchemaExpand;
 import io.fabric8.crd.generator.annotation.SchemaSwap;
 import io.fabric8.crd.generator.utils.Types;
 import io.fabric8.kubernetes.api.model.Duration;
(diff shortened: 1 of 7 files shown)


andreaTP:
Hi @shawkins , a couple of questions to better understand the provided example:
- `RealmRepresentation` -> is it omitted by chance or intentionally?
- `ScopeRepresentation` -> here there are two fields that are causing the cycle to happen, are they going to be terminated at a depth of 3 in the current example? If I understand correctly the desired outcome should be produced by something like:

```
@SchemaExpand(originalType = ScopeRepresentation.class)
```
as both `policies` and `resources` can be omitted.

> One thing that I noticed that if you use void as the replacement there are a couple of behaviors depending on where the class is used. If it's a field or a collection, then it will be omitted. If it's a map value however, it will instead become object - it might be better to omit the map instead.

If this is reproducible it should be filed as a bug, I agree that your interpretation is the expected result.

shawkins:
To match the existing schema more closely it would use:

```
@SchemaExpand(originalType = ScopeRepresentation.class, depth = 1)
```

That will allow the ScopeReprentation to appear and terminate any child ScopeRepresentation as void / omit.

If the exact representation where the given fields are omitted it would require the annotation:
```
@SchemaExpand(originalType = ScopeRepresentation.class, terminalType = ScopeRepresentationOverride.class, depth = 0)
```
where ScopeRepresentationOverride omits the fields.

shawkins:
Closing in favor of #5234

Issue #5255, opened 2023-06-15
Omit empty string values by default


shawkins:
It was observed with #5212 that empty strings may cause server side apply to create unnecessary revisions.  This is likely due to the api server being only tested aginst go clients that would not serialize an effectively null string.  We should update our KubernetesSerialization to omit empty strings to match this effective go expectation.

shawkins:
On second thought it's too risky to do this: https://github.com/fabric8io/kubernetes-client/pull/5256#issuecomment-1593755077

PR #5256, opened 2023-06-15
draft of omitting null / empty by default

shawkins:
## Description
Addresses #5255 It turns out that the "easiest" fix for omitting empty strings would also address any remaining collections that lack a NON_EMPTY jsoninclude.

This approach would be to have the jsonschema2pojo delegate the authority over the include to the KubernetesSerializer by setting USE_DEFAULTS.  Then the KubernetesSerializer will specify NON_EMPTY.

Pros:
- gives us more control in the furture over serialization from the serializer, rather than having it baked into the pojos.  This means it could also be made a top level option for the KubernetesSerializer if desired.
- also covers the collection case, but of course users with custom classes can annotate their collection fields with JsonInclude NON_NULL if they want to preserve empty collections (keeping in mind that the builders will initialize to empty).
- we can go further and remove the existing JsonIncludes we are adding to the fields

Cons:
- will require an update to every generated model class and will also mean that serializing the values outside of using the KubernetesSerializer will change (I'm not sure how explicit of a goal that is).
- any users relying on the current built-in empty collection (which is inconsistent, just up to whether the go class is marked as omitempty - with the same caveat about builders) and they won't have an option to override with their own jsoninclude.

@manusa @metacosm 

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesSerialization.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesSerialization.java
@@ -91,8 +91,8 @@ public class KubernetesSerialization {
   protected void configureMapper(ObjectMapper mapper) {
     mapper.registerModules(new JavaTimeModule(), unmatchedFieldTypeModule);
     mapper.disable(DeserializationFeature.FAIL_ON_INVALID_SUBTYPE);
-    // omit null fields, but keep null map values
-    mapper.setDefaultPropertyInclusion(JsonInclude.Value.construct(Include.NON_NULL, Include.ALWAYS));
+    // omit null / empty fields, but keep null map values for patches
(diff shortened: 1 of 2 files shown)


shawkins:
Following forward with this change fails almost immediately as the model tests are writen to directly use objectmappers, rather than serialization (which isn't available at that level) - so that means all of the null expectations are then off.  The chances that users are also directly serializing our pojos is probably high as well.

If I instead change the pom to set the default as NON_EMPTY things will get much further, but we'll still eventually hit tests directly using objectmappers with different expectations.  It seems that kubernetes will send an empty string in the json responses for non-nullable fields that aren't marked as omitempty, the best example is from any list metadata:

```
    "metadata": {
        "resourceVersion": "",
        "selfLink": ""
    }
```

We have tests expecting that round trip exactly as is.  There also have been issues reported kube treating missing properties differently than those populated by an empty string. 

So I'm thinking that we probably don't want to move forward with this broad of a change.

PR #5265, opened 2023-06-19
improve: public access of object mapper in KubernetesSerialization

csviri:
This is needed for cases when the object mapper needs to be reused to explicitly (de)serialize objects.

Particularly needed for in Java Operator SDK, where this way the `ObjectMapper` can be read from the client and used further withing the framework, for (de)serialize or transfer resources for further usage.

### DIFF
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesSerialization.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesSerialization.java
@@ -380,7 +380,7 @@ public class KubernetesSerialization {
     return unmatchedFieldTypeModule;
   }
 
-  ObjectMapper getMapper() {
+  public ObjectMapper getMapper() {
     return mapper;
   }
 


oscerd (approved):

shawkins:
Ideally whatever provided the mapper to KubernetesSerialization would also provide it to the JOSDK, rather than obtaining it from KubernetesSerialization.  The con of doing it this way is that with direct access to the mapper comes great responsbility and it effectively opens up our public serialization api to the entirity of what objectmapper supports.  At the very least within the fabric8 codebase having access to the serialization mapper meant that over time we had quite a few different ways of doing the exact same serdes logic.

Or it could just be wishful thinking that we could hide this...

metacosm:
> > Ideally whatever provided the mapper to `KubernetesSerialization`...
> 
> I've been trying to convince the JOSDK folks for a few hours now with no luck 😅 Maybe you can find more compelling arguments against this.
> 
> My vote is strongly against it but I'm not going to oppose it. I'd really like to consider all of the alternate options.

It just doesn't work. We need the augmented mapper, not the plain one you pass to `KubernetesSerialization` when you create it.

manusa:
> We need the augmented mapper, not the plain one you pass to KubernetesSerialization when you create it.

If by augmented you mean the mapper that the client fine-tunes for its internal purposes, it's the same mapper instance.

csviri:
An alternative would be that in JOSDK `ConfigurationService` would provide the object mapper and set if for the client. (The users will just provide the KubernetesClientBuilder as argument, not the client. That it little hackish too, but a working compromise.

metacosm:
> > We need the augmented mapper, not the plain one you pass to KubernetesSerialization when you create it.
> 
> If by augmented you mean the mapper that the client fine-tunes for its internal purposes, it's the same mapper instance (you pass in).

If this can be guaranteed, that would probably work.

manusa:
> If this can be guaranteed, that would probably work.

:) I think I've been saying this ever since the Quarkus PR (https://github.com/quarkusio/quarkus/pull/33767). I'm obviously not very good at explaining things :sweat_smile: 

https://github.com/manusa/kubernetes-client/blob/4dba8cf359f2ca9f70da30cd497cc38bc555c0cb/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesSerialization.java#L85-L145

This is the Bring Your Own ObjectMapper (BYOO) pattern. In one of the initial iterations of the Quarkus PR this was the solution offered to be able to customize the mapper either before or after passing it in. The PR later evolved to support both patterns (BYOO) and the customization pattern.

TL;DR, yes that mapper is the one that the client will use. Our customizations, however, might not be compatible with any other stuff that the non-naked mapper brings in, but that would be a problem in any case.

metacosm:
This is even more important in a Quarkus setting where the client would be most likely automatically injected and ensuring (via injection) that the ObjectMapper that is propagated is the correct one seems rather tricky/awkward, since Quarkus uses its own mapper as well.

A compromise would be to make the `getMapper` method protected so that downstream projects could still access it if needed by subclassing `KubernetesDeserialization` for their own needs and not make the method too visible from the general user-facing API?

metacosm:
Another alternative would be to provide more methods on `KubernetesSerialization` to encapsulate the functionality of the `ObjectMapper` but then that's code that needs to be maintain and would eventually grow as people want to access more and more mapper functionality. I initially thought we could get by by using `convertValue` but we also need `readValue` (the version that deserialises a sub-tree identified by its root key).

shawkins:
> A compromise would be to make the getMapper method protected so that downstream projects could still access it if needed by subclassing KubernetesDeserialization for their own needs and not make the method too visible from the general user-facing API?

We can do that or change it into an interface and make the impl have a public method such that you could cast to the impl for access.  But let's explore some of the other options more.

> Another alternative would be to provide more methods on KubernetesSerialization to encapsulate the functionality of the ObjectMapper but then that's code that needs to be maintained and would eventually grow as people want to access more and more mapper functionality. I initially thought we could get by by using convertValue but we also need readValue (the version that deserialises a sub-tree identified by its root key).

It seems fine to change / add more methods.  It sounds like you just need to change one method to varargs and add another.

From a chat:

And another thought is that you really don't even need the mapper from KubernetesSerialization to handle this logic - the only thing that comes into play is mapper.setDefaultPropertyInclusion(JsonInclude.Value.construct(Include.NON_NULL, Include.ALWAYS)) - but the JOSDK logic would be tolerant to explicit nulls because they would show up in the same place on each side of the final comparison.

The final thing brought up was adding the change detection logic into the fabric8 client instead.  I'm a little squimish that this may not age well - but as long as we can reasonably guarentee that it can't miss a needed update, then it could have some use beyond the josdk.  That would need some clear expectations about when and why users would want to call - in particluar that the desired state is the full state, not partial, and some of the exception handling needs explained or should just result in returing true as you'd want to force appling the operation if you couldn't figure out if there had been a change.

manusa:
> This is even more important in a Quarkus setting

Quarkus problem is already **solved**. https://github.com/quarkusio/quarkus/pull/33767 added support to either provide your own ObjectMapper to be used **exclusively** for the client or to provide an ObjectMapper customizer to customize the one that will be created for you. There's even a specific `@KubernetesClientObjectMapper` qualifier to identify this mapper.

> Another alternative would be to provide more methods on KubernetesSerialization to encapsulate the functionality of the ObjectMapper 

**Yes**, this is the right approach. We need to hide whatever implementation or technical internals we have to the user so we don't compromise any future change.
If direct access to the Mapper is needed because something is missing, this should be identified and added if necessary.

Just as a reminder, the Serialization class previously had 2 mappers (one for JSON one for YAML). We were able to ditch the one for YAML. Exposing internals prevents us from having freedom in our future steps (e.g. replacing the clone method, etc.).



As I said in my previous comment, I won't oppose to this change, but I do think is step back or a step in the wrong direction.

csviri:
Will close this since we took a different direction.

PR #5290, opened 2023-06-28
feat: catch generation errors and specify which CR caused it

metacosm:


crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractCustomResourceHandler.java
```diff
@@ -51,51 +52,55 @@ protected AbstractCustomResourceHandler(Resources resources, boolean parallel) {
...
+        }
+        String description = property.getComments().stream().filter(l -> !l.trim().startsWith("@"))
+            .collect(Collectors.joining(" ")).trim();
+        String format = (String) parameters.get("format");
+
+        resources.decorate(
+            getPrinterColumnDecorator(name, version, path, type, column, description, format));
+      });
+    } catch (Exception e) {
+      throw new RuntimeException("Couldn't generate CRD for " + config.crdName(), e);
```
andreaTP, line 102: I think that you can easily enough add a test 🙏

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractCustomResourceHandler.java
```diff
@@ -24,6 +24,7 @@
 import io.sundr.model.TypeDef;
 import io.sundr.model.TypeDefBuilder;
 
+import java.awt.event.WindowFocusListener;
```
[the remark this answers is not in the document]
metacosm: Ouch, wtf?

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractCustomResourceHandler.java
metacosm: It isn't trivial, actually. I actually cannot replicate the use case for which I added this for some reason (I suspect Lombok might play a role here 🤷🏼).

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractCustomResourceHandler.java
```diff
@@ -51,51 +51,57 @@ protected AbstractCustomResourceHandler(Resources resources, boolean parallel) {
...
+          column = property.getName().toUpperCase();
+        }
+        String description = property.getComments().stream().filter(l -> !l.trim().startsWith("@"))
+            .collect(Collectors.joining(" ")).trim();
+        String format = (String) parameters.get("format");
+
+        resources.decorate(getPrinterColumnDecorator(name, version, path, type, column, description, format));
+      });
+    } catch (Exception e) {
+      if (e instanceof IllegalArgumentException) {
```
sunix, line 100: Hello,
I am not sure to understand the purpose of it, why making it a special case? are you expecting to catch this somewhere else ?

metacosm: I don't recall exactly, tbh, but I think the idea is that there's already client code that expects an `IllegalArgumentException` to be thrown in some cases here, so just letting this bubble up as expected, yes.

manusa: Given this context,
`KubernetesClientException.launderThrowable` is what you want for all cases.

In case that's not what you need, why are we using an if-else instead of adding a different catch clause for IllegalArgumentException?

metacosm:
Let's close this, I don't recall the context and we can probably revisit if needed at some later time.

PR #5320, opened 2023-07-09
All lists to omit empty by default

### LINKED ISSUE #5262: Determine if all built-in collections should omit empty
Some collections are not marked as omitempty in their go classes: https://github.com/openshift/api/blob/master/apps/v1/types.go#L63

The java generator will not mark those fields for JsonInclude non_empty.

However our builders are opinionated about treating an empty and null collections the same (in some places), and users will end up with empty collections in their serialized forms without realizing it.  This is problematic for merge patches which treat the presense of an empty collection as a full replacement for the existing.

_Originally posted by @shawkins in https://github.com/fabric8io/kubernetes-client/issues/5222#issuecomment-1585653308_
           
We have two options to evaluate:

1. Just to be consistent make all collections default to non-empty (can be done in either the generator or the jsonschema2pojo logic).  This will of course open users up to #4606

2.  Try to modify the builder logic to differentaite between empty and null based upon the JsonInclude annotation - this seems like a tall order, and probably not viable.

shawkins:
## Description
Closes #5262 

There was only one real change to Fabric8DefaultRule, and one test updated to SerializationTest.  Everything else is the update of the models - which was surprisingly quite extensive.

To add a little more the root casue of this behavior:
- jsonschema2pojo defaults initializeCollections to true, so all lists in our built-in models already had an empty list (and maps) to being with
- builders will preserve that empty list default, so users would need to call withField(null) to omit the collection.

So the alternative would be to set initializeCollections to false and have the Fabric8DefaultRule take over that responsibility and  not initialize fields that are not set to omit empty.  That of course could also potentially be seen as breaking behavior as users will be required to do null checks that they didn't have to do before.

Changing the getter to use Optional would be even more breaking, but could be considered as well.

@manusa @metacosm @rohanKanojia any thoughts about this?  Whatever the decision, I'll add some appropriate change log entries.

## Type of change

## Checklist


rohanKanojia (approved):

manusa (approved):
Trusting your word that this PR only affects the `@JsonInclude(JsonInclude.Include.NON_EMPTY)` annotation, looks good to me.

shawkins:
> It's my understanding that the current state of the PR is with the minimum set of breaking changes (i.e. collections are still initialized).

This change will be breaking for someone those who were relying upon empty collections 
- I'm not sure where in the standard resources this would make a difference between empty and missing for non-patch operations, our guess is that it does not.
- Updated #4606 to more accurately describe where an empty collection would be used.  With a json merge - an empty collection will cause the patch to remove the list.  With a strategic merge if the list has a patchStrategy of atomic it will remove the list.  Users relying upon that behavior wrt the built-in resource serializing with an empty collection will have a breaking change.  json patches and SSA aren't affected.

> Trusting your word that this PR only affects the @JsonInclude(JsonInclude.Include.NON_EMPTY) annotation, looks good to me.

Sure you can trust me.  I also have some good deals on real estate :)

manusa:
So I guess that we're more or less fine once we add the changelog + breaking note

shawkins:
> So I guess that we're more or less fine once we add the changelog + breaking note

Should be good now.

PR #5323, opened 2023-07-10
[crd-gen] More principled filtering of enum constants for CRD generation

andreaTP:
## Description

With the upgrade of `sundrio` we can now correctly filter enum values.
cc. @shawkins 

## Type of change

## Checklist


crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
```diff
@@ -691,8 +691,9 @@ private T internalFromImpl(String name, TypeRef typeRef, Set<String> visited, In
           // check if we're dealing with an enum
           if (def.isEnum()) {
             final JsonNode[] enumValues = def.getProperties().stream()
-                .filter(property -> property.isStatic() && property.isPublic()
-                    && def.getFullyQualifiedName().equals(property.getTypeRef().toString()))
+                .filter(property -> property.isEnumConstant() && !property.isSynthetic() &&
```
shawkins, line 694: isEnumConstant should be sufficient as the filter.  https://github.com/fabric8io/kubernetes-client/pull/5323/files#diff-0bfefceda9952f447cf1a5965a31e7467320a96f1a1fd8877c49978d98102116R698 should no longer be needed as well.

isSynthetic in general should be a replacement for some of the logic that is looking at a field starting with a $ - but there was a convention used elsewhere that makes a property ignored by rewriting as starting with a $, which would need to be changed.

andreaTP: > isEnumConstant should be sufficient as the filter.

I think that the proposed implementation is the safest, as we can always have a property returning an enum constant of another kind, at least, the check on the type should stay.
Regarding `public` and `static` you are right, we can likely safely remove those.

shawkins: .filter(Property::isEnumConstant) should be sufficient

shawkins: ENUM$VALUES or $VALUES is already not an enumConstant, that's why with the change to remove the $ filtering, $VALUES is not showing back up as a field.

andreaTP: Thanks for the explanation, fine by me so 👍

andreaTP:
Ready to merge for me, optional reviewer @metacosm

metacosm (approved):

rohanKanojia (approved):

rohanKanojia:
I see this being marked as both bug and a chore :confused: . We should try to have tests validating bug fixes.

andreaTP:
@rohanKanojia we knew the previous implementation had some unhandled edge cases and here we are filling the gap.
More specifically those changes:
https://github.com/fabric8io/kubernetes-client/pull/5323/files#diff-7f18323f7e219a1a15b63b7c9eba51907ed29a7afd74567c305935930785ed22R114-R128

are breaking the previous implementation, but they got fixed with the one proposed in this PR.

sunix (approved):

PR #5347, opened 2023-07-20
adds a method to customize KubernetesMockServer clients

shawkins:
## Description

@manusa @jonsie this seems like the quickest fix, the alternative would be more involved - updating the annoation(s) to have a KubernetesSerialization provider.  Does this fix seem sufficient? 

More completely addresses #5293

## Type of change

 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [ ] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [x] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [x] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [ ] I Added [CHANGELOG](https://github.com/fabric8io/kubernetes-client/blob/main/CHANGELOG.md) entry regarding this change
 - [ ] I have implemented unit tests to cover my changes
 - [ ] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/main/doc/CHEATSHEET.md) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [ ] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

PR #5349, opened 2023-07-21
KubernetesMockServer Kubernetes client customization

### LINKED ISSUE #5293: Mock server doesn't take KubernetesSerialization customisation into account
### Describe the bug

This is particularly annoying in Quarkus' context where you'd expect the customised bean to be injected in the client returned by the mock server.

SNAPSHOT

### Steps to reproduce

Customise the KubernetesSerialization, use the mock server, realise that the server doesn't use the custom configuration.

### Expected behavior

The mock server should use the custom serialisation configuration.

### Runtime

Kubernetes (vanilla)

1.25.3@latest

### Environment

macOS

_No response_

### Additional context

_No response_

manusa:
## Description

Supersedes closes #5347 
Fixes #5293 

Built on top of #5347 to provide a user-friendlier approach to customize the `KubernetesClientBuilder` instance used by the `KubernetesMockServer` and the `KubernetesMockServerExtension`.

The easiest way to leverage the new functionality is by using the enhanced `@EnableKubernetesMockClient` where you can now pass a reference to a class implementing `Consumer<KubernetesClientBuilder>` to customize the original `KubernetesClientBuilder` provided by the `KubernetesMockServerExtension`.

The following code snippet is an example of such use case:
```java
@EnableKubernetesMockClient(crud = true, kubernetesClientBuilderCustomizer = CustomBuilder.class)
class TheTest {
  // ...
}
class CustomBuilder implements Consumer<KubernetesClientBuilder> {
  @Override
  public void accept(KubernetesClientBuilder builder) {
    /* customize at will */;
  }
}
```

/cc @jonsie

## Type of change

## Checklist


shawkins:
The NamespaceExtension arguably has the same issue.

Also let's just remove the OpenShiftMockServer rather than update it.

oscerd (approved):

manusa:
> The NamespaceExtension arguably has the same issue.
> 
> Also let's just remove the OpenShiftMockServer rather than update it.

Agreed to both. However, I'd rather tackle those separately (One new issue for each).
If this PR looks good for the scoped #5293 context I'd rather merge and be able to release along with v6.8.0 on Monday.

shawkins:
> Agreed to both. However, I'd rather tackle those separately (One new issue for each).

Let's give a little thought to the NamespaceExtension before proceeding.  Is your plan to similarly change the annotation and add another interface or base class to replace KubernetesMockClientKubernetesClientBuilder?  It might be nice to have them be the same to more easily switch from a crud mock to a kubernetestest - maybe provide the whole Config or ConfigBuilder rather than just the url string to the customizer.

Also https://github.com/fabric8io/kubernetes-client/pull/5349/files#diff-9ef31fbfb8fbafe18e5f18f5daa44de9705f3850a2e086aed2d123030a9448deR128 will need some documentation - if someone is using the mockserver manually (not through the annotation), then usage of this method is non-obvious as they likely need to extend KubernetesMockClientKubernetesClientBuilder

manusa:
I'm not convinced either about the String parameter. Sounds reasonable.
Let me add a further commit with some minor refinements.

manusa:
https://github.com/fabric8io/kubernetes-client/pull/5349/commits/3d178e66ce5283580339262c3b86909cda33b9ed

should address the expressed concerns

shawkins:
> should address the expressed concerns

It seems like the customizer could just be Consumer<KubernetesClientBuilder> - that would work for both the mock server and kubernetestest.  Customizing the Config could be done seperately - you could either also provide a plugable Consumer<ConfigBuilder> or like other builders the KubernetesClientBuilder could directly provide an editConfig method.

manusa:
I'm not really following why the current approach won't work for the `@KubernetesTest` annotation+NamespaceExtension.

In NamespaceExtension we would do something in the lines of `kubernetesClientBuilderInstantiator.apply(new Config())`. This would even enable users to configure their own cluster instead of relying on autoConfig.

Maybe better to have a quick sync.

manusa:
> Customizing the Config could be done separately

As agreed in internal meeting, let's defer the Config overriding/editing to a follow-up.

Tasks:
- Removing `OpenShiftMockServer`
- Provide similar functionality for `KubernetesTest`+`NamespaceExtension`
- Enable `Config` overrides for both extensions

shawkins (approved):
LGTM, thanks for pushing to wire this into the annotation.

jonsie:
Much appreciated, will keep an eye out for the release.

jonsie:
Thanks for the release!

PR #5375, opened 2023-08-07
[java-generator] Expand the documentation regarding options

shubhamvrs:
## Description
<!--
Thank a lot for taking time to contribute to Fabric8 <3!

### DIFF
--- a/java-generator/how toa use
+++ b/java-generator/how toa use
@@ -1,0 +1,74 @@
+
+Java Generator Options and Examples:
+
+Option: setPackageName(String packageName)
+
+Description: Sets the package name for the generated Java code.
+Example:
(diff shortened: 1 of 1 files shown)


andreaTP:
This is very much different from what we expected:

- you should expand an already existing section: https://github.com/fabric8io/kubernetes-client/blob/main/doc/java-generation-from-CRD.md#usage
- the format should be Markdown
- for each option:
  - it should include an example configuration for Maven at least (nice to have also CLI and Gradle plugin)
  - it should include a code snippet showing the relevant difference

shubhamvrs:
i tried sir but im unable to create it actually im new to github thatswhy !

PR #5390, opened 2023-08-14
Fix #5388: Generate deterministic CRDs

### LINKED ISSUE #5388: CRD generator is not deterministic

We're using the K8s client in a Quarkus Operator SDK application. We have configured the CRD generation of Quarkus, that is basically triggered on every test execution.

Since the CRD generation is not deterministic, the CRD file is constantly changed. Complex CRDs (for example, our current CRD is 3399 lines long) lead to multiple problems:
* When using the IntelliJ Kubernetes Plugin, you constantly see a `Reload schema` overlay on YAML files. 
image
* It's very hard to determine if the CRD file contains actual changes, or if its content is just in a different order.
* Since we include default K8s resources in our CRD (Metadata, StatefulSetSpec, ServiceSpec), we cannot assume that the CRD is still the same as long as we don't touch our custom resource classes. On every update of the K8s client we could inherit CRD changes, that could force us to bump our CRD version (which is a fine trade-off for us, since we need to offer the full feature set of a StatefulSet and Service spec configuration).
* We have the generated CRD file in version control, so we constantly have a modified file in our change sets.
* It's very hard to setup automatic tests between, for example, an operator and depending Helm charts repository, to assert that both CRD files are still the same (for example, for an automated CI test that ensured both projects are in sync).

The CRD file should be generated deterministically, so the output is the same when the input hasn't changed. Preferably all elements are sorted alphabetically.

We're using an external YAML sorter, but this is additional tooling we have to setup. This also doesn't solve the issues during development, when the CRD file is constantly changed when executing a simple test.

I have a technical solution for the `crd-generator` that I can offer as PR.

### Additional context

_No response_

Donnerbart:
## Description

Fixes #5388

I'm not sure if you want the provided `Complex` custom resource example as is in your code base, since it has nested `ObjectMetaData`, `StatefulSetSpec` and `ServiceSpec` and is tested against a checked in YAML file. So on every change on those K8s resources, the checked in file needs to be adapted (for example, non-breaking changes like adding new fields).

## Type of change

## Checklist


crd-generator/api/src/main/java/io/fabric8/crd/generator/DeterministicObjectMapper.java
```diff
@@ -0,0 +1,206 @@
...
+      return withConverter(clazz, Comparator.comparing(keyExtractor));
+    }
+
+    <T> DeterministicObjectMapperBuilder withConverter(final Class<T> clazz, final Comparator<T> comparator) {
+      customComparators.addConverter(clazz, comparator);
+      return this;
+    }
+
+    ObjectMapper build() {
+      final ObjectMapper mapper = JsonMapper.builder(new YAMLFactory()
```
andreaTP, line 81: This mapper is configured slightly differently from before, more specifically:

`YAML_MAPPER.configure(SerializationFeature.WRITE_NULL_MAP_VALUES, false)` and `YAML_MAPPER.configure(SerializationFeature.WRITE_EMPTY_JSON_ARRAYS, false);` are missing.

crd-generator/api/src/main/java/io/fabric8/crd/generator/DeterministicObjectMapper.java
```diff
@@ -0,0 +1,206 @@
...
+          .enable(YAMLGenerator.Feature.MINIMIZE_QUOTES)
+          .enable(YAMLGenerator.Feature.ALWAYS_QUOTE_NUMBERS_AS_STRINGS)
+          .disable(YAMLGenerator.Feature.WRITE_DOC_START_MARKER))
+          .configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, true)
```
andreaTP, line 85: Have you attempted to only add those configurations: `MapperFeature.SORT_PROPERTIES_ALPHABETICALLY` and `SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS` to the old mapper without defining an entirely new and custom one?
What are the invariants that we are missing?

andreaTP (commented):
@Donnerbart thanks a lot for taking the time to go through this!

I have a couple of questions but let me know your thoughts!

crd-generator/api/src/main/java/io/fabric8/crd/generator/DeterministicObjectMapper.java
Donnerbart: Both are deprecated API, so I looked how to configure it with up to date API.

`YAML_MAPPER.configure(SerializationFeature.WRITE_NULL_MAP_VALUES, false)` is `.withConfigOverride(Map.class, configOverride -> configOverride.setInclude(construct(NON_NULL, NON_NULL)))`.

`YAML_MAPPER.configure(SerializationFeature.WRITE_EMPTY_JSON_ARRAYS, false);` is `.serializationInclusion(NON_EMPTY)`.

crd-generator/api/src/main/java/io/fabric8/crd/generator/DeterministicObjectMapper.java
Donnerbart: The problem is that a lot of relevant elements of a CRD are not maps, but sets or other unsorted collections. The custom code is only taking care of collections.

`MapperFeature.SORT_PROPERTIES_ALPHABETICALLY` and `SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS` is already configured in my code, but it won't affect structures like enums, the additional printer columns section etc.

crd-generator/api/src/main/java/io/fabric8/crd/generator/DeterministicObjectMapper.java
andreaTP: Nice 👍 , thanks for the explanation!

andreaTP:
LGTM but I wish @metacosm can do a final review here.

andreaTP:
@shawkins input would be great too, at least to verify that those changes are not affecting somehow the KC operator.

crd-generator/api/src/main/java/io/fabric8/crd/generator/DeterministicObjectMapper.java
Donnerbart: With all tests of the `crd-generator` the following collections are serialized:
```
com.fasterxml.jackson.databind.node.TextNode
io.fabric8.kubernetes.api.model.apiextensions.v1beta1.CustomResourceColumnDefinition
io.fabric8.kubernetes.api.model.apiextensions.v1beta1.CustomResourceDefinitionVersion
io.fabric8.kubernetes.api.model.apiextensions.v1.CustomResourceColumnDefinition
io.fabric8.kubernetes.api.model.apiextensions.v1.JSONSchemaProps
```
I've added custom converters for all of them, including v1 and v1beta1 variants.

Donnerbart:
I've changed the comparison for `JSONSchemaProps` to use `type` first, since `id` doesn't seem to be populated (but still fallback to the `id` if `type` is the same value).

I also added more JavaDoc to the `DeterministicObjectMapper` and its builder class.

Donnerbart:
The license check is failing for the checked in YAML file in the test resources.
Since the file is used for a byte-wise comparison, it cannot be changed.

I see that there are already some exclusions for test resources:
```
            <exclude>kubernetes-client/src/test/resources/**/*</exclude>
            <exclude>kubernetes-client/src/test/resources/mockito-extensions/*</exclude>
            <exclude>**/src/test/resources/ssl/*</exclude>
            <exclude>**/src/test/resources/io/fabric8/java/generator/approvals/*.*.approved.txt</exclude>
            <exclude>**/it/**/expected/**/*.expected</exclude>
```
I don't want to change `<exclude>kubernetes-client/src/test/resources/**/*</exclude>` to match all modules, so I've added `<exclude>**/src/test/resources/*.yml</exclude>`, which is hopefully narrow enough to not violate the license requirements.

Donnerbart:
@shawkins That's a great finding! I've double checked our complex example, since there were for sure more nondeterministic fields. Looks like these were all properties, that are now sorted via `.configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true)`.

So indeed, with the correctly sorted enums that quite hacky way of adding the collection serializer is not needed anymore.

I also cleaned up the complex example, by adding simplified, but stable versions of the K8s models being used. This still provides enough properties, that were unsorted before, but won't break on future changes on the K8s model classes.

andreaTP:
@Donnerbart the CI failure on Win seems to be relevant, probably the encoding of the file is causing a difference.
I would be completely ok if, instead of comparing byte-by-byte we compare with `String::equal` the full lines.

This approach might also play nicer with the License check, e.g. you can include the License header in the committed file and skip it during the comparison.

Donnerbart:
@andreaTP Yeah, when I saw Windows I had the same thought. Will look into this first thing tomorrow (it's already end of workday for me).

I guess you don't want any helper dependency like Apache Commons or so in the test scope, so I'll write that comparison by hand. Will also look into adding the license header then.

andreaTP:
Thanks for getting back, no rush!

> I guess you don't want any helper dependency like Apache Commons or so in the test scope, so I'll write that comparison by hand.

I (personally) have no issues with adding helper dependencies to the test scope ...

Donnerbart:
* Removed the license check exclusion for YAML files in test resources.
* Added the license header to the checked in YAML file.
* Changed the file comparison to `BufferedReader::readLine` (that should handle all types of newlines).
* Skipped leading comments and newlines for file comparison.
* Improved some JavaDoc on the Complex example classes.

crd-generator/api/src/test/java/io/fabric8/crd/generator/ResourcesTest.java
```diff
@@ -42,14 +42,14 @@ public void shouldSupportMultiplePrinterColumns() {
...
-  public void shouldSupportMultipleSortPrinterColums() {
+  public void shouldSupportMultipleSortPrinterColumns() {
     Resources r = new Resources();
     SortPrinterColumnsDecorator dec1 = new SortPrinterColumnsDecorator("my-crd", "v1");
     SortPrinterColumnsDecorator dec2 = new SortPrinterColumnsDecorator("my-crd", "v2");
     r.decorate(dec1);
     r.decorate(dec2);
     assertEquals(2, r.getDecorators().size());
-    r.getDecorators().stream().forEach(d -> System.out.println(d));
+    r.getDecorators().forEach(System.out::println);
```
shawkins, line 52: It should be fine to start removing a log of the sys outs in tests.

shawkins (approved):
LGTM, just one minor comment.

crd-generator/api/src/main/java/io/fabric8/crd/generator/CRDGenerator.java
```diff
@@ -105,7 +120,10 @@ Map<String, AbstractCustomResourceHandler> getHandlers() {
     return handlers;
   }
 
-  public CRDGenerator customResourceClasses(Class<? extends CustomResource>... crClasses) {
+  // this is public API, so we cannot change the signature, so there is no way to prevent the possible heap pollution
```
metacosm, line 123: Not sure I understand the purpose fo this comment…

metacosm (approved):

crd-generator/api/src/main/java/io/fabric8/crd/generator/CRDGenerator.java
Donnerbart: IntelliJ showed a warning for this method regarding heap pollution ([this blog article explains it quite well](https://arbitrary-but-fixed.net/teaching/java/2018/02/01/java-generic-arrays-varargs.html)). It comes down to the combination of generics and varargs.

I assume this is public API and the method signature cannot be changed, so there is nothing we can do to prevent the possible heap pollution. To get rid of the warning we could either:
1. Make the method `final` and annotate with `@SafeVarargs`.
1. Add a `@SuppressWarnings("unchecked")` annotation.

I wasn't sure if 1. is an option, since it would also change the method signature and may break custom `CRDGenerator` implementations. So I chose 2. as solution.

metacosm: Thanks! Wasn't aware of this issue.

rohanKanojia (approved):

crd-generator/api/src/test/java/io/fabric8/crd/generator/ResourcesTest.java
Donnerbart: I removed the sys out, would be great if you could re-run the tests.

andreaTP (approved):

manusa (approved):
LGTM, thx!

PR #5611, opened 2023-11-25
Fix duplicate enum constant in java generator

iocanel:
**Context**: Trying to create KubeVirt model from CRDs. I am hitting an error as it seems that enum constants defined in the CRD are added twice.

This pull request changes JEnum to use Set instead of List to prevent any chance of having dublicate enum constants.

### DIFF
--- a/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JEnum.java
+++ b/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JEnum.java
@@ -26,8 +26,10 @@ import io.fabric8.java.generator.Config;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import static io.fabric8.java.generator.nodes.Keywords.JAVA_LANG_LONG;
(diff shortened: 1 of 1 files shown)


oscerd (approved):

andreaTP:
Can you share the actual error, or, better, a reproducer for the issue?

This seems to be an effect of something else going wrong, not the root cause. Do we have duplicated fields in the CRD after the name sanitization?

iocanel:
Regardless of where is the root cause, it does make sense to use a `Set` instead of `List` for the obvious reasons.

Regarding a reproducer: 
This repository: https://github.com/iocanel/quarkus-kubevirt-client
When built using: `mvn clean install -Pgenerate-model` can reproduce the issue.
The code generator will work, but will produce invalid code (that contains dublicate enum constants) and therefore will cause compilation to fail.

andreaTP:
> Regardless of where is the root cause, it does make sense to use a `Set` instead of `List` for the obvious reasons.

I agree, but I would really like to understand the underlying root cause before hiding it with this change.

andreaTP:
For reference, the bug is in a KubeVirt CRD IMHO:

https://github.com/kubevirt/cluster-network-addons-operator/blob/3a2b93b4259772257ead6c28ba748936da28e0bf/manifests/cluster-network-addons/0.90.1/network-addons-config.crd.yaml#L1523-L1537

they have duplicated `enum`s.

andreaTP (approved):
Given that the upstream bug has been solved here:
https://github.com/kubevirt/cluster-network-addons-operator/pull/1748

I'm good with those changes, thanks for reporting @iocanel

manusa (approved):

Issue #5327, opened 2023-07-11
io.fabric8.kubernetes.client.informers.impl.cache.Reflector.ReflectorWatcher#reconnecting() {  return true;     


chenylh:
### Describe the bug

io.fabric8.kubernetes.client.dsl.internal.AbstractWatchManager#cannotReconnect（）， always returns false，this will cause the client to keep trying to establish a connection asynchronously

### Fabric8 Kubernetes Client version

6.4.1

### Steps to reproduce

k8s ApiServer rate limit

### Expected behavior

Cancel asynchronous task

### Runtime

Kubernetes (vanilla)

### Kubernetes API Server version

other (please specify in additional context)

### Environment

Linux

### Fabric8 Kubernetes Client Logs

```shell
2023-07-11 18:00:48.727 DEBUG -- [pool-2-thread-3] .i.AbstractWatchManager : Scheduling reconnect task
2023-07-11 18:00:48.727 DEBUG -- [pool-2-thread-3] ckoffIntervalCalculator : Current reconnect backoff is 32000 milliseconds (T336)
2023-07-11 18:00:49.678 DEBUG -- [138.36:6443/...] i.f.k.c.i.i.c.Reflector : Event received MODIFIED Service resourceVersion v1536952902 for v1/services
2023-07-11 18:00:49.678 DEBUG -- [138.36:6443/...] i.f.k.c.i.i.c.Reflector : Event received MODIFIED Service resourceVersion v1536952902 for v1/services
2023-07-11 18:00:50.082 DEBUG -- [ool-2-thread-35] .i.AbstractWatchManager : Watching https://11.50.138.174:6443/api/v1/services?resourceVersion=152868113&allowWatchBookmarks=true&watch=true...
2023-07-11 18:00:50.083 DEBUG -- [ool-2-thread-35] i.f.k.client.Config     : Trying to configure client from Kubernetes config...
2023-07-11 18:00:50.083 DEBUG -- [ool-2-thread-35] i.f.k.client.Config     : Did not find Kubernetes config at: [/root/.kube/config]. Ignoring.
2023-07-11 18:00:50.083 DEBUG -- [ool-2-thread-35] i.f.k.client.Config     : Trying to configure client from service account...
2023-07-11 18:00:50.083 DEBUG -- [ool-2-thread-35] i.f.k.client.Config     : Found service account host and port: 10.233.0.1:443
[...]

PR #5627, opened 2023-12-01
fix: prevents memory accumulation from informers

### LINKED ISSUE #5626: Memory Leak in Kubernetes Client
### Describe the bug

When using Jobs in combination with LogWatches, since version 6.9.0 a memory leak exists.  
I found out, as the memory gradually increases over time. I see the retained heap size in the old gen is getting bigger and bigger. After debugging for a day, I feel like the leak comes from the informer functionality and that the used CompletableFuture is not properly teared down. This creates a lot of back references which the GC can't drop. However, was not fully able to nail it down to the exact location.

I created a demo project which causes the fault to appear quite frequently after a few minutes.  
Please also see the heap dump I attached showing the problem here.

image

[heapdump.zip](https://github.com/fabric8io/kubernetes-client/files/13529474/Main_469235_01_12_2023_16_53_53.zip)

[project.zip](https://github.com/fabric8io/kubernetes-client/files/13529492/memory-leak.zip)

6.9.2

### Steps to reproduce

1. Run demo project
2. Wait 5 mins
3. look at the heap dump

### Expected behavior

no memory leak :D

### Runtime

1.25.3@latest

### Environment

Linux

_No response_

### Additional context

k3d 1.25.3

shawkins:
## Description

I completely missed the need for cleanup with the proactive closing of informers with #5327.  Unfortunately this affects both implicit and explicit usage of informers.  An alternative approach is to track if the client has been closed - I had initially avoided that because it's more work, as you can see from this pr.  This approach though doesn't have much risk other than some code complexity.

The only other possibility would be to track a set of informers on the BaseClient and close them using the completablefuture, but that will also require actively removing entires from the set once they've already been closed.

Fixes #5626

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * Fix #5580: [java-generator] Correctly handle defaults for IntOrString types
 * Fix #5584: Fix CRD generation when EnumMap is used
+* Fix #5626: Prevent memory accumulation from informer usage
 
 #### Improvements
 * Fix #5429: moved crd generator annotations to generator-annotations instead of crd-generator-api. Using generator-annotations introduces no transitive dependencies.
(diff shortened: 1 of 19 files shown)


oscerd (approved):

shawkins:
Correction - closing the client doesn't seem to have enough of a proactive effect for the ResourceTest test to pass.  The next update to the commit adds the explicit tracking of closeables to the BaseClient.  For now I've left in the HttpClient close tracking as well, but it could be removed if desired.

PR #5699, opened 2024-01-09
Reduce CacheImpl lock contention

schlosna:
## Description

When `io.fabric8.kubernetes.client.informers.cache.Lister#list` is used with a namespace under heavy concurrent load, the `CacheImpl#byIndex` becomes a performance bottleneck due to `synchronized` method contention.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/informers/cache/BasicItemStore.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/informers/cache/BasicItemStore.java
@@ -19,13 +19,14 @@ package io.fabric8.kubernetes.client.informers.cache;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.function.Function;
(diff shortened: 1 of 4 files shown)


manusa:
BasicItemStoreTest.java needs a header with the **exact** same content as header.txt (notice that the year is not the current year but the inception year -2015-)

shawkins:
@schlosna thank you for the pr. Just a couple of thoughts:
- I don't think users should notice / care there's a small difference in consistency here between the cache and index state vs. the fully synchronized case. However a javadoc on CacheImpl could mention a serial usage expectation as parallel mutative operations could result in things like orphaned index entries - we currently enforce serialized execution of watch events flow through a SerialExecutor.
- it would probably be better to have CacheImpl hold a special purpose lock object, or we'll need another release not letting implementors of ItemStores know that their object will be the lock used by the CacheImpl.
- there's a few more places where locking may be good - removeIndexer goes against the pattern of locking on mutative index operations, and addIndexers produces an unsychronized copy of indexers.
- it would be great to avoid explicit double locking under addIndexFunc, but it doesn't at first glance seem possible. Also I don't think there would be chance for deadlock - especially if we change the object lock to something held only by CacheImpl.

PR #5731, opened 2024-01-29
fix: ensures kind is set for generic resource lists

### LINKED ISSUE #5729: Kind is `null` in for `GenericKubernetesResource` informer `ResourceEventHandler
### Describe the bug

When I create an informer that uses `GenericKubernetesResource` the kind of the existing resources is `null`, sample:

```java
var informer = client.genericKubernetesResources("v1","ConfigMap").inAnyNamespace().runnableInformer(0);
        informer.addEventHandler(new ResourceEventHandler<GenericKubernetesResource>() {
            @Override
            public void onAdd(GenericKubernetesResource configMap) {
                System.out.println("config map kind "+configMap.getKind());
            }

            @Override
            public void onUpdate(GenericKubernetesResource configMap, GenericKubernetesResource t1) {
            }

            @Override
            public void onDelete(GenericKubernetesResource configMap, boolean b) {
            }
        });
        informer.start();
```
It is already `null` in cache.
It is strange that, if I create a new `ConfigMap` with the client, it gets propagated properly.

6.10.0

### Steps to reproduce

See sample above. Try it with minikube for example.

### Expected behavior

Kind field is propagated

### Runtime

Kubernetes (vanilla)

1.25.3@latest

### Environment

Linux

```shell
No specific logs
```

### Additional context

_No response_

shawkins:
closes #5729

There are two additional thoughts here:
1. we should not need to, but we could do this for types other than generic if we add setKind to HasMetadata.
2. the logic is currently only doing this for lists as other operations and events doen't seem to have the same problem - but we could of course move the logic into the updateApiVersion if needed.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/impl/BaseClient.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/impl/BaseClient.java
@@ -47,6 +47,7 @@ import org.slf4j.LoggerFactory;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
@@ -116,7 +117,7 @@ public abstract class BaseClient implements Client {
 
   BaseClient(final HttpClient httpClient, Config config, ExecutorSupplier executorSupplier,
       KubernetesSerialization kubernetesSerialization) {
-    this.closable = Collections.newSetFromMap(new WeakHashMap<>());
+    this.closable = Collections.synchronizedSet(Collections.newSetFromMap(new WeakHashMap<>()));
     this.closed = new CompletableFuture<>();
     this.config = config;
     this.httpClient = httpClient;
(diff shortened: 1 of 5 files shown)


rohanKanojia (approved):

manusa:
> we should not need to, but we could do this for types other than generic if we add setKind to HasMetadata.

Shouldn't be necessary since the kind is hardcoded during model generation.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/impl/BaseClient.java
```diff
@@ -116,7 +117,7 @@ public void onClose(Executor executor) {
 
   BaseClient(final HttpClient httpClient, Config config, ExecutorSupplier executorSupplier,
       KubernetesSerialization kubernetesSerialization) {
-    this.closable = Collections.newSetFromMap(new WeakHashMap<>());
+    this.closable = Collections.synchronizedSet(Collections.newSetFromMap(new WeakHashMap<>()));
```
manusa, line 120: Do these changes have anything to do with this PR?

manusa (approved):

PR #5756, opened 2024-02-16
feat: use Editable when possible instead of reflection to infer builder

metacosm:
- **feat: use Editable when possible instead of reflection to infer builder**

### DIFF
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesResourceUtil.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesResourceUtil.java
@@ -480,10 +480,6 @@ public class KubernetesResourceUtil {
     return (Class<? extends KubernetesResourceList>) loadRelated(type, "List", DefaultKubernetesResourceList.class);
   }
 
-  public static <T extends HasMetadata, V extends VisitableBuilder<T, V>> Class<V> inferBuilderType(Class<T> type) {
-    return (Class<V>) loadRelated(type, "Builder", null);
-  }
-
   private static Class<?> loadRelated(Class<?> type, String suffix, Class<?> defaultClass) {
     try {
       return Thread.currentThread().getContextClassLoader().loadClass(type.getName() + suffix);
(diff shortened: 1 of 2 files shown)


shawkins:
I think the only reason this was not done initially is that there's no guarentee that custom resources implement Editable. If a user still wants client logic that interacts with visitors, they will need to make it Editable. 

There is also BaseFluent.builderOf that falls back to reflection as well.

manusa:
> I think the only reason this was not done initially is that there's no guarentee that custom resources implement Editable. If a user still wants client logic that interacts with visitors, they will need to make it Editable.

For this case maybe we can rely on a fallback?
I can't recall if users had to do additional setup to register the custom resource types for reflection. If that was the case, then this would be compliant with the current behavior, since those builders should need registering too.

metacosm:
> I think the only reason this was not done initially is that there's no guarentee that custom resources implement Editable. If a user still wants client logic that interacts with visitors, they will need to make it Editable.

Indeed, and I think that's a fair assumption to make. The current implementation attempts to locate a builder via reflection even if it doesn't exist, making users explicitly implement Editable will make them create the associated Builder if they want their custom resources to be editable.

> 
> There is also BaseFluent.builderOf that falls back to reflection as well (we could be calling that method here).

Yes, but the goal is to remove reflection altogether here.

metacosm:
> > I think the only reason this was not done initially is that there's no guarentee that custom resources implement Editable. If a user still wants client logic that interacts with visitors, they will need to make it Editable.
> 
> For this case maybe we can rely on a fallback? I can't recall if users had to do additional setup to register the custom resource types for reflection. If that was the case, then this would be compliant with the current behavior, since those builders should need registering too.

Historically, QOSDK was taking care of registering Custom resources associated types for reflection, now the kubernetes client extension does it in a more complete way (though this is currently only in 3.8.0.CR1, iirc).

As far as fallback, I don't really see the point, to be honest, apart from potential backwards compatibility (but that would be in user-provided code since resources provided by the client should already be implemented to support Editable). Either resources have an associated builder and in this case it's trivial to implement Editable or they don't and no fallback will solve the issue. On the other hand, any fallback implementation would require reflection, which we're trying to remove.

For that matter, we might want to consider a `Listable` interface to provide a way to instantiate the associated list type without requiring reflection as well.

shawkins:
To add more on this the sundio basefluent.builderOf method did not change significantly from the past - it was always using reflection. The change I'm think of was to have the fluent.builder methods call basefluent.builderOf. 

So we may not want to change builderOf, and instead introduce a new method that just checks for editable, which would be used by both the fluent builder methods and by the fabric8 client.

andreaTP:
Probably, we need to solve #5878 along with this.

Issue #5619, opened 2023-11-28
The java generator produces a new for each type occurrence


iocanel:
### Describe the bug

It's not exactly a bug, it's more of a usability thing.

Currently, the java generator will produce a new class for each occurrence of a type in the CRD schema. This means that types that are frequently referenced, e.g. `metadata`, `selector` and more are added multiple times.

One obvious drawback is the size of the generated model (e.g. for kubevrit I got a model of around 3K classes).

Another issue, is that this breaks the visitor concept, since a single visitor is meant to visit all occurrences of type in the hierarchy.   

What we could possibly do, is keep track of the name and structure of the objects we create and allow conditional reuse. I could see that working with a config fragment like:

```xml
<resusableObjects>
   <metadata>io.fabric8.kubernetes.model.ObjectMeta</metadata>
   <selector>io.fabric8.kubernetes.model.Selector</selector>
   <datavolume>io.quarkiverse.kubevirt.blah.blah</datavolume>
</reusableObjects>
```
This is just something that just come into my head, there are probably even better ways to approach this.

### Fabric8 Kubernetes Client version

SNAPSHOT

### Steps to reproduce

1. checkout: https://github.com/iocanel/quarkus-kubevirt-client
2. build with `mvn clean install -Pgenerate-model`
3. check the generated model

### Expected behavior

Types being reused

### Runtime

Kubernetes (vanilla)

### Kubernetes API Server version

1.25.3@latest

### Environment

Linux

andreaTP:
Hi @iocanel , and thanks for opening this super interesting discussion.
I have been tinkering around those things for a while, but couldn't get convinced about one possible implementation and ended up delaying it, thanks for bumping it up!

The big show-stopper for me has been forward compatibility, imagine the following scenario.
V1:
- `class1` -> `class1.metadata1`
- `class2` -> `class2.metadata2`
the 2 metadata fields are identical and we squash them into one object, so it becomes:
- `class1` -> `root.metadata`
- `class2` -> `root.metadata`

This is nice from the deduplication POV, code size etc. etc.
But we don't have control over the shape of metadata, as it's defined by the user, and, for the next V1 release they can add a field to `metadata2` and not to `metadata1` which is a fair non-breaking change.
Unfortunately, we end up being unable to "refactor" the metadata classes, causing a breaking generated code change (package names etc. etc.).

Your proposal is interesting, as it means that the user is still in control of "what to reuse", although it potentially has the same drawback explained above.

[...]

PR #5844, opened 2024-03-30
Support for javaType and existingJavaType extensions in java-generator

matteriben:
## Type of change

## Checklist


oscerd (approved):

andreaTP:
cc. @iocanel  this might help with classes deduplication you have been asking for.

The downside is that, with this approach, you need to be in control of the CRDs ... still seems like a low effort way to start to close the gap.

oscerd (approved):

kubernetes-model-generator/kubernetes-model-apiextensions/src/generated/java/io/fabric8/kubernetes/api/model/apiextensions/v1/JSONSchemaProps.java
```diff
@@ -50,10 +50,12 @@
     "example",
     "exclusiveMaximum",
     "exclusiveMinimum",
+    "existingJavaType",
```
andreaTP, line 53: I haven't realized before that we are changing `JSONSchemaProps` to add those custom extensions.

To stay as close as possible to the k8s representation I think we should not touch this class and rely on an optional [extension](https://swagger.io/docs/specification/openapi-extensions/).
Not sure about the naming, maybe something like:
- `x-javagen-java-type`
- `x-javagen-existing-java-type`

cc. @manusa @rohanKanojia

andreaTP (changes requested):
@matteriben thanks a lot for starting this effort, I added some notes but I'm supportive to this effort 👍 

I have a couple of additional questions:

- I'm not 100% convinced about having 2 different options, I think that `existingJavaType` is enough for 90% of the real-world use cases. Applying the Okkam razor we should only keep one, happy to be convinced otherwise though 🙂 
- I'm interested in exploring(we can do this in a follow-up PR) how to achieve the same result without touching the original CRD, e.g. using a `java-generator` configuration option. It doesn't need to be super "easy" as it's an advanced usage, maybe re-using a logic similar to `packageOverride`?

Happy to have @manusa feedback here 🙏

java-generator/core/src/main/java/io/fabric8/java/generator/Config.java
```diff
@@ -64,60 +66,6 @@ public Config(
     }
   }
 
-  public Config(
```
andreaTP, line 67: low risk, but this is a breaking change, should we go through a round of deprecation before removing those?

andreaTP (approved):
Amazing work @matteriben !
You already covered every aspect of it 🙂  the DX is(as expected) a little clunky but the functionality is there 👍 

I would say that this fixes #5619 too.
LGTM!

matteriben:
@andreaTP @manusa @oscerd @rohanKanojia @shawkins @sunix 

Please let me know if there is anything else I can do to get this change approved and merged.

rohanKanojia (approved):

andreaTP:
@matteriben sorry for the delay, please resolve the conflicts with the changelog and I'll do a final review and eventually merge tomorrow morning.

java-generator/core/src/main/java/io/fabric8/java/generator/Config.java
```diff
@@ -64,6 +66,7 @@ public Config(
     }
   }
 
+  @Deprecated
```
andreaTP, line 69: tnx 👍

andreaTP (approved):
sorry for the delay and thanks a lot for this effort, really appreciated!
Great job here 👍 
LGTM

manusa (approved):
LGTM, thx!

PR #5877, opened 2024-04-10
Address issues from #5866 - improve built-in resolving

### LINKED ISSUE #5866: CRD generation fails when compiled using java version 19 and above
### Describe the bug

CRD generation fails when compiled using java version 19 and above.

SNAPSHOT

### Steps to reproduce

There is an example project here, CRD generation succeeds when built with java version 18 and fails when built with java version 19.

https://github.com/matteriben/crd-generator-date-time-reproducer/actions/runs/8583665846

[job-logs.txt](https://github.com/fabric8io/kubernetes-client/files/14894938/job-logs.txt)

### Expected behavior

CRD generation is expected to succeed when project is built with java version 19 and above.

### Runtime

1.25.3@latest

### Environment

Linux

_No response_

### Additional context

This seems to be related to a change in the implementation of java.time.ZonedDateTime from version 18 to 19 and above.

shawkins:
## Description

Fixes #5866

Addresses issues from #5866

Relies upon jackson to resolve java. / javax. classes.

Does not yet change the default serialization for time stuff, nor does it address being able to pass in the objectmapper.

## Type of change

## Checklist

### DIFF
--- a/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
+++ b/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
@@ -16,7 +16,11 @@
 package io.fabric8.crd.generator;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.module.jsonSchema.JsonSchema;
+import com.fasterxml.jackson.module.jsonSchema.JsonSchemaGenerator;
+import com.fasterxml.jackson.module.jsonSchema.types.ArraySchema.Items;
 import io.fabric8.crd.generator.InternalSchemaSwaps.SwapResult;
 import io.fabric8.crd.generator.annotation.SchemaSwap;
 import io.fabric8.crd.generator.utils.Types;
@@ -24,6 +28,7 @@ import io.fabric8.generator.annotation.ValidationRule;
 import io.fabric8.kubernetes.api.model.Duration;
 import io.fabric8.kubernetes.api.model.IntOrString;
 import io.fabric8.kubernetes.api.model.Quantity;
+import io.fabric8.kubernetes.client.utils.KubernetesSerialization;
 import io.sundr.builder.internal.functions.TypeAs;
 import io.sundr.model.AnnotationRef;
 import io.sundr.model.ClassRef;
@@ -43,6 +48,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.LinkedList;
@@ -122,6 +128,9 @@ public abstract class AbstractJsonSchema<T, B> {
   public static final String JSON_NODE_TYPE = "com.fasterxml.jackson.databind.JsonNode";
   public static final String ANY_TYPE = "io.fabric8.kubernetes.api.model.AnyType";
 
+  private static final JsonSchemaGenerator GENERATOR;
+  private static final Set<String> COMPLEX_JAVA_TYPES = new HashSet<>();
+
   static {
     COMMON_MAPPINGS.put(STRING_REF, STRING_MARKER);
     COMMON_MAPPINGS.put(DATE_REF, STRING_MARKER);
@@ -138,6 +147,10 @@ public abstract class AbstractJsonSchema<T, B> {
     COMMON_MAPPINGS.put(QUANTITY_REF, INT_OR_STRING_MARKER);
     COMMON_MAPPINGS.put(INT_OR_STRING_REF, INT_OR_STRING_MARKER);
     COMMON_MAPPINGS.put(DURATION_REF, STRING_MARKER);
+    ObjectMapper mapper = new ObjectMapper();
+    // initialize with client defaults
+    new KubernetesSerialization(mapper, false);
+    GENERATOR = new JsonSchemaGenerator(mapper);
   }
 
   public static String getSchemaTypeFor(TypeRef typeRef) {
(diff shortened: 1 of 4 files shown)


andreaTP:
Thanks a lot for looking into this @shawkins , I'm not sure about the cost/benefits tradeoff in adding a Jackson dependency just for logging, would it be possible to avoid it somehow?

Alternatively, if it doesn't add much to the complexity of the code, can we consider having it as an optional dependency?

Last, but not least, I'm interested in the output from the `ZonedDateTime` on Java 21 if you have already tried it out.

shawkins:
> Thanks a lot for looking into this @shawkins , I'm not sure about the cost/benefits tradeoff in adding a Jackson dependency just for logging, would it be possible to avoid it somehow?

Jackson is not being used for logging, it's being used to resolve the type for built-in stuff like ZonedDateTime. The changes here allow use to leverage jackson's schema for all of the custom serialiers for java and javax classes. This makes the COMMON_MAPPINGS for things like long and double redundant, but it's fine to leave that logic as is.

> Alternatively, if it doesn't add much to the complexity of the code, can we consider having it as an optional dependency?

That is possible, but we'll then hit the cyclic issue for ZonedDateTime and still have incorrect mappings for a lot of things.

> Last, but not least, I'm interested in the output from the ZonedDateTime on Java 21 if you have already tried it out.

With these changes it's mapped to number.

manusa (approved):
LGTM, thx!

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
```diff
@@ -138,6 +147,10 @@ public abstract class AbstractJsonSchema<T, B> {
     COMMON_MAPPINGS.put(QUANTITY_REF, INT_OR_STRING_MARKER);
     COMMON_MAPPINGS.put(INT_OR_STRING_REF, INT_OR_STRING_MARKER);
     COMMON_MAPPINGS.put(DURATION_REF, STRING_MARKER);
+    ObjectMapper mapper = new ObjectMapper();
+    // initialize with client defaults
+    new KubernetesSerialization(mapper, false);
+    GENERATOR = new JsonSchemaGenerator(mapper);
```
manusa, line 153: This feels dangerous, especially when combined with Quarkus. Not that I can think of a better solution to make this an instance variable and be able to pass in its dependencies, but mentioning it just in case.

Does a `KubernetesSerialization.generateSchema` method make sense (considering the KubernetesSerialization class is already quite coupled with Jackson)?

andreaTP (changes requested):
`date-time` fields should resolve to `string` and not `numeric`.

I'm a bit uncomfortable with the type resolution done using Jackson, do we have any alternative?

shawkins:
> date-time fields should resolve to string and not numeric.

That was my first guess too, but ZonedDateTime is actually a numeric by default. If you want it to be a string, then you do need to use some additional mechanism:

 - Change our ObjectMapper to use mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
 - SchemaFrom
 - the format specification from the other pr. 

Jackson also supports the JsonFormat annotation for this purpose - but obviously the schema generator is not looking for that.

> I'm a bit uncomfortable with the type resolution done using Jackson, do we have any alternative?

Why are you uncomfortable? It's the clearest indication of what will actually happen at runtime when you serialize. In this case if we don't think that numeric handling is a sensible default - then we should disable SerializationFeature.WRITE_DATES_AS_TIMESTAMPS.

I'd like to avoid if at all possible introducing more duplicate annotations with Jackson - since ultimately users will still need to be responsible for those as well.

Try the JacksonTypeTest without this change - all of those fields will resolve to an object type will a variety of properties. Then look at the class heirarchy for StdSerializer - there's quite a bit of type handling that Jackson does that the crd generator is not aware of. I could see we were missing handling for Short - we're also missing Byte, byte, short, and byte[] (the primitives are not covered by this pr). 

The only alternative I can think of would be to restrict to only a blessed set of known types - and if you use anything other than that default to an exception, rather than emitting what could be erroneous schema information.

crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
shawkins: @andreaTP before I take this further, can I get some assurance that you will actually consider this approach?

andreaTP: I'm not going to oppose to this effort, at the same time I'm not sure about it:

- adds an external dependency
- the type resolution logic is currently fully coming from `sundrio`, this approach effectively duplicates the source of truth for specific types
- it binds the type resolution to the serialization format/configuration which is effectively a change from the current behavior

Those details are going to make bug resolution in this codebase more tricky, but happy to be proven wrong.
Possible next steps to be evaluated:

- we resolve all the types(not only the "complex") using Jackson(would it be possible to remove the dependency on `sundrio` in this case?)
- we bake and harmonize this type resolution logic into a fully recursive algorithm(maybe in the `sundrio` codebase?)

shawkins: > adds an external dependency

An additional dependency from a family of dependencies on which we already rely in other parts of the fabric8 client.

> the type resolution logic is currently fully coming from sundrio, this approach effectively duplicates the source of truth for specific types

It's unfortunate that sundrio is coming into the picture here at all. The choice to run this as an annotation processor is forcing that decision. Sundrio is a poor source of truth for this task and adds significant complexity and redundant logic with jackson instead of just using reflection and jackson schema. My vote would still be to eventually push the crd generator to something that runs after compilation and eliminate sundrio entirely from this.

> it binds the type resolution to the serialization format/configuration which is effectively a change from the current behavior

First you should acknowlege the current behavior is lacking or wrong - it is producing a schema that is invalid in many sitiatutions, not just these built-in types that have additional serdes handling in jackson, but in many other circumstances where it doesn't understand what jackson will actually do.

Second that serialization format/configuration is the more well understood construct / contract for a developer. I can create a bean, run Serialization.asJson and see the output and would then expect the crd generator to create an appropriate schema to represent that. Instead I have to be very aware of limitations or issues with the crd generator and deeply cross check that the output will work, and if not go over a separate set of docs about how to make it work.

> Those details are going to make bug resolution in this codebase more tricky, but happy to be proven wrong.

I don't think so. The jackson schema usage here is very limited and to a set of types that we're likely to already handle incorrectly.

> we resolve all the types(not only the "complex") using Jackson(would it be possible to remove the dependency on sundrio in this case?)

Only if we move away from this being run as an annotation processor - which I'm fully in favor of.

> we bake and harmonize this type resolution logic into a fully recursive algorithm(maybe in the sundrio codebase?

I would not involve sundrio at all. We should be able to do crd generation based upon jackson schema and reflection alone. At worst defincencies in jackson schema (I identified several when we were looking at the default issue) should be need to be addressed in jackson - but that would improve jackson schema for everyone, and we can of course workaround it locally with reflection as needed.

andreaTP: Thanks for all the context @shawkins now I have a better picture about where we are going here.

> just using reflection and jackson schema

how this will work in a Quarkus `BuildStep`?
and, just to make sure I understand, how is this going to work in GraalVM `native-image`?

> push the crd generator to something that runs after compilation and eliminate sundrio entirely from this.

I don't disagree but we need to involve other stakeholders in this decision cc. @metacosm @iocanel 

> The jackson schema usage here is very limited and to a set of types that we're likely to already handle incorrectly.

This is causing, anyhow, a divergence in handling types, which is something I would like to avoid.

All in all, a few technical details to be discussed further but my vote goes to either fully rely on sundrio or fully switch to Jackson + reflection.
I'm not up for maintaining both approaches unless there are very compelling reasons for doing so.

shawkins: > how this will work in a Quarkus BuildStep?

It just needs to know how to find the root classes, it's regular reflection and jackson-schema from there.

> and, just to make sure I understand, how is this going to work in GraalVM native-image?

I'm not even sure that comes into the picture. The crd generation is running off of regular java classes prior to creating a native image. Although I suppose you could elect to run it at runtime on a native-image - but I don't see why we'd worry about that yet.

> I'm not up for maintaining both approaches unless there are very compelling reasons for doing so.

That's fine for an overall direction, but please don't block this fix / improvement by arguing that we should already be at a final state.

andreaTP: > That's fine for an overall direction, but please don't block this fix / improvement by arguing that we should already be at a final state.

Given the fact that this fix/improvement involves a scoped issue that already has a viable workaround, it is not compelling enough to introduce the proposed complexity imho.

shawkins: > Given the fact that this fix/improvement involves a scoped issue that already has a viable workaround, it is not compelling enough to introduce the proposed complexity imho.

I disagree. There isn't a high degree of complexity here - it's effectively one method, limited to specific types, and I think we're both saying it's the likely direction for the crd generator in general. The changes to produce a better cycle error message (it's in a separate commit already) can be moved to its own pr if that would help.

You have two options:
1. document a workaround for a subtle problem that affects maybe as many as 100 classes. To know they are affected a user will have to test the field with a non-null value against a live kubernetes server - only then will they see that what they are serializing won't be accepted. That's not great and could lead to publishing an incorrect crd.
2. move forward with this fix that will be correct the vast majority of the time for the affected java / javax types, since we don't know when a full rewrite will be available - which I fear you will be oppositional to as well.

cc @manusa @metacosm

andreaTP: > it's the likely direction for the crd generator in general

We don't have a deal on this just yet.

> document a workaround for a subtle problem that affects maybe as many as 100 classes

How did you get this number?

> with a non-null value against a live kubernetes server

This is surprising, the reported bug happens at generation time, what am I missing?

shawkins: > We don't have a deal on this just yet.

Which is what I fear in general - you say that you are open to removing sundrio, but in practice I doubt it.

> How did you get this number?

It's just a guess - see the comment above about looking at the class hierarchy of StdSerializer.

> This is surprising, the reported bug happens at generation time, what am I missing?

The bug happens at generation time - but when will you be aware of it?

andreaTP: > but in practice I doubt it

Just to make it clear, don't doubt 🙂 , I do believe that Jackson + Reflection can improve over the current situation.

> The bug happens at generation time - but when will you be aware of it?

At generation time, the current status on this bug is that the `crd-generator` throws a StackOverflow Exception.

shawkins: > At generation time, the current status on this bug is that the crd-generator throws a StackOverflow Exception.

Sorry if this is not clear, but this fix addresses more than the stackoverflow - see the associated test class. All of those types are currently wrong.

andreaTP (approved):
As discussed, LGTM.

> The decision seems to be that a backwards compatibility flag is not necessary, but I'm fine with adding one.

Either way fine by me.

shawkins:
@manusa @andreaTP this should be up-to-date now. The description reflects it's not yet changing the default time mappings, and it's still just a static objectmapper - both can be addressed in follow-on issues.

rohanKanojia (approved):

manusa (approved):
LGTM, thx!

Issue #5881, opened 2024-04-11
CRDGenerator: Add support for generic schema definition with @Schema annotation


baloo42:
### Is your enhancement related to a problem? Please describe

At the moment the CRDGenerator does not support some schema properties like `format`, `example`, `externalDocs` (see #5859).

### Describe the solution you'd like

Introduce `@Schema` annotation which allows to define such leftovers:

```java
package io.fabric8.crd.generator.annotation;

import java.lang.annotation.ElementType;
import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;
import java.lang.annotation.Target;

/**
 * Java representation of JSONSchemaProps.
 *
 * <p>The annotation may be used to define a Schema for a set of elements of the OpenAPI spec, and/or
 * to define additional properties for the schema. The annotation has the highest precedence,
 * which means it always wins if the class is annotated with other annotations.</p>
 */
@Target({ ElementType.ANNOTATION_TYPE, ElementType.FIELD, ElementType.METHOD })
@Retention(RetentionPolicy.RUNTIME)
public @interface Schema {

  /**
   * The name of the schema or property.
   *
   * @return the name of the schema
   **/
  String name() default "";

  /**
   * A title to explain the purpose of the schema.
   *
   * @return the title of the schema
   **/
  String title() default "";

  /**
   * A description of the schema.
   *
   * @return the schema's description
   **/
  String description() default "";

  /**
   * Constrains a value such that when divided by the multipleOf, the remainder must be an integer.
   * Ignored if the value is 0.
   *
   * @return the multiplier constraint of the schema
   **/
  double multipleOf() default 0;

  /**
   * Provides an optional override for the format.
   * <p>The following formats are validated:</p>
   * <ul>
   * <li>{@code bsonobjectid}: a bson object ID, i.e. a 24 characters hex string</li>
   * <li>{@code uri}: an URI as parsed by Golang net/url.ParseRequestURI</li>
   * <li>{@code email}: an email address as parsed by Golang net/mail.ParseAddress</li>
   * <li>{@code hostname}: a valid representation for an Internet host name, as defined by RFC 1034, section 3.1 [RFC1034].</li>
   * <li>{@code ipv4}: an IPv4 IP as parsed by Golang net.ParseIP</li>
   * <li>{@code ipv6}: an IPv6 IP as parsed by Golang net.ParseIP</li>
[...]

Issue #5867, opened 2024-04-06
Generate a CRD with {type: string, format: date-time}


matteriben:
### Is your enhancement related to a problem? Please describe

I'd like to generate a CRD with a timestamp of type=string, and format=date-time.

I can generate code from a CRD with a string using date-time format, but when I generate a CRD from that code it generates something else entirely. For example:

Original CRD snippet:
```
                timestamp:
                  type: string
                  format: date-time
```

Java code snippet generated from above CRD snippet:
```
    @com.fasterxml.jackson.annotation.JsonProperty("timestamp")
    @com.fasterxml.jackson.annotation.JsonSetter(nulls = com.fasterxml.jackson.annotation.Nulls.SKIP)
    private java.time.ZonedDateTime timestamp;

    @com.fasterxml.jackson.annotation.JsonFormat(pattern = "yyyy-MM-dd'T'HH:mm:ssVV")
    public java.time.ZonedDateTime getTimestamp() {
        return timestamp;
    }

    @com.fasterxml.jackson.annotation.JsonFormat(pattern = "yyyy-MM-dd'T'HH:mm:ss[XXX][VV]")
    public void setTimestamp(java.time.ZonedDateTime timestamp) {
        this.timestamp = timestamp;
    }
```

CRD snippet generated from above Java code snippet:
```
              timestamp:
                properties:
                  dateTime:
                    properties:
                      date:
                        properties:
                          year:
                            type: integer
                        type: object
                      time:
                        properties:
                          nano:
                            type: integer
                        type: object
                    type: object
                  offset:
                    properties:
                      id:
                        type: string
                      totalSeconds:
                        type: integer
                    type: object
                  zone:
                    type: object
                type: object
            type: object
```

### Describe the solution you'd like

I'd like for CRD fields of {type=string, format=date-time} to roundtrip from CRD to Java and back to CRD.

baloo42:
I don't think that roundtripping is supported at the moment.

The CRDGenerator for example does not use Jackson's `JsonFormat` annotation at the moment and there is no other way available to configure the `format` property in the JSONSchema.

See https://github.com/fabric8io/kubernetes-client/issues/5859 to get an overview about the missing JSONSchema properties.

andreaTP:
From a historical POV when we started the `java-generator` we wished to have a full-blown roundtrip with the `crd-generator` working.
We stopped aiming for it due to the amount of work and the slightly different nuances of the 2 generators.

[...]

PR #5885, opened 2024-04-12
Add crd-generator and java-generator support for Format annotation

matteriben:
## Type of change

## Checklist

### DIFF
--- a/doc/CRD-generator.md
+++ b/doc/CRD-generator.md
@@ -231,6 +231,30 @@ The field will have the `maximum` property in the generated CRD, such as:
             type: object
 ```
 
+### io.fabric8.generator.annotation.Format
+
+If a field or one of its accessors is annotated with `io.fabric8.generator.annotation.Format`
+
+```java
+public class ExampleSpec {
+  @Format("password")
+  String someValue;
+}
+```
+
+The field will have the `format` property in the generated CRD, such as:
+
+```yaml
+          spec:
+            properties:
+              someValue:
+                format: password
+                type: string
+            required:
+            - someValue
+            type: object
+```
+
 ### io.fabric8.generator.annotation.Pattern
 
 If a field or one of its accessors is annotated with `io.fabric8.generator.annotation.Pattern`
--- a/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
+++ b/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
@@ -249,6 +249,12 @@ public class JObject extends AbstractJSONSchema2Pojo implements JObjectExtraAnno
                   new Name("io.fabric8.generator.annotation.Min"),
                   new DoubleLiteralExpr(prop.getMinimum())));
         }
+        if (prop.getFormat() != null) {
+          objField.addAnnotation(
+              new SingleMemberAnnotationExpr(
+                  new Name("io.fabric8.generator.annotation.Format"),
+                  new StringLiteralExpr(StringEscapeUtils.escapeJava(prop.getFormat()))));
+        }
         if (prop.getPattern() != null) {
           objField.addAnnotation(
               new SingleMemberAnnotationExpr(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 * Fix #5867: (crd-generator) Imply schemaFrom via JsonFormat shape (SchemaFrom takes precedence)
 * Fix #5867: (java-generator) Add JsonFormat shape to date-time
 * Fix #5954: (crd-generator) Sort required properties to ensure deterministic output
+* Fix #5867: (crd-generator) Add support to define `format` from `@Format` annotation
+* Fix #5867: (java-generator) Add support to define `@Format` annotation from `format`
 
 #### Dependency Upgrade
 
--- a/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
+++ b/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java
@@ -119,6 +119,7 @@ public abstract class AbstractJsonSchema<T, B> {
   public static final String ANNOTATION_DEFAULT = "io.fabric8.generator.annotation.Default";
   public static final String ANNOTATION_MIN = "io.fabric8.generator.annotation.Min";
   public static final String ANNOTATION_MAX = "io.fabric8.generator.annotation.Max";
+  public static final String ANNOTATION_FORMAT = "io.fabric8.generator.annotation.Format";
   public static final String ANNOTATION_PATTERN = "io.fabric8.generator.annotation.Pattern";
   public static final String ANNOTATION_NULLABLE = "io.fabric8.generator.annotation.Nullable";
   public static final String ANNOTATION_REQUIRED = "io.fabric8.generator.annotation.Required";
@@ -177,6 +178,7 @@ public abstract class AbstractJsonSchema<T, B> {
     final String defaultValue;
     final Double min;
     final Double max;
+    final String format;
     final String pattern;
     final boolean nullable;
     final boolean required;
@@ -187,6 +189,7 @@ public abstract class AbstractJsonSchema<T, B> {
       defaultValue = null;
       min = null;
       max = null;
+      format = null;
       pattern = null;
       nullable = false;
       required = false;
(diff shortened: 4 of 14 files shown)


java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
```diff
@@ -231,6 +231,14 @@ public GeneratorResult generateJava() {
           fieldSetter.addAnnotation(new SingleMemberAnnotationExpr(
               new Name("com.fasterxml.jackson.annotation.JsonFormat"),
               new NameExpr("pattern = \"" + config.getDeserDatetimeFormat() + "\"")));
+          objField.addAnnotation(
+              new SingleMemberAnnotationExpr(
+                  new Name("io.fabric8.crd.generator.annotation.SchemaFrom"),
```
andreaTP, line 236: I would like to avoid engineering workarounds if possible

andreaTP:
related #5881

baloo42 (changes requested):
The implementation looks good to me regarding the CRD-Generator. It also fits to #5881. (My prototype looks the same but I hold it back because I prefer `@Schema`  to avoid annotation explosion). But it's possible to implement both: If `format` is defined by `@Format` and `@Schema`, the format value from `@Schema` would win.

I only would suggest to make it more clear that `format` can be used for other formats than `date-time`.

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
matteriben: @andreaTP

With respect to https://github.com/fabric8io/kubernetes-client/issues/5866 (bug in crd-generator) I agree this appears to be a workaround.

However, with respect to https://github.com/fabric8io/kubernetes-client/issues/5867 (crd-generator doesn't understand date-format) I don't consider this a workaround at all.

doc/CRD-generator.md
```diff
@@ -231,6 +231,30 @@ The field will have the `maximum` property in the generated CRD, such as:
...
+}
+```
+
+The field will have the `format` property in the generated CRD, such as:
+
+```yaml
+          spec:
+            properties:
+              someValue:
+                format: "date-time"
```
baloo42, line 251: nitpick: 
```suggestion
                format: "password"
```

baloo42 (changes requested):
Beside of the nitpick in the docs it looks now good to me regarding the CRD-Generator

andreaTP (approved):
Although I don't see much value for the `java-generator` part, I'm fine with those changes, thanks 👍

rohanKanojia (approved):

baloo42 (approved):

manusa (approved):
LGTM, thx!

shawkins:
@manusa @andreaTP ideally we would not add support a custom Format annotation.

baloo42:
> @manusa @andreaTP ideally we would not add support a custom Format annotation.

@shawkins Does this mean you would prefer a general annotation like described in #5881?

andreaTP:
> ideally we would not add support a custom Format annotation.

I see the use case for this when using the CRD Generator, at the moment you cannot inject a custom format.
Is just another "free form string field" that accepts any value.

shawkins:
> I see the use case for this when using the CRD Generator, at the moment you cannot inject a custom format.
> Is just another "free form string field" that accepts any value.

There is built-in support for the format within Jackson, which is now wired up into the prospective changes in the next state pr. There may be some usability limitations with that, so it's possible that a Format annotation will still make sense. If not we'll just deprecate it in 7 like some of the other annotations.

@matteriben and others - the current progress on what I'd like to see as the 7.0 state of the crd generator is at 
https://github.com/fabric8io/kubernetes-client/compare/main...shawkins:kubernetes-client:iss5866-b?expand=1

Ideally we could start to minimize the 6.x version changes, or at least get a first class branch going for the next state so that it can be targeted with pull requests for support like this as well.

baloo42:
> > I see the use case for this when using the CRD Generator, at the moment you cannot inject a custom format.
> > Is just another "free form string field" that accepts any value.
> 
> There is built-in support for the format within Jackson, which is now wired up into the prospective changes in my next state branch. There may be some usability limitations with that, so it's possible that a Format annotation will still make sense. If not we'll just deprecate it in 7 like some of the other annotations.
> 
> @matteriben and others - the current progress on what I'd like to see as the 7.0 state of the crd generator is at https://github.com/fabric8io/kubernetes-client/compare/main...shawkins:kubernetes-client:iss5866-b?expand=1
> 
> Ideally we could start to minimize the 6.x version changes, or at least get a first class branch going for the next state so that it can be targeted with pull requests for support like this as well.

After looking at the planned changes it doesn't make sense to me to add features / annotations to v6 of the CRD-Generator anymore. Any new feature would have to be replaced by a new implementation based on Jackson.

I also see, that the crd-generator-apt module will be removed. Do you plan a replacement?

Can I help you with v7 instead? e.g. with some feedback on comments or implementing some of the other missing features. I would suggest to create a crd-generator-v7-development branch in the kubernetes-client project.

shawkins:
> I also see, that the crd-generator-apt module will be removed. Do you plan a replacement?

Yes, we have discussed adding a maven plugin and a gradle plugin for non-quarkus / josdk  based usage. Maybe something utilizing jandax for discovery?

> Can I help you with v7 instead? e.g. with some feedback on comments or implementing some of the other missing features. I would suggest to create a crd-generator-v7-development branch in the kubernetes-client project.

Absolutely. I'll create that branch, and we can how it can be adapted to your needs.

I had just sent a message to other team members desribing the state of things, let me know what you may want to work on and we can sync up our efforts - 

Where things stand:

- there are a couple of hacks, noted with TODOs, due to choices we made about where to pick up annotations from. Nothing too crazy.
- the usage of CustomResource has been made optional
- the core logic is no longer static. It's not quite wired all the way up to accept an ObjectMapper, but that shouldn't be too hard
- parallelism has not been wired back in. It will be straight-forward to add that back if needed (either separate ResolvingContexts, or make sure that it's thread-safe)
- there are some generation differences, there are three tests failing due to: 
  - The required list was previously unordered, now it is.
  - The printer column type is taken directly from the property - but in the test it's expecting a string instead of an array type for one of the columns. What is the expectation here?
  - It now better understands the anygetter / setter logic, so more properties are marked as preserve unknown
- all sundrio has been removed including the apt module
  - the crd-generator/test tests have not yet been rewired to work. In the interim I could just manually perform generation.

Next steps:

- create a formal development branch? We'll probably want the flurry of work that is being done to target both 6.x and the next state.
- maven and gradle plugins
- PR to retarget the quarkus logic once we can allow for the objectmapper to be passed in

Related:

- change the default time related serialization to a string, rather than number / array

shawkins:
@matteriben @baloo42 @andreaTP here's an upstream development branch https://github.com/fabric8io/kubernetes-client/tree/crd-generator-v7-development

baloo42:
Thanks for the detailed answer. I think we should move the further discussion out of this PR:

https://github.com/fabric8io/kubernetes-client/discussions/5942

manusa:
So let's merge this one then??

shawkins:
> > So let's merge this one then??
> 
> @shawkins, sorry, it was unclear but the question was specifically for you. It's unclear to me if this should be moved forward as is or not given your previous comments.

Let's have @baloo42 and @matteriben check if their use cases are satisfied by the v2 logic already. If so, then we can close this pr. If not, then we'll probably want to retarget to v2, rather than introducing additional v1 support.

matteriben:
> Let's have @baloo42 and @matteriben check if their use cases are satisfied by the v2 logic already. If so, then we can close this pr. If not, then we'll probably want to retarget to v2, rather than introducing additional v1 support.

My use case does appear to be handled as desired in the v2 branch.

manusa:
OK, so let's try to merge #5949 (#5948) first and then add the functionality on top

baloo42:
After testing v2 it seems like the new implementation covers a lot more cases for format but unfortunatly not all.

Not covered:

- ipv4
- ipv6
- email
- mac
- ...

But is that worth to introduce an own annotation? Maybe we can cover that with a single `@Schema` annotation... (#5881)

shawkins:
@baloo42 in addition to any general proposal, it would help to capture where each of those could have additional support in jackson, or with the v2 generator as well. Anywhere the type alone is sufficient, we should easily cover.

baloo42:
> @baloo42 in addition to any general proposal, it would help to capture where each of those could have additional support in jackson, or with the v2 generator as well. Anywhere the type alone is sufficient, we should easily cover.

If I understand you correct, you mean that e.g. Inet4Address could be mapped to:

```
type: string
format: ipv4
```

I think this can be a solution but we must be aware of that Jackson contains sometimes some overriding logic with a shape:

https://github.com/FasterXML/jackson-databind/blob/master/src/main/java/tools/jackson/databind/ser/jdk/InetAddressSerializer.java#L41

This might require some more code but I don't expect too much of such cases, so I'm fine with that strategy.

shawkins:
> I think this can be a solution but we must be aware of that Jackson contains sometimes some overriding logic with a shape:

Sure, ideally all of this should be considered in the serializer. The InetAddressSerializer should be overriding acceptJsonFormatVisitor and passing the relevant JsonValueFormat - this is what DateTimeSerializer is doing. 

We can of course add handling on our side for it, but I'd prefer to push things upstream where possible.

metacosm (approved):
LGTM

Issue #5850, opened 2024-04-01
CRDGenerator: Replace assertions with ApprovalTests


baloo42:
### Is your task related to a problem? Please describe

At the moment we have implemented some tests with self-developed assertions.

### Describe the solution you'd like

Those tests can be replaced by [ApprovalTests](https://github.com/approvals/ApprovalTests.Java).

This test approach is also used for the Java-Generator and requires less self developed code.

PR #5846, opened 2024-03-30
fix(crd-generator): Fail if multiple versions are marked as stored

baloo42:
## Description

Fixes a bug where the CRDGenerator allows generating CRDs with multiple versions marked as stored. 

Additional changes:
- Extract some assertions to a seperate class (reusing is planned in the next PRs)
- Allow debugging decorator sequence by changing the log level to trace
  (refer to [simplelogger.properties](https://github.com/fabric8io/kubernetes-client/blob/main/crd-generator/api/src/test/resources/simplelogger.properties#L22) during test runs)

Fixes #5845 

## Type of change

 - [x] Bug fix (non-breaking change which fixes an issue)
 - [ ] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [ ] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [x] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [x] I Added [CHANGELOG](https://github.com/fabric8io/kubernetes-client/blob/main/CHANGELOG.md) entry regarding this change
 - [x] I have implemented unit tests to cover my changes
 - [x] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/main/doc/CHEATSHEET.md) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [ ] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

Issue #5794, opened 2024-03-11
CRDGenerator: Allow to configure conversion webhooks


baloo42:
### Is your enhancement related to a problem? Please describe

At the moment there is no easy way to configure conversion webhooks for the generated CRDs:

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#configure-customresourcedefinition-to-use-conversion-webhooks

### Describe the solution you'd like

A solution could be to create one or more additional annotations which allows a user to configure the webhook:

Annotation `@WebhookConversion`:
```java
/**
 * Decorates the CRD with the "Webhook" conversion strategy.
 * <p>
 * If this strategy is used, the Kubernetes API server will request an external webhook to do the conversion.
 * </p>
 * <p>
 * Note that only one strategy can be used at the same time for a single CRD.
 * If the "Webhook" strategy doesn't fit and no field has changed, use the {@link NoneConversion} strategy.
 * </p>
 *
 * @see <a href=
 *      "https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#webhook-conversion">
 *      Kubernetes Docs - CRD Versioning - Webhook Conversion
 *      </a>
 */
@Target(ElementType.TYPE)
@Retention(RetentionPolicy.RUNTIME)
public @interface WebhookConversion {
  /**
   * The name of this strategy.
   */
  String NAME = "Webhook";

  /**
   * The ConversionReviewVersions.
   * <p>
   * ConversionReviewVersions is an ordered list of preferred `ConversionReview` versions the Webhook expects.
   * The API server will use the first version in the list which it supports. If none of the versions specified
   * in this list are supported by API server, conversion will fail for the custom resource. If a persisted Webhook
   * configuration specifies allowed versions and does not include any versions known to the API Server,
   * calls to the webhook will fail.
   * </p>
   *
   * @return the conversion review versions.
   */
  String[] versions();

  /**
   * The location of the webhook, in standard URL form (`scheme://host:port/path`).
   * Exactly one of `url` or `svcNamespace` / `svcName` must be specified.
   *
   * @return the URL of the webhook.
   * @see <a href=
   *      "https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#url">
   *      Kubernetes Docs - CRD Versioning - Webhook Conversion - URL
   *      </a>
   */
  String url() default "";

[...]

PR #5896, opened 2024-04-15
feat(crd-generator): Allow to configure CRDGenerator if running as annotation processor, add approval and compiler tests

baloo42:
## Description

- Allow to configure CRDGenerator if running as annotation processor:
  - Read options from compiler options and crd-generator.properties
  - Allow to configure...
    - parallel generation
    - CRD versions (v1, v1beta1)
    - outputPath
    - header in output files
    - describeTypes (debug output)
    - additional labels
    - additional annotations
- Add approval tests to replace custom assertions for some kind of approval tests 
  #5850
- Add compiler tests with approvals for E2E tests (includes parallel testing)
  see https://github.com/fabric8io/kubernetes-client/pull/5846#discussion_r1546157571
  

## Type of change

## Checklist


baloo42:
TBD:

- I used [Approval Tests](https://github.com/approvals/ApprovalTests.Java) in two cases: In unit tests of the api module and  for the compiler tests in the apt module. Please review the setup in detail as I haven't used it before.
- There are now two modules which test the CustomResourceAnnotationProcessor in an own maven module. 
  This approach was really error prone to me and that's the main reason why I wrote the compiler integration tests with google's [compile-testing](https://github.com/google/compile-testing). In my opinion we can remove the `test` and `test-parallel` module. What do you think?
- There are some public static final flags: Are they required? e.g. `DESCRIBE_TYPE_DEFS`
- There are now some duplicate unit tests because those cases are also tested by the approval tests. Should I remove them in this PR or do you want me to wait with the deletion?

andreaTP:
I skimmed through it and the changes look to be going in a nice direction, but this is massive, either you manage to chunk it into more digestible bites or it will take some time to properly review.

cc. @shawkins asking for help here 🆘

baloo42:
A third of the changes come from the copy of the test module to the new test-parallel module. (It's exactly the same content). *So the first question is, do we want to keep both test modules?* ...if we have an alternative:

I don't like both because it requires a lot (test case) duplications, but it's some kind of a real end-2-end test and it was there before I started contributing. My suggestion would be to replace those both modules by appropriate "compile-testing" tests in the apt module. With a small abstraction we can build easily tests which run always with parallel mode on and off (and I think we can also run them in parallel via failsafe plugin because they are fully isolated).

Regarding the options: 
Implementing more apt processor options requires tests. We would require a lot of modules to test them if we don't change the test strategy. And in the end, I need it for #5794 because here we have to implement a way to configure a ca certificate.

I'm happy to hear any further suggestions or guidence ;)

baloo42:
Can probably be closed. See https://github.com/fabric8io/kubernetes-client/pull/5885#issuecomment-2073002518

Issue #5948, opened 2024-04-24
[EPIC] CRD-Generator v2


shawkins:
### Discussed in https://github.com/fabric8io/kubernetes-client/discussions/5942

We've decided to commit a v2 implementation of the crd generator along side the existing one so that it can be part of fabric8 6.

#### Tasks
- https://github.com/fabric8io/kubernetes-client/issues/5950
- https://github.com/fabric8io/kubernetes-client/issues/5951
- https://github.com/fabric8io/kubernetes-client/issues/5952
- https://github.com/fabric8io/kubernetes-client/issues/5944
- https://github.com/fabric8io/kubernetes-client/issues/5945

manusa:
closing since I mistakenly duplicated it in #6616 which is no more up to date.

PR #5943, opened 2024-04-23
chore(crd-generator): Add approval tests for crd-generator

baloo42:
## Description

Add approval tests for crd-generator

Fixes #5850

## Type of change
 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [ ] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [x] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [x] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [x] I have implemented unit tests to cover my changes
 - [x] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/main/doc/CHEATSHEET.md) accordingly
 - [x] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report

PR #5949, opened 2024-04-24
feature: add a v2 crd generator based on jackson

shawkins:
## Description
closes: #5948

also removes v1beta1 from the new code

This is not yet ready to merge. For the tests to pass we need to:
- confirm the JsonFormat shape behavior from https://github.com/fabric8io/kubernetes-client/discussions/5942#discussioncomment-9214458 - for now the types were changed to string
- The required list was previously unordered, now it is. @andreaTP are there any objections to making that the new expected behavior?
- The printer column type is taken directly from the property - but in the test it's expecting a string instead of an array type for one of the columns. What is the expectation here - is string required, or was that just a limitation of the previous logic? cc @andreaTP 
- It now better understands the anygetter / setter logic, so more properties are marked as preserve unknown causing another test to fail. Thinking about this more, this one is not clear cut. This is coming by the way of core model objects, which we include additionalProperties on as a way of letting them be foward compatible - it's not as a way to indicate what the schema actually should be. It could be using the custom PreseveUnknown annotation is the way to go here, rather than detecting the situation ourselves. cc @andreaTP 

I'll add a changelog entry as well once we're ready.

cc @matteriben @baloo42

## Type of change

## Checklist


shawkins:
I went with crdv2 as the package name - but that can be changed of course.

baloo42:
> The required list was previously unordered, now it is.
I think we should handle this as a bug for the v1 generator and also fix it. If not, the new approval tests (#5943) would also fail sometimes. The outcome should be always deterministic.

shawkins:
@baloo42 I've updated the test for v2 to keep this pr moving forward.

@matteriben based upon https://github.com/fabric8io/kubernetes-client/discussions/5942#discussioncomment-9217399 it sounds like we're moving forward with leaving the type as string for those shape tests, which puts us in a similar situation as sorting the required list - the v1 implementation should be updated to match.

That just leaves two differences (that we currently know of) to account for.

andreaTP:
> The required list was previously unordered, now it is.

This is a feature, great!

> was that just a limitation of the previous logic?

IIRC that's it, in any case printer columns are just for human consumption, we can solve any issue at a later stage without a significant risk of breaking user code badly.

> It could be using the custom PreseveUnknown annotation is the way to go here, rather than detecting the situation ourselves.

Looking at how ppl are using things I would not be surprised if we take the opinionated decision to add `x-kubernetes-preserve-unknown` everywhere by default ...
That said, I would not forcefully try to follow the current detection logic and just be sure we document the behavior appropriately.

Currently on the phone and I'll be out until Mon, but let me know if any additional input from me is needed to progress here and I'll try to answer asap.

shawkins:
> IIRC that's it, in any case printer columns are just for human consumption, we can solve any issue at a later stage without a significant risk of breaking user code badly.

Ok, to keep this pr moving forward I'll update the results here to expect the actual type.

> That said, I would not forcefully try to follow the current detection logic and just be sure we document the behavior appropriately.

The logic here is a little broader - so I'm wondering if it needs to be more limited. The concern is that the object model's intent with using additional properties is to increase compatibility - not that the real schema of ObjectMeta etc. actually calls for preserve unknown. I'm thinking that we'll either need to just look for the PreserveUnknownFields annotation or let it be user configurable in some other way to elect whether anyGetter / anySetter means preserve unknown fields.

shawkins:
Marking as ready for review as all current tests pass. Let me know what needs addressed before this gets committed. This also wires up parallel execution (should likely now default to true)  #5952 and exposes two new options for the crd generator:
- the ability to supply an ObjectMapper (this should be the same instance that is passed to the KubernetesSerialization. If we want to add more of backdoor for exposing the ObjectMapper from an instance of KubernetesSerialization, then we can pass the KubernetesSerialization instead.
- the ability to control whether we implicitly treat anyGetter / anySetter as preserve unknown fields - we will probably want a better story here eventually.

Recap of some other changes:
- it works at the property level, rather than the field level, so some crds may have fields removed - but those weren't actually being used. This is also the reason test code made fields public or used the lombok Data annotation - Jackson doesn't consider a private field without accessors serializable.
- more support for format and pattern based upon what is already supported by jacksonschema. There is still more that could be done here around their validation extension logic
- RawExtension and HasMetadata interfaces will be mapped to embedded resources
- the logic resolve the jackson schema for nested classes only once, then we'll reuse that schema from the ResolvingContext
- the abstraction to support multiple CRD versions has been left in, but could be removed if we want to revisit it later when needed
- also removed the decorators as we can do the same thing more directly with builders.

@metacosm this should be good now to test out with quarkus

cc @matteriben @baloo42 @manusa @andreaTP

crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/AbstractCustomResourceHandler.java
```diff
@@ -0,0 +1,32 @@
...
+
+import io.fabric8.kubernetes.api.model.HasMetadata;
+
+import java.util.stream.Stream;
+
+/**
+ * This class encapsulates the common behavior between different CRD generation logic. The
+ * intent is that each CRD spec version is implemented as a sub-class of this one.
+ */
+public abstract class AbstractCustomResourceHandler {
```
manusa, line 26: Is there any reason why this is a class instead of an interface?

It's my understanding that we can have a  `interface CustomResourceHandler` and then an implementation `V1CustomResourceHandler` or `CustomResourceHandlerV1`, right?

crd-generator/api-v2/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>crd-generator-parent</artifactId>
+    <groupId>io.fabric8</groupId>
+    <version>6.13-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>crd-generator-api-v2</artifactId>
```
manusa, line 29: Note that this can potentially be implemented using Java 11 source (in case this simplifies implementations)

crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/CRDGenerator.java
```diff
@@ -0,0 +1,281 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(CRDGenerator.class);
+  private final Map<String, AbstractCustomResourceHandler> handlers = new HashMap<>(2);
+  private CRDOutput<? extends OutputStream> output;
+  private boolean parallel;
+  private boolean implicitPreserveUnknownFields;
+  private ObjectMapper objectMapper;
+  private Map<String, CustomResourceInfo> infos;
+
+  // TODO: why not rely on the standard fabric8 yaml mapping
+  public static final ObjectMapper YAML_MAPPER = JsonMapper.builder(new YAMLFactory()
```
manusa, line 64: Yes, why not?

crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/CRDInfo.java
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  /**
+   * @deprecated Use {@link #getCrdSpecVersion()} instead
+   * @return the CRD spec version
+   */
+  @Deprecated
+  public String getVersion() {
+    return getCrdSpecVersion();
+  }
```
manusa, line 40: Should we add a deprecated method to the new API?

crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/AbstractCustomResourceHandler.java
shawkins: > Is there any reason why this is a class instead of an interface?

Not really. It used to hold common state and have a common method for dealing with printer columns. My intent was to add back the method for printer columns, but didn't bother yet.

> It's my understanding that we can have a interface CustomResourceHandler and then an implementation V1CustomResourceHandler or CustomResourceHandlerV1, right?

Correct, it seemed fine to leave in the abstraction layer, used here and for AbstractJsonSchema, but it's not really needed at the moment.

crd-generator/api-v2/pom.xml
shawkins: I'm not aware of anything that would change off hand.

crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/CRDGenerator.java
shawkins: A comment was lost from @metacosm about how this format seemed different. Haven't looked at it more deeply.

crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/CRDInfo.java
metacosm: Probably not but I think it's just moving in stages :)

crd-generator/pom.xml
```diff
@@ -33,17 +33,7 @@
...
-    <profile>
-      <id>tests</id>
-      <activation>
-        <jdk>(,17]</jdk>
-      </activation>
-      <modules>
-        <module>test</module>
-      </modules>
-    </profile>
-  </profiles>
```
manusa, line 48: :+1:

crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/CRDInfo.java
shawkins: No, I'll remove it.

crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/CRDGenerator.java
shawkins: The two main changes I see are the standard generator includes the explicit start ---, and it quotes string values. The ordering changes aren't really needed as they should be handled by the construction of the crd object.

I agree that we don't need explicit start, the question is does it matter if there's additional quoting? It shouldn't but we can account for one or both of these differences with an KubernetesSerialization.asYaml method that takes some style options - we are just creating the Dump / DumpSettings on every call anyway.

baloo42 (commented):
Tested it with a few internal projects and looks very good so far.

Beside of the printer columns and the ordering of the required properties list I haven't found any differences to v1.

shawkins:
> Tested it with a few internal projects and looks very good so far.
> 
> Beside of the printer columns and the ordering of the required properties list I haven't found any differences to v1.

@baloo42 thank you for the code suggestions and double checking the output.

crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/CRDGenerator.java
shawkins: Looking at it more, I don't think we should add yaml options. They really aren't needed for this case either. We could just remove this yaml mapper - the only practical issue is that string comparison of the existing crd output yaml won't work.

baloo42:
More general feedback:

#### x-kubernetes-embedded-resource
> RawExtension and HasMetadata interfaces will be mapped to embedded resources

:+1: Nice

#### Printer Columns
> The printer column type is taken directly from the property - but in the test it's expecting a string instead of an array type for one of the columns. What is the expectation here - is string required, or was that just a limitation of the previous logic? 

Kubernetes allows for printer columns only the following types:
boolean,date,integer,number,string

I would suggest to silently filter printer colums on other types. At the moment it produces an invalid CRD.

"Arrays" are only supported if we generate something like the following, which would be an own story:

```
JSONPath: ".spec.hardware.devices.dynamicDirectPathIODevices[*].deviceID"
type: string
```
(the target property must by one of the above types)

See also upcoming "FieldSelectors" feature:
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-selectors
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#type

#### Preserve Unknown Fields

I like the current behaviour of v2 regarding this topic. It behaves as I would expect it.

> if we take the opinionated decision to add x-kubernetes-preserve-unknown everywhere by default ...

In my opinion using x-kubernetes-preserve-unknown everywhere could be "dangerous" because Kubernetes behaves differently. Removed fields in a CR won't get cleaned up (see [control-pruning](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#controlling-pruning))

shawkins:
> I would suggest to silently filter printer colums on other types. At the moment it produces an invalid CRD.

I'd be more in favor of producing a hard error or emitting a warning.

crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/CRDInfo.java
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.fabric8.crdv2.generator;
+
+public class CRDInfo {
```
metacosm, line 18: This is a no-go for QOSDK: this class is missing the `getDependentClassNames` support we need.

metacosm (changes requested):

crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/CRDInfo.java
shawkins: I can add this back, but just to double check - is there a quarkus utility for detecting the transitive graph of a given class?

shawkins: Also - if this is needed, is it okay to just provide the dependent classes across everything passed into the crd generator? The ResolvingContext.uriToJacksonSchema values will contain all of the introspected classes, but it's not tracked by CRD.

shawkins:
> > I would suggest to silently filter printer colums on other types. At the moment it produces an invalid CRD.
> 
> I'd be more in favor of producing a hard error or emitting a warning.

Added the warning and extracted common printer column handling.

crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/CRDInfo.java
metacosm: > I can add this back, but just to double check - is there a quarkus utility for detecting the transitive graph of a given class?

I'm not sure if there's such a thing… I'll ask but, iirc, there wasn't one when I implemented this feature.

metacosm: > > I can add this back, but just to double check - is there a quarkus utility for detecting the transitive graph of a given class?
> 
> I'm not sure if there's such a thing… I'll ask but, iirc, there wasn't one when I implemented this feature.

So I checked, and, no, there still isn't such a thing in Quarkus

metacosm: > Also - if this is needed, is it okay to just provide the dependent classes across everything passed into the crd generator? The ResolvingContext.uriToJacksonSchema values will contain all of the introspected classes, but it's not tracked by CRD.

I'm not sure, I would need to look into things in deeper details. Obviously, this would probably mean more work on my side. Additionally, I'd rather not expose these internals to outside code, if possible. The `getDependentClassNames` method was implementation-agnostic so I think it would be useful to restore the functionality if possible.

shawkins: > So I checked, and, no, there still isn't such a thing in Quarkus

Ok, and do you need it tracked per crd, or is across everything good enough?

shawkins: @metacosm check now, it should map to what you were using before.

crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/CRDGenerator.java
shawkins: I can't find any reference going back in the code for that usage of a yaml mapper. The combination of the comment and the start marker are still legal yaml, so I'm not sure what the issue could would have been.

crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/CRDInfo.java
metacosm: > @metacosm check now, it should map to what you were using before.

It doesn't work completely as expected. In particular, the previous version was overeager and returned too many things (like the `Group` annotation) but the new version doesn't return enough, in particular when there's a class hierarchy being involved (e.g. if a CR extends an intermediate class instead of extending CustomResource directly, the new version won't provide the intermediate class in `getDependentClassNames` even though changes to that class could potentially impact the CRD).

shawkins: @metacosm check now. It's not filtering java level stuff, but it will get the interfaces and super classes as well.

metacosm: Yep, this seems to work nicely now. Will now do a full CI run with this PR to check.

crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/CRDGenerator.java
metacosm: Can confirm that the start marker is causing issues, in particular with OLM validation, at least.  I haven't tried applying the CRDs to a cluster, yet, though.

metacosm: Applying the CRD to the cluster seems to work OK, as expected.

shawkins: Thanks for checking @metacosm - the explicit start has been removed

metacosm: Stripping the marker solved the issue as expected. Thank you! 
For reference, here's the successful QOSDK build with the changes from this PR: https://github.com/quarkiverse/quarkus-operator-sdk/actions/runs/8903462629/job/24459014817

shawkins: Great, thank you for testing. The only other thing that could change is supplying the ObjectMapper / KuberenetesSerialization so that we can pick up any changes the user may have made to the mapper configuration

cc @manusa

manusa:
https://github.com/fabric8io/kubernetes-client/pull/5949#discussion_r1586456252

Is this ready to be merged then?

shawkins:
> Is this ready to be merged then?

I think so. We can certianly refine things from here. There's quite a few follow-on tasks that have already been logged and more that are mentioned in the PR description.

metacosm (approved):

manusa (approved):
LGTM, thx!

PR #5956, opened 2024-04-25
replace

wind57:


shawkins:
Can you be more explicit - are you experiencing a performance issue related to biased locking?

rohanKanojia:
@wind57 : We really appreciate the fact that you're helping us out by providing patches. However, it would be nice to provide the reviewers with some additional context about these changes.

It's a bit hard for us to infer required context just by looking at the changes. Unfortunately PR description and commit message doesn't seem to help either.

shawkins:
> first is biased locking deprecation. I have not had any issues with it as it is now, but the general idea is that biased locking being deprecated can potentially induce performance problems as described in the JEP. At my work place, we have already discovered such tiny performance gaps and are going though the process of moving to locks. 

Generally speaking explicit usage of lock is still more expensive than even unbiased usage of the synchronized keyword for low contention scenarios. In most cases where fabric8 is using the synchronized keyword it is to ensure correctness and is not critical to high-contention performance.

> The second argument, is that not in the so distance future, when project valhalla is going to be a thing, the "identity" of an object, in general, is going to be seen as an anti-pattern. As a matter of fact the current proposal is to have something like : Objects::instanceWithoutIdentity or something like that, where an instance is provided without an "identity" (== mark header). synchronized obviously break this pattern, so its a good habit to grow, in the long run, imho.

I haven't followed the proposals closely, but from what I can gather this would only apply to value objects, not all objects. We are in this case not synchronizing on a primitive wrapper or optional. Or are you concerned that at some point this class would get turned into a value class at which point synchronization won't work?

> Again, if you don't see any value in this, I can close it, I was just trying to get a feeling of where you are regarding it. Thank you.

It's best to start with the issue first so that we can have these discussions there, then open a PR. I'm not saying we can't or won't switch to explicit locks - but it's best to properly identify why and then we can assess what parts of the code need addressed.

PR #5632, opened 2023-12-04
PoC: MockWebServer based on Vert.x

manusa:
## Description

**WIP** Still working on this, but it seems that there is a viable solution with not that many breaking changes. However, it seems reasonable that this will deserve a major version change, or some tooling to help users upgrade.

Replaces the `io.fabric8:mockwebserver` module backend from OkHTTP to Vert.x.

Relates to:
- #2632
- https://github.com/quarkusio/quarkus/pull/14662
- #2764 
- #4193
- #4290
- #5449
- #5485
- #5613 
- https://github.com/quarkusio/quarkus/issues/29520

## Type of change

 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [x] Feature (non-breaking change which adds functionality)
 - [x] Breaking change (fix or feature that would cause existing functionality to change
 - [ ] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [x] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [ ] I Added [CHANGELOG](https://github.com/fabric8io/kubernetes-client/blob/main/CHANGELOG.md) entry regarding this change
 - [ ] I have implemented unit tests to cover my changes
 - [ ] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/main/doc/CHEATSHEET.md) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [ ] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

Issue #4164, opened 2022-05-22
Use one resource annotation


shawkins:
### Is your enhancement related to a problem? Please describe

We have separate annotations for things like group, kind, plural, singular, and version.

### Describe the solution you'd like

For those that don't auto-generate their model, it would be better if all of the single valued annotations could easily be combined to a single annotation:

Resource(group="x", kind="y" ...)

You could even include builder=Something.class, and list=Something.class - if we think that the conventions may not hold at some point.

manusa:
This would make the code look more organized, and reduce the code-base by a few hundred lines.

However, won't this break our current model backwards compatibility support (i.e. consuming a model from v5.9 in v5.12)?

shawkins:
> However, won't this break our current model backwards compatibility support (i.e. consuming a model from v5.9 in v5.12)?

You'd have to keep and deprecate the old style for a couple of releases.

shawkins:
A full proposal would look like:

```
public @interface Resource {
  
  String group();
  
  String kind() default "";  // empty would default to class name
  
  String plural() default ""; // empty would default to plualize kind
  
  boolean namespaced() default true;
  
  String version();
}
```
This would also mean deprecating the Namespaced interface - it won't be needed once this annotation is required.

The only sticking point are how this meshes with the crd related annotations - in particular Version can accept storage and served for that purpose.  The simplest thing is that you could just flatten these attributes into the Resource annotation as well.   The advantages of doing this are:
- is much more straight-forward for those starting with the object model
- is more straight-forward for annotation scanning.  As of now you need to scan for Version - as that's the only one that's effectively required
- generated code will be a little simpler

On the other hand it will be a fair amount of work and introduce some confusion while both sets of annotations are supported.

On versions, an additional thought - for multi-version support from a single object model, we'd need to support something like:

client.resources(Something.class, "v2") - under the covers we'd create a ResourceDefinitionContext off of Something, but change the version to v2.  I'm not sure if this is a common enough use case to add another entry method.

manusa:
> A full proposal would look like:
> 
> ```
> public @interface Resource {
>   
>   String group();
>   
>   String kind() default "";  // empty would default to class name
>   
>   String plural() default ""; // empty would default to plualize kind
>   
>   boolean namespaced() default true;
>   
>   String version();
> }
> ```
> 
> This would also mean deprecating the Namespaced interface - it won't be needed once this annotation is required.

[...]

PR #5979, opened 2024-05-03
feat(crd-generator): Add CRD-Generator Maven plugin

### LINKED ISSUE #5944: CRD-Generator v2: Maven Plugin

Because the CRD-Generator annotation processor is now deprecated, we need a replacement.
(see https://github.com/fabric8io/kubernetes-client/discussions/5942)

A CRD-Generator maven plugin could search for a top level annotation (or interface) and generate the CRDs using the CRD-Generator api.

_No response_

### Additional context

- [Jandex](https://smallrye.io/jandex/) could be used to find and index the relevant annotations/classes.
- The following options should be supported:
  - Enable / disable parallel generation
  - CRD versions (v1, v1beta1)  
    (TBD: There are thoughts to drop v1beta1 which makes this option irrelevant)
   - outputPath
   - header in output files
   - Allow to add additional labels
   - Allow to add additional annotations
   - Enable debug output

The also planned CRD-Generator gradle plugin (https://github.com/fabric8io/kubernetes-client/issues/5945) should support the same options.

baloo42:
## Description

Add CRD-Generator Maven plugin (using api-v2).

Fixes #5944 

## Type of change

## Checklist


baloo42:
@manusa Can you have second look on this? Is this now going into the right direction?

The logic to collect and load custom resource classes is now included in an own module "crd-generator-collector" intended to be shared between maven-plugin, gradle-plugin, cli etc.

TBD:

- Do we have to care about Zip Bombs, if we collect custom resource classes from a jar file? (I have added some safeguards but sonar is not recognizing it)
- Do we need reset methods? (the collector classes contain a state)
- Do we want to include the examples in some way? --> https://github.com/baloo42/crd-generator-maven-examples

A prototype for the CLI which is using the same base can be found here: https://github.com/baloo42/kubernetes-client/pull/9

baloo42:
@shawkins, @andreaTP  Maybe you can have a look on this too. 

Marc has mentioned that he is not sure about the final / future purpose of the new module (me too), so I'll try to explain a few thoughts I had.

To use the CRD-Generator from CLI, Maven, Gradle etc. we need in general two main features which are not implemented in the api-v2 module:

- Search / Collect / Filter custom resources classes
- Customizable class loading

The new crd-generator-collector module implements both features via the main component CustomResourceCollector which can be used in the following way:

```java
CustomResourceCollector customResourceCollector = new CustomResourceCollector()
        .withClasspathElements(<additional classpath elements>)
        .withFilesToIndex(<e.g. directory with class files or jar file with class files>)

CustomResourceInfo[] customResourceInfos = customResourceCollector.findCustomResources();

CRDGenerator crdGenerator = new CRDGenerator()
        .customResources(customResourceInfos)
        .inOutputDir(outputDirectory);
        
crdGenerator.detailedGenerate()
```

In conclusion this approach is some kind of "pre-processor" for the CRDGenerator.

Another approach would be to implement it as a facade ("processor"):

```
CRDGeneratorFacade facade = new CRDGeneratorFacade()
        .withClasspathElements(<additional classpath elements>)
        .withFilesToIndex(<e.g. directory with class files or jar file with class files>)
        .inOutputDir(outputDirectory)

facade.detailedGenerate()
```
This approach is nicer from a user perspective but we need to passthrough every option (or use some kind of nested builder pattern).

A third approach would be to implement both features in the api-v2 module itself and make jandex an optional dependency (or allow to extend it by using a service loader).

```
CRDGenerator crdGenerator = new CRDGenerator()
        .withClasspathElements(<additional classpath elements>)
        .withFilesToIndex(<e.g. directory with class files or jar file with class files>)
        .inOutputDir(outputDirectory)

crdGenerator.detailedGenerate()
```

This approach might be the nicest but it would also be the most advanced: the api-v2 module has now to deal somehow with additional dependencies like Jandex and the code base of the api-v2 would grow.

Which approach do you prefer? Other thoughts?

shawkins:
> Which approach do you prefer? Other thoughts?

I'd like to keep the core api as simple as possible - whatever is needed to satisfy Quarkus should be good.

The first option seems good enough. I don't really see much need for coupling between the collector and the crd generator - it doesn't really need to provide CustomResourceInfo, just classes. I mentioned on a pr somewhere that the collector to me just looks like something that you could provide a list of interfaces / base classes, and a list of annotations, and it will just find everything that matches that.

baloo42:
> I'd like to keep the core api as simple as possible - whatever is needed to satisfy Quarkus should be good.
:+1: 

> The first option seems good enough. I don't really see much need for coupling between the collector and the crd generator - it doesn't really need to provide CustomResourceInfo, just classes.

Yes, you are right. Providing only classes could basically do the job but then we cannot implement high level filtering (e.g. by group or version). Do we need high level filtering? If yes: High level filtering could also be implemented in api-v2 so that we can stay by the strategy to return only classes instead of CustomResourceInfo's.

> I mentioned on a pr somewhere that the collector to me just looks like something that you could provide a list of interfaces / base classes, and a list of annotations, and it will just find everything that matches that.

Yes, at the moment the base interface (HasMetadata) and the additional required annotations (Version, Group) to identify a custom resource class are hardcoded. But this can be easily changed to be configurable (with sane defaults) if you think this is useful. Do we need to search for other Interfaces/Annotations?

andreaTP:
Completely agree with what @shawkins said, let's keep the core API as simple as possible.
I don't think that filtering is required, and can be implemented in the tooling when necessary.

shawkins:
> Do we need high level filtering? If yes: High level filtering could also be implemented in api-v2 so that we can stay by the strategy to return only classes instead of CustomResourceInfo's.

Right I'd vote for putting that into the api if needed.

It would probably be good to consider whether having a public constructor on CustomResourceInfo makes sense - that opens up a different usage model for the CRDGenerator that isn't bound to the typical annotations. 

> Do we need to search for other Interfaces/Annotations?

Not yet. I was just highlighting that it doesn't really need to be seen as that specific to CRs.

shawkins:
> Try to generalize the used interfaces / annotations to identify a custom resource

Don't worry about this step just yet, what you currently have should work fine for the first iteration.

crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/CustomResourceCollector.java
```diff
@@ -0,0 +1,158 @@
...
+    List<Class<? extends HasMetadata>> customResourceClasses = customResourcesClassNames.stream()
+        .filter(classNamePredicate)
+        .map(customResourceClassLoader::loadCustomResourceClass)
+        .collect(Collectors.toList());
+
+    log.debug("Found {} custom resource classes after filtering", customResourceClasses.size());
+    return customResourceClasses;
+  }
+
+  public Class<? extends HasMetadata>[] findCustomResourceClasses() {
```
baloo42, line 151: This method is required because the CRDGenerator accepts only an array.

Can we extend the CRDGenerator with an additional method?

```java
  public CRDGenerator customResourceClasses(Collection<Class<? extends HasMetadata>> crClasses) {
    return customResources(crClasses.stream().map(CustomResourceInfo::fromClass).toArray(CustomResourceInfo[]::new));
  }
```

crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/JandexCustomResourceClassScanner.java
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  /**
+   * Creates the base index required to scan for Custom Resource classes.
+   *
+   * @return the base index.
+   *
+   * @see JandexCustomResourceClassScanner#findCustomResourceClasses(IndexView)
+   */
+  private Index createBaseIndex() {
```
baloo42, line 139: This can be optimized if required. 
Existing indices of the fabric8 modules can be used or a specialized index can be created at build time.

Do we need further optimization within this PR?

crd-generator/maven-plugin/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven.plugin-tools</groupId>
+      <artifactId>maven-plugin-annotations</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.fabric8</groupId>
+      <artifactId>crd-generator-collector</artifactId>
+      <version>${project.version}</version>
```
baloo42, line 52: Add managed dependency to parent pom

crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/JandexIndexer.java
```diff
@@ -0,0 +1,230 @@
...
+class JandexIndexer {
+
+  private static final Logger log = LoggerFactory.getLogger(JandexIndexer.class);
+
+  private static final String CLASS_FILE_SUFFIX = ".class";
+  private static final String JAR_FILE_SUFFIX = ".jar";
+
+  private int maxJarEntries = 100000;
+  private long maxBytesReadFromJar = 100000000; // 100 MB
+  private long maxClassFileSize = 1000000; // 1 MB
```
baloo42, line 53: Do we need limits?

crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/JandexCustomResourceClassScanner.java
shawkins: We probably don't need optimization here, but I agree that we should reuse the existing index is possible - since users wil need to modify their pom / gradle file to use these plugins, it doesn't seem out of the question to just document also requiring the jandex plugin be configured as well for projects that want to create crds.

crd-generator/collector/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>crd-generator-collector</artifactId>
+  <name>Fabric8 :: CRD generator :: Collector</name>
+
+  <packaging>jar</packaging>
+
+  <properties>
+    <jandex.version>3.1.7</jandex.version>
```
rohanKanojia, line 35: I see jandex property defined in parent pom. 
https://github.com/fabric8io/kubernetes-client/blob/e3fa9c82a2c5a5ebdbca41334a2181a5b4df1253/pom.xml#L168
Can we reuse this property here ?

baloo42: Seems possible. The jandex maven plugin and core module are released together. 

Should I change the property name to `jandex.version` and move it to "Core versions" section?

rohanKanojia: Sounds good to me

crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/JandexCustomResourceClassScanner.java
baloo42: I'm currently writing some usage docs about the different use cases, including when additional configuration is required. This should help to get an overview from the user perspective.

Some notes on a more technically perspective:

- Jandex is only required to "scan" for Custom Resource classes. If the list of classes is given, no scanning at all is required. (The classloader does the heavy lifting)
- The implementation detects existing indices in the scan targets automatically and prefers those. Only if a scan target does not contain an index, an own in-memory index will be created. This means having an existing index is just optimization, never a requirement.
- Custom Resource classes must be loaded before they can be used by the CRDGenerator itself. Any dependency which is required to load those classes must be in the classpath.
- A scan target must be also in the classpath. But not every classpath element is a scan target.
- Additional indexed classes beside the "base-index" are only required, if the user wants to scan a target with Custom Resource classes which don't implement the CustomResource or HasMetadata interface directly.  In this case, the user must ensure that the used intermediate class is included somehow in the underlying composite index. For example, he can add the dependency which includes the class to the list of scan targets, so that an index is automatically created if needed.

In conclusion, a user doesn't need to know anything about Jandex. Even for the edge case from the last bullet point, the user only has to care about including all "relevant" classes to the list of scan targets and not Jandex specifically.

(But if his project already uses Jandex, the scans will be a little bit faster.)

Keeping this in mind, optimizing the creation of the base index is mostly about performance.
Only if there are other "intermediate" classes like "CustomResource", then we should add them.

crd-generator/maven-plugin/src/test/java/io/fabric8/crd/generator/maven/plugin/ClasspathTypeTest.java
```diff
@@ -0,0 +1,97 @@
...
+import org.mockito.Mockito;
+
+import java.util.Collections;
+import java.util.Set;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.when;
+
+@EnabledForJreRange(min = JRE.JAVA_17)
```
rohanKanojia, line 34: Why do we want to run these tests in Java 17 and above only? Is it possible to remove it?

crd-generator/maven-plugin/README.md
```diff
@@ -0,0 +1,180 @@
+# CRD-Generator - Maven Plugin
+
+Maven plugin for the CRD-Generator: Generate CRDs from Java model during Maven builds.
```
rohanKanojia, line 3: This document should be either linked in https://github.com/fabric8io/kubernetes-client/blob/main/doc/CRD-generator.md or we should move it in `doc/` folder like the rest of the documentation files.

baloo42: Good point, we havn't talked about where the documentation for the generator-v2 should be placed, how it should be structured and especially how it should be distinguished to generator-v1.

My intention was to write first for each generator-v2 extension (maven-plugin, gradle-plugin, cli) an own README until at least two of them are merged. 
Afterwards, it should be easier to write some common notes about generator-v2 including migration paths and deprecations / roadmap into the central docs. And of course the content from the README's can be moved again, too.

If I should do this within this PR, I need some guidence on how to introduce generator-v2. Otherwise we should probably create a dedicated issue for the common part of the generator-v2 docs.

crd-generator/maven-plugin/src/test/java/io/fabric8/crd/generator/maven/plugin/ClasspathTypeTest.java
baloo42: I get some mockito related errors and havn't figured out yet why this is only the case on a lower versions than 17. (see https://github.com/fabric8io/kubernetes-client/actions/runs/9052896736/job/24871094873#step:4:36770)
Any ideas? 

But in general I'm not really happy with both tests in the maven-plugin module. A lot stuff must be mocked, only to test some simple things...

I'm open to rewrite both tests enterily if someone can point me into the right direction. e.g. writing e2e tests with tool x. (The java-generator maven plugin doesn't include any own tests, so I cannot use it as template)

rohanKanojia: Do you think you can use [maven invoker plugin](https://maven.apache.org/plugins/maven-invoker-plugin/) here? We use it for [JKube  integration tests](https://github.com/eclipse-jkube/jkube/blob/master/kubernetes-maven-plugin/it/pom.xml#L67) to verify generated YAML files.

baloo42: Added the first integration tests, is this the right direction?

rohanKanojia: In my opinion, it looks okay, but we might want to get opinions from other team members. 

Maybe we can make the verification check more strict (it might help us find the differences in generated YAML). In Eclipse JKube we compare whole YAML manifests in `expected/` directory with generated output.

rohanKanojia: It's up to you. If you think they're redundant feel free to remove them.

crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/JandexCustomResourceClassScanner.java
baloo42: I wrote an integration test for the "intermediate class" case (last bullet point from above):

https://github.com/fabric8io/kubernetes-client/pull/5979/commits/ddc3bec8dda01d0ee053d44c6240bb7840f6552d#diff-0488fc4e2f91d3e88ce94d75d1317631a14f2854194a2cf68354be4712e9c2cb

In this case, the user must add the dependency, which contains the intermediate class, to the list of dependencies to scan. Under the hood the classes are now added to the composite index and Custom Resources Classes will be found by the collector.

crd-generator/maven-plugin/README.md
baloo42: @rohanKanojia @manusa 

Are you ok with this strategy or do we need more in this MR? (Can we resolve this conversion?)

rohanKanojia: I'm okay with this strategy. Let's have something up and running first then we can interate	over how to better organize documentation.

crd-generator/maven-plugin/src/test/java/io/fabric8/crd/generator/maven/plugin/ClasspathTypeTest.java
baloo42: All `@EnabledForJreRange(min = JRE.JAVA_17)` annotations are removed, no more java version dependent tests.
I think this conversion can be resolved.

crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/JandexCustomResourceClassScanner.java
baloo42: Are any questions on this open or can we resolve this conversion?

crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/CustomResourceCollector.java
shawkins: @baloo42 yes, feel free to add whatever method works best.

metacosm: Note that we could possibly make API changes in the v2 version if needed… as long as it is possible to generate CRDs programmatically, incrementally registering classes to generate from.

metacosm:
> > Which approach do you prefer? Other thoughts?
> 
> I'd like to keep the core api as simple as possible - whatever is needed to satisfy Quarkus should be good.
> 
> The first option seems good enough. I don't really see much need for coupling between the collector and the crd generator - it doesn't really need to provide CustomResourceInfo, just classes. I mentioned on a pr somewhere that the collector to me just looks like something that you could provide a list of interfaces / base classes, and a list of annotations, and it will just find everything that matches that.

I need to look deeper at this PR but one thing that is raising my interest is the use of Jandex. The Quarkus extension for the Java Operator SDK is using classes to get to the needed `CustomResourceInfo`… If we'd have a generator that worked directly off of Jandex' `ClassInfo`, that'd be a win. That said, the Jackson-based approach requires classes so… 🤷🏼 
What we really need is a Jackson processor that works from Jandex indices! 😀

crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/CustomResourceCollector.java
baloo42: I added two additional input methods for collections (both nullsafe) and kept the existing ones, so that upstream projects can now choose between a collection, an array or a single CustomResourceInfo or CustomResource class.

> Note that we could possibly make API changes in the v2 version if needed… as long as it is possible to generate CRDs programmatically, incrementally registering classes to generate from.

Good to know, but IMHO we should keep both possibilities for now, so that we don't introduce unecessary changes. Please let me now if I should mark the array methods as deprecated, otherwise this thread can be marked as resolved in my opinion.

metacosm:
> > I need to look deeper at this PR but one thing that is raising my interest is the use of Jandex. The Quarkus extension for the Java Operator SDK is using classes to get to the needed `CustomResourceInfo`… If we'd have a generator that worked directly off of Jandex' `ClassInfo`, that'd be a win. That said, the Jackson-based approach requires classes so… 🤷🏼
> > What we really need is a Jackson processor that works from Jandex indices! 😀
> 
> I agree, using directly such ClassInfo objects would be a nice solution. As you said, the underlying JsonSchemaGenerator from Jackson requires classes atm. We would need Jackson to directly work with Jandex or create synthetically classes from ClassInfo objects. But I'm really unsure if this is possible or worth it.

Yes, it's probably not worth it at this point as that would be a major undertaking, not to mention that we want the CRD generator to follow Jackson's conventions…

> 
> For now Jandex is only used to "scan" for Custom Resource classes. If a project already uses Jandex to create an index (or a dependency, which should be scanned, contains a Jandex index) then those existing indices will be used instead of generating an own index, so that finding Custom Resource classes should be little bit faster.

In a Quarkus extension, like is the case for the Quarkus extension for the Java Operator SDK, we already have an index, so being able to pass the index to the generator would be good, indeed.

baloo42:
> > For now Jandex is only used to "scan" for Custom Resource classes. If a project already uses Jandex to create an index (or a dependency, which should be scanned, contains a Jandex index) then those existing indices will be used instead of generating an own index, so that finding Custom Resource classes should be little bit faster.
> 
> In a Quarkus extension, like is the case for the Quarkus extension for the Java Operator SDK, we already have an index, so being able to pass the index to the generator would be good, indeed.

As far as I understand atm the code from QOSDK and Quarkus, I assume the following (please correct me if it's wrong):
- A jandex index will be always generated for the current project (Because it's a core Quarkus feature for CDI)
- In QOSDK, the index is used to find reconcilers and associated CRDS, as well as non-reconciler associated CRDs.

The current code of the CustomResourceCollector already supports contributing existing indices:

https://github.com/baloo42/kubernetes-client/blob/crd-generator-maven-plugin/crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/CustomResourceCollector.java#L98

Therefore it should be possible to reuse the collector module from within QOSDK. At least for the non-reconciler associated CRDs:

https://github.com/quarkiverse/quarkus-operator-sdk/blob/next-fabric8-version/core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/CRDGenerationBuildStep.java#L78

This would result in the exact same behaviour to find a custom resource class if we compare qosdk, crd-generator maven plugin, gradle plugin etc. At the same time it wouldn't need too much changes on QOSDK side.

On CustomResourceCollector side, we only have to ensure that we can disable creating own indices entirely e.g. by adding a flag.

What do you think?

metacosm:
> > > For now Jandex is only used to "scan" for Custom Resource classes. If a project already uses Jandex to create an index (or a dependency, which should be scanned, contains a Jandex index) then those existing indices will be used instead of generating an own index, so that finding Custom Resource classes should be little bit faster.
> > 
> > 
> > In a Quarkus extension, like is the case for the Quarkus extension for the Java Operator SDK, we already have an index, so being able to pass the index to the generator would be good, indeed.
> 
> As far as I understand atm the code from QOSDK and Quarkus, I assume the following (please correct me if it's wrong):
> 
>     * A jandex index will be always generated for the current project (Because it's a core Quarkus feature for CDI)
> 
>     * In QOSDK, the index is used to find reconcilers and associated CRDS, as well as non-reconciler associated CRDs.
> 
>     * The results of the CRD generation are important to further generate the CSV manifests
> 
> 
> The current code of the CustomResourceCollector already supports contributing existing indices:
> 
> https://github.com/baloo42/kubernetes-client/blob/crd-generator-maven-plugin/crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/CustomResourceCollector.java#L98
> 
> Therefore it should be possible to reuse the collector module from within QOSDK. At least for the non-reconciler associated CRDs:
> 
> https://github.com/quarkiverse/quarkus-operator-sdk/blob/next-fabric8-version/core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/CRDGenerationBuildStep.java#L78
> 
> This would result in the exact same behaviour to find a custom resource class if we compare qosdk, crd-generator maven plugin, gradle plugin etc. At the same time it wouldn't need too much changes on QOSDK side.
> 
> On CustomResourceCollector side, we only have to ensure that we can disable creating own indices entirely e.g. by adding a flag.
> 
> What do you think?

Sounds reasonable.

baloo42:
I have digged a little bit deeper into QOSDK and tried to integrate the collector from this PR.
It's possible with a few changes but I don't think it's worth it if no further changes for QOSDK are planned.

If most logic in QOSDK should stay, the only "interesting" part for QOSDK from the collector module is the following function (`JandexCustomResourceClassScanner`):

```java
  private List<ClassInfo> findCustomResourceClasses(IndexView index) {
    return index.getAllKnownImplementors(HasMetadata.class)
        .stream()
        .filter(classInfo -> classInfo.hasAnnotation(Group.class))
        .filter(classInfo -> classInfo.hasAnnotation(Version.class))
        .collect(Collectors.toList());
  }
```
In my opinion, that's too less to add the collector as dependency.

@metacosm Do you want to get rid of some specific parts in QOSDK related to finding custom resources?

crd-generator/maven-plugin/pom.xml
```diff
@@ -0,0 +1,157 @@
...
+          </pom>
+          <groovy>
+            <includes>
+              <include>src/it/**/*.groovy</include>
+            </includes>
+            <importOrder/>
+            <greclipse/>
+          </groovy>
+        </configuration>
+      </plugin>
```
manusa, line 154: Why is this needed?
isn't the plugin configuration in the parent pom enough?
can these changes be part of the parent pom instead so that we keep the plugin configuration in a single place?

pom.xml
```diff
@@ -101,6 +101,7 @@
     <maven-plugin-annotations.version>3.13.0</maven-plugin-annotations.version>
     <vertx.version>4.5.8</vertx.version>
     <zjsonpatch.version>0.3.0</zjsonpatch.version>
+    <jandex.version>3.1.8</jandex.version>
```
manusa, line 104: I think this should be bumped to 3.2.0 to align with the current dependency in the project (sorry, this is due to the late review of the PR :pray:)

crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/JandexIndexer.java
manusa: These seem reasonable limits.
We can keep it this way.
If users complain, then we can make this configurable in a future release.

crd-generator/maven-plugin/pom.xml
baloo42: It's required if the formatter should consider groovy files, too.
This isn't configured global and I was unsure if it should be configured global because of the impact.
If we don't care about the groovy files we can remove the whole plugin config and fallback to the defaults.

Is groovy file formatting important for you, so that we should keep it either here or global? Or should I remove it?

crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/JandexIndexer.java
baloo42: Ok, let's keep it.

Just as a note:
I'm a little bit unsure if those checks make sense overall...
I mean, if a project cannot trust it's dependencies, then the project has more problems than those from a zip bomb. But if we think towards the CRD-Generator CLI, it's some kind of own input validation and kind of useful.

crd-generator/maven-plugin/pom.xml
manusa: I understand this is for verify.groovy and other helper scripts (which are also present in other modules).
I don't mind one or the other, but I do think we need to centralize configuration.
If adding this globally implies making changes in other modules due to formatting, then remove (in the end these files aren't very important and that many).
If not, then please add the configuration globally.

baloo42: I have moved the groovy / integration test spotless config to the parent pom.
First with a minimal config required to cover only the new integration test. The second commit increases the scope and includes corresponding formatting fixes. 
If it's too much, I can revert the second commit.

But I discovered a small "problem": It seems like we are affected by bug of the spotless groovy formatter / greclipse. If formatting is checked or applied with the maven plugin an empty temp directory will be created in the project root.

https://github.com/diffplug/spotless/issues/2019

Can we deal with that or do we need a workaround?

crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/CustomResourceCollector.java
```diff
@@ -0,0 +1,195 @@
...
+    Predicate<String> classNamePredicate = classNamePredicates.stream()
+        .reduce(Predicate::and)
+        .orElse(className -> true);
+
+    List<Class<? extends HasMetadata>> customResourceClasses = customResourcesClassNames.stream()
+        .filter(classNamePredicate)
+        .map(customResourceClassLoader::loadCustomResourceClass)
+        .collect(Collectors.toList());
+
+    log.debug("Found {} custom resource classes after filtering", customResourceClasses.size());
```
metacosm, line 180:
```suggestion
    log.debug("Retained {} custom resource classes after filtering", customResourceClasses.size());
```

crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/CustomResourceCollector.java
```diff
@@ -0,0 +1,195 @@
...
+import java.util.Set;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import static java.util.Optional.ofNullable;
+
+/**
+ * Collects Custom Resource class files from various places and loads them by using
+ * {@link CustomResourceCollector#findCustomResourceClasses()}.
+ */
```
metacosm, line 39:
```suggestion
 */
 @SuppressWarnings("UnusedReturnValue")
```

crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/CustomResourceCollector.java
```diff
@@ -0,0 +1,195 @@
...
+        .collect(Collectors.toList());
+
+    log.debug("Found {} custom resource classes after filtering", customResourceClasses.size());
+    return customResourceClasses;
+  }
+
+  /**
+   * Resets all internal states to defaults.
+   */
+  public void reset() {
```
metacosm, line 187: Do we need this method to be public?

crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/JandexCustomResourceClassScanner.java
```diff
@@ -0,0 +1,156 @@
...
+ * resulting composite index by using {@link JandexCustomResourceClassScanner#findCustomResourceClasses()}.
+ * <p>
+ * If not overridden by {@link JandexCustomResourceClassScanner#forceIndex}, the implementation uses an
+ * existing Jandex index if found in the source. Otherwise, the index will be created on the fly.
+ * </p>
+ *
+ * @see Index
+ * @see JandexIndexer
+ * @see JandexUtils
+ */
```
metacosm, line 50:
```suggestion
 */
 @SuppressWarnings("UnusedReturnValue")
```

crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/JandexCustomResourceClassScanner.java
```diff
@@ -0,0 +1,156 @@
...
+   *
+   * @param index the index
+   * @return the Custom Resource classes
+   *
+   * @see JandexCustomResourceClassScanner#createBaseIndex()
+   */
+  private List<ClassInfo> findCustomResourceClasses(IndexView index) {
+    // Only works if HasMetadata and all intermediate classes
+    // (like CustomResource) are included in given index.
+    return index.getAllKnownImplementors(HasMetadata.class)
```
metacosm, line 125: Why are we looking at all implementors of `HasMetadata` instead of `CustomResource`? Is it meant to also include native resources?

metacosm (approved):
Just minor changes and comments. Can proceed without.

crd-generator/maven-plugin/pom.xml
manusa: I guess we can try and live with this for now. If it gets annoying we can disable.
For the second commit, I'd need to check first.

crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/CustomResourceCollector.java
baloo42: Neither the maven plugin nor the CLI requires this method at the moment. But I thought it's a good indicator to communicate that the subcomponents of the CustomResourceCollector have an internal state and cache some things.

We can make this package-private and open it if someone really needs it.

crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/JandexCustomResourceClassScanner.java
baloo42: I have noticed somewhere that you want to get rid of the CustomResource class. That's why I tried to implement it without referencing the `CustomResource` class. It also works if we use here the CustomResource class.

If in the future other interfaces / classes are introduced to indicate a CustomResource, this method must be ajusted.

metacosm: Makes sense. My concern here is that, depending on the index, this might return a lot of results.

baloo42: I have found the discussions:
https://github.com/fabric8io/kubernetes-client/issues/4164#issuecomment-1328718564

baloo42: > My concern here is that, depending on the index, this might return a lot of results.

If it's really a problem we can use CustomResource class for now until the replacement is in place. I have no problem to change it back within this PR.

Do you have some numbers about the biggest known projects?

metacosm: We can revisit if that becomes an issue, for sure. I don't really have a number but I'm thinking if people are importing a lot of Fabric8 libs and these get use in a Quarkus context, they might all end up in the application index, which should probably be used when generating CRDs from a Quarkus app, so there might be lots of candidates there.

baloo42: Now I understand your concerns better. Some clearifications:

**QOSDK Projects**

QOSDK projects won't use the collector module because QOSDK directly integrates with the CRD-Generator v1 and it seems like you want to integrate in the same way the CRD-Generator v2 in https://github.com/quarkiverse/quarkus-operator-sdk/pull/876, too. 
I think this is good and the way for the future, because Quarkus already provides features to search for classes  during build time.

In conclusion the recommendation for QOSDK projects to generate CRDs is at the moment and should be in the future to use the QOSDK native way.

**Quarkus Projects**

If a Quarkus project, which doesn't use QOSDK, wants to generate CRDs, the project might use either the new CRD-Generator Maven or Gradle plugin.
A Quarkus project already builds a Jandex index for the *project classes* and *submodules* automatically. This index contains external classes from libs only if *explicitly* configured e.g. by `@RegisterForReflection`.

The collector module will detect this index by default, skips building an own index and uses it to search for custom resources. Other indices from libraries are only considered *if explicitly configured*.  (e.g. in the maven plugin with `dependenciesToScan`).

> I'm thinking if people are importing a lot of Fabric8 libs and these get use in a Quarkus context, they might all end up in the application index, which should probably be used when generating CRDs from a Quarkus app, so there might be lots of candidates there.

In conclusion, the relevant index is in most projects smaller as you might have initially thought.
The overal recommendation would be to use features to extend the index (e.g. `dependenciesToScan`) wisely and only if necessary.


Coming back to the original topic, I think we should look at this question more on a strategic perspective.

**What interfaces / annotations / classes should be used within this PR to identify custom resource classes?**

Options:

1. A class is a custom resource class if it extends `CustomResource`
2. A class is a custom resource class if it implements `HasMetadata` and is annotated by `@Groupd` and `@Version` (current implementation)

@manusa @shawkins you might have also an opinion on that. What option is preferred?

manusa: I'm not sure how we're excluding the core classes here (i.e. any of the model types that we generate as part of `kuberentes-model-generator` sub modules).
These classes implement HasMetadata and are annotated with Group and Version.

pom.xml
```diff
@@ -1200,6 +1210,11 @@
         <version>${spotless-maven-plugin.version}</version>
         <configuration>
           <java>
+            <includes>
+              <include>src/main/java/**/*.java</include>
+              <include>src/test/java/**/*.java</include>
+              <include>src/it/**/src/main/java/**/*.java</include>
+            </includes>
```
manusa, line 1217: I'll probably send an independent PR with these and the related formatting changes prior to merging this one.
Otherwise it will be hard to rebase other ongoing PRs (namely #5632)

pom.xml
```diff
@@ -1208,6 +1223,15 @@
...
           </java>
+          <groovy>
+            <includes>
+              <include>src/main/groovy/**/*.groovy</include>
+              <include>src/test/groovy/**/*.groovy</include>
+              <include>src/it/**/*.groovy</include>
+            </includes>
+            <importOrder/>
+            <greclipse/>
+          </groovy>
```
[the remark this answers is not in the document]
baloo42: Sure, no problem. I already prepared the extraction of those formatter changes into a seperate PR.
I'm just a little bit busy this week. If you can wait until next week, I can help you with this task.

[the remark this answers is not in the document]
manusa: I'm going on holiday in a few minutes. There's no rush :)

I'll be back on the 26th of August. My first order of business will be to merge this PR once I catch up with whatever comes up during my days off.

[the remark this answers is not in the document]
baloo42: Ok, then I will prepare it next week. No hurries from my side, too. 

Thank you for your work on this project and have a good time! You deserve it!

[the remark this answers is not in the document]
baloo42: I extracted the formatter changes to #6301 and will rebase this PR once #6301 is merged.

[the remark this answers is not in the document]
manusa: thanks!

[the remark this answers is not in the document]
baloo42: No problem, you can take over the branch now if something has to be fixed or let me know it and I will fix it ;)

pom.xml
```diff
@@ -101,6 +101,7 @@
     <maven-plugin-annotations.version>3.14.0</maven-plugin-annotations.version>
     <vertx.version>4.5.9</vertx.version>
     <zjsonpatch.version>0.3.0</zjsonpatch.version>
+    <jandex.version>3.2.1</jandex.version>
```
baloo42, line 104:
```suggestion
    <jandex.version>3.2.2</jandex.version>
```

manusa (approved):
LGTM, thx!

crd-generator/maven-plugin/src/main/java/io/fabric8/crd/generator/maven/plugin/Dependency.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.fabric8.crd.generator.maven.plugin;
+
+public class Dependency {
```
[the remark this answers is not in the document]
baloo42: I think I didn't choose this existing model because it allows own filtering and I tried to avoid filtering at too many levels.

I currently investigate in other ways of filtering, mainly because of your other comment (https://github.com/fabric8io/kubernetes-client/pull/5979#discussion_r1717034119).

If I can find a way to improve it, should I push it to this branch or is the current state good enough to get merged?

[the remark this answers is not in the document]
manusa: This branch is good to go, let's leave it as is. I'll merge it on Monday (I'm leaving some time in case someone else wants to review).

rohanKanojia (approved):

crd-generator/collector/src/main/java/io/fabric8/crd/generator/collector/JandexCustomResourceClassScanner.java
baloo42: > Is this even possible?

I'm not sure (because I don't know Quarkus in depth), but I don't think this is a problem.

I have started to write some docs and diagrams, so that we can better talk about this:
https://github.com/baloo42/crd-generator-v2-docs/blob/main/crd-generator-maven-plugin.md

The first example shows a default setup where custom resource classes are only located in the current project.
The second example shows a more advanced setup, where CRDs should be generated for custom resource classes in the project **and** for custom resource classes in a dependency.

In both examples, having an jandex index, is optional. But if it exist, it will be used and will therefore result in a small performance improvement.

If an index exists, it is assumed, that this index is not messy and contains only classes of the project (`target/**`) or of the dependency itself. (I'm a little bit unsure if we can assume this always).

> In any case, if that happens, the plugin should provide the option for inclusion/exclusion of classes/packages. This I understand it already does in the current implementation Inclusions, used to filter Custom Resource classes after scanning..
> However, if what metacosm describes is possible, the troublesome part would be the after scanning since the initial loading would require an increased overhead, so filtering should probably happen at this point.

At the moment there are two options to control inclusions/exclusions:

1. By defining the scope for scanning. (Which is something like "inclusions")
2. By filtering after the scan with package inclusions/exclusions.

To explain the first option more in detail, let's focus on the [second example in the draft docs](https://github.com/baloo42/crd-generator-v2-docs/blob/main/crd-generator-maven-plugin.md#scan-scope-project-and-dependency) and assume the "Jar Archive 1" is the kubernetes-client library (and all its dependencies / submodules). If I understand you both correct, your concerns for such a scenario are, that the list of found custom resource classes would be big.
I don't think this is the case, because as shown in the second example, only explicitly defined dependencies are used to build up the internal composite index. Therefore only custom resource classes of explicitly defined dependencies would show up in the list.

I also think, that even the second option is very efficient, because it is also processed before the class is loaded and handed over to the crd-generator.

In conclusion, I think a user won't get into troubles as long as the user creates conventional jandex indices.

The only thing I can imagine as a future improvement is that we allow a user to provide an index and allow to disable detecting existing indices as well as to disable creating indices automatically.
This would give the user full control over the internal index of the maven-plugin, so that he can optimize the scope.

---
Sidenotes:

Beside it makes sense or not, I have searched for a way to get classes from an index filtered by multiple criterias (like by package and by implementing interface). But I always ended up to stream the first results and filter further. I don't think there is a native way to do this with jandex.

I also explored if it's possible to create subindices with a subset of the classes of an index. But I haven't found an efficient way.

manusa: >  only explicitly defined dependencies are used to build up the internal composite index

I think this resolves my concerns. Just to make sure, here I'm repeating what I get from your explanation (+ detailed diagrams and use/config cases) _Users will opt-in to scan their dependencies in their pom.xml plugin configuration, by adding these dependencies explicitly they should be aware of any potential performance issues created by huge model artifacts_

baloo42: > Users will opt-in to scan their dependencies in their pom.xml plugin configuration, by adding these dependencies explicitly they should be aware of any potential performance issues created by huge model artifacts

That's correct. In addition we should add the following warning:

> A serialized jandex index in the project (`target/classes/META-INF/jandex.idx` will be detected and used. Be aware of performance issues if this index contains (many) classes from dependencies. (This shouldn't be the case if you create a conventional jandex index by including the project files only).

manusa (approved):
LGTM, thx!

Issue #5951, opened 2024-04-24
Align crd generator v1 and v2 output


shawkins:
The v1 output should sort the required list.

Also the v1 handling for JsonFormat shape needs to leave the type as string in most cases.

v2 will use set the printer column type to be the actual property type. v1 will continue to use string in some cases.

matteriben:
> Also the v1 handling for JsonFormat shape needs to leave the type as string in most cases.

I think these lines in v1 need to be removed to align handling for JsonFormat shape with v2:
https://github.com/fabric8io/kubernetes-client/blob/a8fba97b2a22bf805894ec33e9982ec37e4e8a8d/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java#L159-L162
Thank you!

shawkins:
I believe we've addressed all known issues.

PR #5991, opened 2024-05-07
test(crd-generator): approval tests for crd-generator-api-v2

manusa:
## Description

Part of #5951 / #5948

Supersedes closes #5978

test(crd-generator): approval tests for crd-generator-api-v2

Relates to:
- #5948
- #5951
- #5850
- #5978
- #5980
- #5986
- #5989

## Type of change

## Checklist


manusa:
Hi @shawkins,

Maybe you can help out here.

This PR adds the tests to ensure _feature parity_ for both versions of the crd-generator.

However, there's a misbehavior, but I'm unsure if this is because something is missing in the new implementation or there's a bug in the old one.

The approval file `CRDGeneratorApprovalTest.approvalTest.containingjsons.sample.fabric8.io.v1.approved.yml` contains the following lines:

https://github.com/fabric8io/kubernetes-client/blob/e3f2ae9e8bcf4338210479e74e651552ebcd66f1/crd-generator/test/src/test/resources/io/fabric8/crd/generator/approvaltests/CRDGeneratorApprovalTest.approvalTest.containingjsons.sample.fabric8.io.v1.approved.yml#L22-L30

which are not generated with the new implementation.

This fragment looks instead like:
```yaml
              foo:
                type: "object"
```

shawkins:
@manusa the new one is correct with the default settings. We now have a CRDGenerator property to control whether x-kubernetes-preserve-unknown-fields is implicitly inferred from anyGetter/Setter - it defaults to false. It would need to use the crd annotation for preserve unknown or turn on the implicit setting for this to be seen as x-kubernetes-preserve-unknown-fields.

At the very least, the old one should not have been adding both a configAsMap property and setting x-kubernetes-preserve-unknown-fields - because that was derived from configAsMap.

manusa:
Thx, yes, it looked more like a bug from the former implementation.
However, I'm not sure it's an easy fix (might break other stuff while fixing it). It might be better to just remove this part from the test (Class).

Thoughts? 

/cc @metacosm @andreaTP @baloo42

baloo42:
IMHO fixing the test for now is better here than changing something in v1 or v2.

Maybe we can add some test cases later to show the differences or a Wiki page to document such things.

manusa:
:+1: I removed the conflicting field + class.
The test case is still valid since it has a JsonNode field that properly maps to `x-kubernetes-preserve-unknown-fields: true`

rohanKanojia (approved):

shawkins (approved):
LGTM

PR #6013, opened 2024-05-16
feat: add factory methods to help with SSA

### LINKED ISSUE #6012: Add utility method to easily create a named/namespaced empty copy of a given resource

When using Server-Side Apply, it is often required to send only the modified state instead of the full resource. This requires creating copies of resources with only the desired fields set. While this is perfectly doable now, having a utility method to remove some of the boiler plate of creating such a new instance (i.e. instantiation and metadata creation + setting) would be helpful.

New `HasMetadata` factory methods to perform such operations.

_No response_

### Additional context

_No response_

metacosm:
## Description

Fixes #6012

## Type of change

## Checklist

### DIFF
--- a/kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
+++ b/kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
@@ -48,10 +48,9 @@ public interface HasMetadata extends KubernetesResource {
   Pattern FINALIZER_NAME_MATCHER = Pattern.compile(
       "^((" + DNS_LABEL_REGEXP + "\\.)+" + DNS_LABEL_START + 2 + DNS_LABEL_END + ")/"
           + DNS_LABEL_REGEXP);
-
-  ObjectMeta getMetadata();
-
-  void setMetadata(ObjectMeta metadata);
+  String REQUIRES_NON_NULL_METADATA = "requires non-null metadata";
+  String REQUIRES_NON_NULL_NAME = "requires non-null name";
+  String REQUIRES_NON_NULL_NAMESPACE = "requires non-null namespace";
 
   /**
    * Retrieves the kind associated with the specified HasMetadata implementation. If the implementation is annotated with
@@ -63,10 +62,6 @@ public interface HasMetadata extends KubernetesResource {
   static String getKind(Class<?> clazz) {
     final Kind kind = clazz.getAnnotation(Kind.class);
     return kind != null ? kind.value() : clazz.getSimpleName();
-  }
-
-  default String getKind() {
-    return getKind(getClass());
   }
 
   /**
@@ -113,12 +108,6 @@ public interface HasMetadata extends KubernetesResource {
     return version != null ? version.value() : null;
   }
 
-  default String getApiVersion() {
-    return getApiVersion(getClass());
-  }
-
-  void setApiVersion(String version);
-
   /**
    * Retrieves the plural form associated with the specified class if annotated with {@link
    * Plural} or computes a default value using the value returned by {@link #getSingular(Class)} as
@@ -131,11 +120,6 @@ public interface HasMetadata extends KubernetesResource {
     final Plural fromAnnotation = clazz.getAnnotation(Plural.class);
     return (fromAnnotation != null ? fromAnnotation.value().toLowerCase(Locale.ROOT)
         : Pluralize.toPlural(getSingular(clazz)));
-  }
-
-  @JsonIgnore
-  default String getPlural() {
-    return getPlural(getClass());
   }
 
   /**
@@ -153,11 +137,6 @@ public interface HasMetadata extends KubernetesResource {
         .toLowerCase(Locale.ROOT);
   }
 
-  @JsonIgnore
-  default String getSingular() {
-    return getSingular(getClass());
-  }
-
   static String getFullResourceName(Class<?> clazz) {
     final String plural = getPlural(clazz);
     final String group = getGroup(clazz);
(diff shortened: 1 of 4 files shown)


kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
```diff
@@ -493,4 +496,26 @@ default void removeOwnerReference(HasMetadata owner) {
   default Optional<ObjectMeta> optionalMetadata() {
     return Optional.ofNullable(getMetadata());
   }
+
+  static <T extends HasMetadata & Namespaced> T emptyCopyOf(T original, Supplier<T> constructor) {
```
shawkins, line 500: Instead of emitCopyOf (copy could be missleading here) could this be named something like withNewMetadata? That would make it consistent with what it is doing and existing builder methods.

metacosm: Well, it is a copy. Not sure how that would be misleading?

metacosm: How about `newNamedInstance` and `newNamedAndNamespacedInstance`?

shawkins: It's only a copy or a new instance if the Supplier provides a new instance / copy. Even then if it's just the constructor, then it's a shallow copy.

If you want a full copy maybe specifying Editable would be the way to go.

metacosm: The goal is to get a shallow copy with only name/namespace set. Indeed, that will depend on the provided Supplier but I guess that also provides some flexibility as to what people want to initialize their instance with. Another "interesting" issue is that it seems we can't constraint the Supplier to return the exact object type as defined by the parameter type, which is rather annoying.

shawkins: In that case maybe simplify to just:

```
  static ObjectMeta newObjectMetaLike(HasMetadata original) {
    final ObjectMeta metadata = Objects.requireNonNull(original.getMetadata(), REQUIRES_NON_NULL_METADATA);
    ObjectMeta objectMeta = new ObjectMeta();
    objectMeta.setName(Objects.requireNonNull(metadata.getName(), REQUIRES_NON_NULL_NAME));
    if (Utils.isResourceNamespaced(original.getClass()) {
      objectMeta.setNamespace(Objects.requireNonNull(metadata.getNamespace(), REQUIRES_NON_NULL_NAMESPACE));
    }
    return objectMeta;
  }
```

And leave it up to the caller how they associate it with an instance.

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
```diff
@@ -493,4 +496,26 @@ default void removeOwnerReference(HasMetadata owner) {
...
+    final String name = Objects.requireNonNull(metadata.getName(), REQUIRES_NON_NULL_NAME);
+    final String ns = Objects.requireNonNull(metadata.getNamespace(), REQUIRES_NON_NULL_NAMESPACE);
+    t.setMetadata(new ObjectMetaBuilder()
+        .withName(name)
+        .withNamespace(ns)
+        .build());
+    return t;
+  }
+
+  static <T extends HasMetadata> T emptyClusterWideCopyOf(T original, Supplier<T> constructor) {
```
rohanKanojia, line 512: Umm, sorry for the naive question. Why is the visibility of this method kept to the default scope?

metacosm: Because this is on an interface so it is by default public.

rohanKanojia (approved):

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
metacosm: Makes sense but then I'd argue that this method should be put on `ObjectMeta` itself.

shawkins: We'd have to remove the ObjectMeta from automatic generation - we don't yet have the ability to mix-in additional methods.

metacosm: If it's not on `ObjectMeta`, I'd be concerned about discoverability. That said, we could also maybe have a `initNameAndNamespaceFrom` method that would set a new metadata on the passed object and retain said object?

```java
static HasMetadata initNameAndNamespaceFrom(HasMetadata original, HasMetata instanceToInit) {
    final ObjectMeta metadata = Objects.requireNonNull(original.getMetadata(), REQUIRES_NON_NULL_METADATA);
    ObjectMeta objectMeta = new ObjectMeta();
    objectMeta.setName(Objects.requireNonNull(metadata.getName(), REQUIRES_NON_NULL_NAME));
    if (original instanceof Namespaced) {
      objectMeta.setNamespace(Objects.requireNonNull(metadata.getNamespace(), REQUIRES_NON_NULL_NAMESPACE));
    }
    instanceToInit.setMetadata(objectMetadata);
    return instanceToInit;
  }
```

metacosm: So which version should I keep?

shawkins: @metacosm sorry I missed the previous message. I'm fine with what you have. initNameAndNamespaceFrom could also be done directly on HasMetadata:

```
default void initNameAndNamespaceFrom(HasMetadata original) {
    final ObjectMeta metadata = Objects.requireNonNull(original.getMetadata(), REQUIRES_NON_NULL_METADATA);
    ObjectMeta objectMeta = new ObjectMeta();
    objectMeta.setName(Objects.requireNonNull(metadata.getName(), REQUIRES_NON_NULL_NAME));
    if (original instanceof Namespaced) {
      objectMeta.setNamespace(Objects.requireNonNull(metadata.getNamespace(), REQUIRES_NON_NULL_NAMESPACE));
    }
    this.setMetadata(objectMeta);
  }
```

The only other question or thought is whether you are typically starting with builders instead, which would probably look like:

```
static ObjectMetaBuilder initNameAndNamespaceFrom(ObjectMetaBuilder builder, HasMetadata original) {
  ...
}
```

metacosm: Ah yes, good points!

metacosm: Should be good now, I think…

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
```diff
@@ -493,4 +495,55 @@ default void removeOwnerReference(HasMetadata owner) {
...
+   * resource) provided by the specified ObjectMeta instance. This is a convenience method to avoid boilerplate, notably when
+   * using Server-Side Apply, when creating a new instance with only some fields of the original one. This method
+   * <em>doesn't</em> set the metadata on this object, since it assumes that further configuration will occur on the newly
+   * created ObjectMetaBuilder.
+   *
+   * @param original an ObjectMeta instance from which to retrieve the name and namespace
+   * @return a new ObjectMetaBuilder instance initialized with the name and namespace (if the resource it is called on is a
+   *         namespaced one) of the specified ObjectMeta
+   */
+  default ObjectMetaBuilder initMetadataBuilderNameAndNamespaceFrom(ObjectMeta original) {
```
shawkins, line 540: Let's make this one static - it does not need access to the current instance.

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
```diff
@@ -493,4 +495,55 @@ default void removeOwnerReference(HasMetadata owner) {
...
+   * Initializes this {@link ObjectMeta} field with name and namespace (if this instance represents a namespaced resource)
+   * provided by the specified HasMetadata instance. This is a convenience method to avoid boilerplate, notably when using
+   * Server-Side Apply, when creating a new instance with only some fields of the original one. Calls
+   * {@link #setMetadata(ObjectMeta)} when done, if you want to further configure this instance's metadata, please use
+   * {@link #initMetadataBuilderNameAndNamespaceFrom(ObjectMeta)} instead, which <em>doesn't</em> sets the metadata, leaving it
+   * up to the user once configuration is finished.
+   *
+   * @param original a HasMetadata instance from which to retrieve the name and namespace
+   */
+  default void initNameAndNamespaceFrom(HasMetadata original) {
```
shawkins, line 509: Can we omit this signature? It duplicates a lot of javadoc, but add's only a minor amount of convenience.

shawkins (commented):
Looks good, just two more thoughts.

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
metacosm: I'd rather keep this one than the `ObjectMeta` one, actually since, when using SSA, that's what you'll most likely start with (a resource).

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/HasMetadata.java
metacosm: Makes sense. I originally hoped that I could set the metadata on the current instance as well.

shawkins:
```
static ObjectMetaBuilder initMetadataBuilderNameAndNamespaceFrom(ObjectMeta original)
```
gives you the usage:

```
new SomethingBuilder().withMetadata(HasMetadata.initMetadataBuilderNameAndNamespaceFrom(original)....build())...
```

An alternative which doesn't quite set the metadata, but allows you to do so more fluently:

```
static <A extends ObjectMetaFluent<A>> A withNameAndNamespace(A value, HasMetadata original) {
```

gives you the usage:

```
HasMetadata.withNameAndNamespace(new SomethingBuilder().withNewMetadata(), original)...endMetadata()...
```

metacosm: Actually, it does access the current instance to check if it is `Namespaced` or not to decide whether to use and set the namespace…

shawkins: @metacosm I think I was assuming based upon your other comment though it could instead be:

```
static ObjectMetaBuilder initMetadataBuilderNameAndNamespaceFrom(HasMetadata original)
```

But didn't make that change in the signature shown above.

metacosm: Gotcha. Sorry for being slow, it's been a long day. 😓

shawkins (approved):
LGTM

manusa (approved):
LGTM, thx!

csviri (approved):

PR #6038, opened 2024-05-31
Remove project from task extension

Sineaggi:
In order to support Gradle's configuration cache, having the Project object in extensions/tasks is not allowed https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements:disallowed_types. We can avoid that by injecting the ProjectLayout object directly which is legal.

> There are a number of types that task instances must not reference from their fields. 
> These types fall into some categories as follows:
> - Live JVM state types
> - Gradle model types
> - Dependency management types
>
> In all cases the reason these types are disallowed is that their state cannot easily be stored or recreated by the configuration cache.

Specifically
> Gradle model types (e.g. Gradle, Settings, Project, SourceSet, Configuration etc…​) are usually used to carry some task input that should be explicitly and precisely declared instead.

### DIFF
--- a/java-generator/gradle-plugin/src/main/java/io/fabric8/java/generator/gradle/plugin/JavaGeneratorPlugin.java
+++ b/java-generator/gradle-plugin/src/main/java/io/fabric8/java/generator/gradle/plugin/JavaGeneratorPlugin.java
@@ -23,7 +23,7 @@ public class JavaGeneratorPlugin implements Plugin<Project> {
   @Override
   public void apply(Project project) {
     // create an extension for the plugin
-    project.getExtensions().create(JavaGeneratorPluginExtension.NAME, JavaGeneratorPluginExtension.class, project);
+    project.getExtensions().create(JavaGeneratorPluginExtension.NAME, JavaGeneratorPluginExtension.class);
     // register tasks
     project.getTasks().register(JavaGeneratorCrd2JavaTask.NAME, JavaGeneratorCrd2JavaTask.class,
         JavaGeneratorPluginExtension.class);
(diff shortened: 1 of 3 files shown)


andreaTP (approved):
I don't know much about Gradle plugins, but looks good to me.

cc. @manusa

rohanKanojia:
@Sineaggi : Could you please fix style errors with `mvn spotless:apply`?

oscerd (approved):

rohanKanojia:
@Sineaggi : Do you know if this configuration cache requirement was added since Gradle 8.x ?

rohanKanojia (approved):

Sineaggi:
> @Sineaggi : Do you know if this configuration cache requirement was added since Gradle 8.x ?

I think since a minor version of Gradle 6.x, not sure which exactly. This does not prevent the plugin from running on Gradle 8.x, but it does break projects that attempt to enable configuration cache.

Sineaggi:
Without these changes, we get the following error
```
2 problems were found storing the configuration cache.
- Task `:app:crd2java` of type `io.fabric8.java.generator.gradle.plugin.task.JavaGeneratorCrd2JavaTask`: cannot deserialize object of type 'org.gradle.api.Project' as these are not supported with the configuration cache.
  See https://docs.gradle.org/8.7/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
- Task `:app:crd2java` of type `io.fabric8.java.generator.gradle.plugin.task.JavaGeneratorCrd2JavaTask`: cannot serialize object of type 'org.gradle.api.internal.project.DefaultProject', a subtype of 'org.gradle.api.Project', as these are not supported with the configuration cache.
  See https://docs.gradle.org/8.7/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
```
No errors with this change.

Sineaggi:
Normally I'd be able to create a test (via GradleRunner) that starts up Gradle with the plugin, but I wasn't having much luck getting that to work from maven.

Sineaggi:
Good to merge?

manusa:
> Good to merge?

Yes it is.

However, I'd like to check a few things locally. I'll probably merge on Friday.

Sorry for the delay.

manusa (approved):
LGTM, thx!

Issue #6008, opened 2024-05-14
remove bouncycastle dependency for EC key support


coopstah13:
### Is your task related to a problem? Please describe

Newer version of BouncyCastle doesn't work with BCFIPS provider

### Describe the solution you'd like

Remove dependency on BouncyCastle and rely on JDK only. The dependency is only used for EC keys, but they are supported since at least JDK8.

### Describe alternatives you've considered

I have pinned the transitive bouncycastle dependency to an older version (1.68) in our repository as it seems to be the last one that doesn't have the problem.

shawkins:
@coopstah13 do you have suggested changes for this? The test case that was added for bcfips does not work with just the jdk - it fails with "java.security.spec.InvalidKeySpecException: java.security.InvalidKeyException: IOException : version mismatch: (supported:     00, parsed:     01 ..."

shawkins:
> We probable need to add a specific case for
[CertUtils#handleECKey](https://github.com/fabric8io/kubernetes-client/blob/2f11261da823bc858658b3218c6f913af36a5342/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/internal/CertUtils.java#L180) so that we don't require the usage of BouncyCastle libraries for Elastic keys.

Yes, I understand what could be done for eliptical curve keys, but nothing simple worked. Taking another look the best I could come up with was to borrow some code from vertx. I'll open a pr with that.

martinwunderlich-celonis:
I've had the same issue and was to remove the Bouncycastle library after upgrading the fabric8 client from 6.13.1 to 7.0-SNAPSHOT (built locally). 
Does anyone have an idea when 7.0.0 will get released?

manusa:
> Does anyone have an idea when 7.0.0 will get released?

End of September or October if everything goes well.

martinwunderlich-celonis:
Hello @manusa , just out of curiosity, is the release for 7.0.0 on track? Thank you

PR #6051, opened 2024-06-12
fix: remove bc optional dependency

shawkins:
## Description

Removes the bc optional dependency

closes: #6008

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client-api/pom.xml
+++ b/kubernetes-client-api/pom.xml
@@ -160,12 +160,10 @@
     <dependency>
       <groupId>org.bouncycastle</groupId>
       <artifactId>bcprov-jdk18on</artifactId>
-      <optional>true</optional>
     </dependency>
     <dependency>
       <groupId>org.bouncycastle</groupId>
       <artifactId>bcpkix-jdk18on</artifactId>
-      <optional>true</optional>
     </dependency>
 
     <!-- Compile Only Dependencies -->
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Bugs
 
 #### Improvements
+* Fix #6008: removing the optional dependency on bouncy castle
 
 #### Dependency Upgrade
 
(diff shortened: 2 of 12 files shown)


rohanKanojia:
@coopstah13 : Could you please review it to see if it meets your expectations?

shawkins:
@rohanKanojia unfortunately it still does have an issue. The remaining failures are related to: 
```
java.security.InvalidAlgorithmParameterException: Curve not supported: secp256k1 (1.3.132.0.10)
```
JDK 15+ dropped support for secp256k1, but we're using a key from that algorithm for KubernetesClientLoadWithFipsProviderTest.

We can either:
- Bring back a little bit of our reliance on bouncy castle - however we don't need an optional dependency. If the user installs bc / bcfips as a provider this will work if we instead call KeyPairGenerator.getInstance("EC", bc_provider)
- Drop support for this EC - do we know what is in use for supported versions of kubernetes?

coopstah13:
I think it would be acceptable to leave fall-back code with the optional dependency for supporting that. Basically, first you'd try to do it using JDK internals, but maybe catch the exception and fall back to the Callable code which lets it dynamically link to the BC dependency and leave BC as an optional dependency defined in your pom.

Appreciate the effort on this.

shawkins:
I've opted for suggesting that BC be installed as a provider if there's a problem with the algorithm. This approach means that we won't have any direct compile / runtime dependency on BC.

kubernetes-client-api/pom.xml
```diff
@@ -160,12 +160,10 @@
     <dependency>
       <groupId>org.bouncycastle</groupId>
       <artifactId>bcprov-jdk18on</artifactId>
-      <optional>true</optional>
     </dependency>
     <dependency>
       <groupId>org.bouncycastle</groupId>
       <artifactId>bcpkix-jdk18on</artifactId>
-      <optional>true</optional>
```
coopstah13, line 168: this change seems odd to me

shawkins: Why? It's now a test dependency.

manusa (approved):
LGTM, nice job.
Thx!

PR #6053, opened 2024-06-14
Remove Generex dependency

### LINKED ISSUE #6052: Remove no longer maintained dependency com.github.mifmif:generex

Auditing health and licenses for dependencies of Keycloak Operator this dependency was detected. The last release of the dependency was [2016-10-30](https://central.sonatype.com/artifact/com.github.mifmif/generex/versions), and can be considered abandoned by its maintainers as there has not been any activity. There's even a rather old open issue asking if the [project is still alive](https://github.com/mifmif/Generex/issues/53), but no response in nearly 5 years.

generex also has a transitive [dependency on automaton that is more than 10 years old](https://central.sonatype.com/artifact/dk.brics.automaton/automaton/versions), with an open [request to update](https://github.com/mifmif/Generex/issues/61).

Remove the use of the dependency that is no longer actively maintained.

_No response_

### Additional context

_No response_

andreaTP:
## Description
Fix #6052 

## Type of change

## Checklist

### DIFF
--- a/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/core/TemplateOperationsImpl.java
+++ b/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/core/TemplateOperationsImpl.java
@@ -183,6 +183,8 @@ public class TemplateOperationsImpl
             parameterValue = valuesMap.get(parameterName);
           } else if (Utils.isNotNullOrEmpty(parameter.getValue())) {
             parameterValue = parameter.getValue();
+          } else if (EXPRESSION.equals(parameter.getGenerate())) {
+            parameterValue = parameter.getFrom();
           } else if (parameter.getRequired() == null || !parameter.getRequired()) {
             parameterValue = "";
           } else {
--- a/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/core/TemplateOperationsImpl.java
+++ b/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/core/TemplateOperationsImpl.java
@@ -16,7 +16,6 @@
 package io.fabric8.openshift.client.dsl.internal.core;
 
 import com.fasterxml.jackson.core.type.TypeReference;
-import com.mifmif.common.regex.Generex;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.KubernetesList;
 import io.fabric8.kubernetes.api.model.KubernetesListBuilder;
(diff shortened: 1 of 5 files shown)


oscerd (approved):

andreaTP:
Seems to be not used :crossed_fingers: let see if the CI confirms.

andreaTP:
Back to draft, but we have reproducers now.

openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/core/TemplateOperationsImpl.java
```diff
@@ -185,8 +184,7 @@ public KubernetesList processLocally(Map<String, String> valuesMap) {
           } else if (Utils.isNotNullOrEmpty(parameter.getValue())) {
             parameterValue = parameter.getValue();
           } else if (EXPRESSION.equals(parameter.getGenerate())) {
-            Generex generex = new Generex(parameter.getFrom());
-            parameterValue = generex.random();
+            parameterValue = parameter.getFrom();
```
andreaTP, line 187: This is not going to generate a valid template instance, at the same time:
- generating a random string can easily lead to errors
- the random string cannot be tracked

Alternatives are in:
1. substitute with an alternative library, but I'm not convinced about any of those:
   - https://github.com/curious-odd-man/RgxGen?tab=readme-ov-file
   - https://github.com/fmselab/mutrex
   - https://github.com/six2six/bfgex
   - https://github.com/agarciadom/xeger
 2. Roll our own implementation, either as an external micro library or within this project: the code is not going to be much, but OpenShift Templates are deprecated, so, we can even deprecate/remove this feature

andreaTP: cc. @manusa @shawkins @metacosm

manusa: > This is not going to generate a valid template instance

Why is this?

> - generating a random string can easily lead to errors
> - the random string cannot be tracked

The purpose of this method is really to process locally the templates as if they were being processed by the OpenShift cluster.

It really doesn't matter if the string can't be tracked or if the random string is invalid (maybe because the provided expression is wrong).
In the end, this would be the user's responsibility.

```yaml
parameters:
- description: Password used for Redis authentication
  from: '[A-Z0-9]{8}'
  generate: expression
  name: REDIS_PASSWORD
```
https://docs.openshift.com/container-platform/4.15/openshift_images/using-templates.html#templates-writing_using-templates


What we want is to be able to support this.

From the [docs](https://docs.openshift.com/container-platform/4.15/rest_api/template_apis/template-template-openshift-io-v1.html) we can read the constraints for this expression:

> generate specifies the generator to be used to generate random string from an input value specified by From field. The result string is stored into Value field. If empty, no generator is being used, leaving the result Value untouched. Optional.
>
> The only supported generator is "expression", which accepts a "from" value in the form of a simple regular expression containing the range expression "[a-zA-Z0-9]", and the length expression "a{length}".
> 
> Examples:
>
> | from | value |
> |------|-------|
> "test[0-9]{1}x" | "test7x"
> "[0-1]{8}" | "01001100"
> "0x[A-F0-9]{4}" | "0xB3AF"
> "[a-zA-Z0-9]{8}" | "hW4yQU5i"


It seems that a simple library or maybe even rolling our own utility method should suffice to handle this.

andreaTP:
Superseded by #6060

Issue #6097, opened 2024-07-02
Release Fabric8 Kubernetes Client 6.13.2


manusa:
## Backports

- [x] #6101
- [x] #6102
- [x] #6104
- [x] #6108
- [x] #6114
- [x] #6088
- [x] #6089
- [x] #6148
- [x] #6157
- [x] #6217
- [x] #6213 
- [x] #6197
- [x] #6153 
- [x] #6144

Issue to track the required tasks to release Fabric8 Kubernetes Client 6.13.2.
- [x] https://github.com/fabric8io/kubernetes-client/milestone/76
- [x] Prepare release
  https://github.com/fabric8io/kubernetes-client/releases/tag/v6.13.2
- [x] Quarkus Version Bump
  https://github.com/quarkusio/quarkus/pull/42450
- [x] Create issue for next release (and milestone if needed)
  https://github.com/fabric8io/kubernetes-client/issues/6242

PR #6144, opened 2024-07-17
Expands the HTTP interceptor API to include a call back for failed connection attempts

### LINKED ISSUE #6143: Interceptor not invoked when connection refused
### Describe the bug

I'm trying to migrate the Apache Flink kubernetes operator from using an [OkHttp interceptor](https://github.com/apache/flink-kubernetes-operator/blob/main/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/KubernetesClientMetrics.java) to using the new fabric8 http interceptor API. Which seems to have gone reasonably smoothly apart from one final test: https://github.com/apache/flink-kubernetes-operator/blob/aaf26aec8bdfedff0116c5897bd89b17b4454606/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/metrics/KubernetesClientMetricsTest.java#L376

Which is failing because `afterFailure` is not invoked when there is no http response (I think). I've added a failing [unit test](https://github.com/SamBarker/kubernetes-client/blob/a60dc211cb67cfa358be9ec568483c8cae444f1d/kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractInterceptorTest.java#L175-L195) in kubernetes-client which demonstrates what I expected to work.

So I'm wondering if the test I've written should work? Or have I missed something in how the API is expected to be used

SNAPSHOT

### Steps to reproduce

1. checkout https://github.com/SamBarker/kubernetes-client/tree/afterFailureInvocation
1. run `io.fabric8.kubernetes.client.http.AbstractInterceptorTest#afterHttpFailureRemoteOffline` (I've been testing with the OkHttp client but I don't think it matters)
1. see test fail

### Expected behavior

The test passes

### Runtime

Kubernetes (vanilla)

1.25.3@latest

### Environment

macOS

_No response_

### Additional context

_No response_

SamBarker:
## Description

Expands the HTTP interceptor API to include a call back for failed connection attempts as discussed in #6143 

Fixes #6143

## Type of change

Depends on the definition of breaking change as it adds a new `default` method to an interface which could be seen as a breaking change for any implementation that already has that method defined but practically speaking this is backwards compatible. 

## Checklist

### DIFF
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/Interceptor.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/Interceptor.java
@@ -19,6 +19,13 @@ import java.nio.ByteBuffer;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 
+/**
+ * A collection of callback methods invoked through the various stages of the HTTP request lifecycle.
+ * Each invocation of {@link Interceptor#before(BasicBuilder, HttpRequest, RequestTags)} will be matched with a call to one of
+ * {@link Interceptor#afterConnectionFailure(HttpRequest, Throwable)} or
+ * {@link Interceptor#after(HttpRequest, HttpResponse, AsyncBody.Consumer)}.
+ * Callbacks that lead to a request being sent allow for that request to be customised.
+ */
 public interface Interceptor {
 
   interface RequestTags {
@@ -63,7 +70,10 @@ public interface Interceptor {
   }
 
   /**
-   * Called after a websocket failure or by default from a normal request
+   * Called after a websocket failure or by default from a normal request.
+   * <p>
+   * Failure is determined by HTTP status code and will be invoked in addition to {@see Interceptor#after(HttpRequest,
+   * HttpResponse, AsyncBody.Consumer)}
    *
    * @param builder used to modify the request
    * @param response the failed response
@@ -75,7 +85,10 @@ public interface Interceptor {
 
   /**
    * Called after a non-websocket failure
-   *
+   * <p>
+   * Failure is determined by HTTP status code and will be invoked in addition to {@see Interceptor#after(HttpRequest,
+   * HttpResponse, AsyncBody.Consumer)}
+   * 
    * @param builder used to modify the request
    * @param response the failed response
    * @return true if the builder should be used to execute a new request
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/Interceptor.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/Interceptor.java
@@ -84,4 +84,15 @@ public interface Interceptor {
     return afterFailure((BasicBuilder) builder, response, tags);
   }
 
+  /**
+   * Called after a connection attempt fails.
+   * <p>
+   * This method will be invoked on each failed connection attempt so can be invoked multiple times for the same request.ID.
+   *
+   * @param request the HTTP request.
+   * @param failure the Java exception that caused the failure.
+   */
+  default void afterConnectionFailure(HttpRequest request, Throwable failure) {
+  }
+
 }
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/Interceptor.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/Interceptor.java
@@ -100,7 +100,7 @@ public interface Interceptor {
   /**
    * Called after a connection attempt fails.
    * <p>
-   * This method will be invoked on each failed connection attempt so can be invoked multiple times for the same request.ID.
+   * This method will be invoked on each failed connection attempt.
    *
    * @param request the HTTP request.
    * @param failure the Java exception that caused the failure.
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/Interceptor.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/Interceptor.java
@@ -72,7 +72,7 @@ public interface Interceptor {
   /**
    * Called after a websocket failure or by default from a normal request.
    * <p>
-   * Failure is determined by HTTP status code and will be invoked in addition to {@see Interceptor#after(HttpRequest,
+   * Failure is determined by HTTP status code and will be invoked in addition to {@link Interceptor#after(HttpRequest,
    * HttpResponse, AsyncBody.Consumer)}
    *
    * @param builder used to modify the request
@@ -86,7 +86,7 @@ public interface Interceptor {
   /**
    * Called after a non-websocket failure
    * <p>
-   * Failure is determined by HTTP status code and will be invoked in addition to {@see Interceptor#after(HttpRequest,
+   * Failure is determined by HTTP status code and will be invoked in addition to {@link Interceptor#after(HttpRequest,
    * HttpResponse, AsyncBody.Consumer)}
    * 
    * @param builder used to modify the request
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/StandardHttpClient.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/StandardHttpClient.java
@@ -177,24 +177,29 @@ public abstract class StandardHttpClient<C extends HttpClient, F extends HttpCli
               }
             }
           } else {
-            final Throwable finalThrowable;
-            if (throwable instanceof CompletionException) {
-              finalThrowable = throwable.getCause();
-            } else {
-              finalThrowable = throwable;
-            }
-            builder.interceptors.forEach((s, interceptor) -> interceptor.afterConnectionFailure(request, finalThrowable));
-            if (finalThrowable instanceof IOException) {
+            final Throwable actualCause = unwrapCompletionException(throwable);
+            builder.interceptors.forEach((s, interceptor) -> interceptor.afterConnectionFailure(request, actualCause));
+            if (actualCause instanceof IOException) {
               // TODO: may not be specific enough - incorrect ssl settings for example will get caught here
               LOG.debug(
                   String.format("HTTP operation on url: %s should be retried after %d millis because of IOException",
                       uri, retryInterval),
-                  finalThrowable);
+                  actualCause);
               return true;
             }
           }
           return false;
         });
+  }
+
+  static Throwable unwrapCompletionException(Throwable throwable) {
+    final Throwable actualCause;
+    if (throwable instanceof CompletionException) {
+      actualCause = throwable.getCause();
+    } else {
+      actualCause = throwable;
+    }
+    return actualCause;
   }
 
   static long retryAfterMillis(HttpResponse<?> httpResponse) {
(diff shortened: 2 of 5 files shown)


kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/StandardHttpClient.java
```diff
@@ -108,6 +108,13 @@ private CompletableFuture<HttpResponse<AsyncBody>> consumeBytesOnce(StandardHttp
     final Consumer<List<ByteBuffer>> effectiveConsumer = consumer;
 
     CompletableFuture<HttpResponse<AsyncBody>> cf = consumeBytesDirect(effectiveRequest, effectiveConsumer);
+    cf.exceptionally(throwable -> {
+      builder.interceptors.forEach((s, interceptor) -> interceptor.afterConnectionFailure(effectiveRequest, throwable));
+      if (throwable instanceof CompletionException) {
+        throw (CompletionException) throwable;
+      }
```
SamBarker, line 115: sonar is complaining this branch is untested, but I can't work out a way to inject a `CompletionException`.  Any pointers gratefully received.

SamBarker: Came up with https://github.com/SamBarker/kubernetes-client/commit/43f88fab5597a3d85bfedda2dab1033259f238de  which does trigger the branch using the Jetty client.

SamBarker: apparently not according to sonar. It seemed to work in the debugger locally.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/Interceptor.java
```diff
@@ -84,4 +97,15 @@ default CompletableFuture<Boolean> afterFailure(HttpRequest.Builder builder, Htt
     return afterFailure((BasicBuilder) builder, response, tags);
   }
 
+  /**
+   * Called after a connection attempt fails.
+   * <p>
+   * This method will be invoked on each failed connection attempt so can be invoked multiple times for the same request.ID.
```
manusa, line 103: I'm not sure if this information might end up being misleading.
What's the motivation to add it?

What I mean is that users (who don't read carefully) might assume the opposite (for a given operation, for each connection failure, request id will be the same) which isn't true.

httpclient-jetty/src/main/java/io/fabric8/kubernetes/client/jetty/JettyAsyncResponseListener.java
```diff
@@ -103,6 +103,11 @@ public void onContent(Response response, ByteBuffer content, Callback callback)
     }
   }
 
+  @Override
+  public void onFailure(Response response, Throwable failure) {
+    asyncResponse.completeExceptionally(failure);
```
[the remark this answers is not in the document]
manusa: I can't remember, but I do recall using the Adapter to avoid implementing all methods.
It's certainly strange that the onFailure wasn't implemented before. However, the HttpClient behavior did change a little after the original implementation (especially the retries part).

Anyway, it does seem OK to have this.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/StandardHttpClient.java
SamBarker: Urgh yeah I hadn't thought about the multiple modules pushing in different directions. Sonar only reports a single number on the `StandardHttpClient` so a bit hard to know. I suspect on balance its probably better to have the test than not but I only added it to try and pass the coverage stats for the PR so not wedded to it as the core flows are tested.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/Interceptor.java
SamBarker: Yeah I think its so I've removed it. 

While trying to update it I attempted to write a test that caused an unclean disconnect from a websocket connection but wasn't able to. If you think thats important to have do you have any thoughts on how I might trigger it?

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/Interceptor.java
```diff
@@ -63,7 +70,10 @@ default AsyncBody.Consumer<List<ByteBuffer>> consumer(AsyncBody.Consumer<List<By
   }
 
   /**
-   * Called after a websocket failure or by default from a normal request
+   * Called after a websocket failure or by default from a normal request.
+   * <p>
+   * Failure is determined by HTTP status code and will be invoked in addition to {@see Interceptor#after(HttpRequest,
```
manusa, line 75: This Javadoc is not ok, you probably meant `@link` instead of `@see`?

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/Interceptor.java
```diff
@@ -75,7 +85,10 @@ default CompletableFuture<Boolean> afterFailure(BasicBuilder builder, HttpRespon
 
   /**
    * Called after a non-websocket failure
-   *
+   * <p>
+   * Failure is determined by HTTP status code and will be invoked in addition to {@see Interceptor#after(HttpRequest,
```
manusa, line 89: This Javadoc is not ok, you probably meant `@link` instead of `@see`?

manusa (changes requested):
Besides the Javadoc issue I think everything else looks OK.

@shawkins any comments?

BTW, even though this is adding a feature, I'm going to consider it as a bug fix since it's addressing the lack of feature parity of Fabric8-OkHttp interceptors.

This way we can include it as part of #6097 and release it in 6.13.2

httpclient-jetty/src/main/java/io/fabric8/kubernetes/client/jetty/JettyAsyncResponseListener.java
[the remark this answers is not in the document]
shawkins: It looks like overriding onComplete ensured timely completion - but does not have appropriate exception handling. 

What we are probably looking for here instead is using onComplete Result.getRequestFailure - that maps most closely to these types of exceptions. There is also Result.getResponseFailure which should be the same as what is passed to onFailure.

shawkins (commented):
Overall looks good, but we may need more refinement of Jetty exception handling.

httpclient-jetty/src/main/java/io/fabric8/kubernetes/client/jetty/JettyAsyncResponseListener.java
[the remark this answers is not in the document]
SamBarker: Oh this gets fun 😅 

Updating `onComplete` to respect `Result.isSucceeded()` or `Result.isFailed()` causes [the expectContinue](https://github.com/SamBarker/kubernetes-client/blob/25ecf8460eb132ae2945342b33f64a4240a2b91c/kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractHttpPostTest.java#L130) test to fail, because Jetty considers a 404 in response to a request with `Expect=100-continue` to be a [failure](https://github.com/jetty/jetty.project/blob/2881fe9ab569946f36ec9dac6c59205b4b969007/jetty-client/src/main/java/org/eclipse/jetty/client/ContinueProtocolHandler.java#L84). So i've tried adding `server.expect().post().withPath("/post-expect-continue").andReply(100, request -> "").always();` which causes the server to reply appropriately to the `POST` request. However that breaks `expectContinue` for `jetty`, `OkHttp` and `VertX` (there doesn't seem to be a JDK version 👀 ) clients tests 😢 .

In the jetty case the test now fails because the response is considered pending and there is no further `demand` so the request future never completes. I haven't debugged the others to work out whats going wrong but I suspect its something similar. 

Reading [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/100) on `100-continue` suggests the jetty implementation is wrong as it attaches the body to the `POST` request at the same time as the `Expect=100-continue` header. Whereas MDN suggests that it should send `Expect=100-continue` and wait for the response to  that before continuing by sending the body. 

I'm nothing like close enough to the details of the Kube API server to understand if 100-continue is commonly used or if it has a problem with including the `POST` body on the initial request. However this is starting to feel like a rabbit hole and not really in scope for this PR and thus what the implications of this are in practice. 

Thoughts @manusa, @shawkins ?

SamBarker:
> BTW, even though this is adding a feature, I'm going to consider it as a bug fix since it's addressing the lack of feature parity of Fabric8-OkHttp interceptors.
This way we can include it as part of https://github.com/fabric8io/kubernetes-client/issues/6097 and release it in 6.13.2

Sounds great from my POV

httpclient-jetty/src/main/java/io/fabric8/kubernetes/client/jetty/JettyAsyncResponseListener.java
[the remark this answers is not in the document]
SamBarker: Ahh the JDK client explicitly disables the `expectContinue` test.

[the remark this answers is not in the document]
SamBarker: > Reading [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/100) on 100-continue suggests the jetty implementation is wrong as it attaches the body to the POST request at the same time as the Expect=100-continue header. Whereas MDN suggests that it should send Expect=100-continue and wait for the response to that before continuing by sending the body.

After further debugging the jetty Impl is right, jetty handles delaying the body if the Expect header is set but I'm still not seeing the request complete.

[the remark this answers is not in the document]
SamBarker: Hmm turns out I was very wrong about what was going on after much debugging I have 
https://github.com/fabric8io/kubernetes-client/pull/6197/files 

Which changes the default socketPolicy for the mockWebserver to `EXPECT_CONTINUE` from `KEEP_OPEN`. I haven't found a way to make it conditional on the headers as its too late by the time dispatch is called. I'm not at all sure what implications that change has.

I'm happy to merge that PR into this one if preferred but I think they are really independent. I'm happy to clean up that PR and its description and or raise an issue to cover it if the consensus is to keep it independent.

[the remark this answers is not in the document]
manusa: Nice work, I think it's better to keep them separate.

SamBarker:
I plan to clean this up once https://github.com/fabric8io/kubernetes-client/pull/6197 is merged, but I think the coverage issues will go away once thats merged.

manusa:
Hi @SamBarker,
#6197 is finally in. Is there something else left to do on this PR? Shall we merge it too?

SamBarker:
> Hi @SamBarker,
> 
> #6197 is finally in. Is there something else left to do on this PR? Shall we merge it too?

I've added an extra commit to get the sonar coverage report passing so yes! Let's get this merged.

Any info on a timeline for the next 6.13 release?

rohanKanojia (approved):

manusa:
> Any info on a timeline for the next 6.13 release?

I was just waiting for this one (#6097), so probably in a couple of hours or early Monday.

manusa (approved):
LGTM, thx!

PR #6197, opened 2024-07-25
Fix jetty completion handling to forward failures

SamBarker:
## Description
When fixing #6143 it was observed that the Jetty client was not handling failures well. This PR fails the appropriate futures when failures are encountered.

## Type of change

## Checklist

### DIFF
--- a/httpclient-jetty/src/main/java/io/fabric8/kubernetes/client/jetty/JettyAsyncResponseListener.java
+++ b/httpclient-jetty/src/main/java/io/fabric8/kubernetes/client/jetty/JettyAsyncResponseListener.java
@@ -72,7 +72,11 @@ public abstract class JettyAsyncResponseListener extends Response.Listener.Adapt
 
   @Override
   public void onComplete(Result result) {
-    asyncBodyDone.complete(null);
+    if (result.isSucceeded()) {
+      asyncBodyDone.complete(null);
+    } else {
+      asyncBodyDone.completeExceptionally(result.getRequestFailure());
+    }
   }
 
   public CompletableFuture<HttpResponse<AsyncBody>> listen(Request request) {
--- a/httpclient-jetty/src/main/java/io/fabric8/kubernetes/client/jetty/JettyAsyncResponseListener.java
+++ b/httpclient-jetty/src/main/java/io/fabric8/kubernetes/client/jetty/JettyAsyncResponseListener.java
@@ -79,6 +79,11 @@ public abstract class JettyAsyncResponseListener extends Response.Listener.Adapt
     }
   }
 
+  @Override
+  public void onFailure(Response response, Throwable failure) {
+    asyncResponse.completeExceptionally(failure);
+  }
+
   public CompletableFuture<HttpResponse<AsyncBody>> listen(Request request) {
     request.send(this);
     return asyncResponse.thenApply(HttpResponse.class::cast);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Fix #6038: Support for Gradle configuration cache
 * Fix #6110: VolumeSource (and other file mode fields) in Octal are correctly interpreted
 * Fix #6215: Suppressing rejected execution exception for port forwarder
+* Fix #6197: JettyHttp client error handling improvements. 
 
 #### Improvements
 * Fix #6008: removing the optional dependency on bouncy castle
--- a/httpclient-jdk/src/test/java/io/fabric8/kubernetes/client/jdkhttp/JdkHttpClientPostTest.java
+++ b/httpclient-jdk/src/test/java/io/fabric8/kubernetes/client/jdkhttp/JdkHttpClientPostTest.java
@@ -18,6 +18,8 @@ package io.fabric8.kubernetes.client.jdkhttp;
 import io.fabric8.kubernetes.client.http.AbstractHttpPostTest;
 import io.fabric8.kubernetes.client.http.HttpClient;
 
+import java.io.IOException;
+
 @SuppressWarnings("java:S2187")
 public class JdkHttpClientPostTest extends AbstractHttpPostTest {
   @Override
@@ -30,5 +32,10 @@ public class JdkHttpClientPostTest extends AbstractHttpPostTest {
     // Disabled
     // Apparently the JDK sets the Expect header to 100-Continue which is not according to spec.
     // We should consider overriding the header manually instead.
+  }
+
+  @Override
+  protected Class<? extends Exception> getConnectionFailedExceptionType() {
+    return IOException.class;
   }
 }
--- a/httpclient-jdk/src/test/java/io/fabric8/kubernetes/client/jdkhttp/JdkHttpClientPutTest.java
+++ b/httpclient-jdk/src/test/java/io/fabric8/kubernetes/client/jdkhttp/JdkHttpClientPutTest.java
@@ -18,10 +18,17 @@ package io.fabric8.kubernetes.client.jdkhttp;
 import io.fabric8.kubernetes.client.http.AbstractHttpPutTest;
 import io.fabric8.kubernetes.client.http.HttpClient;
 
+import java.io.IOException;
+
 @SuppressWarnings("java:S2187")
 public class JdkHttpClientPutTest extends AbstractHttpPutTest {
   @Override
   protected HttpClient.Factory getHttpClientFactory() {
     return new JdkHttpClientFactory();
+  }
+
+  @Override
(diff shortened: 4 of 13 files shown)


manusa (approved):
LGTM, thx!

httpclient-jetty/src/main/java/io/fabric8/kubernetes/client/jetty/JettyAsyncResponseListener.java
```diff
@@ -72,7 +72,11 @@ public void onHeaders(Response response) {
 
   @Override
   public void onComplete(Result result) {
-    asyncBodyDone.complete(null);
+    if (result.isSucceeded()) {
+      asyncBodyDone.complete(null);
+    } else {
+      asyncBodyDone.completeExceptionally(result.getRequestFailure());
```
shawkins, line 78: Is the intention to also have onFailure overriden? Because there could still be a response failure. From the javadocs, the advantage of onFailure is that it can be called sooner in some circumstances, like large file uploads.

junit/mockwebserver/src/main/java/io/fabric8/mockwebserver/dsl/ReturnOrWebsocketable.java
```diff
@@ -15,5 +15,6 @@
  */
 package io.fabric8.mockwebserver.dsl;
 
-public interface ReturnOrWebsocketable<T> extends Returnable<T>, WebSocketable<WebSocketSessionBuilder<T>>, Replyable<T> {
+public interface ReturnOrWebsocketable<T>
+    extends Returnable<T>, WebSocketable<WebSocketSessionBuilder<T>>, Replyable<T>, Failable<T> {
```
SamBarker, line 19: This was the minimal change to make `io.fabric8.kubernetes.client.http.AbstractHttpPostTest#expectFailure` possible without shutting down the mock webserver.

junit/mockwebserver/src/main/java/io/fabric8/mockwebserver/internal/MockServerExpectationImpl.java
```diff
@@ -224,6 +234,21 @@ private void enqueue(ServerRequest req, ServerResponse resp) {
...
+          return repeatable;
+        }
+
+        @Override
+        public MockResponse toMockResponse(RecordedRequest recordedRequest) {
+          MockResponse mockResponse = new MockResponse();
+          mockResponse.setResponseCode(500);
+          mockResponse.setSocketPolicy(SocketPolicy.DISCONNECT_AFTER_REQUEST);
+          return mockResponse;
+        }
```
SamBarker, line 250: This is quite blunt but and feels limited (there are many other possible socket policies) but I don't see a clean way of exposing them through the DSL and I don't have a specific use case for any others at the moment.

SamBarker (commented):
I'd like to call attention to 9c33911d284a4f82905a1a142cac217331cbc8c3 as it makes testing the failure call backs possible but I'm not sure if thats a direction the project wants to take the `mockwebserver` in. So I have kept that as isolated commit so it can be dropped without compromising the core change.

The fact that each `expectFailure` test takes 20s to complete gives me pause but to change that involves exposing configuration for the retry behaviour which seems like a bigger deal than a slow test.

manusa:
Some notes about the MockWebServer.

For version 7.0.0 we're moving it away from OkHttp.
You can check a spike/PoC of the new implementation in #5632

Regarding the failure testing (I still haven't checked exactly what you're doing), but I kind of recall writing failure expectations without the need to modify the MockWebServer.

I think these are the ones:

https://github.com/eclipse-jkube/jkube/blob/0f530e0a6340dcb3dbc3b5ea3c4af40516dd1a17/jkube-kit/helm/src/test/java/org/eclipse/jkube/kit/resource/helm/oci/OCIRegistryClientTest.java#L107-L110

https://github.com/eclipse-jkube/jkube/blob/0f530e0a6340dcb3dbc3b5ea3c4af40516dd1a17/jkube-kit/helm/src/test/java/org/eclipse/jkube/kit/resource/helm/TestMockResponseProvider.java

As I said, I haven't checked (yet) what you're trying to accomplish, but maybe this approach could be vaild for you.

SamBarker:
Thanks @manusa 
>Regarding the failure testing (I still haven't checked exactly what you're doing), but I kind of recall writing failure expectations without the need to modify the MockWebServer.
Hmm I'm maybe using a different definition of failure, I want the connection to fail rather than the request itself. 

I think the current impl is broken if you specify `once` instead of `always` as it doesn't implement equals and hashcode... I've got a some changes locally but they are still failing my test so I haven't pushed them (but I've run out of time today as the kids are climbing the walls.)

manusa:
AFK
I think you could still use a lower-level expectation for the MockWebServer, i.e. provide your own response, but I'd need to check.
Also, if you want the connection to fail, maybe not use the mock web server at all?
(e.g. just a plain Java HTTP server)

junit/mockwebserver/src/main/java/io/fabric8/mockwebserver/dsl/ReturnOrWebsocketable.java
SamBarker: I've killed off the changes to the mockwebserver and replaced them with a dead socket so the initial connection always fails.

junit/mockwebserver/src/main/java/io/fabric8/mockwebserver/internal/MockServerExpectationImpl.java
SamBarker: I've killed off the changes to the mockwebserver and replaced them with a dead socket so the initial connection always fails.

rohanKanojia (approved):

rohanKanojia:
@SamBarker : I might be wrong but this test failure on [windows gh action job](https://github.com/fabric8io/kubernetes-client/actions/runs/10159183949/job/28092852231?pr=6197) looks related to your changes. 

```
Error:  Failures: 
Error:    OkHttpPostTest>AbstractHttpPostTest.expectFailure:179 
Expecting a throwable with cause being an instance of:
  java.net.ConnectException
but was an instance of:
  java.io.InterruptedIOException
Throwable that failed the check:

Error:    OkHttpPutTest>AbstractHttpPutTest.expectFailure:115 
Expecting a throwable with cause being an instance of:
  java.net.ConnectException
but was an instance of:
  java.io.InterruptedIOException
Throwable that failed the check:
```

Could you please check if this is related?

SamBarker:
>@SamBarker : I might be wrong but this test failure on [windows gh action job](https://github.com/fabric8io/kubernetes-client/actions/runs/10159183949/job/28092852231?pr=6197) looks related to your changes.

Yeah sorry its taken a few days to organise a windows box and get it into a state to reproduce the failure. The TL;DR the exceptions propagated are completely different on windows to macos/linux. So the choices are:
1. Disabled the tests on windows (see https://github.com/fabric8io/kubernetes-client/pull/6197/commits/2ee636f8088974310b11f41aa67a048cf51d437b)
2. Relax the test to just assert that the future fails. 

I lean slightly to option 1 as that gives a better test for the other platforms but I'm open to better suggestions or other ways to test the behaviour.

SamBarker:
> > @SamBarker : I might be wrong but this test failure on [windows gh action job](https://github.com/fabric8io/kubernetes-client/actions/runs/10159183949/job/28092852231?pr=6197) looks related to your changes.
> 
> Yeah sorry its taken a few days to organise a windows box and get it into a state to reproduce the failure. The TL;DR the exceptions propagated are completely different on windows to macos/linux. So the choices are:
> 
> 1. Disabled the tests on windows (see [2ee636f](https://github.com/fabric8io/kubernetes-client/commit/2ee636f8088974310b11f41aa67a048cf51d437b))
> 2. Relax the test to just assert that the future fails.
> 
> I lean slightly to option 1 as that gives a better test for the other platforms but I'm open to better suggestions or other ways to test the behaviour.

I came up with a better option in https://github.com/fabric8io/kubernetes-client/pull/6197/commits/ff287b70ff3222120d1fc6c13f4add367093890e realising I could use the Junit to detect the current platform and thus return the right exception type.

I've now tested the jetty client against Openshift /CRC. So I think this is good to go. 

@manusa @rohanKanojia anything else you need from me before merging?

rohanKanojia:
@SamBarker : Thanks, Marc is on PTO this week. We'll try to get this merged next week.

SamBarker:
@shawkins @manusa Anything I can do to move this forward?

shawkins (approved):
LGTM, thank you @SamBarker

PR #6230, opened 2024-08-08
feat: add functionality to multiply Quantity by a scalar value

bachmanity1:
## Description
The Go client has added the ability to multiply a Quantity by a scalar value: kubernetes/kubernetes#117411. I think it would be useful to have this feature in the Java client as well.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/Quantity.java
+++ b/kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/Quantity.java
@@ -408,6 +408,18 @@ public class Quantity implements Serializable, Comparable<Quantity> {
     return op(y, BigDecimal::subtract);
   }
 
+  /**
+   * Multiplies the quantity by the specified scalar multiplicand.
+   *
+   * @param multiplicand the scalar value to multiply by
+   * @return a new Quantity resulting from the multiplication of this quantity by the scalar multiplicand
+   */
+  public Quantity multiply(int multiplicand) {
+    BigDecimal numericalAmount = getNumericalAmount();
+    numericalAmount = numericalAmount.multiply(BigDecimal.valueOf(multiplicand));
+    return fromNumericalAmount(numericalAmount, format);
+  }
+
   Quantity op(Quantity y, BiFunction<BigDecimal, BigDecimal, BigDecimal> func) {
     BigDecimal numericalAmount = this.getNumericalAmount();
     numericalAmount = func.apply(numericalAmount, y.getNumericalAmount());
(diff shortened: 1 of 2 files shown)


rohanKanojia (approved):

kubernetes-model-generator/kubernetes-model-core/src/main/java/io/fabric8/kubernetes/api/model/Quantity.java
```diff
@@ -408,6 +408,18 @@ public Quantity subtract(Quantity y) {
...
+   * Multiplies the quantity by the specified scalar multiplicand.
+   *
+   * @param multiplicand the scalar value to multiply by
+   * @return a new Quantity resulting from the multiplication of this quantity by the scalar multiplicand
+   */
+  public Quantity multiply(int multiplicand) {
+    BigDecimal numericalAmount = getNumericalAmount();
+    numericalAmount = numericalAmount.multiply(BigDecimal.valueOf(multiplicand));
+    return fromNumericalAmount(numericalAmount, format);
+  }
```
manusa, line 421: Looks inconsistent with the rest of the operation methods defined in the class (namely add and subtract).

Wouldn't a `public Quantity multiply(Quantity y);` suffice for the described needs?

bachmanity1: In the Go client reviewers suggested that multiplicand should be a scalar rather than a Quantity. I'm not sure how to handle cases like `1G * 1n`, what should be the format of the result? Besides that, I don't think there are any practical use cases for Quantity by Quantity multiplications. When a multiplicand is a scalar, one the possible use cases may be to calculate total memory requests of a deployment, which would be a pod memory request multiplied by the replica count.

manusa: > I don't think there are any practical use cases for Quantity by Quantity multiplications. When a multiplicand is a scalar, one the possible use cases may be to calculate total memory requests of a deployment, which would be a pod memory request multiplied by the replica count.

Agreed, this makes sense from a semantic perspective.

> In the Go client reviewers suggested that multiplicand should be a scalar rather than a Quantity

For reference: https://github.com/kubernetes/kubernetes/pull/117411/files#diff-c3c7dcd71c7c415cbf21c3d732510021ca5a11b46af7b633673aab30c33c717d

> I'm not sure how to handle cases like 1G * 1n, what should be the format of the result?

This situation is mostly the same for addition and subtraction of completely different types(CPU, memory)

The current implementation **should** be capable of taking care of this.

```java
  public Quantity multiply(Quantity y) {
    return op(y, BigDecimal::multiply);
  }
```

Anyway, it really doesn't make sense to compute a Quantity*Quantity from the user's perspective.

manusa (approved):
LGTM, thx!

bachmanity1:
can we get this merged?

manusa:
> can we get this merged?

Yes, but note that 7.0.0 will take a while to get released.

Issue #5859, opened 2024-04-03
CRDGenerator: Overview about validation constraints / Kubernetes OpenAPI schema properties


baloo42:
### Is your enhancement related to a problem? Please describe

At the moment the CRD-Generator does not allow to set all possible details to the resulting CRDs.

### Describe the solution you'd like

The CRD-Generator supports all Kubernetes features to describe a CRD.

### Additional context

This issue aims to provide an overview of the current state and what we plan to implement.

- Probably superseeds #3768.
- Using custom annotations instead of reusing existing annotations has been introduced in 
  #4348 
  (Discussion: #4298)

> [!IMPORTANT]
> This overview is now targeting api-v2!

#### Current state

| Annotation                                                      | Description                                                                                         |
|-----------------------------------------------------------------|-----------------------------------------------------------------------------------------------------|
| `com.fasterxml.jackson.annotation.JsonProperty`                 | The field is named after the provided value instead of looking up the java field name               |
| `com.fasterxml.jackson.annotation.JsonPropertyDescription`      | The provided text is be embedded in the `description` of the field                                  |
| `com.fasterxml.jackson.annotation.JsonIgnore`                   | The field is ignored                                                                                |
| `io.fabric8.crd.generator.annotation.PreserveUnknownFields`     | The field have `x-kubernetes-preserve-unknown-fields: true` defined                                 |
| `com.fasterxml.jackson.annotation.JsonAnyGetter`                | The corresponding object have `x-kubernetes-preserve-unknown-fields: true` defined                  |
| `com.fasterxml.jackson.annotation.JsonAnySetter`                | The corresponding object have `x-kubernetes-preserve-unknown-fields: true` defined                  |
| `io.fabric8.generator.annotation.Min`                           | The field's `minimum` value                                                                         |
| `io.fabric8.generator.annotation.Max`                           | The field's `maximum` value                                                                         |
| `io.fabric8.generator.annotation.Size`                          | The field (string, list/array, map) has size limits (`minLength`, `minItems`, `minProperties`, ...) |
| `io.fabric8.generator.annotation.Pattern`                       | The field defines a validation `pattern`                                                            |
| `io.fabric8.generator.annotation.Nullable`                      | The field is marked as `nullable`                                                                   |
| `io.fabric8.generator.annotation.Required`                      | The field is marked as `required`                                                                   |
| `io.fabric8.generator.annotation.ValidationRule`                | The field or object is validated by a CEL rule                                                      |
[...]

Issue #3069, opened 2021-04-30
JSONPath selection in PrinterColumn


lburgazzoli:
As today the @PrinterColumn annotation can only be set on a field, but in some case it would be nice to have the option to set in on a type, i.e. on a CustomResource

with the go sdk, one can do something like:

```
kubebuilder:printcolumn:name="provider",type=string,JSONPath=`.spec.runtime.provider`,description="provider"
```

but I don't a similar option with the java operator sdk

My use case is that I have a Java class used in both the spec and the status and I want to print the same field from spec and status but with a different name.

```java
public class ConnectorSpec {    
    private DeploymentRef deployment;
}
public class ConnectorStatus {    
    private DeploymentRef deployment;
}
```

with the go sdk, I could write something like

```
kubebuilder:printcolumn:name=name="foo",type=string,JSONPath=`.spec.deployment.field`"
kubebuilder:printcolumn:name=name="bar",type=string,JSONPath=`.status.deployment.field`
```

I can workaround this by in-lining classes in the spec/status or having two distinct classes but since the information are exactly the same, it would put more maintenance

metacosm:
I can think of two options at the moment: either a class level annotation allowing you to specify which path you want to add as a printer column, similarly to what the go sdk does, or make the `@PrinterColumn` annotation multiple so that a given field could be annotated with different values, adding a predicate field to determine how to activate a given version.

The first option is most likely simpler to implement but that decouples the annotation from the field it's targeting making it more difficult to maintain. The second option doesn't have that disadvantage but it is probably more complex to implement and possibly to understand depending on how complex the activation predicate could be.

What do you think, @iocanel @manusa?

lburgazzoli:
@metacosm @iocanel this is how it is implemented din strimzi: https://github.com/strimzi/strimzi-kafka-operator/blob/a89f9b466a79b36d49b6b7fcdd120ad9b1c6cec4/api/src/main/java/io/strimzi/api/kafka/model/KafkaConnector.java#L53-L78

lburgazzoli:
still relevant

stoetti:
Are there any plans on implementing this enhancement in the near future?

In order to build CRDs in respect common api conventions (https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) it would be really necessary to provide printer columns that are defined using a JsonPath like this .status.conditions[?(@.type=="Ready")].status

[...]

PR #6315, opened 2024-08-29
CRD generator implementation that supports Jakarta and Swagger

adriansuarez:
## Description

This change adds an alternative CRD generator implementation that is based on
[victools/jsonschema-generator](https://github.com/victools/jsonschema-generator).

The victools generator allows modules to be registered to customize schema generation, and there are modules that provide support for Jakarta Validation and Swagger annotations.

These changes are being contributed into test scope as a prototype, since no effort has been made to integrate it with the existing CRD generator API. It is a totally separate implementation that is not at parity with the existing generator. The intent is that someone with more knowledge of the Fabric8 codebase can incorporate some aspects of it into the real product.

## Type of change

## Checklist

### DIFF
--- a/crd-generator/api-v2/pom.xml
+++ b/crd-generator/api-v2/pom.xml
@@ -77,5 +77,46 @@
       <artifactId>lombok</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.github.victools</groupId>
+      <artifactId>jsonschema-generator</artifactId>
+      <version>${victools-jsonschema.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.github.victools</groupId>
+      <artifactId>jsonschema-module-jackson</artifactId>
+      <version>${victools-jsonschema.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.github.victools</groupId>
+      <artifactId>jsonschema-module-jakarta-validation</artifactId>
+      <version>${victools-jsonschema.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.github.victools</groupId>
+      <artifactId>jsonschema-module-swagger-2</artifactId>
+      <version>${victools-jsonschema.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>jakarta.validation</groupId>
+      <artifactId>jakarta.validation-api</artifactId>
+      <version>${jakarta-validation.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.swagger.core.v3</groupId>
+      <artifactId>swagger-annotations</artifactId>
+      <version>${swagger-annotations.version}</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 </project>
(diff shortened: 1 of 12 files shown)


shawkins:
> It is a totally separate implementation that is not at parity with the existing generator. The intent is that someone with more knowledge of the Fabric8 codebase can incorporate some aspects of it into the real product.

It would be good if you could provide an overview of what doesn't work from the current generator, and if there is some other / different mechanism that could be used instead. SchemaSwap in particular seems like may be difficult to support.

adriansuarez:
> It would be good if you could provide an overview of what doesn't work from the current generator, and if there is some other / different mechanism that could be used instead. SchemaSwap in particular seems like may be difficult to support.

It's a completely separate implementation, so nothing works that is not explicitly implemented. I added the Fabric8Module which adds support for several Fabric8 CRD annotations, but not `@SchemaSwap`.

The following are supported:

- `@io.fabric8.generator.annotation.Default`
- `@io.fabric8.generator.annotation.Max`
- `@io.fabric8.generator.annotation.Min`
- `@io.fabric8.generator.annotation.Nullable`
- `@io.fabric8.generator.annotation.Pattern`
- `@io.fabric8.generator.annotation.Required`
- `@io.fabric8.generator.annotation.ValidationRule`/`@io.fabric8.generator.annotation.ValidationRules`

Everything else is unsupported.

The `@AdditionalPrinterColumns` annotation that allows specifying arbitrary JSONPaths replaces `@PrinterColumn` (see #3069). It would be straightforward to add support for `@AdditionalPrinterColumns` to the current generator or support `@PrinterColumn` in the new generator, but I had no use for the latter in my own code.

The important difference between this implementation and the current one is the replacement of jackson-module-jsonSchema with victools/jsonschema-generator. It should be possible to consolidate the functionality here with the current implementation by replacing the schema generator only and keeping everything else, but when I attempted that I had a hard time breaking the jackson-module-jsonSchema dependency.

baloo42:
Thanks for submitting this. I will dig into it because I also have some concerns about the future of jackson-jsonSchema. Especially because of [this section](https://github.com/FasterXML/jackson-module-jsonSchema?tab=readme-ov-file#future-plans-lack-thereof) in the readme.

Expect some feedback from me in the next days.

shawkins:
> but when I attempted that I had a hard time breaking the jackson-module-jsonSchema dependency.

Perhaps that would be a better starting point - could you open a pr with whatever conversion you had already gotten to? For the most part it seems like it will need to be approached similar to the v2 logic. Start with the schema conversion, then walk that structure along with the object graph to make adjustment to the final form emitted by fabric8.

> Especially because of [this section](https://github.com/FasterXML/jackson-module-jsonSchema?tab=readme-ov-file#future-plans-lack-thereof) in the readme.

Keep in mind this does not seem to be time sensitive as Jackson 3.0 has been under works for 7 years and there are no definitive release timelines that I have found.

baloo42:
> Keep in mind this does not seem to be time sensitive as Jackson 3.0 has been under works for 7 years and there are no definitive release timelines that I have found.

Sure. It's definitily not a top priority, but for me it's interesting to explore because I'm more familiar with victools than jackson-jsonschema.

shawkins:
> Specifically, com.fasterxml.jackson.module.jsonSchema.JsonSchema is used in several places rather than an opaque form like JsonNode or ObjectNode.

We're not directly emitting jackson's JsonSchema, rather we're walking that along with with the object structure and base Jackson logic to emit the final form, which is based upon the CRD object model generated by fabric8.

Please take another pass at using victools but starting instead with the current state of the v2 generator. If you run into issues, that can be discussed on the PR.

> Jakarta Validation and Swagger annotations (which would solve several open issues; see https://github.com/fabric8io/kubernetes-client/issues/5859)

I wasn't deeply involved in earlier discussions, but the gist of what I gathered is that we've been reluctant to support additional annotations either because of the additional dependencies or the possible mismatch in semantics. So much of this support would likely end up being considered optional / plugin driven regardless.

baloo42:
@adriansuarez sorry for the delay, I digged into it and started my own experiment based on your ideas which went too far. The results can be found here: https://github.com/baloo42/crd-generator-victools
Feedback welcome ;)

Regarding `@AdditionalPrinterColumn` please have a look at https://github.com/fabric8io/kubernetes-client/pull/6390.

PR #6333, opened 2024-09-05
feat(model): use Object instead of KubernetesResource for raw types

### LINKED ISSUE #6332: Usage of KubernetesResource in models creates massive builder classes
## Description

Part of #6130

As part of the model generation refactor, one of the initiatives is to unify and make consistent the treatment of raw fields.

In this regard, the initial approach was to generate these fields as:
```java
class ModelClass implements HasMetadata, KubernetesResource {
  @JsonProperty("theField")
  private KubernetesResource theField;
  @JsonProperty("theArrayField")
  private List<KubernetesResource> theArrayField;
  @JsonProperty("theMapField")
  private Map<String, KubernetesResource> theMapField;
}
```

This way, the KubernetesDeserializer kicks in for these fields and any Raw object or HasMetadata object will be detected and deserialized accordingly.

However, this approach brings a huge problem when Sundrio generates the builders for model modules with many classes.
Since Sundrio generates a `withNewXxx` (and nested) builder methods for each `KubernetesResource` interface implementation, we end up with massive builders that in most cases are useless.

## Solution

I discussed the problem internally with @shawkins.

Many alternatives were considered, but most of them involve tampering with sundrio builder generation.
For example: 
- We could add flags to limit selectively or completely turn off the interface implementation detection to generate the subtypes.
- We can also limit the number of subtypes by disabling the logic if more than one type of the implementation or root class is detected.

All of these imply work on sundrio and still might have unwanted side-effects.

The other alternative we considered and the one that we **agreed** to implement is generating the class such as:
```java
  @JsonProperty("theField")
  @JsonDeserialize(using = KubernetesDeserializer.class)
  private Object theField;
  @JsonProperty("theArrayField")
  @JsonDeserialize(using = KubernetesDeserializerForList.class)
  private List<Object> theArrayField;
  @JsonProperty("theMapField")
  @JsonDeserialize(using = KubernetesDeserializerForMap.class)
  private Map<String, Object> theMapField;
`` 

This way sundrio won't generate the `withNewXxx` fields (sundrio doesn't consider Object extensions).
The deserialization still works because the `KubernetesDeserializer` (and its derivatives) are manually configured in the field.

manusa:
## Description

Fixes #6332 

## Type of change

## Checklist


oscerd (approved):

rohanKanojia (approved):

shawkins:
Just wanted to talk broadly on the usage of AnyType, RawExtension, GenericKuberenetesResource, and what is being done here with object and the KubernetesDeserializer. 

AnyType - can be a replacement for where we had previously been using JsonNode and Map types, but is not currently directly used in our models. Also serves as a base type for RawExtension and IntOrString. Is not directly needed by the KubernetesDeserializer. 
RawExtension - provides a way to bridge between KubernetesResource and any content. This is due to the previous mapping of kube's raw extension type to KubernetesResource. Generally not expected to appear directly in user object models. Used directly by the KubernetesDeserializer when a KubernetesResource is not an object or has no api/version / kind information. The kube docs imply that raw extension only allows for complete kube resources, but the schema is not actually restricted to that.
GenericKubernetesResource - generally provides a way to represent kube resource for which there is no class definition. Like RawExtension, but restricts to HasMetadata and not deserialize to the real type - only generic.

v2 CRD Mappings:
AnyType - untyped property that preserves unknown.
RawExtension - maps to raw
KubernetesResource - maps to object - this is probably not correct, if the field is exactly KubernetesResource it should map to raw instead.
GenericKubernetesResource / HasMetadata - maps to raw.

It's clear that these mappings are not bi-directional (I haven't check if the logic knows about AnyType), so round-tripping can result in a different object model.

With this pr we're mapping kube's raw extension to Object directly with the KubernetesDeserializer. If we're good with this approach we can consider deprecating the RawExtension class and recommending instead that users with custom resources take the same approach. The one thing that will look bad about this is KubernetesDeserializer is in an internal package. We'll also need to update the CRD generator to understand this convention. Also if user for whatever reason a user puts an untyped value in the Object field, it will be deserialized as a RawExtension - which doesn't work well if we want to put that on the path to deprecation.

We should also update the CRD generation logic to understand this convention, and consider making a similar change to the java generator.

AnyType could similarly be deprecated if we consider its usage synonomous with Object without the KuberenetsDeserializer - currently that will just map to an object property without preserve uknown.

manusa:
I'm extracting your comment into an issue for further discussion (especially for CRDv2).
I need to merge this to be able to proceed with the rest of model conversions.

> If we're good with this approach we can consider deprecating the RawExtension class and recommending instead that users with custom resources take the same approach

I'm not 100% sure that removing the intermediate RawExtension class will allow for a complete deserialization-serialization roundtrip (this is working now, see tests in kustomize module -and others-)

Issue #6350, opened 2024-09-17
Proxy authentication does not work with vertx client and https endpoint


shawkins:
### Describe the bug

The code is using the generic proxy authentication, but vertx needs the username and password passed to it via the proxy options.

### Fabric8 Kubernetes Client version

6.13.3

### Steps to reproduce

Use proxy.usernam and proxy.password with the vertx client

### Expected behavior

Proxy authentication should work

### Runtime

Kubernetes (vanilla)

### Kubernetes API Server version

1.25.3@latest

### Environment

Linux

shawkins:
I need to be more specific, the problem appears to be with https handling.

shawkins:
| End point  | Proxy | Autenticated | Tested |
| ------------- | ------------- | ------------- | ------------- |
| http  | https  | y | n - only Jetty seems to differentiate secure proxies
| http  | https  | n | n - only Jetty seems to differentiate secure proxies
| http  | http  | y | y
| http  | http  | n | n - but should work
| https  | https  | y | n - only Jetty seems to differentiate secure proxies
| https  | https  | n | n - only Jetty seems to differentiate secure proxies
| https  | http  | y | n - this issue. Probably not a high severity because you are still sending unencrypted credentials
| https  | http  | n | n

socks 4/5 are also untested.

PR #6352, opened 2024-09-18
fix: allowing for the usage of authenticated http proxies for https endpoints

shawkins:
## Description

closes: #6350

Switched from using the full proxy authorization to the username and password - this could be viewed as a breaking change, but it is more straight-forward than adding a routine the parse those values back out.

The jdk client works as is with the existing code, but the system property -Djdk.http.auth.tunneling.disabledSchemes="" must be set for the header to get propogated to the tunneling request so I didn't add that test subclass yet.

The ability to set the socket policy per response ended up not being needed so that can be removed if desired.

## Type of change

## Checklist

### DIFF
--- a/junit/mockwebserver/src/main/java/io/fabric8/mockwebserver/internal/MockDispatcher.java
+++ b/junit/mockwebserver/src/main/java/io/fabric8/mockwebserver/internal/MockDispatcher.java
@@ -32,14 +32,19 @@ public class MockDispatcher extends Dispatcher {
 
   private final Map<ServerRequest, Queue<ServerResponse>> responses;
   private final Collection<WebSocketSession> webSocketSessions = new ConcurrentLinkedQueue<>();
+  private SocketPolicy defaultSocketPolicy = SocketPolicy.EXPECT_CONTINUE;
 
   public MockDispatcher(Map<ServerRequest, Queue<ServerResponse>> responses) {
     this.responses = responses;
   }
 
+  public void setDefaultSocketPolicy(SocketPolicy defaultSocketPolicy) {
+    this.defaultSocketPolicy = defaultSocketPolicy;
+  }
+
   @Override
   public MockResponse peek() {
-    return new MockResponse().setSocketPolicy(SocketPolicy.EXPECT_CONTINUE);
+    return new MockResponse().setSocketPolicy(defaultSocketPolicy);
   }
 
   @Override
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Fix #6247: Support for proxy authentication from proxy URL user info
 * Fix #6281: use GitHub binary repo for Kube API Tests
 * Fix #6282: Allow annotated types with Pattern, Min, and Max with Lists and Maps and CRD generation
+* Fix #6350: Allowing authenticated http proxy usage with Jetty, vertx, and JDK for https endpoints
 * Fix #5480: Move `io.fabric8:zjsonpatch` to KubernetesClient project
 
 #### Dependency Upgrade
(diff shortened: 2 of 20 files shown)


manusa:
Jetty tests are failing

junit/mockwebserver/src/main/java/io/fabric8/mockwebserver/internal/MockDispatcher.java
```diff
@@ -32,14 +32,19 @@ public class MockDispatcher extends Dispatcher {
...
   }
 
+  public void setDefaultSocketPolicy(SocketPolicy defaultSocketPolicy) {
+    this.defaultSocketPolicy = defaultSocketPolicy;
+  }
+
   @Override
   public MockResponse peek() {
-    return new MockResponse().setSocketPolicy(SocketPolicy.EXPECT_CONTINUE);
+    return new MockResponse().setSocketPolicy(defaultSocketPolicy);
```
manusa, line 47: In #5632 this is removed since it's not necessary for the Vert.x server.

Is this required? I'm not sure how to port these changes to the other PR once we merge this.

shawkins:
> Jetty tests are failing

Should be addressed now.

junit/mockwebserver/src/main/java/io/fabric8/mockwebserver/internal/MockDispatcher.java
shawkins: > Is this required?

As far as I can tell that is needed to make the https tunnel actually work.

The test doesn't need to fully succeed to establish the tunnel - I've updated it to capture if the authentication header was sent we could just check for that and let the request fail fast.

manusa:
This PR is modifying some of our exposed interfaces.
This prevents me from back-porting the changes to a 6.13.4 version (which is a requirement by other downstream projects -LTS-).
Is it possible to fix the issue without modifying the interfaces and leaving the existentmethod signatures intact?

cescoffier:
As @manusa it would be super appreciated if we can get a fix that can be backported in a micro-release.

shawkins:
> This PR is modifying some of our exposed interfaces.

Although unlikely it is possible that users are directly setting up the proxy authentication for their http clients - I wasn't sure how much of an api guarentee we would want to extend there. At least for 7.0+ it seems much more straight-forward to drop the old method.

> Is it possible to fix the issue without modifying the interfaces and leaving the existentmethod signatures intact?

Leaving the existing signature in a straight-forward way means that we'll need to support parsing the username and password. Not a huge deal, just a pain. I would add that specifically to the back port.

shawkins:
To clarify things more. Supporting proxyAuthorization is more generally than just supporting basic. However only the OkHttp and JDK client (with an additional system property) seem to handle this situation as expected with https endpoints. Jetty and vertx form their connect requests without consulting the current headers. 

So this pr is a minimal alignment of support that is common across all the clients, but we do loose the general proxyAuthorization support. Any other authentication types are no longer supported. I'm not aware however of needing to support anything other than basic.

So the full range of options for compatiblity are:
- Add limited support for proxyAuthorization such that it maps to this username password support
- Add full support for proxyAuthorization - that will require these changes + the old code and some additional (java)docs explaining what takes presedence and what is supported by which clients and when
- Get vertx to change it's logic to not be based specifically on the username/password, but allow for an existing header to be passed to the connect request. Authenticated proxy access to https endpoints still wouldn't work for Jetty under this scenario. @cescoffier is vertx open to change here?

cescoffier:
I think it would make sense. But I'm not sure when this could be done. (Cc @vietj )

shawkins:
> Add full support for proxyAuthorization - that will require these changes + the old code and some additional (java)docs explaining what takes presedence and what is supported by which clients and when

#6371 retains proxyAuthorization and does not yet introduce new method(s). So closing with further work to be done under #6372 #6373

> I think it would make sense. But I'm not sure when this could be done. (Cc @vietj )

Took more of a look at the changes involved with this and our potential direction here and determined this isn't worth pursuing. We'll go down the path of deprecating proxyAuthorization.

PR #6390, opened 2024-09-30
fix: adds an additional printer column annotation

shawkins:
## Description

closes: #3069

I believe the is the most straight-foward appraoch to adding general additional printer columns.

Other thoughts:
- we could reuse the existing PrinterColumn annotation. However I didn't like about that is that it mixes the usage models, so the javadocs / validations would need to be more extensive.
- allow the annotation to appear in other places. I think we'd at least need to auto-prefix the paths with the current path within the structure - we do have that as we're navigating over the object model. This can be easily added later if needed.
- it's not being extensively validated. For example we could check that the path provided is valid / actually exists.

## Type of change

## Checklist

### DIFF
--- a/crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/AbstractCustomResourceHandler.java
+++ b/crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/AbstractCustomResourceHandler.java
@@ -46,7 +46,7 @@ public abstract class AbstractCustomResourceHandler {
 
   protected void handlePrinterColumns(AbstractJsonSchema<?, ?> resolver, PrinterColumnHandler handler) {
     TreeMap<String, AnnotationMetadata> sortedCols = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
-    resolver.getAdditionalPrinterColumns().forEach(apc -> sortedCols.put(apc.path(), new AnnotationMetadata(apc, null)));
+    resolver.getAdditionalPrinterColumns().forEach(apc -> sortedCols.put(apc.jsonPath(), new AnnotationMetadata(apc, null)));
     sortedCols.putAll(resolver.getAllPaths(PrinterColumn.class));
     sortedCols.forEach((path, property) -> {
       if (property.annotation instanceof AdditionalPrinterColumn) {
--- a/crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/AbstractCustomResourceHandler.java
+++ b/crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/AbstractCustomResourceHandler.java
@@ -16,6 +16,7 @@
 package io.fabric8.crdv2.generator;
 
 import io.fabric8.crd.generator.annotation.AdditionalPrinterColumn;
+import io.fabric8.crd.generator.annotation.AdditionalPrinterColumn.Format;
 import io.fabric8.crd.generator.annotation.PrinterColumn;
 import io.fabric8.crdv2.generator.AbstractJsonSchema.AnnotationMetadata;
 import io.fabric8.kubernetes.api.model.HasMetadata;
@@ -48,42 +49,32 @@ public abstract class AbstractCustomResourceHandler {
     resolver.getAdditionalPrinterColumns().forEach(apc -> sortedCols.put(apc.path(), new AnnotationMetadata(apc, null)));
     sortedCols.putAll(resolver.getAllPaths(PrinterColumn.class));
     sortedCols.forEach((path, property) -> {
-      String column;
-      String type;
-      String format;
-      int priority;
-      String description;
       if (property.annotation instanceof AdditionalPrinterColumn) {
         AdditionalPrinterColumn printerColumn = ((AdditionalPrinterColumn) property.annotation);
-        column = printerColumn.name();
-        format = printerColumn.format();
-        priority = printerColumn.priority();
-        type = printerColumn.getType();
-        description = printerColumn.getDescription();
+        String column = printerColumn.name();
+        String format = printerColumn.format() == Format.NONE ? null : printerColumn.format().getValue();
+        String type = printerColumn.type().getValue();
+        int priority = printerColumn.priority();
+        String description = printerColumn.getDescription();
+        handler.addPrinterColumn(path, column, format, priority, type, description);
       } else {
         PrinterColumn printerColumn = ((PrinterColumn) property.annotation);
-        column = printerColumn.name();
-        format = printerColumn.format();
-        priority = printerColumn.priority();
-        type = property.schema.getType();
+        String column = printerColumn.name();
+        String format = printerColumn.format();
+        format = Utils.isNotNullOrEmpty(format) ? format : null;
+        String type = property.schema.getType();
+        if ("object".equals(type) || "array".equals(type)) {
+          LOGGER.warn("Printer column '{}' has a type '{}' that is not allowed, will use string intead", column, type);
+          type = "string";
+        } else if ("string".equals(type) && "date".equals(property.schema.getFormat())) {
+          type = "date";
+        }
+        int priority = printerColumn.priority();
 
         // TODO: add description to the annotation? The previous logic considered the comments, which are not available here
-        description = property.schema.getDescription();
+        String description = property.schema.getDescription();
+        handler.addPrinterColumn(path, column, format, priority, type, description);
       }
-
-      if (Utils.isNullOrEmpty(column)) {
-        column = path.substring(path.lastIndexOf(".") + 1).toUpperCase();
-      }
-      format = Utils.isNotNullOrEmpty(format) ? format : null;
-      if ("object".equals(type) || "array".equals(type)) {
-        LOGGER.warn("Printer column '{}' has a type '{}' that is not allowed, will use string intead", column, type);
-        type = "string";
-      } else if ("string".equals(type) && "date".equals(property.schema.getFormat())) {
-        type = "date";
-      }
-      description = Utils.isNotNullOrEmpty(description) ? description : null;
-
-      handler.addPrinterColumn(path, column, format, priority, type, description);
     });
   }
 
--- a/crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/AbstractCustomResourceHandler.java
+++ b/crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/AbstractCustomResourceHandler.java
@@ -15,6 +15,7 @@
  */
 package io.fabric8.crdv2.generator;
 
+import io.fabric8.crd.generator.annotation.AdditionalPrinterColumn;
 import io.fabric8.crd.generator.annotation.PrinterColumn;
 import io.fabric8.crdv2.generator.AbstractJsonSchema.AnnotationMetadata;
 import io.fabric8.kubernetes.api.model.HasMetadata;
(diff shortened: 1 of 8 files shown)


baloo42:
:+1: worked on the same last weekend, but no tests yet: https://github.com/baloo42/kubernetes-client/pull/10

Looks very similar, but I used enums for format and type. I think using enums is better here, because not all formats and types are allowed to be used here. What do you think on enums?

> - we could reuse the existing PrinterColumn annotation. However I didn't like about that is that it mixes the usage models, so the javadocs / validations would need to be more extensive.

I also wouldn't like it if we reuse the PrinterColumn. Too many differences to explain.

> - allow the annotation to appear in other places. I think we'd at least need to auto-prefix the paths with the current path within the structure - we do have that as we're navigating over the object model. This can be easily added later if needed.

I think there are pro's and con's about this. It is possible and might be explainable 
but if we don't implement it, the difference would be better:

`@PrinterColumn` Uses *magic* to find the path
`@AdditionalPrinterColumn` No magic involved, you as user are fully responsible.

:+1: to postpone this to next iteration at least.

> - it's not being extensively validated. For example we could check that the path provided is valid / actually exists.

:+1: same, as a above. If we validate it, it must support all cases because CRDGenerator is now fully responsible instead of the user.

shawkins:
> worked on the same last weekend, but no tests yet: https://github.com/baloo42/kubernetes-client/pull/10

Whoops, sorry for the duplication of effort.

> What do you think on enums?

My only thought is the mismatch with the PrinterColumn annotation. To evolve the other annotation we'd have to accept Object for a major release before restricting back to an enum.

I've added your enums to the this pr.

> to postpone this to next iteration at least.

I'm guessing that it may never be needed - what are the chances that someone wants the same set of printer columns in two different CRDs based upon a common object?

baloo42:
> > worked on the same last weekend, but no tests yet: [baloo42#10](https://github.com/baloo42/kubernetes-client/pull/10)
> 
> Whoops, sorry for the duplication of effort.

no worries, I havn't published or noted it somewhere. My fault ^^

> > What do you think on enums?
> 
> My only thought is the mismatch with the PrinterColumn annotation. To evolve the other annotation we'd have to accept Object for a major release before restricting back to an enum.
> 
> I've added your enums to the this pr.

:+1: 
For `@PrinterColumn` we could also phase it out:

1. Add additional enum param `formatStrict`, deprecate `format`
2. Replace `format` with enum param, deprecate `formatStrict` 
3. Remove `formatStrict`

This would require the same amount of major releases but avoids Object or do I miss something?

> > to postpone this to next iteration at least.
> 
> I'm guessing that it may never be needed - what are the chances that someone wants the same set of printer columns in two different CRDs based upon a common object?

Full ack. There is no need for additional ways. If someone wants to configure a printer column deep in the hierarchy and maybe shared between CRDs, he should and can already use PrinterColumn.

generator-annotations/src/main/java/io/fabric8/crd/generator/annotation/AdditionalPrinterColumn.java
```diff
@@ -22,27 +22,74 @@
...
+  // https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#format
+  enum Format {
+
+    NONE(""),
+    INT32("int32"),
+    INT64("int64"),
+    FLOAT("float"),
+    DOUBLE("double"),
+    BYTE("byte"),
+    BINARY("binary"),
```
baloo42, line 62: Sorry, havn't checked it twice: `binary` is not allowed. Only `byte`.

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#format

baloo42 (changes requested):

shawkins:
> This would require the same amount of major releases but avoids Object or do I miss something?

Actually Object won't work at all for an annotation property. So it either needs to be a breaking change for 7.0 or done over 3 major releases with the additional field. I'd opt for just doing the breaking change instead.

generator-annotations/src/main/java/io/fabric8/crd/generator/annotation/AdditionalPrinterColumn.java
shawkins: Updated. Altenatively, I'm fine if you want to take over ownership of this feature from your branch.

generator-annotations/src/main/java/io/fabric8/crd/generator/annotation/AdditionalPrinterColumn.java
```diff
@@ -0,0 +1,120 @@
...
+ * Defines an additional printer column. Must be placed at the root of the
+ * custom resource.
+ */
+@Repeatable(AdditionalPrinterColumns.class)
+@Target({ ElementType.TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+public @interface AdditionalPrinterColumn {
+
+  //https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#type
+  enum Type {
```
metacosm, line 34: Might be useful to move this to the top-level in case we want to reuse this in a different context in the future?

generator-annotations/src/main/java/io/fabric8/crd/generator/annotation/AdditionalPrinterColumn.java
```diff
@@ -0,0 +1,120 @@
...
+      this.value = value;
+    }
+
+    public String getValue() {
+      return value;
+    }
+  }
+
+  // https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#format
+  enum Format {
```
metacosm, line 54: Same as Type

generator-annotations/src/main/java/io/fabric8/crd/generator/annotation/AdditionalPrinterColumn.java
baloo42: Yes it could be useful (in rare cases), e.g. as a final check on the type if `@PrinterColumn` is used and the type is derived. But we have to be careful with naming. This enum represents types, which are only allowed in printer columns. The name should make this very clear, like `PrinterColumnType`.

metacosm: Ah, missed that it was restricted to printer columns… (trying to follow too many things at once). Then it makes sense to keep it as an inner type.

generator-annotations/src/main/java/io/fabric8/crd/generator/annotation/AdditionalPrinterColumn.java
baloo42: If you want, you can assign it to me but it doesn't make a lot sense too me as yours is almost ready.
Your approach is able collect the path, my approach tries to distinguesh between top level annotations and annotations in the hierarchy a little bit more but without the ability to collect the path.
So it depends on which direction we want to go.

crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/AbstractCustomResourceHandler.java
```diff
@@ -44,29 +46,35 @@ void addPrinterColumn(String path, String column, String format,
...
-      format = Utils.isNotNullOrEmpty(format) ? format : null;
-      int priority = printerColumn.priority();
-      String type = property.schema.getType();
-      if ("object".equals(type) || "array".equals(type)) {
-        LOGGER.warn("Printer column '{}' has a type '{}' that is not allowed, will use string intead", column, type);
-        type = "string";
-      } else if ("string".equals(type) && "date".equals(property.schema.getFormat())) {
-        type = "date";
-      }
+      if (property.annotation instanceof AdditionalPrinterColumn) {
```
baloo42, line 52: At the moment the `@AdditionalPrinterColum` is collected from everywhere in the spec or status type. Is this intended?

We should either clarify this in the description of the annotation or pick it only from the custom resource class.

crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/AbstractCustomResourceHandler.java
```diff
@@ -44,29 +46,35 @@ void addPrinterColumn(String path, String column, String format,
...
+        String format = printerColumn.format() == Format.NONE ? null : printerColumn.format().getValue();
+        String type = printerColumn.type().getValue();
+        int priority = printerColumn.priority();
+        String description = printerColumn.getDescription();
+        handler.addPrinterColumn(path, column, format, priority, type, description);
+      } else {
+        PrinterColumn printerColumn = ((PrinterColumn) property.annotation);
+        String column = printerColumn.name();
+        String format = printerColumn.format();
+        format = Utils.isNotNullOrEmpty(format) ? format : null;
```
baloo42, line 64: We need those lines quite often.  Can we add a method to CRDUtils?

Something like:
```
 public static String emptyToNull(String s) {
    return Utils.isNullOrEmpty(s) ? null : s;
 }
```

baloo42 (changes requested):
FYI: I also tried to add an approval test (only for `@PrinterColumns`) to compare if something breaks with this change.

Your change in this PR looks good, but it seems there is a bug in api-v1 which blocks us to add an approval test at the moment:

```java
@Data
public class PrinterColumnSpec {
  @PrinterColumn
  private String fromLevel0;

  private DeepLevel1 deepLevel1;

  @Data
  public static class DeepLevel1 {

    @PrinterColumn
    private Integer fromLevel1;
  }
}
```

Result with api-v2 (this MR):

```yaml
    - jsonPath: ".spec.deepLevel1.fromLevel1"
      name: "FROMLEVEL1"
      priority: 0
      type: "integer"
    - jsonPath: ".spec.fromLevel0"
      name: "FROMLEVEL0"
      priority: 0
      type: "string"
```

Result with api-v1:

```yaml
    - jsonPath: ".spec.deepLevel1.fromLevel1"
      name: "FROMLEVEL1"
      priority: 0
      type: "integer"
    - jsonPath: ".spec.fromLevel0"
      name: "FROMLEVEL0"
      priority: 0
      type: "string"
    - jsonPath: ".spec.fromLevel1"
      name: "FROMLEVEL1"
      priority: 0
      type: "integer"
```

Should we fix this within this PR or create another one and add an approval test later?

crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/AbstractCustomResourceHandler.java
shawkins: > At the moment the @AdditionalPrinterColum is collected from everywhere in the spec or status type.

It should not be: https://github.com/fabric8io/kubernetes-client/pull/6390/files#diff-e5491c474236ab12f8e6f5d9895464ecf77754b17ddb0e182682add6db8620ceR146

> We should either clarify this in the description of the annotation or pick it only from the custom resource class.

It is documented as intended for the root: https://github.com/fabric8io/kubernetes-client/pull/6390/files#diff-669d381122146d717c360f17fcfe5eeaab88b5c84dd4bfcf81b6de71aa5a75c1R25

crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/AbstractCustomResourceHandler.java
shawkins: Yes, a utility method could be added in a follow-up.

shawkins:
> Should we fix this within this PR or create another one and add an approval test later?

Do you mean fix the v1 implementation? I don't think we need to bother doing that.

generator-annotations/src/main/java/io/fabric8/crd/generator/annotation/AdditionalPrinterColumn.java
```diff
@@ -0,0 +1,120 @@
...
+   * @return the priority or 0 if no priority is specified.
+   */
+  int priority() default 0;
+
+  /**
+   * The JSON Path to the field
+   * 
+   * @return
+   */
+  String path();
```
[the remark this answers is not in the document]
shawkins: Yes, I'll change it to jsonPath.

baloo42:
> > Should we fix this within this PR or create another one and add an approval test later?
> 
> Do you mean fix the v1 implementation? I don't think we need to bother doing that.

Yes I mean fixing v1 because of differences in the approval tests. If we don't want to fix v1, then we should start splitting the approval tests so that we can add tests for new or breaking features. (not in this PR)

See also https://github.com/fabric8io/kubernetes-client/pull/6447#discussion_r1801978479

shawkins:
> Yes I mean fixing v1 because of differences in the approval tests. If we don't want to fix v1, then we should start splitting the approval tests so that we can add tests for new or breaking features. (not in this PR)

I'm good with that.

crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/AbstractCustomResourceHandler.java
baloo42: Sorry, seems like it was a problem on my side. I have checked out your branch again and could not reproduce it. Looks fine now!

baloo42 (approved):

manusa (approved):
LGTM, thx!

PR #6454, opened 2024-10-15
DO NOT MERGE: Generate CRD from Immutables.org generated classes

teabot:
## Description
Demonstrates that CRD generation does not pick up Jackson annotations from interfaces or generated [Immutables.org](https://immutables.github.io) classes.

Run: `io.fabric8.crd.generator.approvaltests.CRDGeneratorApprovalTest`
Then check for difference between expected/actual for test cases `viaimmutables.samples.fabric8.io*`.

Expected:
```
    schema:
      openAPIV3Schema:
        properties:
          spec:
            properties:
              value:
                type: "string"
            type: "object"
          status:
            type: "object"
```

Actual:

```
    schema:
      openAPIV3Schema:
        properties:
          spec:
            type: "object" <-- 'value' property was not generated
          status:
            type: "object"
```

Note that JSON annotations are on both `ViaImmutableSpec` (although should not be needed), and also on generated class `ImmutableViaImmutableSpec`. 

## Type of change
Reproduces possible error. Do not merge. 

## Checklist


manusa:
I'm copying the discussion in the official Gitter Channel for future reference:
_https://matrix.to/#/!rDlRZKYupRPQnkjTvm:gitter.im/$yKx5jaEpy1SfAoSu5uwnUT3sj4YY5KAPjKBrzK5W6Qg?via=gitter.im&via=matrix.org_

_@teabot said:_
> I am wanting to implement my Status and Spec classes as immutables, with associated fluent builders. I am trying to do this with [Immutables.org](http://immutables.org/), generating classes from interfaces. However, when the CRDs are generated, the spec and status are empty – as if no properties have been detected in the types.
>
> What should I expect from the generator? I notice all examples use POJOs with setters and getters? Are there known constraints? Is this achievable?

_@manusa said:_
> AFAIR it shuldn't be necessary to add getters/setters for the fields, but they should be accessible
>
> Do you have any reproducer project or can you build a quick one to analyze what the problem might be?
>
> Also note that if these classes are going to be later handled by the Kubernetes Client they should be serializable and deserializable by Jackson, so similar rules apply.

_@teabot said:_
> I've reproduced the issue here: https://github.com/fabric8io/kubernetes-client/pull/6454

_@manusa said:_
> OK, I'm checking the reproducer
>
> I don't think that the CRD generator is compatible with [immutables.org](http://immutables.org/), at least in the way that you're using them
>
> I've never used [immutables.org](http://immutables.org/), but I guess that the maven plugin generates some sort of impl classes in the target generated-sources directory
>
> In this case, `ViaImmutable` should reference the impls instead of the interfaces
with that change it maybe works
> From your example I understand is `ImmutableViaImmutableSpec`
>
> so replace:
> ```java
> public class ViaImmutable extends CustomResource<ViaImmutableSpec, Void> {
> ```
>
> with:
> ```java
> public class ViaImmutable extends CustomResource<ImmutableViaImmutableSpec, Void> {
> ```

_@teapot said:_
> Thank you. This does appear to work. It's interesting that the generator is concerned with the specific implementation rather than the contract. I wonder if this holds true for simpler class hierarchies such as interface+implementing class, or class+subclass.

_@manusa said:_
> Well, the generator in the end is introspecting the classes.
>
> Getters in the interface might or might not be implemented as fields.
>
>If the generator had a notion of [immutables.org](http://immutables.org/) internals, then it could probably infer the structure. But currently, this is not possible.
>
> If referencing the impl works, that's how you should proceed.
>
>Note that in the end you're doing the same with the `@JsonDeserialize` annotation

manusa:
Closing, since the reproducer is no longer needed and there is a workaround for the issue.

If further support is required for immutables.org, please open an issue where this can be further discussed and analyzed.

Issue #6616, opened 2024-11-18
[EPIC] CRD Generator v2


manusa:
## Description

Original discussion: https://github.com/fabric8io/kubernetes-client/discussions/5942

As part of the new [v7.0.0 release](https://github.com/fabric8io/kubernetes-client/issues/5778) we need to make the CRD Generator (especially its API) as complete as possible to avoid any breaking changes in future v7 minor version releases.

manusa:
Hi @baloo42,
Once the pending PRs for the CRDGenerator are merged I will probably start with the release process of v7.0.0.
Is there something else that we're missing that should go into v7? I'm talking about any breaking change that will be hard to introduce in any subsequent 7.x version release

baloo42:
Ok, let's focus on stabilizing 7.0.0 from now on. My planned contributions can wait until the merge window for the next release is open.

In my opinion we should at least discuss the following potential candidates of breaking changes:

## Mark spec always as required

This change was initially discussed in terms of introducing new interfaces e.g. `HasSpec`, `HasStatus` to replace CustomResource class.

If the introduction/change of those interfaces must be postponed we should talk at least of hardcoding this into the CRD-Generator. Marking the spec always as required can be easily implemented if hardcoded, it just needs some work on fixing most of the tests.

Can you provide an update on the plans regarding the CustomResource class?

If we introduce this, it would be definitly a breaking change.

https://github.com/fabric8io/kubernetes-client/issues/3096

## Change of value type in `@Min` / `@Max`

At the moment those annotations use double as value type. Is this really safe and lossless?
Other comparable annotations from openapi or jakarta validations have other strategies in place to avoid rounding errors.

If we need to change the type, this would be a breaking change.

Mentioned in PR https://github.com/fabric8io/kubernetes-client/pull/6447#issuecomment-2489465820.

#5868 itself might also contain some breaking changes, but this is already listed in your list.

## `@ValidationRule` detection

At the moment `@ValidationRule` will be detected and merged together for a lot cases. Maybe this was too much and this is the chance to reduce some cases.

Example:
If a field and the corresponding getter have `@ValidationRule`s, then all rules are merged and end up in the resulting CRD. 

Should we stay to this behaviour or should we either pick this annotation from the field or the getter similar to other annotations.

https://github.com/fabric8io/kubernetes-client/blob/main/crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/AbstractJsonSchema.java#L201

[...]

PR #6672, opened 2024-11-28
test(crd-generator): spec and status can be marked as required (opt-in)

### LINKED ISSUE #3096: Mark crd spec as required
When extending CustomResource I would like to mark the spec as required.  The NotNull annotation however only seems to work on fields, and not methods, so this does not seem possible.

cc @metacosm

manusa:
## Description

Closes #3096 

Test to verify that the `CustomResource` `spec` and `status` fields can be marked as required in the generated `CRD`.

By using the fabric8-provided `@Required` annotation in an overridden `getSpec` or `getStatus` method, the fields can be easily marked as required.

In #6616 we've been discussing about making fields required by default and providing an opt-out mechanism.
This can still be done. However, after my initial analysis, I think that for the time being it's better to provide this opt-in mechanism and consider the other one for a future release.
Doing it opt-out with the current logic in place might make it sort of difficult for users, since the only viable annotation to counter-effect this is the `@Nullable` and it's not really designed for this purpose.
Other options could be to add configuration flags for the generator, but then we defeat the simplicity purpose. I think it's fine to enforce good practices, but not even K8s does this.

/cc @baloo42 @shawkins 

## Type of change

## Checklist


baloo42 (approved):
Looks good and can confirm that this opt-in solution worked since a long time.

I think this is worth to mention it somewhere in the docs. Either in the CRD-Generator docs or in the [Java Operator SDK FAQ](https://javaoperatorsdk.io/docs/faq/).

manusa:
> I think this is worth to mention somewhere in the docs. Either in the CRD-Generator docs or in the [Java Operator SDK FAQ](https://javaoperatorsdk.io/docs/faq/).

Agreed. I added a section in the docs with instructions on how to set the fields required.

I'll leave the JOSDK side to @metacosm and  @csviri.

PR #6835, opened 2025-01-27
fix(vertx): VertxHttpClient uses exclusive Vert.x instance by default

### LINKED ISSUE #6792: Shutdown hook in VertxHttpClientFactory prevents client usage in other shutdown hook
### Describe the bug

This library should not assume that it is in charge of shutting down the Http connection. Instead, this should be left to the client app.
In camel k8s we have we have a 'run' option which can be stopped with ctrl+c, upon which all k8s resources are also deleted from the cluster.

This code causes regression, such that these resources can no longer be deleted when executing our shutdown hook ...

```
    public VertxHttpClientFactory() {
        this(VertxHttpClientFactory.VertxHolder.INSTANCE);
        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
            if (this.vertx != null) {
                this.vertx.close();
            }

        }));
    }
```

I checked the other http client variants and it seems that vertx is the only one that installs a shutdown hook. Please consider removing this code and instead rely on higher layers to close the k8s client when appropriate.

7.0.1

### Steps to reproduce

See: https://issues.apache.org/jira/browse/CAMEL-21621

### Expected behavior

Using this k8s client in a shutdown hook should be valid

### Runtime

Kubernetes (vanilla)

1.25.3@latest

### Environment

Linux

```shell

```

### Additional context

_No response_

manusa:
## Description

Fixes #6792

A shared Vert.x instance can still be provided to the VertxHttpClientFactory.

This instance will be shared across the different VertxHttpClient instances.

It's responsibility of the user to handle the Vert.x shared instance lifecycle

/cc @shawkins 

## Type of change

## Checklist

### DIFF
--- a/httpclient-vertx/src/main/java/io/fabric8/kubernetes/client/vertx/VertxHttpClient.java
+++ b/httpclient-vertx/src/main/java/io/fabric8/kubernetes/client/vertx/VertxHttpClient.java
@@ -41,13 +41,25 @@ import static io.fabric8.kubernetes.client.vertx.VertxHttpRequest.toHeadersMap;
 
 public class VertxHttpClient<F extends io.fabric8.kubernetes.client.http.HttpClient.Factory>
     extends StandardHttpClient<VertxHttpClient<F>, F, VertxHttpClientBuilder<F>> {
+
   private final Vertx vertx;
   private final HttpClient client;
+  private final boolean closeVertx;
 
-  VertxHttpClient(VertxHttpClientBuilder<F> vertxHttpClientBuilder, HttpClient client, AtomicBoolean closed) {
+  /**
+   * Create a new VertxHttpClient instance.
+   *
+   * @param vertxHttpClientBuilder the builder that created this client.
+   * @param closed a flag to indicate if the client has been closed.
+   * @param client the Vert.x HttpClient instance (will be closed alongside the client).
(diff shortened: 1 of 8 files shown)


httpclient-vertx/src/main/java/io/fabric8/kubernetes/client/vertx/VertxHttpClient.java
```diff
@@ -122,6 +134,9 @@ public CompletableFuture<HttpResponse<AsyncBody>> consumeBytesDirect(StandardHtt
   @Override
   public void doClose() {
     client.close();
```
shawkins, line 136: Just in case, a try finally would be good here.

shawkins (commented):
Looks good, just one additional thought.

httpclient-vertx/src/main/java/io/fabric8/kubernetes/client/vertx/VertxHttpClient.java
manusa: I'm not completely following, do you mean like this:
```
    try {
      client.close();
    } finally {
      if (closeVertx) {
        vertx.close();
      }
    }
```

shawkins (approved):
LGTM

Issue #6654, opened 2024-11-25
Invalid CRD with when no string default value is used


mfvitale:
### Describe the bug

With the generator v2 the default are not working as expected. If you define for example

```java
@Default(defaultValue = "60000")
private long flushMs = 60000L;
```

the generated crd will put the default value as a string causing the crd to not be valid. 

```shell
Error from server (Invalid): error when creating "k8/debeziumservers.debezium.io-v1.yml": CustomResourceDefinition.apiextensions.k8s.io "debeziumservers.debezium.io" is invalid: [spec.validation.openAPIV3Schema.properties[spec].properties[source].properties[offset].properties[flushMs].default: Invalid value: "string":  in body must be of type integer: 
```

### Fabric8 Kubernetes Client version

6.13.4

### Steps to reproduce

Maybe adding a test in https://github.com/fabric8io/kubernetes-client/blob/ef5944094035a90942c044a513d82cbe384ddb0d/crd-generator/api-v2/src/test/java/io/fabric8/crdv2/generator/v1/JsonSchemaTest.java#L70 will be a good starting point to easily reproduce it. 

### Expected behavior

The correct default value should be used. 

### Runtime

Kubernetes (vanilla)

### Kubernetes API Server version

1.25.3@latest

### Environment

Linux

mfvitale:
@andreaTP

manusa:
### Additional context:

The issue is applicable for both the `@Default` annotation and the `@JsonProperty` `defaultValue`:

https://github.com/fabric8io/kubernetes-client/blob/4beec6259060fd4adffcf8c9dca90bd6101c9dee/crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/AbstractJsonSchema.java#L244


https://github.com/fabric8io/kubernetes-client/blob/4beec6259060fd4adffcf8c9dca90bd6101c9dee/crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/AbstractJsonSchema.java#L274

For CRD v1 this is only applicable to the `@Default` annotation, the `@JsonProperty` `defaultValue` config is not taken into account:

https://github.com/fabric8io/kubernetes-client/blob/1e952e068ac4160235e34f5399beea4acf5f5c1b/crd-generator/api/src/main/java/io/fabric8/crd/generator/AbstractJsonSchema.java#L451-L453


[...]

PR #6903, opened 2025-02-26
fix: allow defaults to be full json or yaml

shawkins:
## Description

The v1 implemenation accepted both YAML or JSON as the default, and converted to a JsonNode.

The initial v2 implemenation was simply converting a String to a JsonNode - so it was always just a text node, and thus https://github.com/fabric8io/kubernetes-client/issues/6654

The fix for that inferred some additional types, but it lost the ability to accept arbitary YAML or JSON.

I have several concerns with this fix:
- we don't have it sufficiently documented what we expect for the default. 
- this implementation is more rigidly typed than what was in v1
- CRDs expect only JSON, so I'm not sure we should be accepting yaml

closes: #6908

## Type of change

## Checklist


crd-generator/api-v2/src/test/java/io/fabric8/crdv2/generator/v1/JsonSchemaDefaultValueTest.java
```diff
@@ -55,24 +58,28 @@ private static final class ClassInTest {
...
     double defaultValueForDouble;
+
+    @JsonProperty
+    @Default("[]")
+    double[] defaultValueForDoubleArray;
   }
 
   @Test
-  @DisplayName("JsonProperty default value should take precedence over Default annotation")
+  @DisplayName("Default annotation should take precedence over JsonProperty")
```
manusa, line 76: https://github.com/fabric8io/kubernetes-client/pull/6666#discussion_r1861130328

We're changing the order of precedence here.
I'm fine with this, but just want to make sure why we're doing this and to have some paper trail in case we want to trace this back in the future.

shawkins: > We're changing the order of precedence here.

Yes back to what it was in the initial port, and inline with the other annotations - the fabric8 ones are dominant.

Looking back over the last fix the switch in ordering may have been done because I was initially wondering if we should deprecate at least some of our fabric8 annotations as you can do some of the same things now with the jackson annotations - but in many cases there's a potential semantic mismatch where the jackson annotation value may not really be appropriate for a crd.  That brings up another thought - if we don't trust the JsonProperty default to be yaml or json and don't want a hard error, then this fix will need refined a little so that there will be a warning instead.

manusa (approved):
LGTM, thx!

PR #7000, opened 2025-04-10
feat(crd-generator): introduce `@JSONSchema` annotation

### LINKED ISSUE #6999: CRDGenerator: JSON Schema annotation for full control

There are a number of JSON schema properties that we are not able to represent with the code-first approach. Several issues are open in this repository to individually add support for things like `oneOf` schema assertions or particular Kubernetes extensions like `x-kubernetes-list-type`.

I'd like to propose introducing an annotation `@JSONSchema` that maps directly to `io.fabric8.kubernetes.api.model.apiextensions.v1.JSONSchemaProps`. When present, this annotation would be used instead of any other processing/introspection for an element (class, field, method). This would work similar to how Swagger or MicroProfile OpenAPI schema annotations work, giving full control to the user. Since Java annotations cannot be recursive, any nested schema references (e.g. `anyOf`) would have a type of `Class`, pointing to the type which would be further scanned to fill in that schema's information.

_No response_

### Additional context

_No response_

MikeEdgar:
## Description

Closes #6999

This is a POC to demonstrate the idea behind #6999 (it may also close several other issues related to the CRD generator's capabilities). The new `@JSONSchema` annotation may be used on a class or property to short-circuit other schema scanning for the target. The exception is when the annotation's `implementation` is used. In that case, the `implementation` class will be scanned as normal, then the `@JSONSchema` annotation attributes will be used to override those from the scan.

The annotation also includes a `structural` boolean attribute that enables certain handling for [structural schemas](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema). This includes copying properties from `allOf`, `anyOf`, `oneOf`, and `not` to the schema with the annotation, and also removing the restricted attributes without those sub-schemas.

Several of the attributes of the annotation may not actually be valid for a Kube CRD schema. However, the intent here is simply to expose the CRD schema model entirely to give that control to the user.

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 * Fix #6880: LogWatch interface provides listeners on close stream event
 * Fix #6971: Exposed Istio v1 models in Istio Client DSL
 * Fix #6998: Removed unneeded dependency on javax.annotation:javax.annotation-api
+* Fix #6999: (crd-generator) introduce JSONSchema annotation for increased control of schema output
 
 #### Dependency Upgrade
 * Fix #6829: Sundrio was upgraded to 0.200.3. In some rare circumstances nested method names will need to be changed.
@@ -54,7 +55,7 @@
 
 #### New Features
 * Fix #5993: Support for Kubernetes v1.31 (elli)
-* Fix #6767: Support for Kubernetes v1.32 (penelope) 
+* Fix #6767: Support for Kubernetes v1.32 (penelope)
 * Fix #6777: Added Javadoc comments to all generated models
 * Fix #6802: (java-generator) Added support for required spec and status
 
@@ -225,7 +226,7 @@
 * Fix #5357: adding additional Quantity methods
 * Fix #5635: refined LeaderElector lifecycle and logging
 * Fix #5787: (crd-generator) add support for deprecated versions for generated CRDs
-* Fix #5788: (crd-generator) add support for Kubernetes validation rules 
+* Fix #5788: (crd-generator) add support for Kubernetes validation rules
 * Fix #5735: Replace usages of `UUID.randomUUID()` with UUID created via AtomicLong
 
 #### New Features
@@ -309,7 +310,7 @@
 * Fix #5220: refinements and clarifications to the validation of names
 
 #### Dependency Upgrade
-* Fix #5286: Update Fabric8 OpenShift Model as per OpenShift 4.13.12 
+* Fix #5286: Update Fabric8 OpenShift Model as per OpenShift 4.13.12
 * Fix #5373: Gradle base API based on v8.2.1
 * Fix #5401: Upgrade Fabric8 Kubernetes Model to Kubernetes v1.28.2
 
--- a/crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/AbstractJsonSchema.java
+++ b/crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/AbstractJsonSchema.java
@@ -167,14 +167,6 @@ public abstract class AbstractJsonSchema<T extends KubernetesJSONSchemaProps, V 
         resolvingContext.objectMapper.getSerializationConfig().constructType(definition), schema, null);
   }
 
-  private <A extends Annotation> T mapAnnotation(A annotation,
-      Function<A, T> mapper) {
-    if (annotation != null) {
-      return mapper.apply(annotation);
-    }
-    return null;
-  }
-
   /**
    * Walks up the class hierarchy to consume the repeating annotation
    */
--- a/crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/AbstractJsonSchema.java
+++ b/crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/AbstractJsonSchema.java
@@ -74,7 +74,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -84,6 +83,7 @@ import java.util.Set;
 import java.util.TreeMap;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -170,7 +170,7 @@ public abstract class AbstractJsonSchema<T extends KubernetesJSONSchemaProps, V 
   private <A extends Annotation> T mapAnnotation(A annotation,
       Function<A, T> mapper) {
     if (annotation != null) {
-        return mapper.apply(annotation);
+      return mapper.apply(annotation);
     }
     return null;
   }
@@ -419,8 +419,8 @@ public abstract class AbstractJsonSchema<T extends KubernetesJSONSchemaProps, V 
 
   private T resolveObject(LinkedHashMap<String, String> visited, InternalSchemaSwaps schemaSwaps, JsonSchema jacksonSchema,
       String... ignore) {
-    Set<String> ignores = ignore.length > 0 ? new LinkedHashSet<>(Arrays.asList(ignore)) : Collections.emptySet();
 
+    Set<String> ignores = Set.of(ignore);
     schemaSwaps = schemaSwaps.branchAnnotations();
     final InternalSchemaSwaps swaps = schemaSwaps;
 
@@ -434,8 +434,11 @@ public abstract class AbstractJsonSchema<T extends KubernetesJSONSchemaProps, V 
     Class<?> rawClass = gos.javaType.getRawClass();
     collectDependentClasses(rawClass);
 
-    JSONSchema schemaAnnotation = resolvingContext.ignoreJSONSchemaAnnotation ? null : rawClass.getDeclaredAnnotation(JSONSchema.class);
-    T classSchema = mapAnnotation(schemaAnnotation, schema -> fromAnnotation(rawClass, true, schema));
+    T classSchema = resolveSchemaAnnotation(
+        rawClass.getDeclaredAnnotation(JSONSchema.class),
+        rawClass,
+        true,
+        resolvingContext.ignoreJSONSchemaAnnotation);
 
     if (classSchema != null) {
       return classSchema;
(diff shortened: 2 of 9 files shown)


MikeEdgar:
Still a work-in-progress, needs lots of testing and feedback/comments on whether this seems generally viable :)

shawkins:
Thanks @MikeEdgar this looks promising.

It does seem highly related to SchemaSwap - at least for the usage of implemenation or type.

SchemaSwap could be altered to work at a field level, but one of the usage models it supports is for Keycloak, and maybe other projects, where the classes are defined in some other module and don't need or want CRD related annotations - so you are forced to define the swaps at a higher level than directly on the intended fields.

cc @andreaTP I know you have mentioned in the past wanted to see a general schema option - did you think of it working vis SchemaSwap or a separate annotation?

MikeEdgar:
> It does seem highly related to SchemaSwap - at least for the usage of implemenation or type.

Yes, I think the `implementation` attribute would be somewhat similar to `SchemaSwap`, but as you say the location is potentially different. The swaps are still limited to whatever schema can be obtained from the referenced class(es) using the existing scanning however, correct?

shawkins:
> The swaps are still limited to whatever schema can be obtained from the referenced class(es) using the existing scanning however, correct?

That is correct, you can't specify an arbitrary type in string form. I also forgot to mention SchemaFrom (the localized limited form of SchemaSwap). So instead of:

```
  @JSONSchema(type = "string")
  private Object customizedType;
```

You could do either:

```
  @SchemaFrom(type = String.class)
  private Object customizedType;
```

or 

```
  @SchemaSwap(originalType = JsonSchemaAnnoSpec.class, targetField = "customizedType", targetType = String.class)
  ...
```

With a similar comparison for implementation.

andreaTP:
Hi @MikeEdgar ! Sorry I completely missed the notification on this one :sweat: 

This proposal is highly related to `SchemaSwap` and `SchemaFrom`, have you attempted to implement the required functionality using one of the above?

> I know you have mentioned in the past wanted to see a general schema option - did you think of it working vis SchemaSwap or a separate annotation?

I'd be interested in a "general schema option" where we can swap in plain json or yaml without limiting the user to a fully typed interface to be as forward compatible as possible.
I have no strong opinion on `SchemaSwap` or another annotation, but, this is a big hammer, I'll leave the power to the user and let them inject whatever instead of trying to provide a safe typed access that will require maintenance and updates over time.

MikeEdgar:
> This proposal is highly related to `SchemaSwap` and `SchemaFrom`, have you attempted to implement the required functionality using one of the above?

I have not. Ultimately the problem with those annotations is that they redirect to another class/type to perform the same scanning logic, so there is no way to specify the schema attributes not already supported.

> I'd be interested in a "general schema option" where we can swap in plain json or yaml without limiting the user to a fully typed interface to be as forward compatible as possible.

I think allowing a JSON or YAML string in the annotation would be a nice feature, but it still needs to be parsed into `io.fabric8.kubernetes.api.model.apiextensions.v1.JSONSchemaProps`, so the type enforcement would just happen at runtime anyway. The new annotation in this PR could support it with a `value` property that is mutually exclusive with all other properties on the annotation.

>  I have no strong opinion on `SchemaSwap` or another annotation, but, this is a big hammer, I'll leave the power to the user and let them inject whatever instead of trying to provide a safe typed access that will require maintenance and updates over time.

Exactly, the intent with the annotation is to hand the control (big hammer) entirely to the user, with the "exit" of incorporating the default scanning using the `implementation` attribute. As mentioned earlier, type safety can't really be avoided in the model, so why not expose it in the annotation as well?

andreaTP:
> The new annotation in this PR could support it with a value property that is mutually exclusive with all other properties on the annotation.

Probably better to have a new annotation if a field is going to change the semantics.

> so the type enforcement would just happen at runtime anyway

Thanks for explaining! Makes sense!

> why not expose it in the annotation as well?

`io.fabric8.kubernetes.api.model.apiextensions.v1.JSONSchemaProps` is generated by `ModelGenerator`, in this proposal there is a 1:1 mapping manually defined, if it was my choice(and is not) I'd avoid it as it's a source of additional maintenance.

That said, the proposal and overall design makes sense, maybe let's try to fish for feedback on @manusa

shawkins:
>  Ultimately the problem with those annotations is that they redirect to another class/type to perform the same scanning logic, so there is no way to specify the schema attributes not already supported.

Correct me if I'm wrong, but is being proposed (and what is already supported) emcompasses:

- swap to
  - a class / implementation
  - to plain text
- overlay from the annotation (although this is possibly exclusive to the class / implementation mode)
- define at
  - the class level 
    - for the whole class
    - transitively applies to a particular class / field, for the case when you cannot modify the object model
  - at the field level 

If the overlay behavior is distinct to swapping to a class / implementation, then it would make the most sense to me to  put the overlay behavior in the existing SchemaSwap annotation, and add a new annotation(s) only for the swapping to text. I would be good with allowing SchemaSwap to be applied at the field level as well so that the new functionality does not have to also be added to SchemaFrom.

MikeEdgar:
> Correct me if I'm wrong, but is being proposed (and what is already supported) emcompasses:
> 
> * swap to
>   * a class / implementation
>   * to plain text

The plain text option isn't part of the PR yet, so it might be better handled separately as another enhancement.

> * overlay from the annotation (although this is possibly exclusive to the class / implementation mode)

It is supplemental to the class/implementation mode. This should feel familiar to those using the Swagger `@Schema` annotation [1].

> 
> * define at
>   * the class level
>     * for the whole class
>     * transitively applies to a particular class / field, for the case when you cannot modify the object model

It is only applied transitively to the extent that the class(es) provided for the `properties` map themselves are annotated or provide the desired schema information.

>   * at the field level
> 
> 
> If the overlay behavior is distinct to swapping to a class / implementation, then it would make the most sense to me to put the overlay behavior in the existing SchemaSwap annotation, and add a new annotation(s) only for the swapping to text.

It could be declared separately, but thinking about the `implementation` as the base and the annotation properties as overlaying that base do work hand in hand. And as mentioned above, it's inspired by the Swagger (and MicroProfile OpenAPI) annotation and should feel familiar to users of those.

> I would be good with allowing SchemaSwap to be applied at the field level as well so that the new functionality does not have to also be added to SchemaFrom.

Splitting out the `implementation` and relying on `SchemaSwap` as the "base" and the new annotation only to overlay may work equally well. I haven't looked at how the code would change for that. Would `SchemaFrom` be more similar to how I described `implementation` works?

[1] https://github.com/swagger-api/swagger-core/blob/f728c95fd8d6d9153cdcc18043a15bd6ae0612e3/modules/swagger-annotations/src/main/java/io/swagger/v3/oas/annotations/media/Schema.java

shawkins:
> Would SchemaFrom be more similar to how I described implementation works?

Yes SchemaFrom and SchemaSwap do the same thing as what JSONSchema.implementation does via the type and targetType properties respectively. 

The difference between them is that SchemaFrom is applicable to fields, while SchemaSwap is used at the class level. However I think with additional logic / validation they could be a single annotation.

> Splitting out the implementation and relying on SchemaSwap as the "base" and the new annotation only to overlay may work equally well.

That seems fine too as a separate annotation. Ideally it would be applicable to both the field level and the class level (using originalType / fieldName like SchemaSwap).

andreaTP:
- [SchemaFrom](https://github.com/fabric8io/kubernetes-client/blob/main/doc/CRD-generator.md#iofabric8crdgeneratorannotationschemafrom)
- [SchemaSwap](https://github.com/fabric8io/kubernetes-client/blob/main/doc/CRD-generator.md#iofabric8crdgeneratorannotationschemaswap)

Overall, I’m a bit hesitant to follow the OpenAPI/Swagger model too closely in this context.
While CRDs do use a format that resembles OpenAPI(but doesn't match 1:1), they remain fundamentally different in some important ways. For example, I’ve yet to encounter a CRD that uses `oneOf`, and the Go tooling around these more advanced constructs tends to either break or lead to obscure edge cases.

Because of that, I’d be cautious about offering a high-level, “nice to use” Java API that tries to generate these complex constructs as it may promise more than the underlying platform can reliably deliver.

> The plain text option isn't part of the PR yet, so it might be better handled separately as another enhancement.

I see things a bit differently here. To me, the raw representation,i.e. without strong typing, just pure JSON/YAML interfaces, is actually the most valuable part of this proposal. If we split it out and handle it separately, I worry that the added complexity and overlap in functionality could end up blocking progress rather than helping us move forward.

MikeEdgar:
> While CRDs do use a format that resembles OpenAPI(but doesn't match 1:1), they remain fundamentally different in some important ways. For example, I’ve yet to encounter a CRD that uses oneOf, and the Go tooling around these more advanced constructs tends to either break or lead to obscure edge cases.
>
> Because of that, I’d be cautious about offering a high-level, “nice to use” Java API that tries to generate these complex constructs as it may promise more than the underlying platform can reliably deliver.

I agree completely that the underlying "OpenAPI" implementation in Kube is limiting and even surprising to those familiar with OpenAPI/JSON schema. That being said, the intent with the new annotation is to anyway give control to the user and let them express what they need in the schema. It is up to them to understand those limitations and annotate accordingly. The only place the annotation attempts to be friendly to helping deal with the peculiarities of the Kube requirements is with the `structural` boolean noted in the description. 

Ultimately, the CRD schema will either be rejected or it will not validate as expected - something that an operator developer should discover in testing :-)

> I see things a bit differently here. To me, the raw representation,i.e. without strong typing, just pure JSON/YAML interfaces, is actually the most valuable part of this proposal. If we split it out and handle it separately, I worry that the added complexity and overlap in functionality could end up blocking progress rather than helping us move forward.

It's a valuable option, but it is still constrained by type checking when parsed into the model during CRD generation.

andreaTP:
> It's a valuable option, but it is still constrained by type checking when parsed into the model during CRD generation.

I like it, as the code to handle it is fully generated, I'm happy to handle the trade-off.

andreaTP:
> Ultimately, the CRD schema will either be rejected or it will not validate as expected

As far as I can tell, in this repo, we have always attempted to give ppl directly usable APIs, dealing with the complexity ourselves when possible.
If we need to delegate validation to an external system, I think, we need a better story about it.

MikeEdgar:
> > Ultimately, the CRD schema will either be rejected or it will not validate as expected
> 
> As far as I can tell, in this repo, we have always attempted to give ppl directly usable APIs, dealing with the complexity ourselves when possible. If we need to delegate validation to an external system, I think, we need a better story about it.

It's the same whether using a literal JSON/YAML schema in an annotation or typed properties, no? In other words, if the string literal is favored over the individual properties, it still exposes the users to the complexities directly and delegates the validation to Kube.

andreaTP:
> It's the same whether using a literal JSON/YAML schema in an annotation or typed properties, no?

That's accurate, it boils down to the expectations we settle, this is my reasoning:

- with a typed API -> I expect full support by the tooling
- with a plain string -> I know that I'm using an escape hatch that might not be fully supported

@MikeEdgar have you attempted to work(e.g. generate clients for different languages etc.) with a CRD that contains those advanced constructs? (I mean `anyOf`, `allOf`, `oneOf`) I expect most of the 3rd party tooling will fail in some way (happy to be proven wrong!).
The `java-generator`(in this repo) will not work, and, from a brief look, popular projects like [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder) do not support them too.

MikeEdgar:
At this point it seems unlikely for this PR to be merged as-is. Based on the feedback so far, is the thinking that the annotation should contain simply a string value with the JSON/YAML of the schema?

shawkins:
> The java-generator(in this repo) will not work, and, from a brief look, popular projects like [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder) do not support them too.

There's https://github.com/kubernetes-sigs/controller-tools/issues/461 - you can't start from the go types and generate a crd with structural schema

> At this point it seems unlikely for this PR to be merged as-is. Based on the feedback so far, is the thinking that the annotation should contain simply a string value with the JSON/YAML of the schema?

I agree with the idea of starting with the most general escape hatch first, then separately considering a more complex annotation and its relationship to existing annotations.

Are you proposing an annotation that would replace the entire schema for a given field or type? Would there also be a way to overlay / patch the generated schema?

MikeEdgar:
> Are you proposing an annotation that would replace the entire schema for a given field or type? Would there also be a way to overlay / patch the generated schema?

Essentially, yes. Perhaps with some basic rule that an object's properties are not scanned when `properties` itself is specified in the schema literal value.

For example, `BeanA` would only include `field2`, whereas `BeanB` would have both `field1` and `field2`. Users could omit all properties with `properties: null`, like `BeanC`, perhaps.

```java
@JSONSchema("""
    description: Bean A
    properties:
        field2:
            type: number
            format: int32
    """)
class BeanA {
    String field1;
    int field2;
}

@JSONSchema("""
    description: Bean B
    """)
class BeanB {
    String field1;
    int field2;
}

@JSONSchema("""
    description: Bean C
    properties: null
    """)
class BeanC {
    String field1;
    int field2;
}
```

MikeEdgar:
Closing for now to declutter. Let me know if this is something to explore again in the future.

Issue #5081, opened 2023-04-26
Add support for watching without pagination


shawkins:
### Is your enhancement related to a problem? Please describe

See https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/3157-watch-list/README.md - and related changes in the go-client.  Kubernetes api-servers will soon support better initial watch semantics such that list operations will not be required.

### Describe the solution you'd like

Change the watch parameters and implement support for the annotated bookmark denoting the end of the initial list.

csviri:
Hi @shawkins , do you plan to work on this? I might be able to pick this up, if not.

shawkins:
> Hi @shawkins , do you plan to work on this? I might be able to pick this up, if not.

I don't have plans currently. It would definitely be good to get in there. The pagination logic definitely complicates things, so it eventually should go.

csviri:
Just to make sure I understand what we want to achieve, so this should also replace the generic `list()` logic for resources. (Not just implement it for informer list watch)

csviri:
Putting here the new blogpost just for the record: https://kubernetes.io/blog/2024/12/17/kube-apiserver-api-streaming/

shawkins:
> Just to make sure I understand what we want to achieve, so this should also replace the generic list() logic for resources. (Not just implement it for informer list watch)

Are you thinking that would look starting a watch with a Watcher that collects values, and terminates after seeing sendInitialEvents action? That seems good. Or do you want to have something more async as well?

Regardless we'll need it, and the informer logic, to be driven either by server feature introspection or a config value.

csviri:
> Are you thinking that would look starting a watch with a Watcher that collects values, and terminates after seeing sendInitialEvents action? That seems good. Or do you want to have something more async as well?

Just the first part now, with the old apis in place (no async)

shawkins:
I'll have a pr with the changes to the informer shortly. Just wanted to see what are the thoughts on other usage of this feature. An example higher-level method could look like:

```
public static CompletableFuture<Void> streamingList(ListOptions listOptions, Consumer<T> onItem) {
    CompletableFuture<Void> future = new CompletableFuture<>();
    Watch watch = this.watch(listOptions, new Watcher<T>() {

      @Override
      public void eventReceived(Action action, T resource) {
        switch (action) {
        case ADDED:
          onItem.accept(resource);
          break;
        case BOOKMARK:
          future.complete(null);
          break;
        default:
[...]

PR #7190, opened 2025-07-28
enhance: add support for streaming lists

shawkins:
Adds support for informers to use watch / streaming lists. This is currently enabled via Reflector.setStreamingList

TODO:

- how to allow wiring of streamingList=true 
  - to do this automatically we'd need to check both the server version and feature gates
  - could be a setting on the Config (like the existing watch settings)
  - could also be exposed per informer

I'm leaning toward doing the latter two and not doing this automatically for now.

- additional validation
  - If you set a Limit it will simply be ignored

- other api methods, see https://github.com/fabric8io/kubernetes-client/issues/5081#issuecomment-3124763623

closes: #5081

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/impl/cache/Reflector.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/impl/cache/Reflector.java
@@ -63,7 +63,15 @@ public class Reflector<T extends HasMetadata, L extends KubernetesResourceList<T
   private CompletableFuture<Void> timeoutFuture;
 
   private boolean cachedListing = true;
-
+ 
+  private static class StreamingListState {
+    Set<String> nextKeys = new ConcurrentSkipListSet<>();
+    private CompletableFuture<Void> listDone = new CompletableFuture<>();
+  }
+ 
+  private boolean streamingList;
+  private volatile StreamingListState streamingListState;
+  
   public Reflector(ListerWatcher<T, L> listerWatcher, ProcessorStore<T> store) {
     this(listerWatcher, store, Runnable::run);
   }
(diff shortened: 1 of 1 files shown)


shawkins:
Added the Config wiring to enable using this feature. Checking the server for feature gate enablement seems a bit hackish - a raw request against /metrics and then scraping for the feature and a 1 value - so I'd like to just leave this as the only mechanism initially.

I would also leave off any built-in mechanism to use a watch list under the covers for a regular list operation.

@manusa @rohanKanojia having two similar constructors on Config seems dangerous (it caused me a little trouble here). Are you good with having a static method for the JsonCreator? We can even javadoc that as something that should not be called directly.

@csviri is there anything you would like to see done here?

csviri:
Hi @shawkins , look good to me, thank you!

> Added the Config wiring to enable using this feature. Checking the server for feature gate enablement seems a bit hackish - a raw request against /metrics and then scraping for the feature and a 1 value - so I'd like to just leave this as the only mechanism initially.

I guess you mean this:
https://github.com/kubernetes/kubernetes/issues/87869#issuecomment-1465634605

It is true that it is hacky, maybe we can experiment with it as the next iteration. Later if this would be enabled on K8S by default could be enabled based on version I guess.

Wonder if this listing should be extended by simple list eventually:

`context.getClient().resources(TestCustomResource.class).list();` 

That could be also beneficial.

shawkins:
> It is true that it is hacky, maybe we can experiment with it as the next iteration. Later if this would be enabled on K8S by default could be enabled based on version I guess.

Using the version would be much preferred. I guess in a rare case you could still disable the feature gate in a late release, so users could still be on the hook for manually toggling the Config value.

> Wonder if this listing should be extended by simple list eventually

To match the current behavior we need to return a typed KubernetesResourceList\<T>. 
- We lack a good way of constructing one that is modifable (PodList etc. don't have a common base impl), so we'd need to add setter methods onto KubernetesResourceList
- We won't easily know what the list kind is when dealing with a generic resource, we'd probably just have to resort to "List" - but that is a behavioral change.

manusa:
> @manusa @rohanKanojia having two similar constructors on Config seems dangerous (it caused me a little trouble here). Are you good with having a static method for the JsonCreator? We can even javadoc that as something that should not be called directly

The multiple config constructors are there for backwards compatibility and to support autoconfigure.

The ConfigXxxTests should verify that the current expectations still work.

The static JsonCreator should be a no go since there seems to be users with bizarre uses of Config, it's been a source of trouble for a while. 

We tried to reduce complexity as much as possible here, but it's the legacy users that have made this impossible to simplify even more.

What kind of troubles are you facing? Note there's also SundrioConfig (plus ConfigBuilder and ConfigFluent) that must be changed accordingly for the complete builder functionality to work

shawkins:
> The multiple config constructors are there for backwards compatibility and to support autoconfigure.

Can you elaborate more the backwards compatibility? Since both need modified anytime a new config item is added, how are they to stay compatible? Do you mean for example that if there is a new config item added that there should be 4 constructors instead of 2?

> What kind of troubles are you facing? Note there's also SundrioConfig (plus ConfigBuilder and ConfigFluent) that must be changed accordingly for the complete builder functionality to work

It relates to the last statement. We have two similar constructors. One with n parameters and one with n+1. If I am using the one with n+1 parameters, and the project adds new constructors to accomodate a new option, there will now either be:
- a constructor with n+1 and n+2 parameters
- or a constructor with n parameters, 2 constuctors with n+1 parameters, and a constructor with n+2 parameters

Depending on where in the parameter list the new thing is added there, and whether I'm using typed arguments, my usage of the old n+1 parameter constuctor may either be pointing to the wrong method or may be ambiguous.

I also don't see any path where we use a value for shouldSetDefaultValues other than true. Can you elaborate on how that is being used?

> We tried to reduce complexity as much as possible here, but it's the legacy users that have made this impossible to simplify even more.

I can certainly decouple this change from this pr, but what's here now seems very difficult to maintain.

> The static JsonCreator should be a no go since there seems to be users with bizarre uses of Config, it's been a source of trouble for a while.

Can you elaborate on what that would be? Are we supporting subclasses to Config where users want to have their own overridable constructor behavior? 

Is there an issue / discussion where we could talk about longer run alternatives, such as using some kind of overridable init method method rather than constructors to handle some of this.

manusa:
> Can you elaborate more the backwards compatibility? 

I can't recall exactly the scenarios.

Also, it seems that at some point the public constructor wasn't JavaDoced appropriately.

Users SHOULD NOT use the Config class constructor. They should rely on the Config builder instead or at most with the static Config creator methods.

> Do you mean for example that if there is a new config item added that there should be 4 constructors instead of 2?

No, that shouldn't be the case anymore (that was the case in v6) -however the JavaDoc problem still remains :facepalm:-

At this point, the single constructor should be updated, the SundrioConfig, ConfigBuilder and ConfigFluent should be updated too.
The builder should then work as expected.

> It relates to the last statement. We have two similar constructors. 

We should only have a **single public** constructor for Config, all Config instances should be created via builders or static methods.

This is probably the confusing part, likely because I didn't update the JavaDoc :sweat: 

> I also don't see any path where we use a value for shouldSetDefaultValues other than true. Can you elaborate on how that is being used?


I really can't recall at the moment, there should definitely be a test to verify this.
I recall we had a problem when we needed to have a clean-slate config without the defaults, but I can't remember what were the scenarios.

> Are we supporting subclasses to Config where users want to have their own overridable constructor behavior?

Yes, OpenShiftConfig is one example.

> such as using some kind of overridable init method method rather than constructors to handle some of this.

new ConfigBuilder().withWhatever().build() should be the approach, Config should really remain hidden from the users.
The coupling with Sundrio and the years of legacy are what's making the class hard to maintain and complex.

shawkins:
> new ConfigBuilder().withWhatever().build() should be the approach, Config should really remain hidden from the users.
The coupling with Sundrio and the years of legacy are what's making the class hard to maintain and complex.

See if https://github.com/fabric8io/kubernetes-client/pull/7204 will work for you.

manusa:
> > new ConfigBuilder().withWhatever().build() should be the approach, Config should really remain hidden from the users.
> > The coupling with Sundrio and the years of legacy are what's making the class hard to maintain and complex.
> 
> See if #7204 will work for you.

This seems to simplify things on the Config class by setting all the fields in the base Sundrio class.
For the classes that extend it I'm not sure we're making things simpler or more complicated.

From my side, I'm fine. It doesn't seem to break anything so feel free to mark it as ready and I'll merge

shawkins:
@manusa rebased to pick up the config changes.

manusa:
> @manusa rebased to pick up the config changes.

Thanks!
This is still marked as draft. Are all the changes in and it's ready to merge? shall we mark it as ready?

shawkins:
> This is still marked as draft. Are all the changes in and it's ready to merge? shall we mark it as ready?

Yes, should be ready now.

Issue #4724, opened 2023-01-04
Private configuration classes cause trouble with Java native (reflection)


manusa:
From https://github.com/quarkusio/quarkus/pull/30153

Some configuration classes (referenced issue is about `ExecConfig` but maybe there are more) have private access and need to be referenced by name in order to register them for reflection with GraalVM.

These classes should be exposed. In addition, it would be interesting to catalog these classes with an interface or something else, so that they can be automatically detected (Jandex) to be registered when used.

rohanKanojia:
At the moment I see these private classes in KubernetesClient related modules:

- [Config.ExecCredential](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/Config.java#L861)
- [Config.ExecCredentialSpec](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/Config.java#L869)
- [Config.ExecCredentialStatus](https://github.com/fabric8io/kubernetes-client/blob/d15bb1dbb250c4367ffe1248e9734fdf5b358fbd/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/Config.java#L873)
- [HttpClientUtils.HeaderInterceptor](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/HttpClientUtils.java#L49)
- [RunOperations.RunConfigNestedImpl](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/extended/run/RunOperations.java#L31)
- [URLFromNodePortImpl.NodePortUrlComponents](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/impl/URLFromNodePortImpl.java#L72)
- [KubernetesClientException.RequestMetadata](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/KubernetesClientException.java#L191)
- [Readiness.ReadinessHolder](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/readiness/Readiness.java#L51)
[...]

PR #6153, opened 2024-07-18
fix (kubernetes-client-api) : `Config`'s `autoConfigure` should disable auto configuration

rohanKanojia:
## Description

Fix #6137 

`Config`'s autoConfigure field should behave as per expectations.

We need to make changes to ConfigBuilder and ConfigFluent in order to achieve this.
- Add a new Constructor with additional boolean arguments `autoConfigure` and   `shouldSetDefaultValues`
- Add a new package private class `SundrioConfig` that would extend `Config` for generating the Sundrio Builder for Config
- Move generated classes `ConfigBuilder` and `ConfigFluent` to `src/main/java`
   - `ConfigBuilder` would extend generated SundrioConfigBuilder and override the `build()` method by calling `disableAutoConfig()` if `autoConfigure` is not provided by the user. Earlier `ConfigBuidler` was calling `new Config()` (this was enabling auto configuration)
- Make all fields in Config and ResourceConfig classes use boxed types instead of primitive types so that we can distinguish whether user has configured them.

## Type of change

 - [X] Bug fix (non-breaking change which fixes an issue)
 - [ ] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [ ] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [X] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [X] I Added [CHANGELOG](https://github.com/fabric8io/kubernetes-client/blob/main/CHANGELOG.md) entry regarding this change
 - [X] I have implemented unit tests to cover my changes
 - [ ] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/main/doc/CHEATSHEET.md) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [ ] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

PR #7204, opened 2025-08-02
rationalizing Config constructors / building

shawkins:
## Description

@manusa @rohanKanojia separated this from the pr dealing with watchLists. I realize that using just the swicth to use a static constructor is pretty narrow change. If possible I'd like to shoot for something bigger. The changes here get rid of all of our messiest constructors and hopely make adding new options more straight-forward.

- SundrioConfig -> goes from extending the Config to being a base class (could be renamed BaseConfig or something).  
  - There are two reasons to do this:
    - Creates a clean pojo from which to do sundrio generation.
    - Creates a clean pojo from which to do delegated Json deserialization - see the Config(SundrioConfig config) constructor.
  - Downsides:
    - Probably needs to be public (https://github.com/fabric8io/kubernetes-client/issues/4724) - or maybe just the class and not the constructor. Since all the apis expect Config, not SundrioConfig, there's not much that can be done with it. If we didn't have the need for overriding ConfigFluent methods, we could probably do away with SundrioConfig (see below for similar reasoning with OpenshiftConfigFluent)
  - Latent issues:
    -  additionalProperties and defaultNamespace do not seem to be handled properly (missing) in the current SundrioConfigFluent.
- ConfigFluent can just reuse the base logic, no need to mess with updating fluents when a new option is added
- You do still have to update the Config copy constructor to copy the value from the SundrioConfig.

- For Openshift the paradigm flips. There's now an SundrioOpenShiftConfig extending OpenshiftConfig with the dummy fields moved there.
  - I believe we could get rid of this class and the hardcoded OpenshiftConfigFluent/Builder if the logic in the copy constructor (if x is not null or empty, set x) were moved to the setter methods as that can all be done in the default sundrio logic post construction.

Other latent issues:
- We're subtlely changing the meaning of several of the ConfigBuilder consturctor methods, not sure if we should address that now that the change is already in.
- autoConfigure behavior seems inconsistent. If you do new ConfigBuilder().build() we'll use autoConfigure unless disabled, but if you parse we will only use autoConfigure if the field is true.
- OpenShiftConfig wrapping can rerun autoConfigure.

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/Config.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/Config.java
@@ -15,16 +15,12 @@
  */
 package io.fabric8.kubernetes.client;
 
-import com.fasterxml.jackson.annotation.JsonAnyGetter;
-import com.fasterxml.jackson.annotation.JsonAnySetter;
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonCreator.Mode;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import io.fabric8.kubernetes.api.model.AuthProviderConfig;
-import io.fabric8.kubernetes.api.model.ConfigBuilder;
-import io.fabric8.kubernetes.api.model.NamedContext;
 import io.fabric8.kubernetes.client.http.TlsVersion;
 import io.fabric8.kubernetes.client.internal.CertUtils;
 import io.fabric8.kubernetes.client.internal.KubeConfigUtils;
@@ -47,7 +43,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -59,7 +55,7 @@ import java.util.stream.Collectors;
 @SuppressWarnings({ "LombokGetterMayBeUsed", "LombokSetterMayBeUsed" })
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonIgnoreProperties(allowGetters = true, allowSetters = true)
-public class Config {
+public class Config extends SundrioConfig {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(Config.class);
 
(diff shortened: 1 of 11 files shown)


shawkins:
To reply to https://github.com/fabric8io/kubernetes-client/pull/7190#issuecomment-3150442858

There are three options:
- Have SundrioOpenShiftConfig extend SundrioConfig - this does not work well because then OpenshiftConfig can't easily extend Config.
- Have a SundrioOpenShiftConfig extend OpenShiftConfig (this pr). We still have to take responsibility for creating an OpenShift Fluent and Builder, and any new OpenShift options will be a pain / spread out to add.
- Just have OpenShiftConfig - the main problem here is that the generated builder will not be compatible with our logic. Moving the post default / autoconfigure constructor logic into setters could be breaking for anyone removing defaults via set methods - if we don't think this could or should happen, then we can could take this path instead.

@manusa @rohanKanojia this should be ready, but it looks like there is some flakeyness to one of the jdk tests. See the main pr comment for more of the thinking behind these changes.

manusa:
>  but it looks like there is some flakyness to one of the jdk tests

I'm not sure of the reason why, but tests have become extremely brittle and flaky.
This was not the case a couple of months ago and there have not been significant code changes that justify it.
There must either be issues with the dependencies that are less reliable now, or the GitHub actions environment.

Regarding the changes in general. I'm still unsure why this is necessary but if it helps then I'm happy with the changes.

> ... logic into setters 

The main goal for the config simplification was to make it as immutable as possible and to only allow its interaction via builders. Both are impossible given the original situation and the uses that people have been giving to the class. In any case, I'd still strive to remove the setters and any logic bound into them as much as possible.

shawkins:
> I'm still unsure why this is necessary but if it helps then I'm happy with the changes.

Sorry, I had hoped the value would have been clearer. Primarily it removes the very large constructors and any confusion that may arrise from updating them. Next it removes some of the steps of adding a new config property. There should just be two steps to add something like watchLists for the other pr - add a field on SundrioConfig, then copy the value from the SundrioConfig to the Config in the Config constructor. Finally it highlights issues with a couple of existing options, such as defaultNamespace and additionalProperties.

> In any case, I'd still strive to remove the setters and any logic bound into them as much as possible.

Config.setMasterUrl is probably the biggest exception to having logic in the setter.

For the rest it's mostly a semantic issue with default handling. For example `Config config = new ConfigBuilder().build(); config.setApiVersion(null);` - is expected to force a null api version, while `new ConfigBuilder().withApiVersion(null).build()` - is expected to use the default api version.

manusa:
I've been digging a little and I think the `if (Boolean.TRUE.equals(shouldSetDefaultValues)) {` was added for deserialization purposes only.
At the time https://github.com/fabric8io/kubernetes-client/commit/08b0e9f301fb163d9b98f71cd74fdf132ae7625b / #6153, there wasn't a constructor annotated with `@JsonCreator`. I'm still not quite sure why it was added in the first place.
Anyway, I think we can safely remove that fork and always set the defaults (in the constructor).

> For the rest it's mostly a semantic issue with default handling. For example Config config = new ConfigBuilder().build(); config.setApiVersion(null); - is expected to force a null api version, while new ConfigBuilder().withApiVersion(null).build() - is expected to use the default api version.

This is precisely the issue with the setters.
Config should ideally be managed only through the ConfigBuilder.

> additionalProperties and defaultNamespace do not seem to be handled properly (missing) in the current SundrioConfigFluent.

I can't see a way how defaultNamespace is actually consumed.
Similar to additionalProperties, I don't think these are applicable for this class.

> autoConfigure behavior seems inconsistent. If you do new ConfigBuilder().build() we'll use autoConfigure unless disabled, but if you parse we will only use autoConfigure if the field is true.

autoConfigure is opt-out, can only be skipped by using `Config.empty()`, by using `new ConfigBuilder(Config.empty()` or by `new ConfigBuilder().withAutoConfigure(false)`

Config parsing is the edge case which ideally should be unsupported (however, I'm unsure about who might be actually doing this).

> OpenShiftConfig wrapping can rerun autoConfigure.

This can be an issue if the user has opted out and is expecting to have a clean OpenShiftConfig.

shawkins:
> Anyway, I think we can safely remove that fork and always set the defaults (in the constructor).

Do you want me to do that in this pr?

> I can't see a way how defaultNamespace is actually consumed.

It's used via config.isDefaultNamespace() in OperationContext.

> Similar to additionalProperties, I don't think these are applicable for this class.

They are set upon deserialization. If you have anything unrecognized, or are even just parsing an OpenShiftConfig as a Config, then it will be populated. If you currently do something like parsedConfig.edit().build() - the additionalProperties are lost.

> Config parsing is the edge case which ideally should be unsupported (however, I'm unsure about who might be actually doing this).

I'm a little fuzzy on that too.

> This can be an issue if the user has opted out and is expecting to have a clean OpenShiftConfig.

This is happening specifically with the `public OpenShiftConfig(Config kubernetesConfig)` style constructors. If you have opted out, then auto configure won't be run either time, but in the other scenarios it will either run twice or as highlighted with the parsing case won't run when parsed, but then will run when wrapped.

manusa:
> It's used via config.isDefaultNamespace() in OperationContext.

Yes, I did find that, but it's not set anywhere. It also doesn't make sense to set it in the ConfigBuilder. That's why I'm unsure this is actually ever used.

> They are set upon deserialization. If you have anything unrecognized, or are even just parsing an OpenShiftConfig as a Config, then it will be populated. If you currently do something like parsedConfig.edit().build() - the additionalProperties are lost.

Yes, but is this a real use-case? i.e. something that can actually happen.

> This is happening specifically with the public OpenShiftConfig(Config kubernetesConfig) style constructors. If you have opted out, then auto configure won't be run either time, but in the other scenarios it will either run twice or as highlighted with the parsing case won't run when parsed, but then will run when wrapped.

The parsing scenarios are the ones that worry me the least. They shouldn't really be supported. I sort of recall that this could be used in IDE tooling, but hopefully they might have refactored that part.

shawkins:
> Yes, I did find that, but it's not set anywhere. It also doesn't make sense to set it in the ConfigBuilder. That's why I'm unsure this is actually ever used.

It was supposed to be configurable.

> Yes, but is this a real use-case? i.e. something that can actually happen.

I'd want to keep things as regular and predicable as possible.

manusa:
Seems like Karaf is not very happy with the changes, test suite keeps failing.

shawkins:
> Seems like Karaf is not very happy with the changes, test suite keeps failing.

But only for Java 11? And without any real logging of a problem. Let me see if I can reproduce this locally.

manusa:
> But only for Java 11? And without any real logging of a problem. Let me see if I can reproduce this locally.

Thx!

I don't think these tests run on any other platform besides Java 11.
Probably some missing registration or something similar.

I'd do this myself, but I don't think I'll find time.
I'd like to merge this PR and start with a release process (which will likely consume most of my available time since the Maven Central change is likely going to impact).

shawkins:
Let's see if this is better. Looks like some of the openshift client changes needed tweaked.

manusa:
> Let's see if this is better. Looks like some of the openshift client changes needed tweaked.

Tests are passing now, thank you!

manusa (approved):

Issue #7160, opened 2025-07-14
LeaderElector.canBecomeLeader should read the leaseDuration from the Kube Lease object and not from local configuration


zoliszel:
### Describe the bug

Hi there, 

to decide if the current process can become leader,  io.fabric8.kubernetes.client.extended.leaderelection.LeaderElector infers expiry of a lease using its local lease duration configuration: 

```
  protected final boolean canBecomeLeader(LeaderElectionRecord leaderElectionRecord) {
    return Utils.isNullOrEmpty(leaderElectionRecord.getHolderIdentity())
        || now().isAfter(leaderElectionRecord.getRenewTime().plus(leaderElectionConfig.getLeaseDuration()));
  }
```

This is incorrect, applications within the same cluster can/will use different KubeAPI clients and even then different applications can decide to acquire the lease for their own duration of choice. 

The duration of the lease is maintained in the `lease.spec.leaseDurationSeconds` field inside of the `LeaderElectionRecord` object which should be used to source how long the lease is held by the current owner. 

This can lead to be premature loss of lease if 2 clients using different leaseDurations are deployed into the same cluster.   

### Fabric8 Kubernetes Client version

SNAPSHOT

### Steps to reproduce

* Prepare 2 LeaderElector, 
  * LE1: LeaseDuration of 10s, retry period 1s (LE1), 
  * LE2: LeaseDuration of 30s, retry period, 20s(LE2) 
* Start LE2 first , let it acquire the lease, then start LE1

### Expected behavior

LE2 should keep hold of the lease until its shut down, but LE1 will take over instead. 

### Runtime

Kubernetes (vanilla)

### Kubernetes API Server version

1.33

### Environment

Linux

### Fabric8 Kubernetes Client Logs

```shell

```

shawkins:
I believe this is one of several things that changed in the go client after the initial port to the fabric8 client, but has never been implemented on the fabric8 side.

https://github.com/kubernetes/client-go/blob/3a4ad9c658b1bd280b1ddf397669c9876b05fdba/tools/leaderelection/leaderelection.go#L360

shawkins:
@zoliszel can you open a pr for this?

iamritikbhardwaj:
Hi can i raise a pr since the bug is already explained how about instead of getLeaseDuration() we use getLeaseDurationSeconds() so it gets the actual configuration in kuberneties and not the local client and 
It's my first contribution can you please help me or give pointer if i am wrong or missing something

[...]

PR #7258, opened 2025-08-27
changed leaderElectionConfig.getLeaseDuration to leaderElectionConfig…

iamritikbhardwaj:
Fixes [[#7160]
This pull request addresses a bug in the `LeaderElector` class, specifically in the `canBecomeLeader()` method. Previously, the method used the local `leaderElectionConfig.getLeaseDuration()` value to determine if a lease had expired.

This logic is flawed in scenarios where multiple leader election participants use different `leaseDuration` values. According to Kubernetes leader election semantics, clients should evaluate lease expiration based on the duration stored in the Lease object itself (`leaseDurationSeconds`), not their own configuration.

This discrepancy can lead to premature leadership takeovers and split-brain behavior. For example:

* One client (`LE2`) runs with `leaseDuration = 30s`, `retryPeriod = 20s`.
* Another client (`LE1`) runs with `leaseDuration = 10s`, `retryPeriod = 1s`.

Even if `LE2` is correctly renewing the lease within its 30-second window, `LE1` may incorrectly assume the lease has expired after 10 seconds, simply because it's using its own local config to calculate expiry.

### Fix

The method now uses the lease duration published in the lease record:

```java
Duration.ofSeconds(leaderElectionRecord.getLeaseDurationSeconds())
```

This ensures all clients interpret lease expiry based on the same source of truth — the current lease holder’s declared configuration — which aligns with how Kubernetes itself handles leader election.

---

## Type of change

* [x] Bug fix (non-breaking change which fixes an issue)

---

## Checklist

* [x] The code I am contributing is licensed under the Apache 2.0 License
* [x] A CHANGELOG entry has been added for this change
* [x] Unit or integration tests have been added or updated
* [x] Documentation and/or Javadocs have been added or updated where relevant
* [x] No new issues or regressions are reported by SonarCloud
* [x] The change has been tested on Kubernetes
* [ ] The change has been tested on OpenShift


shawkins:
> The method now uses the lease duration published in the lease record:

The code change does not match this description.

Also a test case is needed.

iamritikbhardwaj:
Okay then, i wiil read it again carefully

shawkins:
Closing in favor of #7368 since this hasn't been updated in a while.

PR #7315, opened 2025-09-27
Allow VertxHttpClientBuilder to be seeded with a base WebClientOptions object, allowing more customization of the Vertx HTTP Client

### LINKED ISSUE #7252: Vert.x HttpClient should provide ability to customize WebClientOptions

The VertxHttpClientFactory provides a stub for adding more options. But unlike the other HttpClient implementations in the project, it is never called. I feel that it is never called because the VertxHttpClientBuilder doesn't force the Factory type like the others do and since the base HttpClient.Factory interface does not require an additionalConfig method to be implemented, there is no guarantee that the factory used for the builder provides such a method.

A user leveraging the existing VertxHttpClientBuilder should have the ability to provide overrides for the underlying WebClientOptions. Vertx provides a lot of configuration knobs, and the defaults are not going to suit everyone.

Using the existing pattern of additionalConfig call from build() would be sweet, but I'm not sure that can be done without a breaking change. So short of that I'm proposing the idea of a base options object that can be provided to the builder. The build() method will use that base object if it was provided. This gives users the ability to modify vertx options if they so choose.

Stoping using generics for the Factory in VertxHttpClientBuilder and instead requiring the VertxHttpClientFactory (or extension of it). This could allow us to call the existing `additionalConfig` method from the Builder.build(). 

The biggest reason against this is that it is a breaking change for any user not using the stock factory.

### Additional context

I'm coming with this because we moved to vert.x HttpClient in Apache Druid. We were using okhttp in the past. Using vert.x we have found that the defaults around the idle timeouts in vert.x are not sufficient and we are running into instances where our KubernetesClient is trying to use connections that have been closed on the server side. We feel that vert.x should be more aggressively monitoring connection health and cleaning up closed connections. To achieve this we need access to override underlying vert.x options.

capistrant:
## Description

fixes #7252 

Allows user to seed the Vert.x Http Client Builder with a custom WebOptions object if they want to modify configs of this object. As the issue specs out, this is not the most ideal solution, I think using additionalConfig like the other HttpClient implementations do would be best. But that seems like a breaking change since it would be more invasive.

## Type of change

## Checklist
 - [X] I Added [CHANGELOG](https://github.com/fabric8io/kubernetes-client/blob/main/CHANGELOG.md) entry regarding this change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 #### Dependency Upgrade
 
 #### New Features
+* Fix #7252: Allow VertxHttpClientBuilder to be seeded with a base WebClientOptions object, allowing more customization of the Vertx HTTP Client
 
 #### _**Note**_: Breaking changes
 
--- a/httpclient-vertx/src/main/java/io/fabric8/kubernetes/client/vertx/VertxHttpClientBuilder.java
+++ b/httpclient-vertx/src/main/java/io/fabric8/kubernetes/client/vertx/VertxHttpClientBuilder.java
@@ -50,6 +50,7 @@ public class VertxHttpClientBuilder<F extends HttpClient.Factory>
 
   final Vertx vertx;
   private final boolean closeVertx;
+  private WebClientOptions customWebClientOptions;
 
   public VertxHttpClientBuilder(F clientFactory, Vertx sharedVertx) {
     this(
@@ -64,13 +65,31 @@ public class VertxHttpClientBuilder<F extends HttpClient.Factory>
     this.closeVertx = closeVertx;
   }
 
+  /**
+   * Use the provided {@link WebClientOptions} as the base options for the client.
+   * <p>
+   *   Note: The Builder overrides some options to ensure the client works properly with the Kubernetes API server.
+   *   These options that the Builder controls will override the values you set in the provided
+   *   {@link WebClientOptions}.
+   * </p>
+   */
+  public VertxHttpClientBuilder<F> withCustomWebClientOptions(WebClientOptions options) {
+    this.customWebClientOptions = options;
+    return this;
+  }
+
   @Override
   public VertxHttpClient<F> build() {
     if (this.client != null) {
       return new VertxHttpClient<>(this, this.client.getClosed(), this.client.getClient(), closeVertx);
     }
 
-    WebClientOptions options = new WebClientOptions();
+    WebClientOptions options;
+    if (customWebClientOptions == null) {
+      options = new WebClientOptions();
+    } else {
+      options = new WebClientOptions(customWebClientOptions);
+    }
 
     options.setMaxPoolSize(MAX_CONNECTIONS);
     options.setMaxWebSockets(MAX_CONNECTIONS);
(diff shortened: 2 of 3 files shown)


capistrant:
> Looks good to me. Can you ready to take it out of draft?

did that just now. Was working on running it a bit in local k8s env and noodling on how we could actually test that the configs are getting applied. Especially in the sense of not having it regress in future and quietly stop taking effect. Running in dev, I did a little bit on Friday and will do some more today. The testing strategy, I'm still stumped on

shawkins:
> did that just now. Was working on running it a bit in local k8s env and noodling on how we could actually test that the configs are getting applied. Especially in the sense of not having it regress in future and quietly stop taking effect. 

Actually I see there is something we originally missed. The VertxHttpClientFactory.additionalConfig method should be called by VertxHttpClientBuilder.build method. That is consistent with the fabric8 factories and somewhat redundant with the ability to pass in a WebClientOptions. If you wire in that method, you should be able to check that your passed in options is in use.

I'm not opposed to having both ways of handling the WebClientOptions as subclassing is probably less obvious than looking for a builder method.

cc @manusa

capistrant:
> > did that just now. Was working on running it a bit in local k8s env and noodling on how we could actually test that the configs are getting applied. Especially in the sense of not having it regress in future and quietly stop taking effect.
> 
> Actually I see there is something we originally missed. The VertxHttpClientFactory.additionalConfig method should be called by VertxHttpClientBuilder.build method. That is consistent with the fabric8 factories and somewhat redundant with the ability to pass in a WebClientOptions. If you wire in that method, you should be able to check that your passed in options is in use.
> 
> I'm not opposed to having both ways of handling the WebClientOptions as subclassing is probably less obvious than looking for a builder method.
> 
> cc @manusa

https://github.com/fabric8io/kubernetes-client/issues/7252 did talk about this as an alternative I considered. The reason I did not propose it was to avoid proposing what would be a breaking change for folks not using the default factory. That is at least how I interpreted the change, that it could be breaking. If that is ok, I could pursue that instead. It does make the most sense in terms of keeping patterns consistent across the project

capistrant:
> > did that just now. Was working on running it a bit in local k8s env and noodling on how we could actually test that the configs are getting applied. Especially in the sense of not having it regress in future and quietly stop taking effect.
> 
> Actually I see there is something we originally missed. The VertxHttpClientFactory.additionalConfig method should be called by VertxHttpClientBuilder.build method. That is consistent with the fabric8 factories and somewhat redundant with the ability to pass in a WebClientOptions. If you wire in that method, you should be able to check that your passed in options is in use.
> 
> I'm not opposed to having both ways of handling the WebClientOptions as subclassing is probably less obvious than looking for a builder method.
> 
> cc @manusa

@shawkins I opened up #7384 as an alternative to this PR, which uses the additionalConfig method already available. As you'll see the generic factory in the builder ties our hands there, but maybe we can at least provide the functionality to folks using the stock factory. Appreciate if you can take a look and offer thoughts!

shawkins:
@capistrant thank you for the additional pr. Let's see if @manusa agrees with that one, then we can probably close this one.

capistrant:
closing in favor of #7384

PR #7346, opened 2025-10-27
null identity when canceled (fix-7343)

### LINKED ISSUE #7343: leader election callbacks triggered twice
### Describe the bug

Here is the scenario step by step:

- leader election uses lease as the lock
- lease is _not_ present in k8s
- leader election is started for a certain pod
- pod is started with `releaseOnCancel`
- pod is shut down in a graceful way
- lock is released and `LeaderElector::stopLeading` is triggered:

```
if (leaderElectionConfig.isReleaseOnCancel()) {
      try {
        if (release()) {
          return;
        }
      }
```

As a result, the record inside the lease is set to be : 

```
    final LeaderElectionRecord newLeaderElectionRecord = new LeaderElectionRecord(
        "",
        Duration.ofSeconds(1),
        now,
        now,
        current.getLeaderTransitions());
```

- pod is started again, `LeaderElector::tryAcquireOrRenew` is hit and : 

```
if (oldLeaderElectionRecord == null) { .. }
```

is _not_ triggered, because that record exists. 

Instead, it goes to the next line on : `updateObserved(oldLeaderElectionRecord);`. Let's see its code:

```
  private void updateObserved(LeaderElectionRecord leaderElectionRecord) {
    final LeaderElectionRecord current = observedRecord.getAndSet(leaderElectionRecord);
    if (!Objects.equals(leaderElectionRecord, current)) {
      final String currentLeader = current == null ? null : current.getHolderIdentity();
      final String newLeader = leaderElectionRecord.getHolderIdentity();
      if (!Objects.equals(newLeader, currentLeader)) {
        LOGGER.debug("Leader changed from {} to {}", currentLeader, newLeader);
        // this will notify even if the newLeader is null or empty, which is the same behavior as the go client
        // but does not seem entirely correct
        leaderElectionConfig.getLeaderCallbacks().onNewLeader(newLeader);
        if (Objects.equals(currentLeader, leaderElectionConfig.getLock().identity())) {
          leaderElectionConfig.getLeaderCallbacks().onStopLeading();
        } else if (Objects.equals(newLeader, leaderElectionConfig.getLock().identity())) {
          LOGGER.debug("Successfully Acquired leader lease '{}'", leaderElectionConfig.getLock().describe());
          leaderElectionConfig.getLeaderCallbacks().onStartLeading();
        }
      }
    }
  }
```

This: `final LeaderElectionRecord current = observedRecord.getAndSet(leaderElectionRecord);` is resolved to `null` since there is yet no entry in `observedRecord`. 

We then go to : 

```
if (!Objects.equals(newLeader, currentLeader)) { .. }
```

`newLeader = ""` and `currentLeader = null`. This results in callback being called - which should not, the leader is not established. 

Same code is going to be triggered _again_ for the same callback, with the correct leader in the next cycle.

----

I've implemented leader election in spring-cloud-kubernetes and noticed that we have such logs, for example, via our callbacks:

```
...
id :  is the new leader
...
id : readiness-passes-simple-it is the new leader
...
```

which got me into debugging the code and figuring out what is going on. 

The fix, imho, should be trivial

```
 final LeaderElectionRecord newLeaderElectionRecord = new LeaderElectionRecord(
        null, // --> HERE null instead of ""
        Duration.ofSeconds(1),
        now,
        now,
        current.getLeaderTransitions());
```

or the equality check in the `LeaderElector` should be enhanced. 

Thank you.

SNAPSHOT

### Steps to reproduce

already provided above

### Expected behavior

already provided above

### Runtime

Kubernetes (vanilla)

1.33

### Environment

macOS

```shell
I've explained and provided the logs above too.
```

### Additional context

_No response_

wind57:
Fixes #7343


manusa (approved):
LGTM, thx!

Issue #7164, opened 2025-07-14
SharedIndexInformer endlessly attempts to reconnect to the down Kube API server


19Serhii99:
### Describe the bug

I am testing behavior of the Fabric8 SharedIndexInformer during having network instability. I am trying to determine whether it is going to fail or recover when a connection is lost so that I could handle this scenario as neeeded. I ran `kubectl proxy` to control connectivity, so my master URL is http://127.0.0.1:8001.

<pre>
try (SharedIndexInformer<Pod> informer = client.pods()
  .inNamespace("$NAMESPACE")
  .withName("$POD_NAME")
  .runnableInformer(0L)) {
    informer.addEventHandler(...);
    informer.exceptionHandler(...);
}
</pre>

When running the program, I terminate the proxy to break the connection, and then run `kubectl proxy` command again. Instead of invoking `retryOnException` we passed to `exceptionHandler(....)`, it's endlessly attempting to reconnect, throwing lots of red logs (see below).

Of course, if you `kubectl poxy` again, it will reconnect and continue monitoring the pod state but if the server is down for long, then I would like it to fail sooner or later (`retryAfterException` would control it)

### Fabric8 Kubernetes Client version

7.3.1

### Steps to reproduce

1. Run `kubectl proxy`
2. Initialize and start a new SharedIndexInformer
3. Wait until it connects and receives first pod events
4. Terminate the proxy (Ctrl + C in the terminal)
5. Observe the logs -  it's endlessly attempting to reconnect

### Expected behavior

`ExceptionHandler.retryAfterException` is invoked

### Runtime

Kubernetes (vanilla)

### Kubernetes API Server version

other (please specify in additional context)

### Environment

Windows, Azure

### Fabric8 Kubernetes Client Logs

```shell
[vert.x-eventloop-thread-1] ERROR io.vertx.core.net.impl.ConnectionBase - Connection reset
[vert.x-eventloop-thread-3] INFO io.fabric8.kubernetes.client.dsl.internal.AbstractWatchManager - Unknown Watch error received 152 times without progress, will reconnect if possible
java.util.concurrent.CompletionException:
io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: no further information: /127.0.0.1:8001 at java.base/java.util.concurrent.CompletableFuture.encode Throwable (CompletableFuture.java:332)
at java.base/java.util.concurrent.CompletableFuture.complete Throwable (CompletableFuture.java:347)
[...]

Issue #7365, opened 2025-11-21
Infinite loop in AbstractWatchManager


fpoirotte:
### Describe the bug

Hi,

I'm not entirely sure this is the right place to report this issue.
If it is not, feel free to close it.

We are using Jenkins with a Kubernetes cluster to run continuous integration builds.
Jenkins uses this library to interact with the cluster.

Our workflow is this one:
* Jenkins triggers a build
* the build provisions Kubernetes (sub)namespaces dynamically (one subnamespace per test) and sets permissions, etc.
* the rest of the build runs each test in its dedicated namespace
* dynamically-created subnamespaces are deleted at the end of the build

In our logs, we see tons of errors such as this one:
```
Nov 21 10:20:44 jenkins.localdomain jenkins[2459376]: 2025-11-21 09:20:44.725+0000 [id=339220]        INFO        i.f.k.c.d.i.AbstractWatchManager#logEndError: Watch connection error received 7259 times without progress, will reconnect if possible
```
(here `test-221-81077638-4924-3557-ac8b-a3aa976f6499` refers to one of the dynamically-provisioned namespaces)

In this case, the error is permanent and keeps on repeating.
As time goes on, more and more of these errors are logged (with varying namespaces).
This causes Jenkins to eat up more and more memory as well, until the UI becomes unresponsive and we have to restart the service.

Note: this issue has also been reported to the maintainers of Jenkins' Kubernetes Plugin here:
https://issues.jenkins.io/browse/JENKINS-73351

### Fabric8 Kubernetes Client version

7.3.1

### Steps to reproduce

1. Create a subnamespace using the hierarchical namespace controller (https://github.com/kubernetes-retired/hierarchical-namespaces) : `kubectl hns create -n "${parentNS}" "${subNs}"`
2. Use the new subnamespace in the Jenkins pipeline:
```
            podTemplate(
                cloud: cloudName,
                namespace: subNs,
                ...
           ) {
               node(POD_LABEL) {
                   ...
               }
           }
```

3. Delete the subnamespace at the end of the build:
```
    post {
        always {
            container('kubectl') {
                script {
                    catchError(buildResult: null) {
                        sh """
                            kubectl delete --ignore-not-found=true --grace-period=1 subns -n '${parentNS}' '${subNamespaces.values().join("' '")}
                        '"""
                    }
                }
            }
        }
    }
```

### Expected behavior

[...]

PR #7367, opened 2025-11-24
fix: allowing the exceptionhandler to see reconnection exceptions

shawkins:
closes: #7164

draft of changes to allow the ExceptionHandler to deal with reconnection exceptions as well.

To keep this compatible with current behavior the WatcherException is converted to a KubernetesClientException when passed into the the ExceptionHandler - by default that will mean we'll keep retrying.

Users would need to update / set their own ExceptionHandler to cause the informer to give up in this case. This is mostly inline with what was requested on #7164

Alternatives to this are:
- Pass the exception in as a WatcherException, and let the informer stop watching - by itself this would be a breaking change.
- Alternatively don't use the ExceptionHandler mechanism, and add Informer configuration for the number of watch reconnections it will attempt before giving up - it would default to unlimited for backwards compatibility. This is more inline with what is requested on #7365

cc @csviri @manusa @metacosm what would you like to see done here?

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/WatcherException.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/WatcherException.java
@@ -31,7 +31,7 @@ public class WatcherException extends Exception {
   }
 
   public WatcherException(String message, Throwable cause, boolean reconnectionError) {
-    super(message);
+    super(message, cause);
     rawWatchMessage = null;
     this.reconnectionError = reconnectionError;
   }


kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
```diff
@@ -211,7 +211,7 @@ final synchronized void cancelReconnect() {
   /**
    * Called to reestablish the connection. Should only be called once per request.
    */
-  void scheduleReconnect(WatchRequestState state) {
+  void scheduleReconnect(WatchRequestState state, Throwable t) {
```
metacosm: Might be better to keep the original, one-arg version and add the two-arg version to which it would delegate so that call sites do not need to change? Looks also cleaner than to pass `null` explicitly… I wish Java had a way to declare optional parameters…

metacosm (changes requested):

metacosm:
Regarding the overall question of which approach to choose, I'm not sure… The more correct `isWatching` behavior is causing issues upstream because it's difficult to figure out, without human intervention that knows of the context, whether or not a watcher getting repeatedly disconnected is a "real" issue or not.

shawkins:
> Regarding the overall question of which approach to choose, I'm not sure… The more correct `isWatching` behavior is causing issues upstream because it's difficult to figure out, without human intervention that knows of the context, whether or not a watched getting repeatedly disconnected is a "real" issue or not.

I understand it's now causing problems, but that's more because it was not implemented correctly - it was nearly identicaly to isRunning, which was not the intent.

As for the correct behavior it's just a question of:
- do we just want alternative behavior to be driven via a custom ExceptionHandler, in which case this pr is fine as is
- or do we want a new property, either on the informer or in the client configuration or both that dictactes how many times an informer is allowed to retry starting a failed watch? It would probably have to default to unlimited to not be breaking.

metacosm:
>     * or do we want a new property, either on the informer or in the client configuration or both that dictactes how many times an informer is allowed to retry starting a failed watch? It would probably have to default to unlimited to not be breaking.

This solution would probably lead down the way of replicating the kube probe configuration 😄

shawkins:
> > ```
> > * or do we want a new property, either on the informer or in the client configuration or both that dictactes how many times an informer is allowed to retry starting a failed watch? It would probably have to default to unlimited to not be breaking.
> > ```
> 
> This solution would probably lead down the way of replicating the kube probe configuration 😄

It's a little different. Here we're talking about what behavior to exhibit when the watch fails to re-establish. Currently it's unlimited retries, but there's been two user requests to refine that behavior.

In any case there's still absolutely no guarentee that restarting the application will help in these circumstances. In older versions of fabric8 there were some issues with watches not re-establishing, or cache sync issues, but I don't believe we have seen an issue along those lines in a while.

metacosm:
Added informer configuration properties in #7525

Issue #5285, opened 2023-06-26
[java-generator] Improve the docs to include examples of the options


andreaTP:
### Is your enhancement related to a problem? Please describe

Currently, the docs are listing the available options for the plugins and the CLI, but there are no examples of "how to use them" and the expected output.

### Describe the solution you'd like

We should describe the available options and include snippets of the usage and how it affects the generation.

andreaTP:
Hi @shubhamvrs and thanks for your interest!

To complete this task we do expect a PR expanding this section of the current docs:
https://github.com/fabric8io/kubernetes-client/blob/main/doc/java-generation-from-CRD.md#usage

Would you be able to provide it?

shubhamvrs:
i tried sir but im unable to create it actually im new to github thatswhy !

rohanKanojia:
@shubhamvrs : Could you please elaborate? What problem are you facing?

shubhamvrs:
im not  able to provide a PR expanding this section: ( https://github.com/fabric8io/kubernetes-client/blob/main/doc/java-generation-from-CRD.md#) usage of the current docs that i have created previously

rohanKanojia:
@shubhamvrs : Do you mean you don't have enough knowledge about java generator to improve existing documentation?

shubhamvrs:
im unable to create new file in given format with given path in new pull request

rohanKanojia:
@shubhamvrs : Why is that so? Are you getting some error?

shubhamvrs:
@rohanKanojia its just im unable to create a file on given format. I have the doc but file needs to be converted in given format !

rohanKanojia:
I'm not able to understand what you're saying. Which document are you talking about?

shubhamvrs:
Enhancement Proposal: Improving Documentation with Examples of Java Generator Options

Problem:
The current documentation for the Java generator lacks practical examples of how to use the available options and how they affect the generated code. Users might find it difficult to understand how to utilize these options effectively without concrete examples.

Solution:
To address this problem, we propose enhancing the documentation for the Java generator by providing clear and concise examples of how to use each option. Each option should be accompanied by code snippets that demonstrate its usage and illustrate the impact on the output.

Java Generator Options and Examples:

Option: setPackageName(String packageName)

Description: Sets the package name for the generated Java code.
Example:
java
Copy code
JavaGenerator generator = new JavaGenerator();
generator.setPackageName("com.example.generated");
[...]

PR #7383, opened 2025-12-16
doc(java-generator): add configuration examples for plugin options

taole33:
## Description

Refs #5285

This PR improves the documentation for the Java Generator by adding configuration example snippets in the Usage section.

Currently, the documentation lists available options but lacks practical examples of how to configure them in `pom.xml` or `build.gradle`.
This update provides concrete examples for Maven, Gradle, and CLI usage, along with the expected generated Java code for the following common options:

- Uppercase Enums (`enumUppercase`)
- Package Overrides (`packageOverrides`)
- Always Preserve Unknown Fields (`alwaysPreserveUnknown`)
- Filter Source Files (`filesSuffixes`)
- Existing Java Types (`existingJavaTypes`)

This addresses the missing link between configuration and output as pointed out in #5285.

## Type of change
 - [x] Chore (non-breaking change which doesn't affect codebase; test, version modification, documentation, etc.)

## Checklist


ash-thakur-rh (dismissed):
I will do small fixes to make the document changes ready

ash-thakur-rh (commented):
Thanks for the contribution! The examples are accurate and match the actual plugin code — verified against the Maven Mojo, Gradle extension, CLI flags, and approval tests.

A few small fixes needed:

1. **Markdown fence typo** — First Maven XML code block closes with four backticks instead of three. This breaks rendering.

2. **`alwaysPreserveUnknown` snippet** — The actual generator puts `@JsonIgnore` on the field, `@JsonAnyGetter` on the getter, and `@JsonAnySetter` on the setter. The snippet currently shows the annotations on the field directly.

3. **Inconsistent heading escapes** — Section 1 uses `1.` but sections 2–5 use `2\.` with a backslash. Minor but worth making consistent.

4. **Section 4 (filesSuffixes)** — Missing a "Generated Code" block or a note that this option only controls which files are processed.

Will update the PR with these small fixes and mark it ready.

ash-thakur-rh (approved):
LGTM!

andreaTP:
:tada: Thanks! Waiting to see a skill come next :stuck_out_tongue:

doc/java-generation-from-CRD.md
```diff
@@ -241,6 +241,211 @@ And the corresponding configurations of the Maven plugin are (output of `mvn hel
...
+```bash
+--enum-uppercase
+```
+
+**Generated Code:**
+
+```java
+public enum Material {
+    @com.fasterxml.jackson.annotation.JsonProperty("plastic")
+    PLASTIC("plastic"), // Uppercase constant
```
andreaTP: nitpick: uppercase enum as compared with default.

doc/java-generation-from-CRD.md
```diff
@@ -241,6 +241,211 @@ And the corresponding configurations of the Maven plugin are (output of `mvn hel
...
+
+```bash
+--always-preserve-unknown
+```
+
+**Generated Code:**
+
+```java
+public class ToySpec implements KubernetesResource {
+
```
andreaTP, line 369: nit:

```
// added to base class to capture un-specified fields and values
```

remove the `private` Map? is an implementation detail.

ash-thakur-rh: done!

doc/java-generation-from-CRD.md
ash-thakur-rh: done!

manusa (changes requested):
Thanks for the iteration! Two things from the latest push went the opposite way from what was needed.

The `enumUppercase` example has the *default* and *with-flag* labels swapped. `DEFAULT_UPPERCASE_ENUM = true` ([`Config.java#L29`](https://github.com/fabric8io/kubernetes-client/blob/3cc2f3caebcf7eb6f1065e276b426df56d126658/java-generator/core/src/main/java/io/fabric8/java/generator/Config.java#L29)) and `isUppercaseEnums()` falls back to that constant when the value is unset ([`Config.java#L164-L166`](https://github.com/fabric8io/kubernetes-client/blob/3cc2f3caebcf7eb6f1065e276b426df56d126658/java-generator/core/src/main/java/io/fabric8/java/generator/Config.java#L164-L166)), so the **uppercase** snippet is actually what you get out of the box. The lowercase one is what `enumUppercase = false` produces. Swapping the two labels would do it. The section intro "Force the generated Enum constants to be uppercase" is also a bit misleading for the same reason — something like "Defaults to \`true\` — set to \`false\` to preserve the case from the CRD" would land better.

The other one is the `alwaysPreserveUnknown` snippet — it doesn't compile any more. The field declaration got replaced with just the `// added to capture un-specified fields and values` comment, but the getter and setter still reference `additionalProperties`. The generator does emit the field, with `@JsonIgnore` and a HashMap initializer (see [`JObject.java#L330-L339`](https://github.com/fabric8io/kubernetes-client/blob/3cc2f3caebcf7eb6f1065e276b426df56d126658/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java#L330-L339)):

```java
@com.fasterxml.jackson.annotation.JsonIgnore
private java.util.Map<java.lang.String, java.lang.Object> additionalProperties = new java.util.HashMap<>();
```

Bringing that field back at the top of the class — alongside the getter and `setAdditionalProperty` method that are already there — should make the snippet match what the plugin actually produces.

Rest looks good!

ash-thakur-rh (approved):

manusa (approved):
LGTM, thx!

PR #7384, opened 2025-12-17
Call additionalConfig when building vertx http client with the stock VertxHttpClientFactory

### LINKED ISSUE #7252: Vert.x HttpClient should provide ability to customize WebClientOptions

capistrant:
fixes #7252 

## Description

Align vert.x with the other supported Http clients in the area of supporting calling `additionalConfig` from the associated Builder, allowing users of the client to do some customization of the base config for the client. Since the vert.x client does not strictly type the factory used by the builder like the other supported clients do, I added type check to only call additionalConfig when the factory in use is the base `VertxHttpClientFactory`. This is not ideal, but since the underlying HttpClient.Factory interface does not have `additionalConfig` in the contract, I don't think there is much we can do that would not be a breaking change, without explicitly checking the factory type. 

I had proposed #7315 as an alternative non-breaking change to this issue, but since that did not get much traction, I wanted to see if there is more appetite for this. I think giving this flexibility to clients using vert.x would be very useful for folks wanting a bit more customization capability like they get with the other backing http clients.

### vert.x 5

With vert.x 5 support being added, this PR also adds an additionalConfig call for that the same as it does for vert.x 4. There are more configuration object types that appear customizable in vert.x 5 than 4. I have added all 3 types to the noop additionalConfig in the stock factory with the assumption that folks overriding in their own factory impls will pick and choose how they want to customize things

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,14 @@
 * Fix #7174: (httpclient-vertx-5) Add runtime check for Vert.x 5 classes to provide clear error when Vert.x 4/5 conflict occurs
 * Fix #7174: (chaos-tests) Fix classpath conflict when testing with Vert.x 5 HTTP client
 * Fix #7415: (java-generator) Fix generic type erasure for array of enums with default values
+* Fix #7422: (okhttp) Remove internal API usage and fix deprecated OkHttp 5 calls
 
 #### Improvements
 * Fix #7422: bump okhttp from 4.12.0 to 5.3.2
 * Fix #7252: call additionalConfig when building VertxHttpClient with a VertxHttpClientFactory typed factory with the builder
 
 #### Dependency Upgrade
+* Fix #7374: bump snakeyaml-engine from 2.10 to 3.0.1
 
 #### New Features
 * Fix #7385: Support for Kubernetes v1.35 (Timbernetes)
--- a/pom.xml
+++ b/pom.xml
@@ -98,7 +98,7 @@
     <vertx.version>4.5.24</vertx.version>
     <vertx5.version>5.0.7</vertx5.version>
     <jandex.version>3.5.0</jandex.version>
-    <snakeyaml.version>2.10</snakeyaml.version>
+    <snakeyaml.version>3.0.1</snakeyaml.version>
     <snakeyaml.bundle.version>2.4</snakeyaml.bundle.version> <!-- Transitive Jackson -->
 
     <!-- API versions -->


httpclient-vertx/src/main/java/io/fabric8/kubernetes/client/vertx/VertxHttpClientBuilder.java
```diff
@@ -136,6 +136,13 @@ public SslContextFactory sslContextFactory() {
         }
       });
     }
+
+    // Since the factory is not explicitly typed as VertxHttpClientFactory, we need to check the type before invoking
```
shawkins: It's perfectly fine to do this instanceOf check. Quarkus is using their own factory type with this builder.

shawkins (approved):
Thank you @capistrant this looks good to me.

@manusa this should be ready to merge.

httpclient-vertx/src/main/java/io/fabric8/kubernetes/client/vertx/VertxHttpClientBuilder.java
shawkins: An alternative to make this available to other factories is to introduce an interface for this specific additionalConfig method signature.

capistrant:
@shawkins thank you for reviewing! @manusa thoughts on this approach? I should be able to do some local testing this week to run it in a k8s env and check off that box.

capistrant:
@ash-thakur-rh Hey Ashish! curious if you've been able to take a look at the proposed approach here yet and have a sense on if there's a path forward for this implementation strategy or I should re-assess. Thanks!

ash-thakur-rh:
Hi @capistrant, changes looks good to me. Lets wait for CI checks complete successfully.

However, we recently included vertx5 client module in the project, which also require these changes regarding client options  in vertx 5 module as well to make sure that we have same facade available with both vertx 4 and vertx 5 clients.

@manusa, do we do it this PR or create a follow up PR for vertx 5 client?

ash-thakur-rh:
Oops 😬! The generate model is failing, I might have look into it. It shouldn't fail as there no change related to it in this PR.

manusa:
Thx for taking care of this!

> Oops 😬! The generate model is failing, I might have look into it. It shouldn't fail as there no change related to it in this PR.

Fixed by https://github.com/fabric8io/kubernetes-client/pull/7432

Will work after rebasing.

> @manusa, do we do it this PR or create a follow up PR for vertx 5 client?

It'd be great if @capistrant could add this to Vert.x 5 in the scope of this PR

capistrant:
> Thx for taking care of this!
> 
> > Oops 😬! The generate model is failing, I might have look into it. It shouldn't fail as there no change related to it in this PR.
> 
> Fixed by #7432
> 
> Will work after rebasing.
> 
> > @manusa, do we do it this PR or create a follow up PR for vertx 5 client?
> 
> It'd be great if @capistrant could add this to Vert.x 5 in the scope of this PR

thanks for the help to you both. I'm checking on the vert.x 5 stuff today

capistrant:
shoot I forgot to run spotless after those last changes 🤦

PR #7390, opened 2025-12-21
feat: support @JsonClassDescription for CRD schema descriptions (#7375)

### LINKED ISSUE #7375: CRD Generator : Adding a top level description for a CRD
When i use kubectl explain, on a custom resource made by the CRD Generator, i always  get:
```
$ kubectl explain Applications
GROUP:      core.idp
KIND:       Application
VERSION:    v1alpha1

DESCRIPTION:
    <empty>
```
This top level description should be here in the CRD yaml generated.
```
    schema:
      openAPIV3Schema:
        description: Here's the custom description
        properties:
```
I try to use a JsonClassDescription annotation, but without success.

```
@JsonClassDescription("Application represents an application managed in the organization")
public class Application extends CustomResource<ApplicationSpec, ApplicationStatus>
```

taole33:
## Description

This PR adds support for the Jackson `@JsonClassDescription` annotation in the v2 CRD generator, addressing #7375.

Previously only property-level descriptions via `@JsonPropertyDescription` were supported. With this change, `@JsonClassDescription` on a class is emitted as the `description` of the corresponding object type in the generated OpenAPI v3 schema. It applies at any level (the root custom resource, `spec`/`status`, and any nested object type), not just the top-level object.

Behavior:

* The class-level description is resolved through the class hierarchy, consistent with the other class-level annotations (`PreserveUnknownFields`, `SchemaSwap`): a description declared on a superclass applies to subclasses that do not declare their own, and the most-derived declaration takes precedence.
* When a field has both a `@JsonPropertyDescription` and a type annotated with `@JsonClassDescription`, the field-level description takes precedence.

### Changes

* `crd-generator-api-v2`: `AbstractJsonSchema` resolves `@JsonClassDescription` through the superclass hierarchy and sets it as the object schema description. The property description is only applied when present, so a class-level description is not overwritten by an absent property description.
* Added an approval-test fixture (`Described` / `DescribedSpec`) and golden CRD covering the top-level, spec, nested, deeply-nested, inherited, and field-overrides-class cases.
* Documented the annotation in `doc/CRD-generator.md` (including the features cheatsheet) and added a CHANGELOG entry.

The v1 generator is intentionally left unchanged.

## Type of change

* [x] Feature (non-breaking change which adds functionality)

## Checklist

* [x] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
* [x] I Added [CHANGELOG](https://github.com/fabric8io/kubernetes-client/blob/main/CHANGELOG.md) entry regarding this change
* [x] I have implemented unit tests to cover my changes
* [ ] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/main/doc/CHEATSHEET.md) accordingly
* [x] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
* [x] I tested my code in Kubernetes
* [ ] I tested my code in OpenShift

Fixes #7375


shawkins (commented):
Thank you for the PR, it looks very good.

Just wanted to add that it's not required to make changes to the older implementation. You can just add a v2 generator specific test.

taole33:
@shawkins 
I'll revert the changes in the v1 implementation to keep the legacy code base untouched and will update the PR to focus solely on the v2 generator and its specific tests. Thanks for the guidance!

taole33:
@shawkins 
I have updated the PR to focus the @JsonClassDescription support exclusively on the v2 generator. The v1 logic has been reverted and restored to its original state.

taole33:
@shawkins 
I have pushed the following updates to address the recent CI errors:

* License Header Fix: Applied `mvn -N license:format` to ensure all new files comply with the project's license requirements.

* Regarding the "Kubernetes Model Generator Build" failure, I have validated the same CI-test in my [repository](https://github.com/taole33/kubernetes-client/actions) and confirmed it passes successfully.

All CI checks should be green now. Thanks for waiting.

shawkins (commented):
This looks great. The only nit would be that the description annotation is used at any level, not just the top-level object. 

Could the changelog and a doc entry in https://github.com/fabric8io/kubernetes-client/blob/main/doc/CRD-generator.md#features-cheatsheet be updated to reflect this?

taole33:
@shawkins 
> Could the changelog and a doc entry in https://github.com/fabric8io/kubernetes-client/blob/main/doc/CRD-generator.md#features-cheatsheet be updated to reflect this?

Sure, I’d be happy to update the changelog and documentation. 
I also noticed the CI failure, so I will investigate the cause.

taole33:
@shawkins Thank you for the feedback.

I have updated the documentation and the features cheatsheet in CRD-generator.md.

While updating the docs, I discovered and fixed a logic issue in AbstractJsonSchema where a class-level description was being overwritten by a null property description when @JsonPropertyDescription was absent. This fix ensures that @JsonClassDescription is correctly preserved for all object types, including nested ones.

I have also synchronized this branch with the latest upstream changes. Please let me know if you have any further comments.

shawkins (approved):
Thank you for the latest changes.

> While updating the docs, I discovered and fixed a logic issue in AbstractJsonSchema where a class-level description was being overwritten by a null property description when @JsonPropertyDescription was absent.

The logic hadn't needed to consider this case before. The update you made is appropriate and effectively makes the class description the default for the field description. I think this makes sense, but if any of the other reviewers feel otherwise we'll revisit it.

taole33:
Thanks for the approval, @shawkins!

Regarding the CI failure in the 'Build' job, it seems unrelated to these changes (appears to be a flaky test).

I have analyzed both the Java 11 and Java 21 CI logs. The failure in ConfigRefreshTest occurs within the kubernetes-client core module, which is upstream of the crd-generator module I modified. Since the build failed before even reaching my changes, it appears to be a pre-existing flaky issue in the CI environment.

Could someone from the team please re-run the failed jobs? Thank you!

metacosm (approved):

csviri (approved):

manusa (approved):
LGTM, thx!

PR #7394, opened 2026-01-07
fix(port-forwarding): Service Port-Forwarding Ignores Service Port Mapping

### LINKED ISSUE #7372: Service Port-Forwarding Ignores Service Port Mapping
### Describe the bug

**Summary:**
Service-based port forwarding fails when the service port differs from the target port, because Fabric8 attempts to connect to the service port inside the pod's network namespace instead of the target port specified in the service definition.

**Expected Behavior:**
When using `service.portForward(servicePort, localPort)`, Fabric8 should connect to the service's `targetPort` inside the pod, similar to how `kubectl port-forward service/name` works.

**Actual Behavior:**
Fabric8 attempts to connect to the `servicePort` inside the pod's network namespace, ignoring the service's port mapping to `targetPort`.

**Service Configuration:**
```yaml
apiVersion: v1
kind: Service
metadata:
  name: my-webapp
  namespace: my-namespace
spec:
  ports:
  - name: http
    port: 80          # Service port
    targetPort: 8080  # Pod port where application actually listens
  selector:
    app: my-webapp
```

**Reproduction Code:**
```java
// This fails - tries to connect to pod's port 80 instead of 8080
LocalPortForward servicePortForward = client.services()
    .inNamespace("my-namespace")
    .withName("my-webapp") 
    .portForward(80, 8080);
```

**Comparison with kubectl:**
```bash
# This works correctly - kubectl properly maps service port 80 to target port 8080
kubectl port-forward service/my-webapp 8080:80 -n my-namespace
```

**Root Cause:**
The service port-forwarding implementation appears to use the service port (80) when establishing the connection inside the pod's network namespace, but should use the service's `targetPort` (8080) instead.

**Workaround:**
Use pod-based port forwarding instead of service-based port forwarding:
```java
Pod targetPod = client.pods()
    .inNamespace("my-namespace")
    .withLabel("app", "my-webapp")
    .list().getItems().get(0);
    
LocalPortForward portForward = client.pods()
    .inNamespace("my-namespace") 
    .withName(targetPod.getMetadata().getName())
    .portForward(8080, 8080);
```

7.4.0

### Steps to reproduce

**Service Configuration:**
```yaml
apiVersion: v1
kind: Service
metadata:
  name: my-webapp
  namespace: my-namespace
spec:
  ports:
  - name: http
    port: 80          # Service port
    targetPort: 8080  # Pod port where application actually listens
  selector:
    app: my-webapp
```

**Reproduction Code:**
```java
// This fails - tries to connect to pod's port 80 instead of 8080, but does not throw any exceptions and reports that the portForward is alive
LocalPortForward servicePortForward = client.services()
    .inNamespace("my-namespace")
    .withName("my-webapp") 
    .portForward(80, "127.0.0.1", 8080);
```

### Expected behavior

When using `service.portForward(servicePort, localPort)`, Fabric8 should connect to the service's `targetPort` inside the pod, similar to how `kubectl port-forward service/name` works.

### Runtime

### Environment

macOS

```shell
io.fabric8.kubernetes.client.KubernetesClientException: Received an error from the remote socket 
error forwarding port 80 to pod [pod-id], uid : failed to execute portforward in network namespace 
"/var/run/netns/cni-[namespace-id]": failed to connect to localhost:80 inside namespace "[pod-id]", 
IPv4: dial tcp4 127.0.0.1:80: connect: connection refused 
IPv6 dial tcp6 [::1]:80: connect: connection refused
```

### Additional context

**Environment:**
- Java version: Corretto-21.0.8.9.1
- Kubernetes version: v1.34.1+k3s1

ash-thakur-rh:
## Description
Fixes #7372 

## Type of change

## Checklist

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/core/v1/ServiceOperationsImpl.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/core/v1/ServiceOperationsImpl.java
@@ -43,8 +43,8 @@ import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.ServiceLoader;
 import java.util.Objects;
+import java.util.ServiceLoader;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
@@ -140,13 +140,13 @@ public class ServiceOperationsImpl extends HasMetadataOperation<Service, Service
   }
 
   @Override
+  public LocalPortForward portForward(int port) {
+    return portForward(port, 0);
+  }
+
+  @Override
   public LocalPortForward portForward(int port, int localPort) {
-    Service s = requireFromServer();
-    Pod m = matchingPod(s);
-    return new PodOperationsImpl(context.getClient())
-        .inNamespace(m.getMetadata().getNamespace())
-        .withName(m.getMetadata().getName())
-        .portForward(getTargetPortAsInt(s, port), localPort);
+    return portForward(port, null, localPort);
   }
 
   @Override
@@ -166,16 +166,6 @@ public class ServiceOperationsImpl extends HasMetadataOperation<Service, Service
         .filter(Objects::nonNull)
         .findFirst()
         .orElse(port);
-  }
-
-  @Override
-  public LocalPortForward portForward(int port) {
-    Service s = requireFromServer();
-    Pod m = matchingPod(s);
-    return new PodOperationsImpl(context.getClient())
-        .inNamespace(m.getMetadata().getNamespace())
-        .withName(m.getMetadata().getName())
-        .portForward(getTargetPortAsInt(s, port));
   }
 
   public class ServiceToUrlSortComparator implements Comparator<ServiceToURLProvider> {


ash-thakur-rh:
WIP! Need to add more tests and add more robust assertions.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/core/v1/ServiceOperationsImpl.java
```diff
@@ -131,38 +130,50 @@ private Pod matchingPod() {
...
-        .portForward(port, localInetAddress, localPort);
+        .portForward(getTargetPortAsInt(s, port), localInetAddress, localPort);
+  }
+
+  private int getTargetPortAsInt(Service service, int port) {
+    return service.getSpec().getPorts().stream()
+        .filter(p -> p.getPort().equals(port))
+        .map(p -> p.getTargetPort().getIntVal())
+        .findFirst()
+        .orElse(port);
```
manusa: Consider adding a null filter to handle named target ports gracefully:

```java
private int getTargetPortAsInt(Service service, int port) {
  return service.getSpec().getPorts().stream()
      .filter(p -> p.getPort().equals(port))
      .map(p -> p.getTargetPort().getIntVal())
      .filter(Objects::nonNull)  // Handle named ports gracefully
      .findFirst()
      .orElse(port);
}
```

If `targetPort` is a named port (String like `"http"`), `getIntVal()` returns `null`. Without this filter, the code would throw a `NullPointerException` when unboxing to `int`.

With this change, named ports would fall back to the original port (same as "port not found"), which is consistent and avoids the NPE. Full support for named ports could be addressed in a separate enhancement.

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/core/v1/ServiceOperationsImpl.java
```diff
@@ -131,38 +130,50 @@ private Pod matchingPod() {
...
+        .portForward(getTargetPortAsInt(s, port), in, out);
   }
 
   @Override
   public LocalPortForward portForward(int port, int localPort) {
-    Pod m = matchingPod();
+    Service s = requireFromServer();
+    Pod m = matchingPod(s);
     return new PodOperationsImpl(context.getClient())
         .inNamespace(m.getMetadata().getNamespace())
```
manusa: Consider chaining the `LocalPortForward` overloads like `PodOperationsImpl` does, to reduce code duplication:

```java
@Override
public LocalPortForward portForward(int port) {
  return portForward(port, 0);
}

@Override
public LocalPortForward portForward(int port, int localPort) {
  return portForward(port, null, localPort);
}

@Override
public LocalPortForward portForward(int port, InetAddress localInetAddress, int localPort) {
  Service s = requireFromServer();
  Pod m = matchingPod(s);
  return new PodOperationsImpl(context.getClient())
      .inNamespace(m.getMetadata().getNamespace())
      .withName(m.getMetadata().getName())
      .portForward(getTargetPortAsInt(s, port), localInetAddress, localPort);
}
```

This way only the most complete overload contains the actual logic, and the simpler ones just delegate. The `PortForward portForward(int port, ReadableByteChannel in, WritableByteChannel out)` method would still need its own implementation since it returns a different type.

PR #7445, opened 2026-02-19
Add javadoc cross-linking for Fabric8 modules and external dependencies

### LINKED ISSUE #1105: Nit: pom.xml should have <links> property
It sure would be helpful for us users of javadoc.io if the `pom.xml` had a `<links>` property in it.  Like this:

    <links>https://static.javadoc.io/io.fabric8/kubernetes-model/2.1.1</links>

(More on the `links` property here: https://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html#links.  The fact that the **User Property** is `links` is what allows you to simply set it in a `<properties>` stanza in your `pom.xml`.)

Then, to resolve `static.javadoc.io` links (as above), you need to make sure that the [`maven-javadoc-plugin` has a user agent set of some kind](https://stackoverflow.com/questions/33866209/linking-to-javadoc-io-using-javadoc-link-option/47891403#47891403).

With this relatively simple change, the Javadocs for `kubernetes-client` hosted automatically at javadoc.io would be a _lot_ more readable.  Thanks!

manusa:
## Summary

Adds `<offlineLinks>` configuration to the `maven-javadoc-plugin` in both the `release` and `javadoc-test` profiles so that generated javadocs contain clickable hyperlinks to referenced types from other Fabric8 modules and external dependencies via javadoc.io.

All links use **offline mode** (`-linkoffline`) with static `element-list` files checked into `doc/javadoc-links/`. This avoids any network fetches during javadoc generation, which is critical because:
- Fabric8 SNAPSHOT versions don't exist on javadoc.io
- CI runners (GitHub Actions) cannot reliably reach javadoc.io

### Cross-linked modules
- **Fabric8 core**: `kubernetes-model-core`, `kubernetes-model-common`, `kubernetes-client-api`, `kubernetes-client`, `openshift-client-api`, `openshift-client`
- **External deps**: `jackson-databind`, `jackson-core`, `slf4j-api`

### Generated link examples
- `KubernetesResource` → `javadoc.io/doc/io.fabric8/kubernetes-model-core/...`
- `ObjectMapper` → `javadoc.io/doc/com.fasterxml.jackson.core/jackson-databind/...`
- `TypeReference` → `javadoc.io/doc/com.fasterxml.jackson.core/jackson-core/...`
- JDK types → `docs.oracle.com/en/java/javase/11/docs/api/...` (unchanged, via `detectJavaApiLink`)

### Files changed
- **`pom.xml`** — `<offlineLinks>` and `<detectOfflineLinks>false</detectOfflineLinks>` in both profiles, plus `<javadoc.links>` property and user-agent header
- **`doc/javadoc-links/*/element-list`** — Static package lists for each linked artifact (9 files)
- **`Makefile`** — `make generate-javadoc-links` target to regenerate element-list files from source directories
- **`CLAUDE.md`** — Documents the new make target
- **`CHANGELOG.md`** — Entry under Improvements

### Notes
- `jackson-annotations` is excluded because its Maven artifact version (`2.20`) differs from `${jackson.version}` (`2.20.0`)
- Only `element-list` files are used (not `package-list`) — having both in the same directory causes JDK 11's javadoc tool to silently fail link resolution
- `failOnError` is left at the default (`true`) so real javadoc errors still break the build
- When external dependency versions are bumped, regenerate their element-list files by extracting from the new javadoc JARs

Fixes #1105

## Test plan
- [x] Build javadoc for `kubernetes-client-api` with `-Pjavadoc-test` — build succeeds with no network errors
- [x] Verify Fabric8 cross-module links (KubernetesResource, UnmatchedFieldTypeModule) resolve to javadoc.io
- [x] Verify Jackson links (ObjectMapper, TypeReference) resolve to javadoc.io
- [x] Verify JDK links (Object, String, InputStream) still resolve to docs.oracle.com
- [x] CI pipeline passes (no `Error fetching URL` failures)


manusa:
@ash-thakur-rh I'm not sure if this is something we want to add or not.

It adds more complexity to the build but it also improves the help discoverability. These days this is tantamount since many are starting to rely on coding agents which require context discovery.

I'll leave this here waiting for feedback.

ash-thakur-rh:
Hi @manusa , I think this change is good to have but we need to check if this works as expected or not. I will check it out and provide feedback.

manusa:
> Hi @manusa , I think this change is good to have but we need to check if this works as expected or not. I will check it out and provide feedback.

Awesome, thanks for taking care of it. Let's wait for your feedback.

ash-thakur-rh (approved):
LGTM! Links for referenced types are now working and redirecting to https://javadoc.io/, this was not the case with existing java docs. This looks to be a good update to the docs.

Issue #7391, opened 2025-12-24
[TEST] Replace Thread.sleep with Awaitility or deterministic waits to prevent flaky tests


manusa:
## Description

Many tests in the codebase are using `Thread.sleep()` with arbitrary wait times. This is a common anti-pattern that leads to:

1. **Flaky tests** - Tests may fail intermittently on slower machines or under heavy load
2. **Slow test execution** - Fixed sleep durations often wait longer than necessary
3. **Non-deterministic behavior** - Race conditions may still occur if the sleep duration is insufficient

The project already has Awaitility as a dependency in multiple modules, making it straightforward to refactor these tests.

## Affected Files

The following test files contain `Thread.sleep()` calls that should be reviewed and replaced:

| File | Module | Usage Context |
|------|--------|---------------|
| `LeaderElectorTest.java` | kubernetes-client-api | Simulating renew deadline timeout (lines 73, 328, 330) |
| `SerialExecutorTest.java` | kubernetes-client | Testing executor behavior with delays (lines 41, 68) |
| `AbstractWatchManagerTest.java` | kubernetes-client | Waiting for reconnect scheduling (line 173) |
| `UtilsTest.java` | kubernetes-client-api | Testing thread interruption handling (line 340) |
| `WatchIT.java` | kubernetes-itests | Integration test with ping timeout (line 124) |

## Expected Behavior

Tests should use one of the following approaches instead of `Thread.sleep()`:

1. **Awaitility** (preferred for most cases):
   ```java
   await().atMost(Duration.ofSeconds(5))
          .untilAsserted(() -> assertEquals(expected, actual));
   ```

2. **CountDownLatch/CompletableFuture** for synchronization points:
   ```java
   CountDownLatch latch = new CountDownLatch(1);
   // ... trigger async operation that calls latch.countDown()
   assertTrue(latch.await(5, TimeUnit.SECONDS));
   ```

## Acceptance Criteria

- [ ] All `Thread.sleep()` calls in test files are reviewed
- [ ] Each usage is replaced with an appropriate deterministic alternative
- [ ] Tests remain stable and pass consistently
- [ ] No new flaky test failures are introduced
- [ ] Test execution time is maintained or improved

## Additional Information

Awaitility is already available as a dependency in the following modules:
- `kubernetes-client-api`
- `kubernetes-client`
- `kubernetes-tests`
- `kubernetes-itests`
- `httpclient-*` modules
- And others

## Tests

[...]

PR #7457, opened 2026-03-02
Changed the Thread.sleep in runShouldAbortRenewDeadlineExpired

adithyanms-debug:
I have only changed the Thread.sleep() in runShouldAbortAfterRenewDeadlineExpired method. Since i am new to open source i don't know if it is correct or not. I would really appreciate your guidence


manusa:
Hi @adithyanms-debug

Thanks a lot for taking the time to work on this. After looking again at #7391, I think the issue itself needs to be reframed — the goal isn't really a 1:1 replacement of Thread.sleep with Awaitility/latches, but rather rethinking each test around the actual invariant it's trying to verify, and in some cases adjusting the production code so tests can hook into deterministic signals instead of waits.

One piece of feedback while I'm here, since you asked for guidance: in `runShouldAbortAfterRenewDeadlineExpired` the original `Thread.sleep(renewDeadlineMillis * 2)` is deliberate — the test needs the mocked `update` call to take longer than the renew deadline so the deadline-expiry path fires. Replacing it with `latch.await(5, TimeUnit.SECONDS)` where nothing ever counts the latch down just blocks for the full 5s timeout, so it's effectively a longer fixed sleep with a `CountDownLatch` wrapped around it rather than a deterministic gate.

Given the rescope, I'm going to close this PR. I appreciate the effort and hope you'll consider contributing to other issues in the project.

PR #7465, opened 2026-03-04
fix(ci): deploy BOMs separately in snapshot workflow

manusa:
## Summary

The `central-publishing-maven-plugin` deploys each module individually for SNAPSHOTs (unlike releases, which are bundled into a single ZIP upload). When BOMs deploy in the same reactor invocation as the modules they reference, they may fetch `maven-metadata.xml` for recently deployed artifacts and hit a brief server-side inconsistency window between the metadata file and its checksum, causing checksum validation failures like:

```
Checksum validation failed, expected '...' (REMOTE_EXTERNAL) but is actually '...'
```

This PR splits the snapshot deploy step into two separate Maven invocations:
1. Deploy all modules except BOMs
2. Deploy BOMs (after metadata has settled on the server)

Also updates the `bom-generator-plugin` README with the correct workflow steps and documents the SNAPSHOT deployment caveat.

### DIFF
--- a/.github/workflows/release-snapshots.yaml
+++ b/.github/workflows/release-snapshots.yaml
@@ -60,5 +60,12 @@ jobs:
         run: ./mvnw ${MAVEN_ARGS} clean install -pl bom-generator-plugin
       - name: Generate BOM
         run: ./mvnw ${MAVEN_ARGS} -Pbom clean validate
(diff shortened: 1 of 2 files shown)


manusa:
This didn't fix the issue. We need to revert these changes.

PR #7468, opened 2026-03-04
fix(ci): use lax checksums for snapshot deploy step

### LINKED ISSUE #7467: Revert changes from fix(ci): deploy BOMs separately in snapshot workflow
This didn't fix the issue. We need to revert these changes.

_Originally posted by @manusa in https://github.com/fabric8io/kubernetes-client/issues/7465#issuecomment-3998074214_

manusa:
## Summary

The snapshot deploy step fails with checksum validation errors like:

```
Checksum validation failed, expected '...' (REMOTE_EXTERNAL) but is actually '...'
```

Remote SNAPSHOT metadata (`maven-metadata.xml`) on the Sonatype server can have persistent
checksum inconsistencies (`.xml` vs `.sha1` file mismatch). The `-C` (`--strict-checksums`)
flag in `MAVEN_ARGS` makes these fatal. Adding `-c` (`--lax-checksums`) to the deploy step
overrides this, allowing the build to warn and continue instead of failing.

Also reverts the split deploy from #7465 (separate BOM invocation) since lax checksums
makes it unnecessary.

Fixes #7467

### DIFF
--- a/.github/workflows/release-snapshots.yaml
+++ b/.github/workflows/release-snapshots.yaml
@@ -60,12 +60,9 @@ jobs:
         run: ./mvnw ${MAVEN_ARGS} clean install -pl bom-generator-plugin
       - name: Generate BOM
         run: ./mvnw ${MAVEN_ARGS} -Pbom clean validate
-      # BOMs are deployed separately to avoid SNAPSHOT metadata checksum errors.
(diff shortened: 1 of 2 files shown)


ash-thakur-rh (approved):
LGTM!

PR #7531, opened 2026-03-06
feat(dsl): add support for server-side content negotiation

### LINKED ISSUE #5716: Add support for partial object metadata
## Description

The Kubernetes API supports HTTP requests that reply with a partial representation of an object.

See:
- https://github.com/kubernetes/website/issues/37448
- https://github.com/kubernetes/kubernetes/commit/33a3e325f754d179b25558dee116fca1c67d353a
- https://kubernetes.slack.com/?redir=%2Farchives%2FC0EG7JC6T%2Fp1666449343298609%3Fthread_ts%3D1666448393.697149%26cid%3DC0EG7JC6T%26name%3DC0EG7JC6T%26perma%3D1666449343298609

In order to achieve this purpose, the user can provide a specially crafted `accept` HTTP header for resource requests:

```
accept: application/json;as=PartialObjectMetadata;g=meta.k8s.io;v=v1
```

And for list requests too:

```
accept: application/json;as=PartialObjectMetadataList;g=meta.k8s.io;v=v1
```

AFAIU this will return the resources with just the metadata field, omitting spec and status. I'm not sure if the list can be further customized.

A DSL method `partialObjectMetadata` could be added so that we can set this in the context and then execute the requests with the modified accept header.

### LINKED ISSUE #7451: Add support for server-side content negotiation (Table and PartialObjectMetadata responses)
## Description

The Kubernetes API supports HTTP requests that return alternative representations of resources via content negotiation using the `Accept` header. Two key response formats are currently not supported by the Fabric8 Kubernetes Client DSL:

### 1. Table format (#6823)

```
Accept: application/json;as=Table;v=v1;g=meta.k8s.io
```

Returns a `meta.k8s.io/v1/Table` object with column definitions and rows — the same format `kubectl get` uses for display. This is useful for tools that need to display resources in a tabular format without downloading the full resource payloads.

### 2. PartialObjectMetadata (#5716)

```
Accept: application/json;as=PartialObjectMetadata;g=meta.k8s.io;v=v1
```

Returns resources with only the `metadata` field populated, omitting `spec` and `status`. This significantly reduces response sizes when only metadata (labels, annotations, owner references, etc.) is needed.

For list operations:
```
Accept: application/json;as=PartialObjectMetadataList;g=meta.k8s.io;v=v1
```

Both features use the **same underlying mechanism** (Accept header content negotiation) and can share the same infrastructure changes.

## Use Cases

- **Memory reduction**: A single Pod descriptor can be >50KB. When listing thousands of pods for status monitoring, PartialObjectMetadata drastically reduces memory consumption (#6823).
- **Tool/UI integration**: Table format provides server-defined columns, useful for CLI tools, dashboards, and MCP servers (e.g., [kubernetes-mcp-server](https://github.com/containers/kubernetes-mcp-server) already uses table format for list output).
- **Efficient informers**: PartialObjectMetadata informers could watch large numbers of resources while storing minimal data (complementing the existing `ReducedStateItemStore`).

### The type-safety challenge

The DSL is strongly typed through generics `<T, L, R>`:

```java
client.pods()                    // → MixedOperation<Pod, PodList, Resource<Pod>>
  .inNamespace("default")        // → NonNamespaceOperation<Pod, PodList, Resource<Pod>>
  .withLabel("app", "web")       // → FilterWatchListDeletable<Pod, PodList, Resource<Pod>>
  .list()                        // → PodList
```

The `list()` return type is baked in as `L` (= `PodList`). There is no way to change the return type to `Table` or `PartialObjectMetadataList` mid-chain without breaking the generic contract.

### No Accept header support

Currently, the HTTP requests built by `OperationSupport.handleResponse()` and `BaseOperation.submitList()` do not set any `Accept` header, defaulting to the server's standard JSON response.

### Missing model classes

`Table`, `TableColumnDefinition`, `TableRow`, `PartialObjectMetadata`, and `PartialObjectMetadataList` do not exist as model classes in the project.

### Approach A: Type-switching method on the DSL chain

Add methods that change the return type of subsequent operations:

```java
client.pods().inNamespace("default")
    .asPartialObjectMetadata()   // returns a differently-typed operation
    .list()                      // returns PartialObjectMetadataList

client.pods().inNamespace("default")
    .asTable()
    .list()                      // returns Table
```

**How it works**: `asPartialObjectMetadata()` returns a new `Listable<PartialObjectMetadataList>` (NOT the full `FilterWatchListDeletable`). Similarly, `asTable()` returns `Listable<Table>`.

**Pros**: Clean fluent API, type-safe chain.
**Cons**: Complex generic type changes. `asPartialObjectMetadata()` on `FilterWatchListDeletable<Pod, PodList, Resource<Pod>>` must return a different generic type, breaking the chain's type parameter flow. Requires new wrapper classes/interfaces to represent the "type-switched" state.

### Approach B: Overloaded list() with response type parameter

```java
client.pods().inNamespace("default")
    .withLabel("app", "web")
    .list(Table.class)               // returns Table

client.pods().inNamespace("default")
    .list(PartialObjectMetadataList.class)  // returns PartialObjectMetadataList
```

Add a new overload to `Listable`:
```java
public interface Listable<T> {
    T list();
    <R> R list(Class<R> responseType);  // new
}
```

**Pros**: Minimal interface changes, simple to implement.
**Cons**: Less type-safe. The relationship between the class and the accept header is implicit. Doesn't naturally extend to `get()`. Users could pass arbitrary classes.

### Approach C: New terminal methods on existing interfaces (Recommended)

Add new methods alongside existing `list()` and `resources()`:

```java
// New interface
public interface MetadataListable {
    Table listAsTable();
    Table listAsTable(ListOptions options);
    PartialObjectMetadataList listAsPartialObjectMetadata();
    PartialObjectMetadataList listAsPartialObjectMetadata(ListOptions options);
}
```

Mix into existing hierarchy:
```java
public interface FilterWatchListDeletable<T, L, R>
    extends Filterable<...>, Listable<L>, WatchAndWaitable<T>,
    DeletableWithOptions, Informable<T>,
    MetadataListable {  // NEW
}
```

Usage:
```java
Table table = client.pods().inNamespace("default")
    .withLabel("app", "web")
    .listAsTable();

PartialObjectMetadataList metadata = client.pods().inNamespace("default")
    .listAsPartialObjectMetadata();
```

For single-resource get (on `Resource`):
```java
PartialObjectMetadata meta = client.pods().inNamespace("default")
    .withName("mypod")
    .getAsPartialObjectMetadata();
```

**Pros**:
- No generic type changes — return types are fixed (`Table`, `PartialObjectMetadataList`)
- Follows existing patterns — similar to how `resources()` (`Stream<R>`) was added alongside `list()` (`L`) in `FilterWatchListDeletable`
- Incrementally adoptable — ship PartialObjectMetadata first, Table later
- No breaking changes to existing APIs
- Clear and explicit method names

**Cons**:
- Adds methods to an already broad interface surface
- Each new response format requires new methods

#### 1.1 Model classes

Add to `kubernetes-model-generator` (or `kubernetes-client-api` as simple POJOs):

**Table** (meta.k8s.io/v1):
```java
public class Table implements KubernetesResource {
    private String apiVersion;  // meta.k8s.io/v1
    private String kind;        // Table
    private ListMeta metadata;
    private List<TableColumnDefinition> columnDefinitions;
    private List<TableRow> rows;
}

public class TableColumnDefinition {
    private String name;
    private String type;
    private String format;
    private String description;
    private int priority;
}

public class TableRow {
    private List<Object> cells;
    private RawExtension object;  // optional partial object
    // conditions omitted for brevity
}
```

**PartialObjectMetadata** (meta.k8s.io/v1):
```java
public class PartialObjectMetadata implements HasMetadata {
    private String apiVersion;
    private String kind;
    private ObjectMeta metadata;
    // No spec, no status - that's the point
}

public class PartialObjectMetadataList extends DefaultKubernetesResourceList<PartialObjectMetadata> {
}
```

#### 1.2 Accept header constants

```java
public static final String ACCEPT_TABLE_V1 =
    "application/json;as=Table;v=v1;g=meta.k8s.io, " +
    "application/json;as=Table;v=v1beta1;g=meta.k8s.io, " +
    "application/json";

public static final String ACCEPT_PARTIAL_METADATA_V1 =
    "application/json;as=PartialObjectMetadata;v=v1;g=meta.k8s.io";

public static final String ACCEPT_PARTIAL_METADATA_LIST_V1 =
    "application/json;as=PartialObjectMetadataList;v=v1;g=meta.k8s.io";
```

#### 2.1 New interfaces in `kubernetes-client-api`

```java
// For list operations
public interface MetadataListable {
    Table listAsTable();
    Table listAsTable(ListOptions options);
    PartialObjectMetadataList listAsPartialObjectMetadata();
    PartialObjectMetadataList listAsPartialObjectMetadata(ListOptions options);
}

// For single-resource get operations
public interface MetadataGettable {
    PartialObjectMetadata getAsPartialObjectMetadata();
}
```

#### 2.2 Integration points

- `FilterWatchListDeletable` extends `MetadataListable` — available after namespace/filter operations
- `Resource` extends `MetadataGettable` — available for single-resource operations

#### 3.1 New internal method

```java
// In BaseOperation.java
protected <M> CompletableFuture<M> submitListAs(
    ListOptions listOptions, TypeReference<M> typeRef, String acceptHeader) {
    URL fetchListUrl = fetchListUrl(getNamespacedUrl(),
        defaultListOptions(listOptions, null));
    HttpRequest.Builder requestBuilder = withRequestTimeout(
        httpClient.newHttpRequestBuilder()
            .url(fetchListUrl)
            .setHeader("Accept", acceptHeader));
    return handleResponse(httpClient, requestBuilder, typeRef);
}
```

This mirrors the existing `submitList()` (line 428) but:
- Accepts a custom `TypeReference` instead of using the hardcoded `listTypeReference`
- Sets the `Accept` header on the request builder

#### 3.2 Public implementations

```java
@Override
public Table listAsTable() {
    return listAsTable(new ListOptions());
}

@Override
public Table listAsTable(ListOptions listOptions) {
    try {
        return waitForResult(submitListAs(listOptions,
            new TypeReference<Table>() {},
            ACCEPT_TABLE_V1));
    } catch (IOException e) {
        throw KubernetesClientException.launderThrowable(forOperationType("list"), e);
    }
}

@Override
public PartialObjectMetadataList listAsPartialObjectMetadata() {
    return listAsPartialObjectMetadata(new ListOptions());
}

@Override
public PartialObjectMetadataList listAsPartialObjectMetadata(ListOptions listOptions) {
    try {
        return waitForResult(submitListAs(listOptions,
            new TypeReference<PartialObjectMetadataList>() {},
            ACCEPT_PARTIAL_METADATA_LIST_V1));
    } catch (IOException e) {
        throw KubernetesClientException.launderThrowable(forOperationType("list"), e);
    }
}
```

### Phase 4 (Future): Informer support

The `ListerWatcher<T, L>` interface currently ties list responses to `L`. PartialObjectMetadata informers would:
- Use `ACCEPT_PARTIAL_METADATA_LIST_V1` for the initial list
- Use `ACCEPT_PARTIAL_METADATA_V1` for watch events
- Complement the existing `ReducedStateItemStore` which already supports reduced-state informers on the storage side

This could be exposed as:
```java
client.pods().informAsPartialObjectMetadata(handler, resyncPeriod);
```

## Key Files Reference

| File | Role |
|------|------|
| `kubernetes-client-api/.../dsl/Listable.java` | `list()` returns `L` |
| `kubernetes-client-api/.../dsl/Gettable.java` | `get()` returns `T` |
| `kubernetes-client-api/.../dsl/FilterWatchListDeletable.java` | Combines Filterable + Listable + Watchable; has `resources()` |
| `kubernetes-client-api/.../dsl/Resource.java` | Single-resource operations; extends `FromServerGettable<T>` |
| `kubernetes-client/.../internal/OperationContext.java` | Carries all operation state |
| `kubernetes-client/.../internal/OperationSupport.java` | HTTP request/response handling |
| `kubernetes-client/.../internal/BaseOperation.java:428-446` | `submitList()` — builds HTTP request, no Accept header |
| `kubernetes-client/.../internal/BaseOperation.java:791-795` | `handleGet()` — single-resource GET |
| `kubernetes-client/.../internal/BaseOperation.java:1114-1116` | `resources()` — precedent for terminal methods |
| `kubernetes-client/.../informers/cache/ReducedStateItemStore.java` | Existing reduced-state informer support |

## Related Issues

- #6823 — Receiving resources as Tables
- #5716 — Add support for partial object metadata

Both issues are addressed by this proposal since they share the same underlying mechanism (Accept header content negotiation).

ash-thakur-rh:
## Description
Adds support for two Kubernetes server-side content negotiation formats via the Accept header, allowing clients to receive lightweight or display-optimized representations of resources without changing the API endpoint. Both 
features work across all resource types: built-in resources, CRDs, and custom resources.                                                                                                                                        

Fixes #7451 
Fixes #5716 

---
### PartialObjectMetadata

Fetches only the metadata of a resource (no spec, no status). The server strips the heavy fields before sending, significantly reducing payload size. Useful in controllers and operators that need to watch or list many resources but only care about labels, annotations, UIDs, or resource versions.

**Accept header:** `application/json;as=PartialObjectMetadata;g=meta.k8s.io;v=v1`

**New model classes (kubernetes-model-core):**
- `PartialObjectMetadata`: implements HasMetadata, contains apiVersion, kind, metadata
- `PartialObjectMetadataList`: extends DefaultKubernetesResourceList\<PartialObjectMetadata\>

**New DSL interfaces (kubernetes-client-api):**
- `MetadataGettable`: getAsPartialObjectMetadata() → returns null on 404, consistent with get()
- `MetadataListable`: listAsPartialObjectMetadata() and listAsPartialObjectMetadata(ListOptions)

**Wired into:**
- `Resource<T>` extends MetadataGettable
- `FilterWatchListDeletable<T,L,R>` extends MetadataListable

---

### Table

Fetches a server-rendered table: the same format used by kubectl get. The server returns columnDefinitions (headers) and rows (one per resource), with each row's cells aligned by index with columnDefinitions. Column names and count are decided by the server based on resource type; priority == 0 columns are the default view, priority > 0 are wide columns (kubectl get -o wide).

**Accept header:** `application/json;as=Table;v=v1;g=meta.k8s.io`

**New model classes (kubernetes-model-core):**
- `Table`: columnDefinitions, rows, metadata (ListMeta); annotated with @JsonIgnoreProperties(ignoreUnknown = true) to tolerate server-side field additions
- `TableColumnDefinition`: name, type, format, description, priority
- `TableRow`(cells): List\<Object\>, optional object: RawExtension; annotated with @JsonIgnoreProperties(ignoreUnknown = true) to handle conditions and other optional server fields

**New DSL interfaces (kubernetes-client-api):**
- `TableGettable`: getAsTable() → returns null on 404
- `TableListable`: listAsTable() and listAsTable(ListOptions)

**Wired into:**
- `Resource<T>` extends TableGettable
- `FilterWatchListDeletable<T,L,R>` extends TableListable

---

### Shared Implementation Details

- `BaseOperation` implements all four methods with the appropriate Accept headers
- `getAsPartialObjectMetadata()` and `getAsTable()` both catch HTTP 404 and return null, matching the behaviour of `get()`
- `ResourceAdapter` delegates `getAsPartialObjectMetadata()` and `getAsTable()` to the wrapped resource
- All methods support the full filter chain: `inNamespace()`, `inAnyNamespace()`, `withLabel()`, `withField()`, `withName()`, and `ListOptions` (pagination, label/field selectors)

---

## Type of change

## Checklist


ash-thakur-rh:
image
tested with k8s, Attaching screenshots

kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/PartialObjectMetadataTest.java
```diff
@@ -0,0 +1,297 @@
...
+
+  KubernetesMockServer server;
+  KubernetesClient client;
+
+  @Test
+  void getPartialMetadataOfPodReturnsOnlyMetadataFields() {
+    server.expect().get()
+        .withPath("/api/v1/namespaces/default/pods/my-pod")
+        .andReturn(200, podPartialMetadata("my-pod", "default", "app", "backend", "version", "v1"))
+        .once();
```
manusa, line 47: AS discussed internally.
The HTTP mock should return different objects depending on the header provided. IF the header matches the one for partial metadata, the server should respond only with metadata fields, if not it should return everything.

(You can create this with a proper function in the expectation that checks the headers -I think-)

kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/PartialObjectMetadataTest.java
```diff
@@ -0,0 +1,297 @@
...
+        .inNamespace("default")
+        .withName("my-pod")
+        .getAsPartialObjectMetadata();
+
+    assertThat(result).isNotNull();
+    assertThat(result.getMetadata().getName()).isEqualTo("my-pod");
+    assertThat(result.getMetadata().getNamespace()).isEqualTo("default");
+    assertThat(result.getMetadata().getLabels())
+        .containsEntry("app", "backend")
+        .containsEntry("version", "v1");
```
manusa, line 59: Once the previous comment is implemented, we should actually have an assertion to check that the other fields in the Pod are empty (spec, status)

kubernetes-tests/src/test/java/io/fabric8/kubernetes/client/mock/PartialObjectMetadataTest.java
manusa: If you switch to use KubeApiTest, you won't even need to mock the response and we can be sure that the Kube API behaves as expected

junit/kube-api-test/client-inject/src/test/java/io/fabric8/kubeapitest/junit/sample/PartialObjectMetadataKubeApiIT.java
manusa, line 1: The test is good, but I'm not sure this is the right place.
Ideally this should be in a different module. Tests here are supposed to be testing kubeapitest functionality and not kubernetes client implementations.

manusa:
The last commit you created is plainly wrong. It removed the Kube API tests and used mocks instead. That's not ok.

@ash-thakur-rh

ash-thakur-rh:
> The last commit you created is plainly wrong. It removed the Kube API tests and used mocks instead. That's not ok.
> 
> @ash-thakur-rh

will take care of it, but @manusa, kubernetes-itests or any modules do not use kube-api-test(which has enable kube sever related utilities) as test dependency. Should I create new module under junit/kube-api-test or import kube-api-test as test dependencies in kubernetes-itests module?

manusa:
> > The last commit you created is plainly wrong. It removed the Kube API tests and used mocks instead. That's not ok.
> > @ash-thakur-rh
> 
> will take care of it, but @manusa, kubernetes-itests or any modules do not use kube-api-test(which has enable kube sever related utilities) as test dependency. Should I create new module under junit/kube-api-test or import kube-api-test as test dependencies in kubernetes-itests module?

Import as test dependency. itests is specifically designed for these tests AFAIR, the fact we don't have kube-api-test is that the module predates by far kube-api-test

manusa (changes requested):
I see why the `KubernetesExtension`/`NamespaceExtension` opt-out ended up here — the autodetected extensions in `kubernetes-junit-jupiter-autodetected` fire on every test class in `kubernetes-itests` and try to build their own `KubernetesClient` and ephemeral namespace from KUBECONFIG, which collides with the embedded apiserver `@EnableKubeAPIServer` brings up. And I see this is exactly what you were flagging when you asked whether to make a new module or pull `kube-api-test` into `kubernetes-itests`. The answer is the new module: the opt-out as written has `kubernetes-junit-jupiter` reach across modules to pattern-match `kube-api-test`'s annotation by FQN, which couples two extension systems that shouldn't know about each other, and we can't take that on.

Could we put the kube-api-test based ITs in their own home — say a new `kubernetes-itests-envtest` (or similar) sibling module — that doesn't depend on `kubernetes-junit-jupiter-autodetected`? `PartialObjectMetadataIT`/`TableIT` move there, no cross-extension coupling, and it becomes the natural landing spot for any future tests that run against the embedded apiserver. The two `KubernetesExtension`/`NamespaceExtension` changes come out entirely.

While we're rearranging tests: `PartialObjectMetadataTest` and `TableTest` in `kubernetes-tests` overlap heavily with the new ITs. Most of what they assert is round-tripping a `PartialObjectMetadata`/`Table` we just serialized seconds earlier — the IT against a real apiserver is a stronger statement of the same fact. The genuinely unique coverage is (a) the exact `Accept` header value the *client* sends, which the ITs can't see — they only observe what comes back — and (b) the 404 → `null` swallow. Could we collapse each class down to a handful of focused tests covering just those two things? Something like:

```java
@EnableKubernetesMockClient(https = false)
class PartialObjectMetadataAcceptHeaderTest {
  KubernetesMockServer server;
  KubernetesClient client;

  @Test void getSendsCorrectAcceptHeader()  { ... }
  @Test void listSendsCorrectAcceptHeader() { ... }
  @Test void getReturnsNullOn404()          { ... }
}
```

Same shape for `Table`. Keeps the value, sheds the tautology, and tightens the Accept-header coverage at the same time — right now only 4 of the 18 mock tests actually assert that header, so a regression on the cluster-scoped or `inAnyNamespace()` paths would slide through. Side note: `https = false` is the convention for new mock-mode tests since HTTPS pays a ~7–8s TLS cold-start per surefire fork.

One coverage gap worth filling on the IT side: the new doc headlines a `do { ... withContinue(token) ... } while (...)` pagination loop as the canonical usage, but nothing actually exercises a continue-token round-trip. One IT per side asserting that the second page's URL carries `continue=<token>` would lock that contract in.

A couple of things on the implementation itself. `getAsPartialObjectMetadata()` and `getAsTable()` in `BaseOperation` skip the "name not specified" guard that `get()` enforces via `requireFromServer()` — verified by probe that calling either without `.withName(...)` silently builds the LIST URL, fires the GET with the single-object Accept header, and returns a `PartialObjectMetadata` (or `Table`) with all-null fields rather than throwing `"name not specified for an operation requiring one."`. Normal DSL flow is gated by `Resource<T>`, but `BaseOperation` (via `ExtensibleResource extends Resource`) exposes the methods on the un-named handle to anyone casting. Worth adding the `Utils.isNullOrEmpty(getName())` check at the top — and since the four new methods share nearly identical scaffolding (URL build → request builder → Accept header → handleResponse → 404-or-throw), this is the natural moment to extract `getAs(accept, type)` / `submitListAs(opts, accept, typeRef)` helpers the issue plan called for. Kills the duplication and gives the guard a single home.

Second one: `ACCEPT_TABLE_V1` declares only `application/json;as=Table;v=v1;g=meta.k8s.io`. kubectl sends comma-separated fallbacks — typically `application/json;as=Table;v=v1;g=meta.k8s.io,application/json;as=Table;v=v1beta1;g=meta.k8s.io,application/json` — so an aggregated apiserver or CR endpoint that hasn't registered a Table converter falls back to plain JSON instead of returning 406. Without the fallbacks, those endpoints throw, and since the swallow logic only catches 404, the user sees a raw `KubernetesClientException`. Same applies to the partial-metadata constants. Worth matching kubectl's behavior — and is one more reason 406 is worth covering in tests.

Couple of small model-class things while we're here: `PartialObjectMetadata` and `PartialObjectMetadataList` are missing `@Group("meta.k8s.io")` and `@Version("v1")` (the `Table` family has them) — direct typed deserialization works either way, but polymorphic resolution via `KubernetesDeserializer` won't find these without the annotations, and it's nice to keep the family symmetric. And `TableColumnDefinition.priority` is a primitive `int` while every neighbour is a boxed wrapper with `@JsonInclude(NON_NULL)` — the primitive defeats the `NON_NULL` filter so 0 always serializes. `Integer` would line up.

@ash-thakur-rh

ash-thakur-rh:
@manusa, I have made changes as per your suggestions and tested them locally, works as expected.

manusa (approved):
LGTM, thx!

PR #7575, opened 2026-03-19
fix: updating indexes in a more consistent manner

### LINKED ISSUE #7265: Temporal resource miss on Informer index
### Describe the bug

If I add an index to the Informer, and later retrieve resource using `byIndex` sometimes not the entire set of resources are received. This is caused by the fact that when the informer updates index first it removes the old item later adds it. So when reading the just in time that event is received, it might not return the related item:

https://github.com/csviri/kubernetes-client/blob/6fe86db49817093168a3f11a5f363374a8d68c48/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/impl/cache/CacheImpl.java#L277-L279

I know that it is noted in the Javadoc that this is because it is not fully synchronized, and we probably don't want that since would affect the performance, but maybe we can find a different solution 

go client:
https://github.com/kubernetes/cri-client

7.3.1

### Steps to reproduce

Not easy to repdouce, easies is just have an event handler that reads the index.

### Expected behavior

all the items are received

### Runtime

Kubernetes (vanilla)

1.33

### Environment

Linux, macOS

```shell

```

### Additional context

_No response_

shawkins:
closes: #7265

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Fix #7543: fix processInlineDuplicateFields to recursively resolve nested inline embeds
 * Fix #7450: StandardHttpClient.shouldRetry() does not retry on Vert.x HttpClosedException
 * Fix #7350: Improper callback timing in leaderelection leads to the dual-leader
+* Fix #7265: fix ephemeral removal of index entries from informer caches
 
 #### Improvements
 * Fix #7662: (mockwebserver) new `MockWebServer#setHttp2ClearTextEnabled(boolean)` setter to opt out of HTTP/2 cleartext (h2c) upgrade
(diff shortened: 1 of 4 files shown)


csviri (commented):
Not sure if following all the aspects but seems to solve the original issue.

Maybe could you elaborate the downside (if any) compared to alternative (ReadWriteLock version).

thank you!

shawkins:
> Maybe could you elaborate the downside (if any) compared to alternative (ReadWriteLock version).

The difference in the approaches comes down to CacheImpl.put. With a ReadWriteLock, a simultaneous read while put is called will block.

Here the read will not block, and you are not guarenteed to see the new index state of the item being put (at least not until the updateIndices call is finished).

In any other circumstance, the behavior is the same in the sense that if the cache update "happens before" the index read, then you get the same results. This is because if an item already exists in an index, there's no ephemal removal. And there's a consistency check that ensures if another revision of that resource already exists in the cache, that we won't inappropriately return something as belonging to the index. On a delete, the item will no longer be in the cache, and nothing will be returned for it.

manusa:
After reviewing both approaches, I think this is the right direction. The lock-free read path is important for performance, and the optimistic consistency approach addresses the core issue from #7265 (items temporarily disappearing during same-key updates) without introducing read contention.

I've asked @Desel72 to scope down #7558 to just the concurrency reproducer test (`CacheImplConcurrencyTest`) landed as `@Disabled`. Once that's merged, this PR should re-enable the test and use it to validate the fix.

@shawkins could you bring this out of draft once we have the reproducer test to verify against?

PR #7633, opened 2026-04-16
fix(java-generator): HTML-escape field descriptions in generated Javadoc

### LINKED ISSUE #7632: Java generator generates invalid javadocs
### Describe the bug

I'm generating Java code from my CRD, using `io.fabric8:java-generator-maven-plugin`. The generated Java docs has Javadocs (❤️), however if I write something like `<foo>` within a `description` property in the CRD's schema that's used _unquoted_ in Javadoc comments in the generated Javadoc. So if I run the javadoc tool over that source code I get an `error: unknown tag: foo`, because the javadoc tool is strict about validating things that look like HTML tags are actually real HTML tags. (I guess because people often use text like <foo> as a placeholder in an explanation, which then disappears in the rendered HTML).

Obviously I could HTML-quote the `<` → `&lt;`, but that isn't really the correct fix, because JSON Schema/CRDs are not HTML documents, and (as far as I know) there's not requirement/expectation that `description` needs to be HTML quoted. The expectation is that `description` is simply plain text (and is assumed to be such by other tools, such as `kubectl explain`). 

7.6.1

### Steps to reproduce

1. Use `<foo> in a `description` for a property in the validation schema of a CRD
2. Generate the java source code using java-generator-maven-plugin
3. Use javadoc to try to generate the API documentation. 

### Expected behavior

HTML quote the `description` so that it can be used in javadoc comments without triggering errors from the javadoc tool. This would mean people could not smuggle HTML into their descriptions. But that's 

### Runtime

Kubernetes (vanilla), other (please specify in additional context)

### Environment

Linux

```shell

```

### Additional context

_No response_

ash-thakur-rh:
## Summary

- CRD `description` fields containing angle-bracket text (e.g. `<foo>`, `<Database Name>`) were embedded raw into Javadoc comments, causing `javadoc` to error with `error: unknown tag: foo`
- Escape `&` → `&amp;`, `<` → `&lt;`, `>` → `&gt;` before inserting descriptions into Javadoc (in correct order to avoid double-encoding), in addition to the existing `*/` escaping
- Updated the Keycloak CRD approval snapshot to reflect the corrected output

Fixes #7632

## Test Plan

- [ ] `testFieldDescriptionWithHtmlTagsIsEscapedInJavadoc` — verifies `<foo>` → `&lt;foo&gt;` in generated Javadoc
- [ ] `testFieldDescriptionWithAmpersandIsEscapedInJavadoc` — verifies `&` → `&amp;`
- [ ] `testFieldDescriptionWithCommentTerminatorIsEscapedInJavadoc` — verifies `*/` is still escaped (regression guard)
- [ ] All 80 tests in `java-generator/core` pass


manusa (approved):
LGTM, thx!

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
```diff
@@ -260,7 +260,11 @@ protected List<GeneratorResult.ClassResult> generateJavaFields(ClassOrInterfaceD
         }
 
         if (Utils.isNotNullOrEmpty(prop.getDescription())) {
-          objField.setJavadocComment(prop.getDescription().replace("*/", "&#042;&#047;"));
+          objField.setJavadocComment(prop.getDescription()
+              .replace("&", "&amp;")
```
andreaTP, line 264: Reading it again, should it be replaceAll?

andreaTP (commented):
Thanks for taking care of this!
The tests are meaningless as they are only testing that 'replace` works.
The changes in the approval are enough to me.

Quick question, have you attempted to use and render the mangled javadocs?

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
ash-thakur-rh: Good question! `String.replace(CharSequence, CharSequence)` already replaces **all** occurrences — it's not like `replaceFirst`. `replaceAll(String, String)` takes a regex pattern, which is unnecessary here and would actually be riskier since characters like `*` in `*/` are regex metacharacters and would need escaping.

So `replace` is the correct choice for literal string substitution.

ash-thakur-rh:
Created a follow-up issue to verify the rendered Javadoc output looks correct in the browser: #7643

The escaping is syntactically correct, but we haven't actually generated and rendered the Javadoc to visually confirm the descriptions display as intended.

java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JObject.java
andreaTP: Thanks for the double check

PR #7655, opened 2026-04-24
chore(skills): add k8s-model-update skill

ash-thakur-rh:
## Description

Adds a new Claude Code skill (`k8s-model-update`) that automates the process of updating
Fabric8 Kubernetes Client models when a new Kubernetes version is released. The skill guides
through downloading the OpenAPI spec, regenerating Java models, analyzing API changes
(graduations, deprecations, removals), updating the client DSL, and raising a PR.

It follows the same patterns as the existing `bump-go-dep` and `gh-issue-release` skills
(scoped `allowed-tools`, `disable-model-invocation`, argument wiring via `$0`/`$1`,
pre-fetched context injection). Includes a download script with version validation and
a context script that surfaces the current schema property, Java version, and issue info.

No CHANGELOG entry — this is tooling-only, no runtime code changes.

## Type of change

## Checklist


manusa (changes requested):
Hey @ash-thakur-rh, thanks a lot for putting this together! 🙌

The skill itself is in good shape and the overall structure works well. Before we move it out of draft, I'd like to align it with the two skills we already have in the repo (`bump-go-dep`, `gh-issue-release`) and a few things from the official [Skills docs](https://code.claude.com/docs/en/skills). Listing them here so we can iterate:

**Frontmatter / skill wiring**
- Add `disable-model-invocation: true` — this skill has real side effects (branch, commits, PR), so we don't want Claude auto-triggering it. Both existing skills do this.
- Scope `allowed-tools` the way the other skills do — comma-separated string, with `Bash(...)` rules (e.g. `Bash(make *)`, `Bash(gh *)`, `Bash(git *)`, `Bash(${CLAUDE_SKILL_DIR}/scripts/*)`) instead of a bare `Bash`. Drop `Agent` if the skill doesn't actually delegate, and either wire `WebSearch`/`WebFetch` into the blog-lookup step or remove them.
- Use `${CLAUDE_SKILL_DIR}` instead of the literal `<skill-dir>` placeholder when invoking the download script.
- Wire `$0` / `$1` into the body so `argument-hint` is actually used (see how `bump-go-dep` and `gh-issue-release` do it).
- Consider front-loading context via a `` !`script` `` injection block, the way the other two skills do — would let us pre-fetch the K8s release blog / current pom property up front.

**`download-k8s-schema.sh`**
- `curl -fSL` + the `HTTP_CODE -eq 200` check is contradictory under `set -euo pipefail` — pick one.
- Add a simple version regex check (`^[0-9]+\.[0-9]+\.[0-9]+$`) so we don't interpolate garbage into the output path/URL.
- Trailing newline missing on both files.

**Skill content**
- Drop the "around line 83" reference in `kubernetes-model-generator/pom.xml` — line numbers rot, just point at the `<properties>` section.
- Mention `make generate-javadoc-links` — CLAUDE.md calls it out as required when packages change, which is exactly what a K8s bump does.
- Quick note that `make format` needs Java 17 (contributors on 21+ hit spotless failures otherwise).
- Add a pointer to also check the OpenShift client side (`openshift-client-api` / `openshift-client`) for any graduated API that's re-exposed there.
- The `gh issue comment` step posts to a public issue — let's move that behind a confirmation checkpoint (or at least a preview) rather than firing before CHECKPOINT 1.

**PR hygiene**
- PR title has a typo (`feat-` → `feat(skills):`), and description/checklist are still template defaults — could you fill those in?
- No CHANGELOG entry; happy to skip it since this is tooling-only, but let's call that out explicitly in the PR body.

Happy to pair on any of these — the bones are solid, mostly a matter of bringing it in line with what's already here. 🙏

ash-thakur-rh:
Done! Took care of suggestions in your above comment @manusa

.claude/skills/k8s-model-update/SKILL.md
```diff
@@ -0,0 +1,315 @@
...
+
+```bash
+gh issue comment $1 --body "..."
+```
+
+---
+
+## Phase 3: DSL Updates
+
+### 11. Present DSL candidates
```
manusa: nit: numbering is wrong, the last documented step is 9, not 10.

.claude/skills/k8s-model-update/SKILL.md
```diff
@@ -0,0 +1,315 @@
...
+Formatting applies license headers and spotless Java formatting (requires JDK 17+). The build verifies everything compiles cleanly.
+
+### 16. Commit
+
+Stage all generated model changes, DSL additions, pom.xml update, and CHANGELOG. **Do not stage** the investigation markdown file.
+
+Commit message pattern:
+
+```
+feat(openapi): support for Kubernetes v$0 (<Release Name>) (#$1)
```
manusa: nit: The commit message should really be major.minor, `$0` as documented includes the patch version too

.claude/skills/k8s-model-update/SKILL.md
```diff
@@ -0,0 +1,315 @@
...
+
+Search for the official release blog at `https://kubernetes.io/blog/` for the target version. The blog covers major themes, graduated features, deprecations, and removals. Use it to:
+
+- Verify your diff analysis matches the highlighted changes
+- Catch behavioral changes not reflected in models
+- Find the release code name (needed for CHANGELOG entry)
+
+### 9. Write the investigation report
+
+Create a detailed markdown file at `k8s-$0-investigation.md` in the project root. This file is for local reference only — **do not commit it**.
```
manusa: Ideally we should add this to a temporary directory or a gitignored directory in the repo so that it doesn't get committed accidentally.

manusa (approved):
This looks very good to me, thx!

I've added a few nits that we might want to tackle now or igore.
Let me know if you want to fix them or if I should merge the branch already.

.claude/skills/k8s-model-update/SKILL.md
ash-thakur-rh: Fixed — steps now continue correctly as 10–16 after step 9.

.claude/skills/k8s-model-update/SKILL.md
ash-thakur-rh: Fixed — commit message and PR title now use major.minor derived from \$0 by dropping the patch version (e.g., 1.36.0 → 1.36).

.claude/skills/k8s-model-update/SKILL.md
ash-thakur-rh: Fixed — investigation report now writes to /tmp instead of the project root, so it can't be committed accidentally.

PR #7662, opened 2026-04-24
feat(mockwebserver): add opt-out for HTTP/2 cleartext (h2c) upgrade

manusa:
## Summary

Adds a `MockWebServer#setHttp2ClearTextEnabled(boolean)` opt-out setter. Default stays at `true` (matching Vert.x's own default) to preserve behaviour for external consumers of `kubernetes-server-mock`. The Kubernetes-client `AbstractAsyncBodyTest` opts out to fix a long-standing CI flake.

## Flake

`JdkHttpClientAsyncBodyTest.consumeBytesProcessesLargeBodies` on Java 11:

```
[ERROR] io.fabric8.kubernetes.client.jdkhttp.JdkHttpClientAsyncBodyTest.consumeBytesProcessesLargeBodies
java.util.concurrent.ExecutionException: java.net.ProtocolException: protocol error: zero streamId for DataFrame
	at jdk.internal.net.http.Http2Connection.protocolError(Http2Connection.java:1184)
	at jdk.internal.net.http.Http2Connection.processFrame(Http2Connection.java:826)
	at jdk.internal.net.http.frame.FramesDecoder.decode(FramesDecoder.java:155)
```

## Root cause

The JDK HttpClient defaults to HTTP/2 and, on some Java 11 updates, sends `Upgrade: h2c` on plain-HTTP requests. Vert.x's `HttpServerOptions.setHttp2ClearTextEnabled` defaults to `true`, so the mock server accepts the upgrade and switches mid-connection to HTTP/2 framing. On 1 MiB bodies this intermittently produces a DATA frame the JDK's H2 decoder rejects as malformed — a framing race in the upgrade path, either on the Netty/Vert.x send side or the JDK decoder.

Small bodies fit in one/few frames and don't expose it. Older Java 11 updates never send the `Upgrade: h2c` header at all, which is why this only bites on CI.

## Why opt-out, not change-the-default

`kubernetes-server-mock` is published externally and may be used by third-party projects that rely on h2c being accepted over cleartext. Flipping the default would be a behaviour change for those consumers. Adding an opt-out lets callers who know h2c doesn't apply to their scenario (e.g. tests that mock the real Kubernetes API, which is HTTPS-only) disable it surgically.

`AbstractAsyncBodyTest` now opts out with an inline comment describing the symptom and the reasoning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Fix #7350: Improper callback timing in leaderelection leads to the dual-leader
 
 #### Improvements
+* Fix #7662: (mockwebserver) new `MockWebServer#setHttp2ClearTextEnabled(boolean)` setter to opt out of HTTP/2 cleartext (h2c) upgrade
 * Fix #7522: improve dependency management for kubernetes-httpclient-okhttp
 * Fix #7550: add a ResourceEventHandler onList method and deprecated onNothing
 * Fix #3396: (mockwebserver) Enhance self-signed certificate generation to include Subject Alternative Names (SANs) for proper TLS verification by modern clients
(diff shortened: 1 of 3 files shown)


ash-thakur-rh (approved):
LGTM!

PR #7733, opened 2026-05-06
test: Replace Thread.sleep() with Awaitility in LeaderElectorTest.loopCancel (Issue #7391)

### LINKED ISSUE #7391: [TEST] Replace Thread.sleep with Awaitility or deterministic waits to prevent flaky tests

lukman48:
Replace flaky Thread.sleep() waits in loopCancel() test with deterministic
Awaitility polling to prevent intermittent test failures on slower machines.

Previously, the test used fixed sleep durations:
- Thread.sleep(100) to await cancellation
- Thread.sleep(100) to verify state doesn't change

Now uses:
- Awaitility.await() with 200ms timeout
- Poll every 25ms to detect when condition is met
- Eliminates race conditions from fixed-duration waits

This improves test stability and execution time on CI systems.

Fixes #7391

Signed-off-by: lukman48 <lukman_uki@yahoo.co.id>

### DIFF
--- a/kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/extended/leaderelection/LeaderElectorTest.java
+++ b/kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/extended/leaderelection/LeaderElectorTest.java
@@ -325,10 +325,10 @@ class LeaderElectorTest {
     cf.cancel(true);
 
     // make sure that the task is no longer running
-    Thread.sleep(100);
     int sample = count.get();
-    Thread.sleep(100);
-    assertEquals(sample, count.get());
+    Awaitility.await().timeout(Duration.ofMillis(200))
+        .pollInterval(Duration.ofMillis(25))
+        .untilAsserted(() -> assertEquals(sample, count.get(), "Count should not increment after cancellation"));
   }
 
   @Test


manusa:
Hi @lukman48

Thanks a lot for taking the time to work on this. After looking again at #7391, I think the issue itself needs to be reframed — the goal isn't really a 1:1 replacement of Thread.sleep with Awaitility/latches, but rather rethinking each test around the actual invariant it's trying to verify, and in some cases adjusting the production code so tests can hook into deterministic signals instead of waits.

Given that, I'm going to close this PR and revisit the issue with a clearer scope. I appreciate the effort and hope you'll consider contributing to other issues in the project.

Issue #7737, opened 2026-05-07
Flaky test PodTest.testExecEphemeralContainer


manusa:
```
Error:  Tests run: 32, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 11.80 s <<< FAILURE! -- in io.fabric8.kubernetes.client.mock.PodTest
Error:  io.fabric8.kubernetes.client.mock.PodTest.testExecEphemeralContainer -- Time elapsed: 10.06 s <<< ERROR!
io.fabric8.kubernetes.client.KubernetesClientException: An error has occurred.
	at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:109)
	at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:102)
	at io.fabric8.kubernetes.client.utils.Utils.waitUntilReadyOrFail(Utils.java:187)
	at io.fabric8.kubernetes.client.dsl.internal.core.v1.PodOperationsImpl.setupConnectionToPod(PodOperationsImpl.java:387)
	at io.fabric8.kubernetes.client.dsl.internal.core.v1.PodOperationsImpl.exec(PodOperationsImpl.java:293)
	at io.fabric8.kubernetes.client.mock.PodTest.testExecEphemeralContainer(PodTest.java:1048)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
Caused by: java.util.concurrent.TimeoutException: not ready after 10000 MILLISECONDS
	... 7 more
```

manusa:
## Investigation summary (Claude Code session, 2026-05-07)

### Root-cause hypotheses tested

**Initial hypothesis (rejected):** `PodOperationUtil.waitUntilReadyOrTerminal` issues a doomed `LIST /…/pods?fieldSelector=metadata.name=…` when `readyWaitTimeout=0` (the default since #606d3590fc), and that wasted call contends with the WS upgrade.

[...]

PR #7742, opened 2026-05-07
fix(kubernetes-client): skip doomed informer when readyWaitTimeout=0

### LINKED ISSUE #7741: PodOperationUtil.waitUntilReadyOrTerminal issues a doomed LIST when readyWaitTimeout=0
## Symptom

Every `client.pods()...exec()`, `attach()`, `getLog()`, etc. call issues a doomed `LIST /api/v1/namespaces/<ns>/pods?fieldSelector=...` before proceeding to the actual operation. In mock-server tests this shows up as a 404 + `Reflector` ERROR stack trace; on real clusters it is one wasted round-trip plus event-loop contention with the follow-up call (the WS upgrade for exec, the GET for log, etc.).

## Root cause

`#606d3590fc` (Nov 2024) (#6665 #6140 #5782) changed `PodOperationsImpl.DEFAULT_POD_READY_WAIT_TIMEOUT_MS` from 5000 to 0. After that, the default path through `PodOperationUtil.waitUntilReadyOrTerminal(podOp, 0)` still:

1. creates an informer + handlers
2. starts a list/watch
3. immediately times out (`Utils.waitUntilReady(future, 0, MS)` returns `false` synchronously)
4. cancels the future, schedules `informer.stop()`

The informer's predicate cannot fire in zero time, so the LIST is always wasted. Caller falls back to `getItemOrRequireFromServer()` for a synchronous GET regardless.

## Fix

Short-circuit `waitUntilReadyOrTerminal` when `logWaitTimeoutMs == 0` and return `null` directly. Negative values keep the existing \"wait indefinitely\" semantics from `Utils.waitUntilReady`.

## Out of scope

This is **not** a fix for #7737. The flaky `PodTest.testExecEphemeralContainer` was the trigger for this audit, but local A/B testing did not show a measurable change in flake rate from this cleanup alone. #7737 stays open.

manusa:
## Summary

Since #606d3590fc (Nov 2024) made `DEFAULT_POD_READY_WAIT_TIMEOUT_MS = 0`, every default `client.pods()...exec()` / `attach()` / `getLog()` call issues a doomed `LIST /…/pods?fieldSelector=metadata.name=…` before the actual operation: a list/watch is started, the wait times out synchronously (`Utils.waitUntilReady(future, 0, MS)` returns `false`), the future is cancelled. In mocks this surfaces as a 404 plus a `Reflector` ERROR stack trace; on real clusters it's one wasted round-trip plus event-loop contention with the follow-up call.

`PodOperationUtil.waitUntilReadyOrTerminal` now short-circuits when `logWaitTimeoutMs == 0` and returns `null` directly. Callers (`PodOperationsImpl` for exec/attach/log, `DeploymentConfigOperationsImpl`/`BuildOperationsImpl` on OpenShift) already tolerate `null` and fall back to `getItemOrRequireFromServer()`. Negative values are still passed through and interpreted by `Utils.waitUntilReady` as "wait indefinitely".

Two new unit tests pin both branches: `0` short-circuits without calling `waitUntilCondition`; `-1` delegates to it as before.

Fixes #7741

## Out of scope

Not a fix for #7737. The flaky `PodTest.testExecEphemeralContainer` was the trigger for this audit, but local A/B testing did not show a measurable change in flake rate from this cleanup alone (17/100 vs 18/100 fail at a 90 ms request timeout, on a 16-core machine where the WS-upgrade race does not surface naturally). #7737 stays open.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * Fix #7350: Improper callback timing in leaderelection leads to the dual-leader
 
 #### Improvements
+* Fix #7741: PodOperationUtil.waitUntilReadyOrTerminal short-circuits when readyWaitTimeout=0 (the default) instead of starting a doomed informer LIST that races with the follow-up call (e.g. exec WebSocket upgrade) and produces 404 ERROR log noise
 * Fix #7662: (mockwebserver) new `MockWebServer#setHttp2ClearTextEnabled(boolean)` setter to opt out of HTTP/2 cleartext (h2c) upgrade
 * Fix #7522: improve dependency management for kubernetes-httpclient-okhttp
 * Fix #7550: add a ResourceEventHandler onList method and deprecated onNothing
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/internal/PodOperationUtil.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/utils/internal/PodOperationUtil.java
@@ -119,6 +119,16 @@ public class PodOperationUtil {
   }
 
   public static Pod waitUntilReadyOrTerminal(PodResource podOperation, int logWaitTimeoutMs) {
+    if (logWaitTimeoutMs == 0) {
+      // Caller does not want to wait (the default since DEFAULT_POD_READY_WAIT_TIMEOUT_MS=0):
+      // starting an informer would issue a list/watch whose predicate cannot fire in zero
+      // time, produce log noise on a failed list, and contend with subsequent calls (e.g.
+      // the exec WebSocket upgrade) on the shared HttpClient/event loop for nothing. Skip
+      // and let the caller fall back to getItemOrRequireFromServer().
+      // Negative values are passed through and interpreted by Utils.waitUntilReady as
+      // "wait indefinitely".
+      return null;
+    }
     AtomicReference<Pod> podRef = new AtomicReference<>();
     try {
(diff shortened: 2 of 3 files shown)


manusa:
Closing — CI surfaced a real behavior change I missed during analysis.

The informer LIST in `waitUntilReadyOrTerminal` is **not** purely wasted work. With the `Runnable::run` executor and a fast HTTP layer (synchronous in tests, possibly fast on real clusters too), the LIST can:

1. Return synchronously / fast enough that the future is already complete by the time `Utils.waitUntilReady(future, 0, MS)` is called
2. The future short-completes via `future.get(0, MS)` returning the cached value (not throwing `TimeoutException`)
3. `waitUntilCondition` returns the pod, so `validateOrDefaultContainerId` skips the follow-up `getItemOrRequireFromServer()` GET

Several tests rely on this fast path: `UploadTest` only mocks the LIST via `informPodReady(...)`, never the per-pod GET. `RequestTimeoutTest.httpLogNoFollowOperationsDefaultTimeout` asserts exactly 2 HTTP calls; my change drops it to 1.

So the "doomed informer LIST" framing was wrong: it's an opportunistic prefetch that succeeds whenever the HTTP layer happens to be fast. My change forces an unconditional extra GET on real clusters where the prefetch does succeed, and breaks tests that lock in the prefetch behavior.

#7741 stays open. Any follow-up needs to either (a) preserve the fast-path semantics, or (b) explicitly redesign that contract.

Issue #7675, opened 2026-04-28
Speed up CI test runs (round 2): -T 1C in CI, forkCount tuning, OpenShift delete tests


manusa:
Follow-up to #7648 (closed by #7673). After landing the first round, audited the build a second time to find the next set of measurable wins. A few items from the original umbrella turned out to be moot once verified against the post-#7673 tree — those are listed under "Explicitly out of scope" with the reason.

## Status

- [x] 1. **CI: add `-T 1C` to multi-module Maven invocations** — local `make` already uses it; CI does not. Largest expected wall-time win on PRs. Risk: low (Sonar profile already opts out). **(Done — Windows in #7768, Linux in #7769.)**
- [x] 2. **`forkCount=1C` on `kubernetes-client-api` and `kubernetes-client`** — measured locally below. Risk: low (state-leakage already addressed by #7667/#7673). **(Done in #7684.)**
- [x] 3. **Fix the OpenShift mock `delete` test pattern** — ~13 tests waste ~16 s each in retry backoff against a single-shot mock. Risk: low; localized to test-side `withGracePeriod(0)` / additional mock expectation. **(Done in #7683 — but kept for hygiene only; the wall-time premise was wrong, see re-diagnosis below.)**
- [ ] 4. **MockedStatic-leak guard (JUnit 5 extension)** — covered by #7673's incident; only 4 test files in the whole repo use `MockedStatic`, so the guard surface is small. Risk: none.
- [x] 5. **Ship-leak fix: `chaosmesh-client` packages `org.mockito.plugins.MockMaker` in its production jar** — out of scope for "speed", but discovered during this audit; flagging because it leaks `mock-maker-inline` to consumers. Risk: none. **(Done in #7692.)**
- [x] 6. **Drop the explicit `mockito-inline:5.2.0` test deps and the in-tree `mockito-extensions` resource** — Mockito 5 (this project uses 5.11.0/5.23.0) bundles the inline maker as the default in `mockito-core`; the separate dep and the resource files are no-ops that confuse readers. Risk: none. **(Done in #7692.)**
[...]

PR #7777, opened 2026-05-13
fix(kubernetes-server-mock): share MockWebServer + Vertx per JVM fork

manusa:
## Status: scoped down after CI measurements

After landing the initial implementation, the actual Java 21 CI job wall time was measured 4× on each side (`gh run rerun` against fixed run IDs — see [comment](https://github.com/fabric8io/kubernetes-client/pull/7777#issuecomment-4438758301) for the full methodology) and the delta sat at ~1% median, well inside the per-run noise floor (main alone swung ~16% across the 4 samples). The shared MockWebServer + Vertx infrastructure was reverted; only the small correctness fixes survived.

## What ships at HEAD

- Marks `MockWebServer.dispatcher` `volatile` so a `setDispatcher(...)` call is reliably visible to the Vert.x request-handler thread without further synchronization. Defense-in-depth for any future dispatcher-swap caller.
- Tightens the `MockWebServer.reset()` Javadoc to make its non-destructive contract explicit (no change to the running server, dispatcher, listeners, SSL/TLS state, port, or protocols).

Net diff at HEAD vs `main`: ~2 functional lines in `MockWebServer.java` plus a single Improvements entry in `CHANGELOG.md`.

## Original investigation (discarded — preserved in commit history)

The first two commits attempted to amortize the Vert.x/Netty cold-start across all `@EnableKubernetesMockClient` tests in a JVM fork:

- Reuse one `MockWebServer` per transport (HTTP/HTTPS) and one `Vertx` per JVM fork in `KubernetesMockServerExtension` instead of creating fresh instances per `@BeforeEach`. Per-test isolation preserved via dispatcher and responses-map swap on each entry, plus `MockWebServer.reset()` to clear request count and queue. Shared resources torn down by a JVM shutdown hook.
- Reflective `(Vertx)`-constructor probe so the shared Vertx flows into either the Vert.x 4 or Vert.x 5 `HttpClient.Factory` without a hard compile-time dep, with a graceful fallback for OkHttp / JDK consumers.

Local 2-core Docker measurements showed per-class wall times dropping by up to ~38% on small classes (e.g. `MachineConfigPoolTest` 0.68s → 0.42s, `DeploymentTest` 4.20s → 3.53s). Those numbers did **not** translate to job-level wall time on the real CI matrix, so the infrastructure was reverted in commit `5e0954a7c5`.

### Consumer-visible behavior changes (discarded — no longer applicable)

These would have been the breaking-change surface if the sharing infrastructure had shipped. Listing for the record so the trade-offs are visible to future contributors:

1. ~~`KubernetesMockServer#getPort()` / `getHostName()` stable across tests within a JVM fork (was: fresh random port per test).~~
2. ~~`MockWebServer.addListener(...)` listeners persist for the lifetime of the fork (no auto-removal API exists).~~
3. ~~Test classes that declare both a `static` and an instance `KubernetesMockServer` field under `@TestInstance(PER_CLASS)` see both fields point to wrappers over the same underlying server (was: two servers on different ports).~~

Refs #7675

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,15 @@
 * Fix #7765: (kubernetes-client) `BaseOperation.informOnCondition` now stops the informer inline when the inner predicate completes the future, closing a CompletableFuture `postComplete` race where a waiter helping drain dependents could fire `informer.stop` after `cf.complete` had already triggered a spurious `?watch=true` HTTP request
 
 #### Improvements
+* Fix #7675: (kubernetes-server-mock) `KubernetesMockServerExtension` now reuses a single `MockWebServer` (one for HTTP, one for HTTPS) and a single `Vertx` per JVM fork across all tests using `@EnableKubernetesMockClient`, amortizing the Vert.x/Netty cold-start cost across the suite. Per-test isolation is preserved by swapping the dispatcher and `responses` map on each `@BeforeEach` and calling `MockWebServer.reset()` to clear the request queue and request count. Wall-time savings depend on suite size; on `kubernetes-tests` per-class wall times drop by up to ~38% on small classes
+* Fix #7675: (mockwebserver) `MockWebServer.dispatcher` field marked `volatile` so the dispatcher swap performed by `KubernetesMockServerExtension` between tests is reliably visible to the Vert.x request handler thread
 
 #### Dependency Upgrade
 
 #### New Features
 
 #### _**Note**_: Breaking changes
+* Fix #7675: (kubernetes-server-mock) Tests using `@EnableKubernetesMockClient` now share a single `MockWebServer` per transport (HTTP/HTTPS) and a single `Vertx` for the duration of the JVM fork instead of constructing a fresh instance per test method. Consumer-visible behavior changes: (1) `KubernetesMockServer#getPort()` and `getHostName()` are stable across tests within a fork (previously a fresh random port per test); (2) `MockWebServer#addListener(...)` listeners persist for the lifetime of the fork (no auto-removal API exists); (3) test classes that declare both a `static` and an instance `KubernetesMockServer` field under `@TestInstance(PER_CLASS)` now see both fields point to wrappers over the same underlying server (previously: two separate servers on different ports). Tests that don't depend on these specifics are unaffected
 
 ### 7.7.0 (2026-05-12)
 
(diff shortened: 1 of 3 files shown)


manusa:
## CI wall-time measurements (Java 21 Build job)

Ran `Java 21 Maven` 4× on each branch via sequential `gh run rerun` against fixed run IDs, so each branch sees the same runner pool, same Maven cache state, same suite. Main and PR were measured in parallel against different refs (`refs/heads/main` vs `refs/pull/7777/merge`); samples within a ref are forced sequential by the `cancel-in-progress` concurrency lock.

### Raw timings

| Sample | main (mm:ss) | PR (mm:ss) |
|---|---|---|
| 1 | 14:22 | 14:36 |
| 2 | 14:38 | 13:42 |
| 3 | 16:40 | 14:11 |
| 4 | 14:26 | 14:40 |
| **min** | **14:22** | **13:42** |
| **median** | **14:32** | **14:23** |
| **mean** | **15:02** | **14:17** |
| **max** | **16:40** | **14:40** |

### Deltas (PR vs main)

- **Median**: −9s (~1% faster)
- **Min**: −40s (~5% faster)
- **Mean**: −44s (~5% faster) — but heavily influenced by main's 16:40 outlier
- **Max**: −120s (~12% faster) — same caveat as mean

### Honest read

1. **Signal sits inside the noise floor.** Main alone swung 138s peak-to-trough across 4 samples (~16% range). The 1% median delta is well within that. The 5% mean delta is essentially a story about one slow main run.
2. **The 38% headline from the PR body doesn't translate to job-level wall time.** That number was per *small* test class where the Vert.x cold-start dominates. At the whole-job level the suite is dominated by long-running classes where the amortised cold-start is a small fraction — the ceiling is more like 3–8%, not 38%.
3. **n=4 tightened the picture, not the gap.** Adding one sample per side dropped the median delta from 27s to 9s. The original n=3 was over-optimistic about the gain.

### Pushback for #7675

The shared-server + shared-Vertx infrastructure (~200 lines incl. shutdown hook, reflective `(Vertx)` constructor probe, dispatcher lifecycle management) buys ~1% median Java 21 job wall time in exchange for permanent maintenance overhead and three documented breaking-change semantics (stable port across tests, persistent listeners, shared underlying server for `@TestInstance(PER_CLASS)` dual-field classes). At this gain magnitude the ROI looks negative.

Two cheaper subsets of this PR are still worth keeping on their own:

- `volatile` on `MockWebServer.dispatcher` — a 1-line correctness fix that's valid for any future dispatcher-swap caller.
- The tightened `reset()` Javadoc — makes the non-destructive contract explicit and load-bearing.

### Methodology notes

- `gh run rerun --job <id>` against successive Java 21 job IDs from `runs/{run_id}/attempts/{attempt}/jobs`. Each rerun creates a new `job_id`, so the rerun must be re-pointed each iteration.
- One PR attempt (attempt 2) was cancelled mid-flight at ~33m by something outside the test loop — most likely a transient GHA hiccup. The recorded 4 PR samples are attempts 3/4/5/6; the cancelled attempt isn't in the data.

Issue #7798, opened 2026-05-15
Speed up CI build phase (round 3): ARM64 runners, --builder smart, Windows job alignment


manusa:
Follow-up to #7675. After landing the test-side wins, the build/compile phase has emerged as the dominant CI wall-time cost — a recent main-branch PR cell on `build.yml` runs ~16 min on Linux and ~22 min on Windows, with the majority spent compiling ~25 model modules under sundrio APT. This issue collects four small, scoped build-phase changes that are independent of the bigger model-caching work tracked separately in #5622, and can land independently of each other.

## Status

- [x] 1. **Try GitHub ARM64 Linux runners for `build.yml`** — landed in #7799 for Java 11 + Java 17 cells (~28–30% wall-time reduction, n=4 samples vs n=7 baseline). Java 21 ARM hit Java-21-specific flakes (`DefaultMockServerWebSocketTest`, `KubeApiServerTest`); reverted to `ubuntu-latest`. Reverting surfaced a third flake on x64 too (`MasterProtocolTest`), so the Java 21 flake surface is broader than ARM and predates this work — see follow-ups.
- [ ] 2. **~~Try `mvn --builder smart -T 1C`~~** — **Discarded.** The bundled Maven 3.9.9 lacks the `smart` core builder; `./mvnw -b smart …` errors with *"The builder requested using id = smart cannot be found"*. `smart` is the Takari `takari-smart-builder` extension (last release 2018, unmaintained), which would need to be registered in `.mvn/extensions.xml`. Not worth the dependency for a speculative 0–10% gain. Revisit if/when we move to Maven 4.
- [x] 3. **Drop `clean` from the Windows job's `mvnw -T 1C clean install`** — landed in #7799 (~1m12–18s saved, matching the predicted 30s–2min).
- [x] 4. **Align Windows `MAVEN_ARGS` with Linux** — landed in #7799. Implementation note: the Windows build step also had to switch to `shell: bash` (Git Bash) because pwsh's native-command argument parser was splitting `-Dhttp.keepAlive=false` into `-Dhttp` + `.keepAlive=false` (one of those passed as a separate argument to mvn, which then read it as a lifecycle phase).

## Results

| Cell | Median (n=4) | x64 baseline (n=7) | **Savings** |
|---|---|---|---|
| Linux Java 11 (`ubuntu-24.04-arm`) | 12m13.5s | 17m25s | **~30%** |
[...]

PR #7802, opened 2026-05-16
test: stabilize three contention-driven CI flakes (#7800)

### LINKED ISSUE #7800: Java 21 CI flakiness: three different timing-sensitive failures surfaced during #7798 benchmark
While benchmarking `ubuntu-24.04-arm` runners for #7798 (PR #7799), the Java 21 Maven cell hit three different intermittent test failures across 8 build attempts (4× ARM, 4× x64). Java 11 and Java 17 cells were clean across the same runs (0/8 failures each), so the pattern is **Java-21-specific, not arch-specific**. ARM amplifies the rate but x64 hits at least one of the failures too.

Filing as a follow-up to #7798 / #7799 (where Java 21 was reverted from ARM to keep that PR green and shippable). This issue captures the evidence so the underlying flakes can be investigated separately from the CI-perf work.

## Failures observed

| # | Run / job | OS | Test | Symptom |
|---|---|---|---|---|
| 1 | [run 25921509324, attempt 1](https://github.com/fabric8io/kubernetes-client/actions/runs/25921509324/job/76191614925) | `ubuntu-24.04-arm` | `io.fabric8.mockwebserver.DefaultMockServerWebSocketTest.andUpgradeToWebSocket, with no events, should emit onClose` | 10 s timeout waiting for `onClose` event. Pre-existing flake history on this file (multiple "improve reliability" commits in `git log junit/mockwebserver/src/test/groovy/io/fabric8/mockwebserver/DefaultMockServerWebSocketTest.groovy`). |
| 2 | [run 25921509324, attempt 3](https://github.com/fabric8io/kubernetes-client/actions/runs/25921509324/job/76200710097) | `ubuntu-24.04-arm` | `io.fabric8.kubeapitest.sample.KubeApiServerTest.creatingClientFromConfigString` | 60 s `ProcessReadinessChecker.ready` timeout. The arm64 binaries downloaded correctly and the sibling `JUnitExtensionTargetVersionTest` started its own API server in 9.6 s on the same run, so the binaries work — it's a contention-driven timing failure under `-T 1C` parallelism on a 2-vCPU runner. |
| 3 | [run 25927012575, attempt 1](https://github.com/fabric8io/kubernetes-client/actions/runs/25927012575/job/76210864881) | `ubuntu-latest` | `io.fabric8.kubernetes.client.server.mock.MasterProtocolTest.testWithoutSSL` | Fast error (0.162 s), not a timeout: `mapping values are not allowed here, in reader, line 16, column 15: client-key: /home/runner/.kubeapitest/clie ...`. Test isolation leak — `kube-api-test` writes to `$HOME/.kubeapitest/` and a parallel module's test reads a partial kubeconfig. (Worth its own follow-up — see [#7798 follow-ups](https://github.com/fabric8io/kubernetes-client/issues/7798#follow-ups-discovered-during-this-work).) |

Across the four samples per cell:

- Java 11 ARM: 4/4 clean
- Java 17 ARM: 4/4 clean
- Java 21 ARM: 2/4 clean (failures 1 and 2 above)
- Java 21 x64: 3/4 clean (failure 3 above)

## What's odd about this

Each failure is a different test, with a different symptom (10 s timeout, 60 s timeout, fast YAML parse error). They share two properties:

1. **Timing-sensitive / contention-prone** — all three are exposed by `-T 1C` reactor parallelism on a 2-vCPU runner, where Surefire forks across modules compete for CPU, file handles, and `$HOME` state.
2. **Java-21-specific failure rate** — the *same tests* run on Java 11 and 17 in the same matrix without failing. This is the part that needs an explanation.

## Open questions

- Is there a Java 21 default that interacts badly with `-T 1C` Surefire forks on 2-vCPU runners? Candidates worth checking:
  - GC default change (G1 tuning, ZGC availability) altering tail-latency under contention.
  - `Thread.sleep` / `LockSupport.parkNanos` precision differences across JDKs on aarch64.
  - Default `-XX:ActiveProcessorCount` resolution under cgroups (GH Actions runners use cgroup v2).
- Are these tests passing on Java 21 *locally* under the same `-T 1C` constraint? (The CI flake repro guide in `CLAUDE.md` covers the docker constrained-runner approach.)
- Is the WS test (`DefaultMockServerWebSocketTest`) reliably reproducible under Java 21 + constrained Docker, or only on ARM CI specifically?
- For the kube-api-test contention: are tests in the `kube-api-test` module supposed to be serialised (i.e. one API server at a time), or is the per-test `KubeAPIServer` startup expected to handle concurrent starts on the same host?

## Reproducer notes (not yet attempted)

Per the local CI-flake-repro recipe in `CLAUDE.md`:

```
docker run --rm \
  --cpus=2 --memory=7g --cpuset-cpus="0,1" \
  --user $(id -u):$(id -g) -e HOME=/tmp \
  -v "$PWD:/work" -v "$HOME/.m2:/host-m2" -w /work \
  maven:3.9-eclipse-temurin-21 \
  mvn -pl junit/mockwebserver,junit/kube-api-test/core,kubernetes-tests \
      -Dmaven.repo.local=/host-m2/repository \
      -T 1C \
      test
```

If 20 iterations of this don't reproduce, fall back to instrument-and-wait on a debug branch.

## Out of scope here

- Fixing `kube-api-test` `$HOME` isolation (separate follow-up).
- Reverting Java 21 to ARM (depends on this investigation landing).

Refs #7798, #7799.

manusa:
Closes #7800. Three intermittent CI failures observed during the #7798 ARM/Java-21 benchmark. Different symptoms; shared contention root cause from the Maven reactor's `-T 1C` parallel module builds + surefire `forkCount` on 2-vCPU runners.

## 1. `MasterProtocolTest.testWithoutSSL` — SnakeYAML "mapping values are not allowed here"

`JUnitExtensionKubeConfigUpdateTest` shells out to `kubectl config set-*`, which writes `$HOME/.kube/config` via `os.WriteFile` (`open(O_TRUNC) + write`) — non-atomic against external readers. A sibling surefire fork's `MasterProtocolTest` calls `new ConfigBuilder().build()` → `Config.autoConfigure(...)` → `KubeConfigUtils.parseConfig(...)`, which can catch a partial write.

A standalone Java reproducer (`Files.newOutputStream(..., TRUNCATE_EXISTING)` writer + `Files.newInputStream` reader + SnakeYAML) emits the exact same `ScannerException` at ~171/1.6M reads in 30 s.

**Fix**: scope `HOME` to `${project.build.directory}` per surefire fork in both `kube-api-test/core` and `kube-api-test/client-inject`, with `KUBE_API_TEST_DIR` pinned back to `${user.home}/.kubeapitest` so the binary/cert cache survives `mvn clean`. `~/.kube/config` is no longer touched by these forks; other modules' `Config.autoConfigure` reads the runner's unmodified file.

**Known limitation (Windows)**: the surefire-level isolation does not work on Windows — three attempted variants (`KUBECONFIG` forward-slash, `KUBECONFIG` `${file.separator}`, and the final `HOME` form) all fail identically with `UnknownHostException: kubernetes.default.svc`. Likely cause is Go's `homedir.HomeDir` preferring `%USERPROFILE%` over `$HOME` on Windows when `$HOME/.kube/config` doesn't already exist. `JUnitExtensionKubeConfigUpdateTest` is `@DisabledOnOs(OS.WINDOWS)` and the Windows fix is tracked in **#7804**.

## 2. `KubeApiServerTest.creatingClientFromConfigString` — 60 s `ProcessReadinessChecker` timeout

`ready()` rejects every status code except 200. The unauthenticated `/readyz` probe returns 403 until kube-apiserver's post-start hook seeds the `system:public-info-viewer` ClusterRoleBinding. On an idle host the window is ~1.2 s; under CI contention it can exceed the 60 s `startupTimeout`.

Confirmed by booting kube-apiserver 1.26 locally and instrumenting the probe (`403 → 200` transition at t≈2.6 s).

**Fix**: add a `waitUntilReady` overload that accepts the project's `CertManager` and uses mTLS for `/readyz`. With the `system:masters` client cert, the probe sees the real readiness state (`500 → 200`) and never hits the 403 path. The original 5-arg signature is preserved for the etcd call site.

**Known limitation (residual)**: the mTLS fix addresses only the 403 sub-case. Under heavy `-T 1C` contention on 2-vCPU runners the apiserver process itself can take >60 s to bind, hitting the same `startupTimeout` boundary in a different test (`KubernetesMutationHookHandlingTest` observed once across 13 reruns). Tracked in **#7807**.

## 3. `DefaultMockServerWebSocketTest, with no events, should emit onClose` — 10 s timeout

Vert.x `closeHandler(...)` throws `IllegalStateException: WebSocket is closed` when registering on an already-closed WebSocket. The previous `isClosed() ? capture : register` order had a race window where the close frame landed between the check and the registration. Vert.x logs the exception as SEVERE but swallows it on the event loop — `closeReason` is never set and the test times out at 10 s.

**Fix**: wrap `closeHandler` registration in `try/catch (IllegalStateException)` and re-check `isClosed()` after. Whichever path actually sees the close populates `closeReason`.

**Known limitation (Windows)**: the same `Condition not satisfied` symptom was observed once on Windows Java 17 across 13 reruns — `closeReason` stayed null even with the try/catch in place. The Windows manifestation appears to be a *different* race than the `IllegalStateException` path captured locally (handler may fire but `closeReason()` returns null, or `ws.result()` is null and the closure NPEs silently). Tracked in **#7806**.

## Stability sample

Ran 13 full workflow attempts on this PR (initial + 12 reruns):
- 10 fully green, 3 with at least one job failure (23% per-attempt — down from ~37% pre-fix baseline reported in #7798).
- 2 of the 3 failures are residuals of #7800 root causes that this PR's fixes don't cover (now tracked in #7806, #7807).
- 1 attempt hit 2 unrelated pre-existing flakes (`LeaderElectorTest.jitterWithNegativeShouldReturnDuration`, `AsyncUtilsTest.withTimeout_timeout`) — out of scope here.

See the [stability summary comment](https://github.com/fabric8io/kubernetes-client/pull/7802#issuecomment-4469398474) for the full per-failure table.

## Notes

- `~/.kubeapitest/` (cert + binary cache) remains shared across kube-api-test/core and kube-api-test/client-inject forks. They serialize via the reactor dep graph (`client-inject` depends on `kube-api-test`), so cert-file generation is not a current race source.
- Internal test stabilization — no CHANGELOG entry.

## Follow-ups

- **#7804** — Windows surefire-level kubeconfig isolation
- **#7806** — Windows WS test still flakes after the try/catch fix
- **#7807** — `kube-api-test` 60 s `startupTimeout` boundary under heavy `-T 1C` contention

### DIFF
--- a/junit/kube-api-test/client-inject/pom.xml
+++ b/junit/kube-api-test/client-inject/pom.xml
@@ -78,7 +78,7 @@
                 <configuration>
                     <environmentVariables combine.children="append">
                         <!-- Isolate kubeconfig writes to this fork; see kube-api-test/core/pom.xml. -->
-                        <KUBECONFIG>${project.build.directory}/test-home/.kube/config</KUBECONFIG>
+                        <KUBECONFIG>${project.build.directory}${file.separator}test-home${file.separator}.kube${file.separator}config</KUBECONFIG>
                     </environmentVariables>
                 </configuration>
             </plugin>
--- a/junit/kube-api-test/client-inject/pom.xml
+++ b/junit/kube-api-test/client-inject/pom.xml
@@ -69,4 +69,19 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <environmentVariables combine.children="append">
+                        <!-- Isolate kubeconfig writes to this fork; see kube-api-test/core/pom.xml. -->
+                        <KUBECONFIG>${project.build.directory}/test-home/.kube/config</KUBECONFIG>
+                    </environmentVariables>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
 </project>
--- a/junit/kube-api-test/client-inject/pom.xml
+++ b/junit/kube-api-test/client-inject/pom.xml
@@ -78,7 +78,8 @@
                 <configuration>
                     <environmentVariables combine.children="append">
                         <!-- Isolate kubeconfig writes to this fork; see kube-api-test/core/pom.xml. -->
-                        <KUBECONFIG>${project.build.directory}${file.separator}test-home${file.separator}.kube${file.separator}config</KUBECONFIG>
+                        <HOME>${project.build.directory}</HOME>
+                        <KUBE_API_TEST_DIR>${user.home}${file.separator}.kubeapitest</KUBE_API_TEST_DIR>
                     </environmentVariables>
                 </configuration>
             </plugin>
--- a/junit/kube-api-test/core/pom.xml
+++ b/junit/kube-api-test/core/pom.xml
@@ -127,7 +127,7 @@
                              catch a partial write, producing a SnakeYAML "mapping values are not
                              allowed here" failure (see #7800). Scoping KUBECONFIG to target/ keeps
                              this fork's kubectl writes from racing with sibling-fork readers. -->
-                        <KUBECONFIG>${project.build.directory}/test-home/.kube/config</KUBECONFIG>
+                        <KUBECONFIG>${project.build.directory}${file.separator}test-home${file.separator}.kube${file.separator}config</KUBECONFIG>
                     </environmentVariables>
                 </configuration>
             </plugin>
--- a/junit/kube-api-test/core/pom.xml
+++ b/junit/kube-api-test/core/pom.xml
@@ -112,4 +112,25 @@
             <artifactId>jackson-dataformat-yaml</artifactId>
         </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <environmentVariables combine.children="append">
+                        <!-- Isolate kubeconfig writes to this fork. JUnitExtensionKubeConfigUpdateTest
+                             shells out to `kubectl config set-*`, which truncate+writes the file
+                             non-atomically (client-go's os.WriteFile). Other modules' tests
+                             concurrently read $HOME/.kube/config in Config.autoConfigure and can
+                             catch a partial write, producing a SnakeYAML "mapping values are not
+                             allowed here" failure (see #7800). Scoping KUBECONFIG to target/ keeps
+                             this fork's kubectl writes from racing with sibling-fork readers. -->
+                        <KUBECONFIG>${project.build.directory}/test-home/.kube/config</KUBECONFIG>
+                    </environmentVariables>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
 </project>
--- a/junit/kube-api-test/core/pom.xml
+++ b/junit/kube-api-test/core/pom.xml
@@ -120,14 +120,22 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
                     <environmentVariables combine.children="append">
-                        <!-- Isolate kubeconfig writes to this fork. JUnitExtensionKubeConfigUpdateTest
-                             shells out to `kubectl config set-*`, which truncate+writes the file
-                             non-atomically (client-go's os.WriteFile). Other modules' tests
-                             concurrently read $HOME/.kube/config in Config.autoConfigure and can
-                             catch a partial write, producing a SnakeYAML "mapping values are not
-                             allowed here" failure (see #7800). Scoping KUBECONFIG to target/ keeps
-                             this fork's kubectl writes from racing with sibling-fork readers. -->
-                        <KUBECONFIG>${project.build.directory}${file.separator}test-home${file.separator}.kube${file.separator}config</KUBECONFIG>
+                        <!-- Scope $HOME to this fork's target/ so kubectl (from
+                             JUnitExtensionKubeConfigUpdateTest) and Config.autoConfigure both
+                             resolve $HOME/.kube/config inside the fork instead of touching the
+                             developer/runner's real ~/.kube/config. kubectl's `config set-*`
+                             writes are non-atomic (client-go's os.WriteFile) and sibling forks
+                             read $HOME/.kube/config concurrently, producing a SnakeYAML
+                             "mapping values are not allowed here" failure (see #7800).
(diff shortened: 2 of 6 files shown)


manusa:
## 13-rerun stability sample

Ran 13 full workflow attempts on this PR (initial run + 12 reruns) to gauge the post-fix flake rate.

| | Count | Rate |
|---|---|---|
| Attempts | 13 | — |
| Fully green | 10 | 77% |
| At least one job failure | 3 | 23% |

Compared to the pre-fix baseline reported in #7800 (~37% per-cell on the #7798 benchmark), the rate is roughly **halved**, but not zero.

### Failure breakdown

| Rerun | Cell | Test | Cause | Scope |
|---|---|---|---|---|
| #6 | Java 11 Linux | `KubernetesMutationHookHandlingTest` | `KubeAPIServer did not start properly` — 60.22s `startupTimeout` boundary hit; mTLS fix only addresses RBAC 403 sub-case, not the broader "apiserver slow to bind under -T 1C contention" mode | Adjacent to #7800 #2 — see #7807 |
| #10 | Java 17 Linux | `LeaderElectorTest.jitterWithNegativeShouldReturnDuration` | `expected: <true> but was: <false>` — jitter-randomness assertion | Pre-existing, unrelated |
| #10 | Java 17 Windows | `AsyncUtilsTest.withTimeout_timeout` | Expected `ExecutionException`, got `TimeoutException` — timing-sensitive exception-type assertion | Pre-existing, unrelated |
| #12 | Java 17 Windows | `DefaultMockServerWebSocketTest, with no events, should emit onClose` | `Condition not satisfied after 10.01 seconds` — same symptom as original #7800 failure #1. The `try/catch (IllegalStateException)` fix in this PR addresses one race path; the Windows manifestation appears to be a different one (handler fires but `closeReason()` returns null, or `ws.result()` is null and the closure NPEs silently) | Same as #7800 #1 — see #7806 |

### Per-failure rate within scope of this PR

- #7800 failures (failures #1, #2, #3): **2 hits in 13 attempts** (1× residual WS test on Windows, 1× apiserver-readiness under heavy contention not in the 403 sub-case)
- Unrelated pre-existing flakes: **2 hits in 1 attempt** (both in rerun #10)

Both in-scope residuals are documented in follow-up issues:
- **#7806** — Windows-specific WS test still flakes after #7802 fix
- **#7807** — apiserver `startupTimeout` boundary hit under heavy reactor contention (`KubernetesMutationHookHandlingTest` and friends)

Calling it stable enough to merge — the three #7800 root causes are addressed for the failure modes I could mechanically verify, and the residuals have isolated follow-ups.

PR #28, opened 2015-07-24
Upgrade model

jimmidyson:

PR #7809, opened 2026-05-17
Support for shard selectors for list and watch - including informers

csviri:
Allows to pass shardSelector for Filterable, this is mainly important for informers, so we can have sharding in JOSDK.
Might be used both for list and watch subsequently in informer implementation.

Signed-off-by: Attila Mészáros <a_meszaros@apple.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 #### Improvements
 * Fix #7675: (mockwebserver) `MockWebServer.dispatcher` field marked `volatile` so a `setDispatcher(...)` call is reliably visible to the Vert.x request handler thread without further synchronization. `MockWebServer.reset()` Javadoc tightened to make its non-destructive contract explicit (no change to the running server, dispatcher, listeners, SSL/TLS state, port, or protocols)
-* Fix #7809: (kubernetes-client) Support for shard selectors for listing and informers
+* Fix #7809: (kubernetes-client) Support for shard selectors for list and watch - including informers
 
 #### Dependency Upgrade
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 #### Improvements
 * Fix #7675: (mockwebserver) `MockWebServer.dispatcher` field marked `volatile` so a `setDispatcher(...)` call is reliably visible to the Vert.x request handler thread without further synchronization. `MockWebServer.reset()` Javadoc tightened to make its non-destructive contract explicit (no change to the running server, dispatcher, listeners, SSL/TLS state, port, or protocols)
+* Fix #7809: (kubernetes-client) Support for shard selectors for listing and informers
 
 #### Dependency Upgrade
 
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/dsl/Filterable.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/dsl/Filterable.java
@@ -76,8 +76,19 @@ public interface Filterable<T> {
   T withLabelSelector(String selectorAsString);
 
   /**
+   * Sets the {@code shardSelector} list option propagated to subsequent list and watch
+   * (including informer) requests. The selector is sent verbatim as a query parameter and
+   * is honored by API servers that support the {@code StorageVersionAPI} / shard-aware
+   * list-watch alpha. On servers that do not understand the parameter it is ignored.
+   *
+   * @param shardSelector the shard selector expression, or {@code null} to clear
+   * @return filtered resource
+   */
+  T withShardSelector(String shardSelector);
+
+  /**
    * Filter with the object that this event is about.
-   * 
+   *
    * @param objectReference {@link ObjectReference} for providing information of referred object
    * @return filtered resource
    */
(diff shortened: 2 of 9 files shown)


manusa (commented):
## Review

**Spec verification:** `Filterable.withShardSelector` is declared (Filterable.java:87), `OperationContext` stores it with copy + identity-on-equal semantics (OperationContext.java:71, 504-515), `BaseOperation` propagates to list, watch, informer via `defaultListOptions` (BaseOperation.java:563-566) and **also** to DELETECOLLECTION via `deleteAll` (628-632), and `FilterNestedImpl` carries it across nested filter mutations. CI is green on the v1.36.0 baremetal cell, so the IT does pass.

### Strengths

- Propagation surface is fully walked: list (`submitList:459`), watch (`submitWatch:750`), watch reconnect (`AbstractWatchManager:309`), informer (Reflector reuses both), and `deleteAll` all carry the new field. No bypass found.
- `appendListOptionParams` serializes `ListOptions` via Jackson, so the existing `shardSelector` field on the model picks it up automatically — no per-call assembly needed.
- `OperationContextTest.testShardSelectorIsPreservedAndOrthogonalToOtherSelectors` exercises orthogonality with `selectorAsString`, the identity-on-equal short-circuit, clear-via-null, and preservation across other mutations.
- IT uses real apiserver + alpha feature gate rather than mocking.

### Issues (Should Fix)

**1. Wrong feature-gate name in Javadoc** — `Filterable.java:81-82`

> *"...honored by API servers that support the `StorageVersionAPI` / shard-aware list-watch alpha."*

`StorageVersionAPI` is an unrelated K8s gate. The gate is `ShardedListAndWatch` (Alpha in 1.36) — same one the IT enables at `ShardingIT.java:35`. This will mislead anyone enabling it on their cluster.

**2. `shardSelector` on DELETECOLLECTION is undocumented and unverified** — `BaseOperation.java:628-632` + `Filterable.java:79-86`

The Javadoc says the field applies to "list and watch (including informer) requests." But the code also writes it onto `ListOptions` for the DELETECOLLECTION URL. The K8s KEP-5866 / alpha is scoped to list/watch in the docs (*"API server filters both list responses and watch event streams server-side"*). Whether the apiserver also honors it on DELETECOLLECTION's internal list traversal is unstated.

Two real risks if a user does `client.configMaps().withShardSelector(SHARD1).delete()`:
- If apiserver honors it → great, deletes only shard.
- If apiserver ignores it → deletes the **whole** label-matched collection while the caller thinks they deleted one shard. Data-loss footgun.

The envtest is already set up — add a third method to `ShardingIT` to nail this down before merge. Sketch:

```java
@Test
void shardedDelete() {
  // UIDs are server-assigned; create enough to populate both shards probabilistically.
  // With N=8, P(all-in-one-shard) ~ 0.78% — low enough; bump if needed.
  for (int i = 0; i < 8; i++) {
    client.resource(new ConfigMapBuilder()
        .withNewMetadata().withName("cm-" + i)
            .withNamespace("default").withLabels(Map.of("test", "true"))
        .endMetadata().build()).create();
  }

  int totalBefore = client.configMaps().withLabelSelector(LABEL_SELECTOR).list().getItems().size();
  int shard1Before = client.configMaps()
      .withLabelSelector(LABEL_SELECTOR).withShardSelector(SHARD1).list().getItems().size();
  int shard2Before = client.configMaps()
      .withLabelSelector(LABEL_SELECTOR).withShardSelector(SHARD2).list().getItems().size();
  assertThat(shard1Before + shard2Before).isEqualTo(totalBefore);
  Assumptions.assumeTrue(shard1Before > 0 && shard2Before > 0,
      "UID hashing didn't populate both shards — re-run");

  client.configMaps()
      .withLabelSelector(LABEL_SELECTOR).withShardSelector(SHARD1).delete();

  // The oracle: SHARD1 is empty, SHARD2 is untouched.
  assertThat(client.configMaps().withLabelSelector(LABEL_SELECTOR)
      .withShardSelector(SHARD1).list().getItems()).isEmpty();
  assertThat(client.configMaps().withLabelSelector(LABEL_SELECTOR)
      .withShardSelector(SHARD2).list().getItems()).hasSize(shard2Before);
}
```

If green → propagation is correct and the Javadoc just needs to mention deletion. If red → drop the assignment at `BaseOperation.java:628-632` and keep `shardSelector` strictly list/watch.

**3. `OperationContext` copy-ctor sidesteps the all-args ctor for one field** — `OperationContext.java:82-89`

```java
public OperationContext(OperationContext other) {
    this(other.client, ..., other.requestConfig);   // 26 fields go through here
    this.shardSelector = other.shardSelector;       // 27th set post-hoc
}
```

`shardSelector` is the only field treated this way. A future contributor adding field #28 via the same pattern compounds the smell, and any future call to the all-args ctor (or a `with*` mutator that forgets the post-assignment) silently loses the value. Either thread it through the all-args ctor (already `@SuppressWarnings(\"java:S107\")`, so one more param doesn't change the lint posture), or leave a one-line comment so the rationale survives.

**4. No unit test for the `deleteAll` propagation** — `BaseOperationTest.java`

`BaseOperationTest` covers `fetchListUrl` URL emission but not the parallel hand-rolled options assembly at `BaseOperation.java:611-636`. The envtest in #2 catches the apiserver-behavior question, but a unit test that captures the recorded delete URL and asserts `shardSelector=...` is present catches the *client-side* regression where a refactor drops the propagation. Both worth having.

**5. IT state leaks across methods** — `ShardingIT.java:45, 60, 81`

`@EnableKubeAPIServer` is class-scoped, so the apiserver is shared across all three test methods (and the new `shardedDelete` in #2 would be a fourth). All three current tests create `cm1` in `default`. Cleanup is reached only on success: `client.resource(cm).delete()` is inside the try-with-resources, *after* `await()`. If the `await()` ever fails, the next test hits `AlreadyExists` on create and we lose the original failure signal. Compare `PaginationIT` which uses a dedicated namespace + `@AfterAll`. Use `@AfterEach` to delete-if-exists, or unique names per test. The new `shardedDelete` makes this more pressing — it creates 8 objects.

csviri:
@manusa I addressed the issues mentioned above.

manusa (approved):
LGTM, thx!

PR #7838, opened 2026-05-19
fix(sonar): remove duplicated logger and resource fields (S1845, S2387)

### LINKED ISSUE #7825: [SONAR] Remove duplicated logger and resource fields in client classes

Two SonarCloud BLOCKER findings concern fields in subclasses that duplicate fields declared in their superclass.

- **[java:S1845]** [`kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/HasMetadataOperation.java:47`](https://github.com/fabric8io/kubernetes-client/blob/main/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/HasMetadataOperation.java#L47) — `private static final Logger logger` differs only in case from `BaseOperation.LOGGER` ([line 109](https://github.com/fabric8io/kubernetes-client/blob/main/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/BaseOperation.java#L109)). The project convention in `CLAUDE.md` is to name loggers `logger` (lowercase), so the duplication is also a convention mismatch.
- **[java:S2387]** [`kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/extension/ExtensibleResourceAdapter.java:33`](https://github.com/fabric8io/kubernetes-client/blob/main/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/extension/ExtensibleResourceAdapter.java#L33) — `protected ExtensibleResource<T> resource` shadows `ResourceAdapter.resource` ([line 61](https://github.com/fabric8io/kubernetes-client/blob/main/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/extension/ResourceAdapter.java#L61)). The `init()` method assigns both `super.resource` and `this.resource` with the same reference, which is confusing.

Filter: https://sonarcloud.io/project/issues?impactSeverities=BLOCKER&issueStatuses=OPEN%2CCONFIRMED&id=fabric8io_kubernetes-client

**S1845 (logger):**
- Rename `BaseOperation.LOGGER` → `logger` (and the two use sites at lines 783, 789) to align with the project convention.
- Delete `HasMetadataOperation.logger` (only used once at line 328) and reuse the inherited field — both classes are in the same package, so package-private access is fine.

**S2387 (resource):**
- Remove the duplicate `resource` field in `ExtensibleResourceAdapter` and rely on the parent's, casting locally where the narrower `ExtensibleResource<T>` type is required.
- Alternatively, rename the child field (e.g. `extensibleResource`) to make the two distinct fields unambiguous and update internal uses.

- Suppress both via `@SuppressWarnings("java:S1845")` / `@SuppressWarnings("java:S2387")` — rejected because both are real readability issues, not Sonar false positives.

## Additional context

Identified during a Sonar BLOCKER backlog grooming pass. Companion PR is handling the other BLOCKER findings (false-positive suppressions and an intentional switch fallthrough).

ash-thakur-rh:
## Summary
- Rename `BaseOperation.LOGGER` → `logger` to align with the project convention and remove the shadowing `logger` field in `HasMetadataOperation` (S1845)
- Remove the duplicate `resource` field in `ExtensibleResourceAdapter` that shadows `ResourceAdapter.resource` (S2387), using a private cast helper instead
- Make `ResourceAdapter.resource` `protected` so subclasses in other packages retain access

Closes #7825

## Test plan
- [x] `make quickly` — BUILD SUCCESS
- [x] `ExtensibleResourceAdapterTest` — 1 test, 0 failures
- [x] `BaseOperationTest`, `BaseOperationWatchTest`, `HasMetadataOperationsImplTest` — 15 tests, 0 failures
- [x] `kubernetes-tests` — 1302 tests, 0 failures
- [x] `extensions/volumesnapshot/tests` — 3 tests, 0 failures
- [x] `spotless:check` — formatting clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Fix #7765: (kubernetes-client) `BaseOperation.informOnCondition` now stops the informer inline when the inner predicate completes the future, closing a CompletableFuture `postComplete` race where a waiter helping drain dependents could fire `informer.stop` after `cf.complete` had already triggered a spurious `?watch=true` HTTP request
 
 #### Improvements
+* Fix #7825: (kubernetes-client, kubernetes-client-api) Remove duplicated logger and resource fields (Sonar S1845, S2387)
 * Fix #7675: (mockwebserver) `MockWebServer.dispatcher` field marked `volatile` so a `setDispatcher(...)` call is reliably visible to the Vert.x request handler thread without further synchronization. `MockWebServer.reset()` Javadoc tightened to make its non-destructive contract explicit (no change to the running server, dispatcher, listeners, SSL/TLS state, port, or protocols)
 * Fix #7809: (kubernetes-client) Support for shard selectors for list and watch - including informers
 * Fix #7837: (kubernetes-client) Follow-ups on shard selector
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/extension/ExtensibleResourceAdapter.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/extension/ExtensibleResourceAdapter.java
@@ -30,7 +30,6 @@ import java.util.function.Function;
  */
 public abstract class ExtensibleResourceAdapter<T> extends ResourceAdapter<T> implements ExtensibleResource<T> {
 
-  protected ExtensibleResource<T> resource;
   protected Client client;
 
   public ExtensibleResourceAdapter() {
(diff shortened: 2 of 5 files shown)


manusa (changes requested):
A couple of things before this is good to go.

**`ResourceAdapter.java:61`, is the `protected` actually needed?** `ExtensibleResourceAdapter` lives in the same package, so package-private would already cover the `this.resource = resource` at `ExtensibleResourceAdapter.java:42` and the cast in `extensibleResource()`. I scanned the current out-of-package subclasses: `NamespaceableResourceAdapter` goes through `super(op.resource(item))` at line 47, the volumesnapshot adapters only override `newInstance()`, and the test fixture doesn't touch the field either. None of them read or write `resource` directly.

The slight worry with widening to `protected` is that this is a published abstract class, so any downstream extension can now assign a non-`ExtensibleResource` `Resource<T>` and the cast in `extensibleResource()` will blow up at runtime. The previous typed field in `ExtensibleResourceAdapter` ruled that out by declaration. If there's a known external consumer that needs this, could you call it out? Otherwise I'd lean toward keeping it package-private.

**Drive-by nit on `BaseOperation.java:784`:** while you're touching that line, `"Websocket hanshake failed"` could become `"Websocket handshake failed"`. Pre-existing and totally optional, just feels cheap to fix in the same diff.

ash-thakur-rh:
@manusa Both points addressed in 4b88158:

1. **`ResourceAdapter.resource`** reverted to package-private — agreed, `ExtensibleResourceAdapter` is in the same package so `protected` is unnecessary, and widening would let downstream subclasses assign a plain `Resource` that breaks the cast in `extensibleResource()`.
2. **Typo** fixed: `"hanshake"` → `"handshake"`.

manusa (approved):
LGTM, thx!

manusa:
These fixes shouldn't include a changelog entry since they are not user-facing. I'll remove the entry myself in main once merged, but make sure not to include them in upcoming PRs.

Issue #7847, opened 2026-05-20
[SONAR] Miscellaneous reliability findings (S899, S3064, S2445, S5998, S2637, S2583)


ash-thakur-rh:
## Is your task related to a problem? Please describe

Remaining miscellaneous SonarCloud reliability findings on production code, split out from #7832 to keep the S2142 `InterruptedException` fixes in a separate PR.

Filter: https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&types=BUG&inNewCodePeriod=true&id=fabric8io_kubernetes-client

### Findings

- **java:S899** [`junit/mockwebserver/.../MockWebServer.java:348`](https://github.com/fabric8io/kubernetes-client/blob/main/junit/mockwebserver/src/main/java/io/fabric8/mockwebserver/MockWebServer.java#L348) — discarded return value of `delete`
- **java:S899** [`junit/mockwebserver/.../MockWebServer.java:349`](https://github.com/fabric8io/kubernetes-client/blob/main/junit/mockwebserver/src/main/java/io/fabric8/mockwebserver/MockWebServer.java#L349) — discarded return value of `delete`
- **java:S3064** [`kubernetes-client-api/.../Serialization.java:90`](https://github.com/fabric8io/kubernetes-client/blob/main/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/Serialization.java#L90) — `YAML_MAPPER` not fully initialized before assignment (publishes a partially-constructed object)
- **java:S2445** [`kubernetes-client/.../BaseClient.java:415`](https://github.com/fabric8io/kubernetes-client/blob/main/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/impl/BaseClient.java#L415) — synchronizes on a method parameter (callers can break mutual exclusion)
- **java:S5998** [`kubernetes-client-api/.../KubernetesResourceUtil.java:62`](https://github.com/fabric8io/kubernetes-client/blob/main/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesResourceUtil.java#L62) — regex with repetition that can stack-overflow on large inputs (potential ReDoS)
[...]

PR #7848, opened 2026-05-20
fix(sonar): re-interrupt in production-code S2142 catch blocks

### LINKED ISSUE #7832: [SONAR] Re-interrupt or rethrow InterruptedException in production code (S2142 + misc)

Remaining SonarCloud reliability findings on production code after the quick-win companion PR landed.

Filter: https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&types=BUG&inNewCodePeriod=true&id=fabric8io_kubernetes-client

### S2142 — `InterruptedException` caught without re-interrupt/rethrow

When `InterruptedException` is caught (directly or via a parent like `Exception`), the handler should either rethrow the exception or call `Thread.currentThread().interrupt()` to preserve the interrupt status. Failing to do so leaves callers unable to detect that the thread was asked to stop.

- [`junit/kube-api-test/core/.../EtcdProcess.java:120`](https://github.com/fabric8io/kubernetes-client/blob/main/junit/kube-api-test/core/src/main/java/io/fabric8/kubeapitest/process/EtcdProcess.java#L120)
- [`junit/kube-api-test/core/.../KubeAPIServerProcess.java:124`](https://github.com/fabric8io/kubernetes-client/blob/main/junit/kube-api-test/core/src/main/java/io/fabric8/kubeapitest/process/KubeAPIServerProcess.java#L124)
- [`kubernetes-client/.../PodOperationsImpl.java:496`](https://github.com/fabric8io/kubernetes-client/blob/main/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/core/v1/PodOperationsImpl.java#L496)
- [`kubernetes-client-api/.../Utils.java:177`](https://github.com/fabric8io/kubernetes-client/blob/main/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/Utils.java#L177)
- [`kubernetes-client/.../BaseOperation.java:642`](https://github.com/fabric8io/kubernetes-client/blob/main/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/BaseOperation.java#L642)
- [`openshift-client/.../TemplateOperationsImpl.java:128`](https://github.com/fabric8io/kubernetes-client/blob/main/openshift-client/src/main/java/io/fabric8/openshift/client/dsl/internal/core/TemplateOperationsImpl.java#L128)
- [`kubernetes-client/.../HasMetadataOperation.java:185`](https://github.com/fabric8io/kubernetes-client/blob/main/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/HasMetadataOperation.java#L185)
- [`kubernetes-client/.../HasMetadataOperation.java:232`](https://github.com/fabric8io/kubernetes-client/blob/main/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/HasMetadataOperation.java#L232)

### Miscellaneous reliability findings

- **java:S899** [`junit/mockwebserver/.../MockWebServer.java:348`](https://github.com/fabric8io/kubernetes-client/blob/main/junit/mockwebserver/src/main/java/io/fabric8/mockwebserver/MockWebServer.java#L348) — discarded return value of `delete`
- **java:S899** [`junit/mockwebserver/.../MockWebServer.java:349`](https://github.com/fabric8io/kubernetes-client/blob/main/junit/mockwebserver/src/main/java/io/fabric8/mockwebserver/MockWebServer.java#L349) — discarded return value of `delete`
- **java:S3064** [`kubernetes-client-api/.../Serialization.java:90`](https://github.com/fabric8io/kubernetes-client/blob/main/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/Serialization.java#L90) — `YAML_MAPPER` not fully initialized before assignment (publishes a partially-constructed object)
- **java:S2445** [`kubernetes-client/.../BaseClient.java:415`](https://github.com/fabric8io/kubernetes-client/blob/main/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/impl/BaseClient.java#L415) — synchronizes on a method parameter (callers can break mutual exclusion)
- **java:S5998** [`kubernetes-client-api/.../KubernetesResourceUtil.java:62`](https://github.com/fabric8io/kubernetes-client/blob/main/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/KubernetesResourceUtil.java#L62) — regex with repetition that can stack-overflow on large inputs (potential ReDoS)
- **java:S2637** [`httpclient-okhttp/.../OkHttpClientBuilderImpl.java:79`](https://github.com/fabric8io/kubernetes-client/blob/main/httpclient-okhttp/src/main/java/io/fabric8/kubernetes/client/okhttp/OkHttpClientBuilderImpl.java#L79) — possible `null` passed to a `@NotNull` parameter
- **java:S2583** [`httpclient-okhttp/.../OkHttpClientImpl.java:260`](https://github.com/fabric8io/kubernetes-client/blob/main/httpclient-okhttp/src/main/java/io/fabric8/kubernetes/client/okhttp/OkHttpClientImpl.java#L260) — condition always evaluates to `true` (dead branch or logic bug)

**S2142** — In each catch block that may catch `InterruptedException` (directly or via `Exception`), either:
- rethrow the exception, OR
- call `Thread.currentThread().interrupt();` before continuing.

The pattern used in the [companion PR](https://github.com/fabric8io/kubernetes-client/pull/XXXX) (catching `InterruptedException` as a separate catch and calling `interrupt()`) is a clean fit for these production-code sites too.

**Misc** — Each requires independent analysis:
- **S5998** is the highest-risk: `KubernetesResourceUtil` parses arbitrary user input, so a stack-overflow regex is a real DoS surface worth addressing first.
- **S2445** in `BaseClient` is a real concurrency bug — synchronizing on a parameter is broken because callers can pass the same reference and break mutual exclusion. Fix by introducing a private lock object.
- **S3064** in `Serialization` should be reviewed to confirm whether the partially-initialized `YAML_MAPPER` is observable from other threads (the JLS allows it for non-final fields).
- **S2583** in `OkHttpClientImpl` is either dead code or a real logic bug; needs a careful read.

- Splitting into smaller per-rule issues — would create issue churn for what is fundamentally one cleanup pass. Keeping them grouped makes it easier to do the work in a single PR.

## Additional context

Identified during Sonar reliability backlog grooming. Companion PR cleared 7 quick-win reliability bugs in `kubernetes-examples/*` and `LeaderElector`. The `volatile`/`S3077` cluster is tracked separately in #7831.

ash-thakur-rh:
## Summary
- Preserve thread interrupt status at 8 production-code sites where `InterruptedException` was caught without calling `Thread.currentThread().interrupt()` (SonarCloud rule S2142)
- **EtcdProcess / KubeAPIServerProcess**: true bug — interrupt status was silently lost when rethrowing as `KubeAPITestException`
- **Utils, BaseOperation, PodOperationsImpl, TemplateOperationsImpl, HasMetadataOperation (×2)**: add a dedicated `catch (InterruptedException)` arm ahead of the broad `catch (Exception)` so the intent is explicit and does not rely on `launderThrowable`'s internal handling
- Remaining misc reliability findings (S899, S3064, S2445, S5998, S2637, S2583) tracked in #7847

## Test plan
- [x] Regression test `UtilsTest#waitUntilReadyPreservesInterruptStatus` verifies interrupt flag is preserved after `InterruptedException`
- [x] All `kubernetes-client-api` tests pass (634 tests, 0 failures)
- [x] All `kubernetes-client` tests pass (272 tests, 0 failures — 1 pre-existing failure in `UploadTest.bigNumbersSupported` unrelated to this change)
- [x] All `openshift-client` tests pass (47 tests, 0 failures)
- [x] `spotless:check` passes on all affected modules

Closes #7832


manusa (commented):
## Review

### Strengths

- **Scope matches issue #7832 line-for-line.** All 8 S2142 sites from the issue are patched (`EtcdProcess:120`, `KubeAPIServerProcess:124`, `Utils:177`, `BaseOperation:642`, `HasMetadataOperation:185` and `:232`, `PodOperationsImpl:496`, `TemplateOperationsImpl:128`). Misc findings explicitly deferred to #7847, which exists.
- **Idiom is consistent.** Every new arm calls `Thread.currentThread().interrupt()` before rethrowing via `KubernetesClientException.launderThrowable(...)`, matching the existing pattern in `HasMetadataOperation.java:195-199`.
- **Catch ordering is correct everywhere.** Each `catch (InterruptedException)` is placed before the broad `catch (Exception)`. In `Utils.waitUntilReady`, the order (`TimeoutException` → `ExecutionException` → `InterruptedException` → `Exception`) is valid (no subtype overlap among the specifics), and unchecked `CancellationException` still routes to the broad arm.
- **No drive-by edits.** Only the CHANGELOG entry, the 8 catch blocks, and a single `java.util.concurrent.Future` import in the test file.
- **Regression test exercises the actual fix path** (`UtilsTest:351-393`): a `Future` whose `get(long, TimeUnit)` throws `InterruptedException` hits the new arm at `Utils.java:177-179` exactly. Reverting the production change flips the post-condition.

### Issues

#### Critical (Must Fix)

None.

#### Important (Should Fix)

1. **`UtilsTest#waitUntilReadyPreservesInterruptStatus` can poison sibling tests on regression.** `kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/internal/UtilsTest.java:386-390`. The `finally` block asserts `isInterrupted().isTrue()` **before** clearing via `Thread.interrupted()`. If the fix regresses, the AssertJ failure throws `AssertionError` and the `Thread.interrupted()` cleanup never runs, leaving the runner thread interrupted. `kubernetes-client-api/pom.xml:229-230` sets `forkCount=1C, reuseForks=true`, so subsequent tests in the same fork would observe an interrupted thread and could fail spuriously. Fix:
   ```java
   } finally {
     boolean wasInterrupted = Thread.interrupted(); // clear first
     assertThat(wasInterrupted)
         .as("interrupt status should be preserved after InterruptedException is caught")
         .isTrue();
   }
   ```

2. **Test doesn't bind to the actual exception type.** Same file, `:381-383` swallows the thrown exception as `// expected` without asserting it's `KubernetesClientException` wrapping `InterruptedException`. The test would still pass if `waitUntilReady` regressed to throw an unrelated exception, as long as the interrupt bit happened to be set. Suggest combining with #1 above using `assertThatThrownBy`:
   ```java
   assertThatThrownBy(() -> Utils.waitUntilReady(interruptingFuture, 10, TimeUnit.SECONDS))
       .isInstanceOf(KubernetesClientException.class)
       .hasCauseInstanceOf(InterruptedException.class);
   boolean wasInterrupted = Thread.interrupted();
   assertThat(wasInterrupted).isTrue();
   ```

#### Minor (Nice to Have)

3. **The explicit `interrupt()` calls are redundant with `launderThrowable`, but worth keeping.** `kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/KubernetesClientException.java:117-118` already calls `Thread.currentThread().interrupt()` when the cause is `InterruptedException`. The new explicit calls at each catch site are technically duplicative, but: (a) Sonar S2142 can't see through the indirection, so the explicit form is required to clear the finding; (b) being explicit at the catch site is the standard idiom and avoids coupling to internal helper behavior. Net: keep as-is, but a brief note in the PR description acknowledging this would help future readers.

4. **`BaseOperation.deleteAll` intentionally bypasses the 404 / HTTP_BAD_METHOD fallback on interrupt.** `kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/BaseOperation.java:642-644`. This is correct (the 405 fallback re-issues HTTP via `list()` + per-item `delete()`, which is the wrong behavior on an already-interrupted thread; the 404 short-circuit is irrelevant since `InterruptedException` has no HTTP code). Worth a one-line inline comment to forestall a future "fix" that merges the arms.

5. **Pre-condition `assertThat(Thread.currentThread().isInterrupted()).isFalse()` at `UtilsTest:380` is defensible but fragile.** No sibling in `UtilsTest` currently leaks the flag onto the runner thread, so it works today. A `Thread.interrupted()` at entry (without asserting) would be more robust and is not what the test is contractually verifying.

6. **PR description claim "true bug" for `EtcdProcess` / `KubeAPIServerProcess` is slightly overstated.** Impact is bounded to test cleanup paths in `kube-api-test`. Reliability fix, but not a production-runtime defect. Cosmetic.

### Assessment: Ready to merge? With fixes

Production changes are correct and minimal. The two Important findings are test-only and easily addressed (reorder cleanup, tighten exception assertion). Worth getting #1 in before merge to prevent the test from poisoning fork-mates on regression; #2 strengthens the regression guarantee and folds cleanly into the same edit.

manusa (changes requested):
Formalizing the two **Important** test-quality concerns from my [prior comment review](https://github.com/fabric8io/kubernetes-client/pull/7848#pullrequestreview-4328530006) as a blocker. Production fix is correct, scope matches #7832 exactly, all 8 S2142 sites addressed. The blockers are test-only:

1. **`UtilsTest#waitUntilReadyPreservesInterruptStatus` can poison sibling tests on regression.** `kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/internal/UtilsTest.java:386-390` asserts `isInterrupted().isTrue()` **before** clearing via `Thread.interrupted()`. Combined with `forkCount=1C, reuseForks=true` (`kubernetes-client-api/pom.xml:229-230`), an `AssertionError` from a regression would leave the runner thread interrupted for every subsequent test in the same fork. Clear first, then assert on the captured boolean.

2. **Test doesn't bind to the actual exception type.** Same file, `:381-383` swallows the thrown exception as `// expected` without asserting it's `KubernetesClientException` wrapping `InterruptedException`. Folding into #1 with `assertThatThrownBy(...).isInstanceOf(KubernetesClientException.class).hasCauseInstanceOf(InterruptedException.class)` is a single-edit fix.

The Minor items in the prior comment are nice-to-have, not blockers.

ash-thakur-rh (commented):
@manusa Both blockers addressed in adc468e:

1. **Fork poisoning**: interrupt flag is now cleared via `Thread.interrupted()` before asserting on the captured boolean — a regression assertion failure no longer leaves the runner thread interrupted.
2. **Weak assertion**: replaced bare `catch (Exception e) { // expected }` with `assertThatThrownBy(...).isInstanceOf(KubernetesClientException.class).hasCauseInstanceOf(InterruptedException.class)`.

Ready for re-review.

manusa (approved):
LGTM, thx!

PR #7851, opened 2026-05-20
fix(sonar): miscellaneous reliability findings (7847)

### LINKED ISSUE #7847: [SONAR] Miscellaneous reliability findings (S899, S3064, S2445, S5998, S2637, S2583)

ash-thakur-rh:
## Summary

Fixes six SonarCloud reliability rules (S899, S3064, S2445, S5998, S2637, S2583) across four modules, addressing [issue #7847](https://github.com/fabric8io/kubernetes-client/issues/7847).

- **S2445** — `BaseClient.addToCloseable` now synchronizes on the internal `closable` set instead of the method parameter, matching the monitor used by `close()`
- **S5998** — `KubernetesResourceUtil.KUBERNETES_SUBDOMAIN_REGEX` uses possessive quantifiers (`?+`, `*+`) on outer groupings to prevent catastrophic backtracking (ReDoS)
- **S3064** — `Serialization.yamlMapper()` uses local-variable double-checked locking to prevent unsafe publication of a partially-constructed `ObjectMapper`
- **S899** — `MockWebServer` uses `Files.deleteIfExists()` instead of `File.delete()` for temp cert/key cleanup, with logged `IOException`
- **S2637** — `OkHttpClientBuilderImpl.initialBuild` guards against null `trustManager` before calling `sslSocketFactory()`, with a warning log when SSL config cannot be applied
- **S2583** — `OkHttpClientImpl.doClose()` removes dead null checks on `dispatcher` and `connectionPool` (always non-null per OkHttp API contract)

## Test plan

- [x] Added tests for S5998: possessive quantifier structural check + valid/invalid subdomain matching
- [x] Added test for S3064: verify `GoCompatibilityModule` is registered after lazy init
- [x] Added test for S2445: concurrency test verifying `addToCloseable` blocks on `closable` lock
- [x] Added tests for S2637: null trustManagers and multi-entry trustManagers avoid NPE
- [x] All affected module tests pass (`kubernetes-client-api`, `kubernetes-client`, `httpclient-okhttp`, `mockwebserver`)
- [x] `mvn spotless:check` passes


manusa (commented):
### Strengths

- All six SonarCloud findings are real reliability issues, and the fixes are mostly correct.
- The **S2445** fix correctly mirrors `close()` by synchronizing on `this.closable` (`kubernetes-client/src/main/java/io/fabric8/kubernetes/client/impl/BaseClient.java:415`). The new test reliably distinguishes the old broken behavior from the fixed one: it holds the `closable` monitor externally and verifies `addToCloseable` blocks. The entire `addToCloseable` body is inside `synchronized (this.closable)`, so there is no residual TOCTOU between the `isDone()` check and `closable.add(...)` against the drain in `close()` (line 161-164).
- The **S3064** fix uses the canonical local-variable DCL idiom; `YAML_MAPPER` is already declared `volatile` (`Serialization.java:48`), so publication is JMM-safe under the patched code (`yamlMapper()` lines 86-97).
- The **S5998** patched regex `[a-z0-9]([-a-z0-9]*[a-z0-9])?+(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?+)*+` is exactly the canonical RFC 1123 subdomain pattern from the Kubernetes [naming docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/), just with possessive outer quantifiers. The 253-char limit matches the existing `name.length() < 254` gate (`KubernetesResourceUtil.java:371`). Possessive `?+` and `*+` on the outer groups are semantically identical to greedy `?`/`*` here because the inner `[-a-z0-9]*[a-z0-9]` cannot consume `.` (not in its char class), so the outer groups never need to give back matched content to let the dot-segment loop proceed. Behavior preserved against valid and invalid K8s names.

### Issues

#### Critical (Must Fix)

None.

#### Important (Should Fix)

1. **S2637 silent trust-config downgrade.** `httpclient-okhttp/src/main/java/io/fabric8/kubernetes/client/okhttp/OkHttpClientBuilderImpl.java:79-84`. When `trustManagers.length != 1`, the patched code skips `builder.sslSocketFactory(...)` entirely. The user's `sslContext` (which `StandardHttpClientBuilder.sslContext(...)` built from those same `trustManagers` via `SSLUtils.sslContext(km, tm)`) is then silently ignored, and OkHttp falls back to its default `SSLSocketFactory` (system truststore + no client key material). Previously this path threw NPE at build time, surfacing the misconfiguration loudly. The patched behavior turns it into a quiet trust-config downgrade. For an advanced caller wiring a multi-CA `TrustManager[]` (or a length-0 array as an error path), TLS now validates against the JVM default truststore instead of their configured trust anchors, and any custom `keyManagers` for mTLS are dropped. Recommend either throwing a clear `KubernetesClientException` when no single `X509TrustManager` can be selected, or filtering `trustManagers` for the first `X509TrustManager` so the user's `sslContext` is still applied.

2. **S2637 warning message inaccurate.** Same site, `OkHttpClientBuilderImpl.java:82-83`. The text "no suitable X509TrustManager was found" misrepresents the multi-entry case: the user did supply trust managers; the limitation is that OkHttp accepts only one. Suggest differentiating the two cases (none vs multiple) so the log helps the user diagnose what they actually misconfigured.

#### Minor (Nice to Have)

1. **S2583 shutdown order flipped.** `httpclient-okhttp/src/main/java/io/fabric8/kubernetes/client/okhttp/OkHttpClientImpl.java:254-257`. The sequence changed from `cancelAll → evictAll → shutdownNow` to `evictAll → cancelAll → shutdownNow`. `evictAll()` only closes idle connections, so this is functionally equivalent and not a regression, but the PR description framed S2583 as removing dead null checks without mentioning the reorder. Worth a one-line note in the CHANGELOG entry or just preserving the original ordering.

2. **S5998 drop the implementation-detail test.** `kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/utils/KubernetesResourceUtilTest.java:372-380`. `subdomainRegex_shouldUsePossessiveQuantifiers` asserts that the literal pattern contains `?+` or `*+`, which is an implementation detail of how the ReDoS fix is currently expressed. The behavioral tests directly below it (`shouldMatchValidSubdomains`, `shouldRejectInvalidSubdomains`) already pin the contract against the canonical K8s RFC 1123 subdomain spec. Please drop this assertion; any future refactor of the regex (e.g., a non-possessive but linear-time rewrite) should not need a test update.

3. **S899 crash-leak window unchanged.** `junit/mockwebserver/src/main/java/io/fabric8/mockwebserver/MockWebServer.java:317-356`. `deleteOnExit()` plus the new `Files.deleteIfExists()` handle clean exits, but JVM crash / SIGKILL still leaves freshly-generated cert and private-key files in `/tmp`. Pre-existing and arguably out of scope, since the keys are self-signed dev material with no production trust value; flagging it as a follow-up only.

### Assessment: Ready to merge? With fixes

The S2637 silent trust downgrade is the one that needs attention before merge: it converts a noisy NPE into a quiet TLS-trust-config change for the affected callers, which is worse than the original defect. Everything else is correct or minor.

httpclient-okhttp/src/main/java/io/fabric8/kubernetes/client/okhttp/OkHttpClientBuilderImpl.java
```diff
@@ -76,7 +76,12 @@ public OkHttpClientImpl initialBuild(okhttp3.OkHttpClient.Builder builder) {
       if (trustManagers != null && trustManagers.length == 1) {
         trustManager = (X509TrustManager) trustManagers[0];
       }
-      builder.sslSocketFactory(sslContext.getSocketFactory(), trustManager);
+      if (trustManager != null) {
+        builder.sslSocketFactory(sslContext.getSocketFactory(), trustManager);
+      } else {
+        logger.warn("sslContext is configured but no suitable X509TrustManager was found; "
+            + "SSL socket factory not applied, OkHttp will use its default trust configuration");
+      }
```
manusa, line 91: Silent trust-config downgrade in this branch. When `trustManagers.length != 1`, `builder.sslSocketFactory(...)` is skipped entirely, and OkHttp falls back to its default `SSLSocketFactory` (system truststore + no client key material). The user's `sslContext`, which `StandardHttpClientBuilder.sslContext(km, tm)` built from those same `trustManagers` via `SSLUtils.sslContext(km, tm)`, is silently discarded along with any `keyManagers` for mTLS.

Previously this path NPE'd at build time, making the misconfiguration loud. The patched behavior turns it into a quiet TLS-trust change. Suggest either throwing a clear `KubernetesClientException` here, or filtering `trustManagers` for the first `X509TrustManager` so the user's `sslContext` is still applied.

Separately, the warning text "no suitable X509TrustManager was found" is inaccurate when `trustManagers.length > 1`: the caller did supply trust managers; the limitation is that OkHttp accepts only one. Differentiating none-vs-multiple in the log would help triage.

httpclient-okhttp/src/main/java/io/fabric8/kubernetes/client/okhttp/OkHttpClientImpl.java
```diff
@@ -254,22 +252,10 @@ public OkHttpClientImpl(OkHttpClient client, OkHttpClientBuilderImpl builder, At
 
   @Override
   public void doClose() {
```
manusa, line 254: Heads up: the shutdown order flipped from `cancelAll → evictAll → shutdownNow` to `evictAll → cancelAll → shutdownNow`. `evictAll()` only touches idle connections, so this is functionally equivalent, but the PR description framed S2583 as dead-code removal and didn't mention the reorder. Either restore the original order or call it out in the CHANGELOG entry.

kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/utils/KubernetesResourceUtilTest.java
```diff
@@ -369,6 +369,41 @@ private final void assertConfigMapContainsData(ConfigMap configMap, MapEntry<Str
         .contains(mapEntries);
   }
 
+  @Test
+  @DisplayName("S5998: KUBERNETES_SUBDOMAIN_REGEX should use possessive quantifiers to prevent ReDoS")
+  void subdomainRegex_shouldUsePossessiveQuantifiers() {
+    String pattern = KubernetesResourceUtil.KUBERNETES_SUBDOMAIN_REGEX.pattern();
+    assertThat(pattern)
+        .as("Regex should use possessive quantifiers (?+ or *+) to prevent catastrophic backtracking")
+        .containsAnyOf("?+", "*+");
```
manusa: Please drop this test. It asserts on the literal pattern string containing `?+` or `*+`, which couples it to an implementation detail of how the ReDoS fix is currently expressed. The behavioral tests directly below already pin the contract against the canonical K8s RFC 1123 subdomain spec, and a future non-possessive but linear-time rewrite of the regex should not need a test update.

ash-thakur-rh:
@manusa Thank you for the thorough review. I've addressed all your feedback:

1. **S2637 trust-config downgrade** — Instead of requiring exactly one `TrustManager`, the code now iterates to find the first `X509TrustManager` from the array, so multi-entry arrays work correctly. The warning message also differentiates between null and non-null `trustManagers` for easier diagnostics.

2. **S2583 shutdown order** — Restored the original `cancelAll → evictAll → shutdownNow` sequence.

3. **S5998 implementation-detail test** — Dropped the `subdomainRegex_shouldUsePossessiveQuantifiers` test; the behavioral tests for valid/invalid subdomains remain.

Ready for re-review when you have a chance.

manusa (approved):
LGTM, thx!

PR #7877, opened 2026-05-28
fix(sonar): verify chaos-mesh install.sh checksum before executing

### LINKED ISSUE #7829: [SONAR] Verify chaos-mesh install.sh before piping to bash in e2e workflow

SonarCloud HIGH-severity finding **githubactions:S8482** — *"Avoid executing downloaded artifacts directly without verification."*

- [`.github/workflows/e2e-chaos-tests.yml:70`](https://github.com/fabric8io/kubernetes-client/blob/main/.github/workflows/e2e-chaos-tests.yml#L70)

```yaml
- name: Install Chaos Mesh
  run: |
    curl -sSL https://mirrors.chaos-mesh.org/v${CHAOS_MESH_VERSION}/install.sh | bash
```

The workflow downloads `install.sh` from `mirrors.chaos-mesh.org` and pipes it directly into `bash` without verifying its integrity. If the mirror were compromised, CI would execute arbitrary code with whatever permissions the GitHub Actions runner has at that step (including `secrets.GITHUB_TOKEN` and any cached credentials).

The risk is bounded — this workflow runs only on the chaos-tests matrix, not on every PR — but the pattern is exactly what S8482 is designed to catch. It matches the upstream chaos-mesh project's official install instructions, so the trust assumption is reasonable, but verification is cheap.

Filter: https://sonarcloud.io/project/issues?impactSoftwareQualities=SECURITY&issueStatuses=OPEN%2CCONFIRMED&id=fabric8io_kubernetes-client

Pin a SHA256 of `install.sh` per chaos-mesh version, verify before executing. Something like:

```yaml
- name: Install Chaos Mesh
  env:
    CHAOS_MESH_INSTALL_SHA256: "<sha256-pinned-per-CHAOS_MESH_VERSION>"
  run: |
    curl -sSL "https://mirrors.chaos-mesh.org/v${CHAOS_MESH_VERSION}/install.sh" -o install.sh
    echo "${CHAOS_MESH_INSTALL_SHA256}  install.sh" | sha256sum -c -
    bash install.sh
```

The SHA256 would need to be bumped whenever `CHAOS_MESH_VERSION` is bumped. A small workflow comment near the constant should make this obvious to future contributors.

- **Vendor `install.sh` into the repo** under `.github/workflows/scripts/` and execute the vendored copy. Pro: no network dependency at install time. Con: drifts from upstream, requires manual sync on version bumps.
- **Suppress via SonarCloud "Safe" marker** with a comment in the workflow file. Pro: zero CI risk. Con: the underlying supply-chain concern remains; the next contributor has nothing to point at.

## Additional context

Last of the 14 open HIGH-severity security findings on the project. The other 13 (all `java:S4830` / `java:S5527` SSL/TLS findings on opt-in `trustCerts` code paths and on `kube-api-test` localhost test infrastructure) are being handled as in-code `@SuppressWarnings` with comments in a companion PR.

ash-thakur-rh:
## Summary
- Addresses SonarCloud HIGH-severity finding **S8482** ("Avoid executing downloaded artifacts directly without verification")
- Downloads `install.sh` to a file, verifies its SHA256 checksum, then executes it — instead of piping `curl` directly to `bash`
- Pins `CHAOS_MESH_INSTALL_SHA256` env var with a comment reminding contributors to bump the hash when bumping `CHAOS_MESH_VERSION`

Fixes #7829

## Test plan
- [x] Verify YAML is syntactically valid
- [x] Confirm SHA256 matches live download of `mirrors.chaos-mesh.org/v2.6.0/install.sh`
- [x] Trigger `e2e-chaos-tests` workflow via `workflow_dispatch` to confirm the install step passes with verification


ash-thakur-rh:
workflow for installing chaos mesh was success: https://github.com/ash-thakur-rh/kubernetes-client/actions/runs/26626708682/job/78465130853

manusa (approved):
LGTM, thx!

PR #7898, opened 2026-06-05
fix: emit WatcherException on silent clean close to prevent stale-rv reconnect loop

### LINKED ISSUE #7896: Event watcher reconnects indefinitiely when k8s doesn't send 410 on stale version
### Describe the bug

This is more of a GKE bug, but it leads to unfortunate behaviours on fabric8 client so I'm reporting here too.
There's an open issue on k8s about `api/v1/events?watch=true` closing abruptly when the provided `resourceVersion` is stale instead of returning an error 410: https://github.com/kubernetes/kubernetes/issues/137089 

Since `Watcher<>` relies on the 410 error to invalidate the `resourceVersion` used as a bookmark this means that the version never gets invalidated leading to an infinite loop where:
1. Watcher tries to open a websocket connection
2. k8s kills the connection abruptly without errors
3. goto 1.

This happens only with `api/v1/events` other resources such as `api/v1/jobs` honor the contract and send back the 410 as expected.

7.7.0

### Steps to reproduce

I created a repo to reproduce: https://github.com/tehaci/fabric8-reconnect-bug

1. Run `./mvnw clean compile exec:java -Pevent` to see the bug 
2. Run `./mvnw clean compile exec:java -Pjob` to see the expected behavior
3. Run `./mvnw clean compile exec:java -Pevent -Pfixed` to see a proposed workaround (I patched io.fabric8.kubernetes.client.dsl.internal.AbstractWatchManager with a fix behind `watch.fix.enabled` toggle and a log to show the frame being sent from the API)

### Expected behavior

When k8s "forgets" to send 410 and closes the connection the client should invalidate the `resourceVersion` is currently using instead of retrying forever

### Runtime

1.35

### Environment

macOS

```shell
11:05:00.560 [com.github.tehaci.EventWatcherReconnectLoopBug.main()] DEBUG Config - Trying to configure client from Kubernetes config...
11:05:03.907 [com.github.tehaci.EventWatcherReconnectLoopBug.main()] DEBUG HttpClientUtils - Using httpclient io.fabric8.kubernetes.client.vertx.VertxHttpClientFactory factory
11:05:03.916 [com.github.tehaci.EventWatcherReconnectLoopBug.main()] DEBUG LoggerFactory - Using io.vertx.core.logging.SLF4JLogDelegateFactory
11:05:04.101 [com.github.tehaci.EventWatcherReconnectLoopBug.main()] INFO EventWatcherReconnectLoopBug - Connected: https://<...>europe-west1.gke.goog/
11:05:04.101 [com.github.tehaci.EventWatcherReconnectLoopBug.main()] INFO EventWatcherReconnectLoopBug - Step 1: opening fresh watch...
11:05:04.112 [com.github.tehaci.EventWatcherReconnectLoopBug.main()] DEBUG AbstractWatchManager - Watching https://<...>.europe-west1.gke.goog/api/v1/namespaces/default/events?allowWatchBookmarks=true&fieldSelector=involvedObject.kind%3DJob&resourceVersion=1&watch=true...
11:05:04.669 [vert.x-eventloop-thread-1] DEBUG Http1xClientConnection - WebSocket handshake complete
11:05:04.671 [vert.x-eventloop-thread-1] DEBUG WatcherWebSocketListener - WebSocket successfully opened
11:05:04.673 [vert.x-eventloop-thread-1] DEBUG WatcherWebSocketListener - WebSocket close received. code: 1000, reason: null
11:05:04.673 [vert.x-eventloop-thread-1] DEBUG AbstractWatchManager - Scheduling reconnect task in 1000 ms
11:05:05.681 [-782539754-pool-1-thread-1] DEBUG AbstractWatchManager - Watching https://<...>.europe-west1.gke.goog/api/v1/namespaces/default/events?allowWatchBookmarks=true&fieldSelector=involvedObject.kind%3DJob&resourceVersion=1&watch=true...
11:05:05.953 [vert.x-eventloop-thread-2] DEBUG Http1xClientConnection - WebSocket handshake complete
11:05:05.953 [vert.x-eventloop-thread-2] DEBUG WatcherWebSocketListener - WebSocket successfully opened
11:05:05.957 [vert.x-eventloop-thread-2] DEBUG WatcherWebSocketListener - WebSocket close received. code: 1000, reason: null
11:05:05.957 [vert.x-eventloop-thread-2] DEBUG AbstractWatchManager - Scheduling reconnect task in 1000 ms
11:05:06.959 [-782539754-pool-1-thread-1] DEBUG AbstractWatchManager - Watching https://<...>.europe-west1.gke.goog/api/v1/namespaces/default/events?allowWatchBookmarks=true&fieldSelector=involvedObject.kind%3DJob&resourceVersion=1&watch=true...
11:05:07.227 [vert.x-eventloop-thread-2] DEBUG Http1xClientConnection - WebSocket handshake complete
11:05:07.228 [vert.x-eventloop-thread-2] DEBUG WatcherWebSocketListener - WebSocket successfully opened
11:05:07.231 [vert.x-eventloop-thread-2] DEBUG WatcherWebSocketListener - WebSocket close received. code: 1000, reason: null
11:05:07.231 [vert.x-eventloop-thread-2] DEBUG AbstractWatchManager - Scheduling reconnect task in 1000 ms
11:05:08.242 [-782539754-pool-1-thread-1] DEBUG AbstractWatchManager - Watching https://<...>.europe-west1.gke.goog/api/v1/namespaces/default/events?allowWatchBookmarks=true&fieldSelector=involvedObject.kind%3DJob&resourceVersion=1&watch=true...
11:05:08.526 [vert.x-eventloop-thread-2] DEBUG Http1xClientConnection - WebSocket handshake complete
11:05:08.526 [vert.x-eventloop-thread-2] DEBUG WatcherWebSocketListener - WebSocket successfully opened
11:05:08.528 [vert.x-eventloop-thread-2] DEBUG WatcherWebSocketListener - WebSocket close received. code: 1000, reason: null
11:05:08.528 [vert.x-eventloop-thread-2] DEBUG AbstractWatchManager - Scheduling reconnect task in 1000 ms
```

### Additional context

Observed on GKE v1.35.3-gke.138900
The bug is not present on minikube v1.38.1

tehaci:
## Description

Fixes #7896.

On GKE, `v1/events` watches with a stale `resourceVersion` are rejected by the GKFE proxy with a bare WebSocket close (code 1000, no body) instead of the standard `{"type":"ERROR","code":410}` response. Without the 410 signal, `AbstractWatchManager` reconnects with the same stale `resourceVersion` indefinitely.

### Root cause

Two compounding issues:
- **Bug A (GKE/upstream):** the GKFE proxy silently closes the WebSocket instead of sending `410 Gone` for `v1/events`. See [k/k#137089](https://github.com/kubernetes/kubernetes/issues/137089).
- **Bug B (fabric8):** `AbstractWatchManager` only resets `resourceVersion` on an explicit `410` — reconnecting with the same stale rv after a clean close is otherwise _correct_ behaviour (legitimate idle disconnects should resume from last known position).

### Fix

In `watchEnded()`, detect a clean close (no error, no messages received) that happened within 2 seconds of the connection being established, and reset `resourceVersion` to `null`. This forces the next reconnect to re-list rather than retry the stale version.

The 2-second bound distinguishes GKE's immediate rejection (~100ms) from legitimate idle-period disconnects, which would have been alive for much longer.

## Type of change

## Checklist

- [x] I have added/updated the javadocs and other documentation accordingly
- [x] I tested my code in Kubernetes (GKE v1.35.3-gke.1389002 — the environment where the bug manifests)
- [x] I tested my code in OpenShift (OpenShift local)

### Evidence

A reproduction PoC is available at https://github.com/tehaci/fabric8-reconnect-bug. It includes:
- `EventWatcherReconnectLoopBug` — reproduces the infinite reconnect loop on GKE
- `JobWatcherControlCase` — control case showing correct behaviour on `batch/v1/jobs` (server sends proper 410)

The integration test (`WatchIT#watchWithStaleResourceVersionRecovers`) was validated against GKE:
- **Without fix:** times out after 30s
- **With fix:** passes in ~3.5s


shawkins:
Thank you for the pr @tehaci. The changes look good - or at least it's a contained way of working around an unfortunate situation. It's not great that the upstream bug, and a similar one before it, don't seem to be getting any traction. It would be nice to at least have an acknowledgement on what flavors / versions this is happening on.

tehaci:
@shawkins currently I have this problem on GKE v1.35.3-gke.1389002 and only appears to happen on the `api/v1/events` my services watches `api/v1/jobs` too and I get the 410 there. Should I update the comment in the code or just the PR description?

shawkins:
@tehaci no, your comment is good - what I mean is that the upstream kubernetes issues should ideally be triaged more.

shawkins (changes requested):
@tehaci please change the logic to emit an http gone watcher exception similar to what https://github.com/fabric8io/kubernetes-client/blob/5613b7a6d2c9005d8ec76a5aa2a97bd22c3d4adb/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java#L481 is doing for ProtocolExceptions. Resetting the last sync version will lead to events that are not expected.

tehaci:
@shawkins done, closing with an exception. btw i don't think there's a risk of relisting when we take this approach, from what i've seen the resourceVersion gets stale when there's no activity in the watcher resulting on the version not being updated

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
```diff
@@ -478,6 +488,20 @@ void watchEnded(Throwable t, WatchRequestState state) {
...
+    // GKE-specific behaviour on v1/events where the GKFE proxy sends a bare close instead of
+    // the standard {"type":"ERROR","code":410} response, causing the watch to loop indefinitely
+    // with the same stale resourceVersion. Emitting a WatcherException mirrors the ProtocolException
+    // path and lets the caller (e.g. an informer) decide how to recover.
+    // See: https://github.com/kubernetes/kubernetes/issues/137089
+    long connectionAgeMs = System.currentTimeMillis() - state.startedAtMs;
+    if (t == null && !state.messageReceived.get() && connectionAgeMs < STALE_RV_CLOSE_THRESHOLD_MS) {
+      logger.debug("Watch ended cleanly with no messages after {}ms — treating as stale resourceVersion rejection",
+          connectionAgeMs);
+      close(new WatcherException("resourceVersion too old"));
```
shawkins: Could you fully mimic a 410 WatcherException here? That would lead to the consistent graceful handling.

tehaci: ok should be done @shawkins

manusa (changes requested):
Thanks for the update, switching to the http-gone exception is the right call and informers will relist as expected now.

There's one blocker before this can go in: CI is fully red, and it's caused by this change rather than being a flake or something unrelated. Every pipeline fails on an existing test, `WatchTest.testErrorAfterReady`, where `assertTrue(latch.await(...))` at line 584 returns false.

Here's what's happening. That test's first watch response is `.andUpgradeToWebSocket().open().done()`, which is an ordinary clean close (code 1000) with no frames, happening within milliseconds of connecting. It then expects the watch to reconnect through a following 503 and finally receive a MODIFIED event. Your new branch in `watchEnded` matches that exact shape (`t == null`, no messages, age under 2s) and force-closes the watch, so it never reconnects and the latch times out. The heuristic currently can't tell GKE's stale-rv silent close apart from a normal immediate clean close, and reconnecting on a clean close is exactly the contract that test protects.

The smallest fix that keeps your GKE handling and turns CI green is to only apply the branch when there is actually a resourceVersion in play, since a stale-resourceVersion rejection is meaningless without one:

```java
if (t == null && resourceVersion.get() != null && !state.messageReceived.get()
    && connectionAgeMs < STALE_RV_CLOSE_THRESHOLD_MS) {
```

`testErrorAfterReady` starts with no resourceVersion, so it will reconnect as before, while the GKE case (which always has a stale rv set) still triggers. Please leave `testErrorAfterReady` itself unchanged, the behaviour it asserts is correct.

One knock-on: your `watchEndedEmitsWatcherExceptionOnQuickCleanClose` test will now skip the branch and fail unless a resourceVersion is set, so give it one to represent the GKE scenario:

```java
awm.resourceVersion.set("12345"); // a stale resourceVersion is in use, as on GKE
awm.watchEnded(null, state);
```

It's worth doing the same `set(...)` in `watchEndedSchedulesReconnectOnOldCleanClose` and `watchEndedSchedulesReconnectWhenMessagesReceived` so they keep exercising the age and message guards instead of passing only because the resourceVersion is null.

manusa (approved):
LGTM, thx!

tehaci:
thank you!

tehaci:
It's me again, I bumped fabric8 version and the problem was partially addressed:
- if i use a watcher with `allowWatchBookmarks=false` the reconnect loop is mitigated
- if I use an informer which hardcodes `allowWatchBookmarks=true` the workaround is invalidated because it treats the bookmark we get when we open the websocket gets treated as a message and don't invaldate the bookmark if we don't receive any actual messages. 

I don't know how I missed this during my testing, but the fix should be easy. [Here](https://github.com/fabric8io/kubernetes-client/pull/7898/changes#diff-7ca5806d751b9d76ce33e1adbef2dabd2ef3c10d499aeb66215451bd8bbff802R405) I should have excluded bookmark messages, should I open a separate issue to work on this?

shawkins:
> if I use an informer which hardcodes allowWatchBookmarks=true the workaround is invalidated because it treats the bookmark we get when we open the websocket gets treated as a message and don't invaldate the bookmark if we don't receive any actual messages.

This seems very odd - what is the resourceVersion on that bookmark? Is it the same as what was used to start the watch?

tehaci:
@shawkins maybe i was too hasty, I tried to replicate again and now behaves as expected

```
08:31:39.644 [-428709297-pool-11-thread-8] DEBUG io.fabric8.kubernetes.client.dsl.internal.AbstractWatchManager {} - Closing the current watch
08:31:39.668 [vert.x-eventloop-thread-5] DEBUG io.fabric8.kubernetes.client.dsl.internal.WatcherWebSocketListener {} - WebSocket close received. code: 1000, reason: null
08:31:39.669 [vert.x-eventloop-thread-5] DEBUG io.fabric8.kubernetes.client.dsl.internal.AbstractWatchManager {} - Scheduling reconnect task in 1000 ms
08:31:40.673 [-428709297-pool-11-thread-8] DEBUG io.fabric8.kubernetes.client.dsl.internal.AbstractWatchManager {} - Watching <redacted>.europe-west1.gke.goog/api/v1/namespaces/<redacted>/events?allowWatchBookmarks=true&resourceVersion=1782893984990111023&timeoutSeconds=600&watch=true...
08:31:40.918 [vert.x-eventloop-thread-7] DEBUG io.fabric8.kubernetes.client.dsl.internal.WatcherWebSocketListener {} - WebSocket successfully opened
08:31:40.922 [vert.x-eventloop-thread-7] DEBUG io.fabric8.kubernetes.client.dsl.internal.WatcherWebSocketListener {} - WebSocket close received. code: 1000, reason: null
08:31:40.923 [vert.x-eventloop-thread-7] DEBUG io.fabric8.kubernetes.client.dsl.internal.AbstractWatchManager {} - Watch ended cleanly with no messages after 249ms — treating as stale resourceVersion rejection
```
i think my ide didn't pick up the dependency bump

PR #7899, opened 2026-06-08
feat(informers): callback before re-list for Informers

csviri:
## Description

Introducing a callback before re-list happens in Informers. Note that we have already a callback when a list is finished:

```java
   
  // before list happening
  default void onBeforeList(String lastSyncResourceVersion) {
  
  }

  // after list finished
  default void onList(String resourceVersion, boolean remainedEmpty) {
    // ...
  }

```

This is both for the WatchList and former pure list approach.

The rational behind this that, if a list happens in informers there might be events lost. So let's say we have a resource  in already in the caches in the informer, if we start to do a re-list because watch connection broke (and cannot continue continue watch from the last resource version). If we just make multiple updates on our resource before re-list while watch is down, after re-list we will see only the final state of the resource.

In Java Operator SDK we are currently doing improvements regarding read-after-write-consistency; there this would help to cover some corner cases.

See: https://github.com/operator-framework/java-operator-sdk/issues/3398

## Type of change

## Checklist

Signed-off-by: Attila Mészáros <a_meszaros@apple.com>

### DIFF
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/informers/ResourceEventHandler.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/informers/ResourceEventHandler.java
@@ -39,10 +39,22 @@ public interface ResourceEventHandler<T> {
   }
 
   /**
+   * Called before a list operation is started, including the initial list and any subsequent re-list
+   * (for example after an HTTP GONE).
+   * <p>
+   * Should not be implemented with long-running logic as that may lead to memory issues.
+   *
+   * @param lastSyncResourceVersion the latest resource version known prior to the list operation, or
+   *        {@code null} if no list has completed yet
+   */
+  default void onBeforeList(String lastSyncResourceVersion) {
+  }
+
+  /**
    * Called after a listing is completed. By default calls {@link #onNothing()} when remainedEmpty is true.
    * <p>
    * Should not be implemented with long-running logic as that may lead to memory issues.
-   * 
+   *
    * @param resourceVersion the latest resource version known to the list operation
    * @param remainedEmpty will be true if the cache remained empty prior to and after the list operation meaning no
    *        other events would have been emitted as part of the relist
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/informers/ResourceEventHandler.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/informers/ResourceEventHandler.java
@@ -56,6 +56,8 @@ public interface ResourceEventHandler<T> {
    *
    * <p>
    * Should not be implemented with long-running logic as that may lead to memory issues.
+   * <p>
+   * This method is not called for re-sync.
    *
    * @param lastSyncResourceVersion the latest resource version known prior to the list operation, or
    *        {@code null} if no list has completed yet
@@ -67,6 +69,8 @@ public interface ResourceEventHandler<T> {
    * Called after a listing is completed. By default calls {@link #onNothing()} when remainedEmpty is true.
    * <p>
    * Should not be implemented with long-running logic as that may lead to memory issues.
+   * <p>
+   * This method is not called for re-sync.
    *
    * @param resourceVersion the latest resource version known to the list operation
    * @param remainedEmpty will be true if the cache remained empty prior to and after the list operation meaning no
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/informers/ResourceEventHandler.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/informers/ResourceEventHandler.java
@@ -40,8 +40,10 @@ public interface ResourceEventHandler<T> {
 
   /**
    * Called once per list/watch cycle, before the list operation is initiated. This fires for the
-   * initial list and for each fresh re-list (for example after an HTTP GONE), but not for retry
-   * attempts that follow a transient failure within the same cycle.
+   * initial list and for every fresh re-list — whether triggered by an HTTP GONE or by any other
+   * watch failure that goes through the reconnect path — but not for retry attempts that follow a
+   * transient failure within the same cycle (i.e. before the matching {@link #onList(String, boolean)}
+   * has been delivered).
    *
    * <p>
    * In the window between this callback and the matching {@link #onList(String, boolean)}:
--- a/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/informers/ResourceEventHandler.java
+++ b/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/informers/ResourceEventHandler.java
@@ -39,8 +39,9 @@ public interface ResourceEventHandler<T> {
   }
 
   /**
-   * Called before a list operation is started, including the initial list and any subsequent re-list
-   * (for example after an HTTP GONE).
+   * Called once per list/watch cycle, before the list operation is initiated. This fires for the
+   * initial list and for each fresh re-list (for example after an HTTP GONE), but not for retry
+   * attempts that follow a transient failure within the same cycle.
    * <p>
    * Should not be implemented with long-running logic as that may lead to memory issues.
    *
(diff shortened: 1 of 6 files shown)


csviri:
@manusa @shawkins @ash-thakur-rh please review, and let me know what do you think, but please do not merge until we finalized our tests in JOSDK

shawkins:
Let's put this on hold until we get a chance to discuss the JOSDK changes.

kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/informers/ResourceEventHandler.java
```diff
@@ -38,11 +38,24 @@ default void onNothing() {
...
+   * Called once per list/watch cycle, before the list operation is initiated. This fires for the
+   * initial list and for each fresh re-list (for example after an HTTP GONE), but not for retry
+   * attempts that follow a transient failure within the same cycle.
+   * <p>
+   * Should not be implemented with long-running logic as that may lead to memory issues.
+   *
+   * @param lastSyncResourceVersion the latest resource version known prior to the list operation, or
+   *        {@code null} if no list has completed yet
+   */
+  default void onBeforeList(String lastSyncResourceVersion) {
```
shawkins, line 67: I'm not sure it's clear from the javadocs how this would be used. It might be good to highlight in the window betwen onBeforeList and onList:
- already deleted resources may still appear in the cache (we don't process that until just before onList is called).
- the last sync version remains frozen - however add / update events will can still be emitted with later resourceVersions

That last condition may be another direction to take this. There is a deletedFinalStateUnknown - but there isn't a corresponding way to tell whether an onAdd or onUpdate is from an event or a list operation.

With something like an ExtendedResourceEventHandler interface the onAdd, onUpdate methods could take an additional parameter that was either the last sync version after this event (if equal to the add or update object's resource version, then this is coming from an event), or just a boolean. There isn't a great way to simply get the last sync version from the informer with the current onAdd / onUpdated as it's being modified asynchronously from the event processing.

csviri: > already deleted resources may still appear in the cache (we don't process that until just before onList is called).
the last sync version remains frozen - however add / update events will can still be emitted with later resourceVersions

Sure will extend that.

> however add / update events will can still be emitted with later resourceVersions

Do I understand that correctly, that those events with larger resource version will always come after this callback. Or that is not guaranteed with the current model? (will also dig deeper)

> With something like an ExtendedResourceEventHandler interface the onAdd, onUpdate methods could take an additional parameter that was either the last sync version after this event (if equal to the add or update object's resource version, then this is coming from an event), or just a boolean. There isn't a great way to simply get the last sync version from the informer with the current onAdd / onUpdated as it's being modified asynchronously from the event processing.

Not sure if following, the `lastSyncResourceVersion` as it is now in the param, comes the watch if I understand correctly, or previous list. But in that regard is always the latest to my understanding (will check the code)

To our use case actually that version does not matter, I actually don't use it, just though I add it for sake of generality.

shawkins: > Do I understand that correctly, that those events with larger resource version will always come after this callback. Or that is not guaranteed with the current model? (will also dig deeper)

Basically - but with that caveat events are being processed asynchronous from how the informer is updating the last sync version.

> Not sure if following, the lastSyncResourceVersion as it is now in the param, comes the watch if I understand correctly, or previous list. But in that regard is always the latest to my understanding (will check the code)

I'm trying to experess that if onDelete is called with deletedFinalStateUnknown == true, you know that the deletion is from a relist - not from an event.

However there isn't a good way to determine that same thing from onAdd or onUpdate - so perhaps we should have similar flags in an extended event handler interface. Something like `onAdd(T newObj, boolean addFirstStateUnknown)`, and `onUpdate(T oldObj, T newObj, boolean updateOldStateUnknown)`

addFirstStateUnknown == true - means may not be event that reflects the creation of the resource (it just wasn't in the cache yet)
updateOldStateUnknown == true - means this may not be the exact prior state for the new (it was just the state that was in the cache).

csviri:
> Basically - but with that caveat events are being processed asynchronous from how the informer is updating the last sync version.

will check it in the code, thx. But if understand correctly the lastSyncVersion is updated first and event is emitted after, what the best what we can do. And that is generic enough.

> However there isn't a good way to determine that same thing from onAdd or onUpdate - so perhaps we should have similar flags in an extended event handler interface. Something like onAdd(T newObj, boolean addFirstStateUnknown), and void onUpdate(T oldObj, T newObj, boolean updateOldStateUnknown)
addFirstStateUnknown == true - meaning may not be event that reflects the creation of the resource (it just wasn't in the cache yet)
updateOldStateUnknown == true - means this may not be the exact prior state for the new (it was just the state that was in the cache).

Makes sense, hmm, good point, but not sure, for our case this granularity is not needed; and not sure if anything outside of operator world would utilize such api. So I have no strong opinion on this. 

For us, having the correct `lastSyncVersion` present from the last event received from the websocket is more than enough, also that is probably good enough to calculate such flags if needed?!

shawkins:
> will check it in the code, thx. But if understand correctly the lastSyncVersion is updated first and event is emitted after, what the best what we can do.

It's updated before the onList is queued, but updated after the add/update/delete events are queued - and again the event processing is acutally happening in a different thread, so the callbacks don't have a guarentee over what last sync version they will see.

> For us, having the correct lastSyncVersion present from the last event received from the websocket is more than enough, also that is probably good enough to calculate such flags if needed?!

Not down to a single event because the event processing is happening asychronously. The cache may have moved ahead, but you could be lagging in event processing.

csviri:
> Not down to a single event because the event processing is happening asychronously. The cache may have moved ahead, but you could be lagging in event processing.

yes, but this is good enough for our use case; and in general is well understood behavior.

> It's updated before the onList is queued, but updated after the add/update/delete events are queued - and again the event processing is acutally happening in a different thread, so the callbacks don't have a guarentee over what last sync version they will see.

makes sense, thank you!

csviri:
Note that to be on safe side of things, we might mark this method as experimental, in case we discover something that we don't see atm.

Or we might have an `ExtendedResoureHandler` (that is experimental ) ?

shawkins:
> Note that to be on safe side of things, we might mark this method as experimental, in case we discover something that we don't see atm.
> 
> Or we might have an `ExtendedResoureHandler` (that is experimental ) ?

I think with the expanded text explaining what information you can glean from the method there shouldn't be any confusion about what it provides. And while having the information in onAdd, or onUpdate seems appealing, it's much more complex of a change, so let's go with what you have here.

The only other thought, which isn't currently covered for onList, is that both methods should document that they won't be called for a resync.

csviri:
> The only other thought, which isn't currently covered for onList, is that both methods should document that they won't be called for a resync.

Noted in javadoc. Thank you!

kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers/impl/cache/Reflector.java
```diff
@@ -374,6 +382,7 @@ public void onClose(WatcherException exception) {
           logger.debug("Watch restarting due to http gone for {}", Reflector.this);
         }
         // start a whole new list/watch cycle
+        pendingBeforeListNotification = true;
```
shawkins: This is covering only the http gone case, and not the onException reconnect. It might be confusing seeing onList without a corresponding onBeforeList.

Could we instead use `Boolean pendingBeforeListNotification = true`, then set it to null when calling onBeforeList, then transition to true only when onList is called?

csviri: @shawkins changed it , not sure if you meant this way, but feel free also to adjust it. thx!

shawkins: Right, that works - no need for a null state.

shawkins (approved):
LGTM

manusa (approved):
LGTM, thx for taking care of this!

PR #7875, opened 2026-05-27
chore(deps): bump vertx5.version from 5.0.12 to 5.1.1

manusa:
## Summary

Bumps `vertx5.version` from 5.0.12 to 5.1.1 and adapts `httpclient-vertx-5` to upstream behaviour changes. Supersedes #7869 (Dependabot's PR for the same bump, whose title was misleading — only `vertx5.version` moves; `vertx.version` stays on 4.5.x).

## Changes

- **SSL handshake fix** (`Vertx5HttpClientBuilder#createSslEngineOptions`): pass `null` instead of `new String[0]` to `JdkSslContext`. The empty array always meant "no protocols enabled" to JSSE, but was harmless before 5.1: Vert.x re-applied the option-level protocols to each engine after creation (`SslContextProvider#configureEngine`). 5.1 removed that override — protocols moved into the `SslContextFactory` SPI (`enabledProtocols(Set)`, a no-op for our lambda factory) — so the empty array now sticks and the handshake fails with `No appropriate protocol`. Fixes `Vertx5SslTest` and `Vertx5HttpClientProxyHttpsTest` against 5.1.
- **Stream-error propagation** (`InputStreamReadStream`): add a re-entry guard and switch `reset(0, cause)` → `reset(0x8)`. In 5.1, `HttpClientRequestBase#mapException` now unwraps the reset cause (and also matches `StreamResetException`, not just `HttpClosedException`), so our `IOException` would surface on the response; `StandardHttpClient`'s retry layer treats `IOException` as retryable, causing a re-issue loop until the user-facing timeout. Code `0x8` (HTTP/2 CANCEL, same as Vert.x's own `cancel()`) with no cause makes the response fail with `StreamResetException` — not retried. The reset itself remains necessary because the pipe (unchanged since 5.0) just `end()`s the request on source failure, sending a truncated-but-valid chunked body and leaving the response pending forever.
- Matching update to `InputStreamReadStreamTest`.
- CHANGELOG entry under *Dependency Upgrade*.

## Notes

This PR originally targeted 5.1.0 and was held in draft because Vert.x 5.1.0's `TcpClientConfig.setConnectTimeout(Duration)` rejected `Duration.ZERO` (eclipse-vertx/vert.x#6131), breaking `Vertx5HttpClientBuilderTest#testZeroTimeouts`. The fix landed in Vert.x 5.1.1, and `testZeroTimeouts` passes again with no local workaround — the bump now goes straight to 5.1.1.

Closes #7869

PR #7908, opened 2026-06-12
fix(httpclient-vertx-5): trust certificate on WebSocket-over-TLS

### LINKED ISSUE #7907: httpclient-vertx-5: WebSocket-over-TLS (exec/attach/portForward/watch) broken with Vert.x 5.1 — fails cert trust and hangs
## Summary

Since Vert.x 5 was bumped from `5.0.12` to `5.1.1` (#7875), every **WebSocket-over-TLS** operation performed with the `httpclient-vertx-5` client fails. The TLS handshake for the WebSocket connection does not trust the API server certificate (`PKIX path building failed: unable to find valid certification path to requested target`), the connect future never completes (`futureDone=false`), and the operation eventually times out at the client-side WebSocket timeout. Regular HTTPS request/response on the same client is unaffected.

This is what has been turning the scheduled [E2E HTTPClient Tests](https://github.com/fabric8io/kubernetes-client/actions/workflows/e2e-httpclient-tests.yml) workflow red every day since 2026-06-05. The failure is isolated to the `vertx-5` column of the matrix; `jdk`, `jetty`, and `okhttp` are green across all Kubernetes versions.

## Impact

With the Vert.x 5 client against a TLS cluster, all WebSocket-backed features are broken:

- `pod.exec()` / `attach()` / interactive shell
- `cp` / file upload and download (these run over exec)
- `portForward()`
- WebSocket-based watches (`Watch`)

The watch breakage is the reason CRD-establishment waits also time out: `BaseOperation` defaults to the WebSocket `WatchConnectionManager` and only falls back to `WatchHTTPManager` in specific cases (`kubernetes-client/.../dsl/internal/BaseOperation.java:754` vs `:815`), so `waitUntilCondition`-style flows fail too.

Plain HTTP(S) request/response operations are not affected, which is why the bulk of functionality still works and why this slipped past the unit suite.

## When it started / bisection

| | Run | Date | Vert.x 5 |
| --- | --- | --- | --- |
| Last green | [26935194187](https://github.com/fabric8io/kubernetes-client/actions/runs/26935194187) | 2026-06-04 | 5.0.12 |
| First red | [26999503566](https://github.com/fabric8io/kubernetes-client/actions/runs/26999503566) | 2026-06-05 | 5.1.1 |

The bump (commit `1e4fc0621a`, PR #7875) landed between the two. It already adapted `httpclient-vertx-5` to two other Vert.x 5.1 behaviour changes (the empty-protocol-array handshake rejection and the `InputStreamReadStream` retry loop), but this WebSocket TLS trust regression was not caught because it only manifests against a real TLS endpoint, which only the `kubernetes-itests` E2E job exercises.

I confirmed the bisection locally by overriding only `-Dvertx5.version` against the isolated reproduction below:

| Vert.x 5 | HTTP-over-TLS | WS-over-TLS |
| --- | --- | --- |
| 5.0.12 | pass | **pass** |
| 5.1.1 | pass | **fail (hang to timeout)** |

## Evidence from CI

Run [27328656190](https://github.com/fabric8io/kubernetes-client/actions/runs/27328656190), job `K8S (v1.36.0, vertx-5)`. `PodIT` alone: `Tests run: 28, Failures: 1, Errors: 16` (the failure is `portForward`; the errors are `exec`, `execExitCode`, `execInteractiveShell`, `attach`, `readFile`, `readFileEscapedParams`, `readFileMissingException`, `readDirMissingException`, `copyFile`, `copyDir`, `uploadFile[1..3]`, `uploadDir`, `uploadBinaryStream`, `uploadBinaryFile`). Plus `PodEphemeralContainersIT.exec`, and the watch-backed CRD waits `TypedCustomResourceIT`, `GenericCustomResourceIT`, `GenericClusterScopedCustomResourceIT`. Module total: `Tests run: 304, Failures: 1, Errors: 20`.

Full stack trace (repeats per WebSocket op):

```
[main] ERROR i.f.k.c.d.i.core.v1.PodOperationsImpl - Exec WebSocket upgrade did not complete after 10000ms (timeout 10000ms). uri=https://.../pods/pod-standard/exec?... futureDone=false thread=main
java.util.concurrent.CompletionException: javax.net.ssl.SSLHandshakeException: Failed to create SSL connection
    at io.fabric8.kubernetes.client.vertx5.Vertx5HttpClient.lambda$buildWebSocketDirect$3(Vertx5HttpClient.java:144)
    at io.vertx.core.impl.future.FailedFuture.addListener(FailedFuture.java:86)
    ...
Caused by: javax.net.ssl.SSLHandshakeException: Failed to create SSL connection
Caused by: javax.net.ssl.SSLHandshakeException: (certificate_unknown) PKIX path building failed:
    sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
Caused by: sun.security.validator.ValidatorException: PKIX path building failed: ...
Caused by: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
```

Workflow command (Java 17): `./mvnw ... -P"httpclient-vertx-5" -Pitests -pl kubernetes-itests test -Dtest="io.fabric8.kubernetes.**"`.

## Isolated reproduction (no cluster required)

Standing up an in-process Vert.x HTTPS server with a `SelfSignedCertificate` and building the client with `sslContext(null, trustManagers)` (exactly as `Vertx5SslTest#testGet` already does for HTTP) reproduces it: the HTTPS GET passes, the WebSocket against the same endpoint hangs (`futureDone=false`, neither succeeds nor fails) until the timeout, on 5.1.1 only. Drop-in test used:

<details>
<summary><code>Vertx5WsSslReproTest.java</code> (httpclient-vertx-5, src/test)</summary>

```java
package io.fabric8.kubernetes.client.vertx5;

import io.fabric8.kubernetes.client.http.HttpClient;
import io.fabric8.kubernetes.client.http.HttpRequest;
import io.fabric8.kubernetes.client.http.HttpResponse;
import io.fabric8.kubernetes.client.http.WebSocket;
import io.vertx.core.Vertx;
import io.vertx.core.http.HttpServer;
import io.vertx.core.http.HttpServerOptions;
import io.vertx.core.net.SelfSignedCertificate;
import org.junit.jupiter.api.AfterEach;
import org.junit.jupiter.api.BeforeEach;

import java.net.URI;
import java.util.concurrent.CompletableFuture;

import javax.net.ssl.TrustManager;
import javax.net.ssl.TrustManagerFactory;

import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertNotNull;

class Vertx5WsSslReproTest {

  private Vertx vertx;
  private HttpServer server;
  private int port;
  private final HttpClient.Factory clientFactory = new Vertx5HttpClientFactory();
  private TrustManager[] trustManagers;

  @BeforeEach
  public void before() throws Exception {
    SelfSignedCertificate cert = SelfSignedCertificate.create();
    vertx = Vertx.vertx();
    server = vertx.createHttpServer(new HttpServerOptions().setSsl(true).setKeyCertOptions(cert.keyCertOptions()))
        .requestHandler(req -> req.response().end("OK"))
        .webSocketHandler(ws -> ws.writeTextMessage("hello"));
    server.listen(0).toCompletionStage().toCompletableFuture().get(20, TimeUnit.SECONDS);
    port = server.actualPort();
    TrustManagerFactory tmf = cert.trustOptions().getTrustManagerFactory(vertx);
    trustManagers = tmf.getTrustManagers();
  }

  @AfterEach
  public void after() throws Exception {
    vertx.close().toCompletionStage().toCompletableFuture().get(20, TimeUnit.SECONDS);
  }

  @Test // passes on both 5.0.12 and 5.1.1
  void testGetOverTls() throws Exception {
    HttpClient client = clientFactory.newBuilder().sslContext(null, trustManagers).build();
    HttpRequest request = client.newHttpRequestBuilder().uri("https://localhost:" + port).build();
    HttpResponse<String> resp = client.sendAsync(request, String.class).get(10, TimeUnit.SECONDS);
    assertEquals(200, resp.code());
  }

  @Test // passes on 5.0.12, hangs to timeout on 5.1.1
  void testWebSocketOverTls() throws Exception {
    HttpClient client = clientFactory.newBuilder().sslContext(null, trustManagers).build();
    CompletableFuture<WebSocket> fut = client.newWebSocketBuilder()
        .uri(URI.create("https://localhost:" + port))
        .buildAsync(new WebSocket.Listener() {
        });
    WebSocket ws = fut.get(15, TimeUnit.SECONDS); // -> TimeoutException, fut.isDone()==false
    assertNotNull(ws);
  }
}
```
</details>

Note on diagnosing locally: the `httpclient-vertx-5` test classpath has no SLF4J binding, so the async `SSLHandshakeException` that CI logs is routed to the NOP logger and does not print; the observable symptom in the unit repro is purely the hang (`fut.isDone()==false`). The handshake failure is not propagated to the connect future, so the client falls through to its own 10s WebSocket timeout instead of failing fast. Both facts match the CI log (`futureDone=false` followed by a separately logged `SSLHandshakeException`).

### What the client does

The client does **not** set Vert.x `trustOptions`/`keyCertOptions`. Instead it injects the cluster trust material through a custom `JdkSSLEngineOptions` whose `sslContextFactory()` returns a fully built Netty `JdkSslContext` wrapping the JSSE `SSLContext` (`Vertx5HttpClientBuilder#createSslEngineOptions`, `httpclient-vertx-5/.../Vertx5HttpClientBuilder.java:265-290`). The trust lives entirely inside that custom factory's `create()`.

The same `createSslEngineOptions(...)` is applied to both clients:
- HTTP: `build()` -> `httpOptions.setSslEngineOptions(...)` (`:176-179`) -> `vertx.createHttpClient(...)` — **works**.
- WebSocket: `createWebSocketClientOptions()` -> `wsOptions.setSslEngineOptions(...)` (`:207-210`) -> `vertx.createWebSocketClient(wsOptions)` — **fails on 5.1**.

The connect/failure plumbing is `Vertx5HttpClient#buildWebSocketDirect` (`Vertx5HttpClient.java:111-147`); line 144 is the `onFailure` that completes the future exceptionally, which is where the CI stack trace is rooted.

### What changed in Vert.x 5.1

Verified against the `5.0.12` and `5.1.1` tags of `eclipse-vertx/vert.x`:

1. `WebSocketClientImpl` was rewritten. In 5.0 it called `super(vertx, options)` (HttpClientBase set up TLS from the full options). In 5.1 it does not; it derives SSL from a copy-constructed `HttpClientOptions` and a per-connection `ClientSSLOptions`:
   - `VertxImpl#createWebSocketClientImpl` (`io/vertx/core/impl/VertxImpl.java:438-456`) does `HttpClientOptions o = new HttpClientOptions(wsOptions)` and builds the `NetClient` with `.sslOptions(wsOptions.getSslOptions()).sslEngineOptions(wsOptions.getSslEngineOptions())`.
   - `WebSocketClientImpl` constructor reads `options.getSslOptions()` into `defaultSslOptions` and calls `configureSSLOptions(...)`; `webSocket(...)` then resolves `sslOptions(options.isVerifyHost(), connectOptions, defaultSslOptions)` and passes that `ClientSSLOptions` into the connection (`HttpConnectParams`).
2. The `io.vertx.core.spi.tls.SslContextFactory` SPI signatures changed: `forClient(boolean)` -> `forClient(SNIServerName, String)`, `clientAuth(ClientAuth)` -> `forServer(ClientAuth)`, removed `serverName(String)`, added `enabledProtocols(Set<String>)`.
3. `SslContextProvider` was split into abstract base + `ClientSslContextProvider` / `ServerSslContextProvider` (and `ClientSslContextManager`/`ServerSslContextManager`). `ClientSslContextProvider#createClientContext` now calls `provider.get().forClient(serverName, endpointIdentificationAlgorithm).enabledProtocols(...).enabledCipherSuites(...).useAlpn(...).applicationProtocols(...)` and then `factory.create()`. `SslContextManager` still captures `supplier = sslEngineOptions::sslContextFactory`.
4. `TCPSSLOptions` SSL fields were reworked ("Provide setters for ssl options on TCPSSLOptions / Remove specific SSL config fields"). The copy constructor copies `sslEngineOptions` via `other.sslEngineOptions.copy()`, and `JdkSSLEngineOptions#copy()` returns `this` for subclasses in 5.1, so the custom factory survives the `new HttpClientOptions(wsOptions)` copy.

### The open question for the implementer

The custom factory's default SPI methods (`forClient`, `enabledProtocols`, `trustManagerFactory`, etc.) are all no-ops returning `this`, so `create()` returns the same trust-bearing `JdkSslContext` regardless. Reading the source, both the HTTP and WebSocket paths appear to funnel through `ClientSslContextProvider#createClientContext` -> `factory.create()`, which should yield fabric8's trust-bearing context for both. Yet empirically the WebSocket path falls back to default JVM trust (PKIX failure) while the HTTP path does not.

So the remaining work is to pinpoint *where* on the 5.1 WebSocket path the custom factory's context is dropped or bypassed in favour of a default-trust context. Strong candidate: the per-connection `ClientSSLOptions` resolution in the rewritten `WebSocketClientImpl` (the `defaultSslOptions` derived from the copy-constructed `HttpClientOptions`) takes a different code path than the direct `createHttpClient` setup, and because the trust is carried only inside the engine-options `sslContextFactory()` (no `trustOptions`), the WebSocket resolution ends up using the default trust managers. Confirming this with a breakpoint in `ClientSslContextProvider#createClientContext` (compare `provider` identity and resolved `trustManagers` for HTTP vs WS) would settle it.

## Suggested directions (for discussion)

1. Supply the trust/key material to the WebSocket client in a form 5.1 propagates through its new `ClientSSLOptions` path — i.e. set Vert.x `trustOptions`/`keyCertOptions` (or a `TrustManager`/`KeyManager`-backed `KeyStoreOptions`) in addition to, or instead of, carrying everything inside the custom `sslContextFactory()`. This removes the dependency on the engine-options factory being the single source of trust, which is the part that regressed.
2. Add a WebSocket-over-TLS regression test to `httpclient-vertx-5` (the repro above is a starting point; bind a test SLF4J provider so the handshake failure is visible, and assert the connect future completes rather than relying on a timeout). The same coverage gap likely exists in the other HTTP client modules, so consider lifting it into the shared `kubernetes-client-api` HTTP-client TCK.

## References

- Bump: #7875 (commit `1e4fc0621a`), `vertx5.version` 5.0.12 -> 5.1.1
- Failing workflow: `.github/workflows/e2e-httpclient-tests.yml`
- Client code: `httpclient-vertx-5/src/main/java/io/fabric8/kubernetes/client/vertx5/Vertx5HttpClientBuilder.java` (`createSslEngineOptions` `:265-290`, `createWebSocketClientOptions` `:198-213`), `Vertx5HttpClient.java:111-147`
- Vert.x upstream (diff `5.0.12..5.1.1`): `WebSocketClientImpl`, `VertxImpl#createWebSocketClientImpl`, `spi/tls/SslContextFactory`, `internal/tls/ClientSslContextProvider`, `net/TCPSSLOptions`

manusa:
## Description

Since the Vert.x 5 bump from 5.0.12 to 5.1.1 (#7875), every WebSocket-over-TLS operation performed with the `httpclient-vertx-5` client failed: the TLS handshake did not trust the API server certificate, the connect future never completed, and the operation timed out. This broke `exec`/`attach`/`portForward` and WebSocket-backed watches (and the CRD-establishment waits that depend on them) against any TLS cluster. Plain HTTPS request/response was unaffected, which is why it slipped past the unit suite and only surfaced in the scheduled E2E HTTPClient job.

### Root cause

Vert.x 5.1 rewrote `WebSocketClientImpl` to resolve TLS through a per-connection `ClientSSLOptions` derived from the client options, which does not invoke the custom `JdkSSLEngineOptions.sslContextFactory()` that `Vertx5HttpClientBuilder` used as its only carrier of trust material. With neither `trustOptions` nor `keyCertOptions` set, the WebSocket path fell back to the default JVM trust store and failed PKIX validation.

### Fix

Configure both the HTTP and WebSocket clients uniformly through a shared helper that supplies the trust/key material to Vert.x as `TrustOptions`/`KeyCertOptions` (wrapping the builder's trust/key managers), instead of carrying everything inside a custom `SslContextFactory`. This removes the dependency on the engine-options factory being the single source of trust, which is the part that regressed.

### Tests

- Lifted a WebSocket-over-TLS regression test into the shared HTTP-client TCK (`AbstractWebSocketTlsTest`), subclassed across all five client modules (JDK, Jetty, OkHttp, Vert.x 4, Vert.x 5). It hangs to timeout on the pre-fix Vert.x 5 client and passes on every client after the fix.
- Added `Vertx5MutualTlsTest` covering client-certificate (mTLS) auth over both HTTP and WebSocket, since the unified configuration also routes key material through `KeyCertOptions`.
- Added an HTTPS over-trust guard to `Vertx5SslTest` asserting an untrusted certificate is rejected rather than silently accepted.

Closes #7907


ash-thakur-rh (approved):
LGTM

PR #7911, opened 2026-06-15
feat: add JPMS module descriptors to all published artifacts (#1996)

### LINKED ISSUE #1996: Module based library ( support for Java 9+ )

austek:
Fix #1996.

### Description
Adds a real `module-info` to all 74 published modules via the moditect plugin (post-compile injection, since the model/builder classes are generated by the Sundrio annotation processor and can't be exported by inline `javac`). Module names follow `io.fabric8.<artifactId>`. Supersedes the abandoned #2110 — the original blockers are gone: split packages (resolved by the model split #2108) and Sundrio coupling (now compile-only).

### Notable changes
- **parent pom**: `moditect-maven-plugin` in `pluginManagement`; removed the `Automatic-Module-Name` profile (real descriptors replace it).
- **openshift-model**: tightened the OSGi `Export-Package` to stop embedding `config.*` — otherwise bnd copies `openshift-model-config` classes into the jar, creating a split package on the module path.
- **kube-api-test/client-inject**: moved `Fabric8ClientInjectionHandler` to `io.fabric8.kubeapitest.junit.inject` (the package was shared with core).

### Migration notes (consumers)
- HTTP-client and extension selection still work on the module path (SPI is translated to module `provides`/`uses`).
- `openshift-model` no longer re-exports `io.fabric8.openshift.api.model.config.*` — depend on `openshift-model-config` directly if you relied on this transitively.
- `Fabric8ClientInjectionHandler` moved to package `io.fabric8.kubeapitest.junit.inject`; new module name `io.fabric8.kube.api.test.client.inject`.

### Verification
Full reactor builds; all 74 descriptors validate (`--validate-modules`) with no split packages and unique names; runtime `ServiceLoader` (HTTP `Factory`, `ExtensionAdapter`, `ClientInjectionHandler`) confirmed working on the module path.

### Open question for maintainers
The `openshift-model` OSGi `Export-Package` change is the only behavior change — please confirm it's acceptable for existing OSGi consumers.

### DIFF
--- a/extensions/certmanager/client/pom.xml
+++ b/extensions/certmanager/client/pom.xml
@@ -68,4 +68,29 @@
     </dependency>
   </dependencies>
 
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.moditect</groupId>
+        <artifactId>moditect-maven-plugin</artifactId>
+        <configuration>
+          <module>
+            <moduleInfoSource>
+              open module io.fabric8.certmanager.client {
+                exports io.fabric8.certmanager.client;
+                exports io.fabric8.certmanager.client.dsl;
+
+                requires transitive io.fabric8.certmanager.model;
+                requires transitive io.fabric8.kubernetes.client;
+                requires transitive io.fabric8.kubernetes.client.api;
+                requires static builder.annotations;
+                provides io.fabric8.kubernetes.client.extension.ExtensionAdapter with
+                    io.fabric8.certmanager.client.CertManagerExtensionAdapter;
+              }
+            </moduleInfoSource>
+          </module>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
 </project>
--- a/extensions/certmanager/model/pom.xml
+++ b/extensions/certmanager/model/pom.xml
@@ -60,6 +60,33 @@
     </dependency>
   </dependencies>
 
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.moditect</groupId>
+        <artifactId>moditect-maven-plugin</artifactId>
+        <configuration>
+          <module>
+            <moduleInfoSource>
+              open module io.fabric8.certmanager.model {
+                exports io.fabric8.certmanager.api.model.acme.v1;
+                exports io.fabric8.certmanager.api.model.meta.v1;
+                exports io.fabric8.certmanager.api.model.v1;
+
+                requires transitive io.fabric8.kubernetes.model.apiextensions;
+                requires transitive io.fabric8.kubernetes.model.core;
+                requires transitive io.fabric8.kubernetes.model.gatewayapi;
+                requires transitive com.fasterxml.jackson.annotation;
+                requires com.fasterxml.jackson.databind;
+                requires static builder.annotations;
+              }
+            </moduleInfoSource>
+          </module>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
   <profiles>
     <profile>
       <id>generate</id>
--- a/extensions/chaosmesh/client/pom.xml
+++ b/extensions/chaosmesh/client/pom.xml
@@ -72,6 +72,26 @@
   <build>
     <plugins>
       <plugin>
+        <groupId>org.moditect</groupId>
+        <artifactId>moditect-maven-plugin</artifactId>
+        <configuration>
+          <module>
+            <moduleInfoSource>
+              open module io.fabric8.chaosmesh.client {
+                exports io.fabric8.chaosmesh.client;
+
+                requires transitive io.fabric8.chaosmesh.model;
+                requires transitive io.fabric8.kubernetes.client;
+                requires transitive io.fabric8.kubernetes.client.api;
+                requires static builder.annotations;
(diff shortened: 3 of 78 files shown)


kelemen:
Given that the minimum Java version is 11, wouldn't it be better and safer to add the `module-info.java` files explicitly?

austek:
Thanks for the review! I did start with explicit `module-info.java`, but it doesn't compile here: the model classes and all builders (`io.fabric8.kubernetes.api.builder.*` and every `*Builder`/`*Fluent`) are generated by the Sundrio annotation processor during compilation. `javac` validates a hand-written `module-info.java`'s `exports`/`uses`/`provides` against packages that exist at compile time, and those generated packages don't exist yet — so it fails with `package is empty or does not exist` (e.g. `io.fabric8.kubernetes.api.builder` in `kubernetes-model-common`).

moditect injects the descriptor *after* compilation, once the processor has produced everything, which is the standard approach for code-generated modules. I've added a comment in the parent pom documenting this rationale (and why explicit `moduleInfoSource` is used instead of moditect's jdeps auto-generation, which trips over a split package in the provided-scope Sundrio toolchain).

kelemen:
Interesting, this is a problem I never even considered. That said, this sounds like something I would consider a bug in javac because javac should consider the fact that an annotation processor might generate new classes (in new packages). Maybe worth reporting?

austek:
I had a look — I couldn't find an OpenJDK issue for this exact case (an inline `module-info.java` exporting a package that only exists as annotation-processor output). The closest are adjacent: [JDK-8176148](https://bugs.openjdk.org/browse/JDK-8176148) (an annotation-processing-time "reads package from both" split-package error) and [JDK-8225164](https://bugs.openjdk.org/browse/JDK-8225164) (`java.lang.module.FindException` with an annotation processor). The fact that tooling like [moditect](https://github.com/moditect/moditect) exists specifically to inject descriptors post-compile suggests it's treated as a known limitation rather than something with a javac fix. Happy to file a fresh report with a minimal reproducer if you think it's worth it — but it doesn't block this PR.

kelemen:
The main use of these plugins generating `module-info` is to be able to add them to Java 8 jars (as many libraries did so), the documentation of moditect lists a couple of other potential benefits but doesn't mention annotation processing, so I don't think it was even intended to workaround this.

I think it is very worth-while to report this on <https://bugreport.java.com/bugreport/>. If you don't, then I guess eventually if I'm not too lazy, I might do so.

Anyway, this PR is good to have even if it doesn't get merged because one can use the module-info declaration in them if they want to patch these dependencies. I have found this PR simply because I wanted to use jlink, and of course that needs module-info. The most difficult thing about patching it is to figure out all the `requires` but this PR helps that effort.

austek:
@manusa when you have a moment, would you be able to take a look (or kick off CI)? This is the JPMS work from #1996 — real `module-info` descriptors injected via moditect across all published artifacts (model, client, openshift, httpclient, extensions), since the model/builder classes are generated by Sundrio and can't be exported by an inline `module-info.java`.

Verified locally: full reactor builds on JDK 11/17, `--validate-modules` is clean across all jars (no split packages, unique names), and runtime `ServiceLoader` for `HttpClient.Factory` / `ExtensionAdapter` works on the module path. Happy to adjust naming or split it into smaller PRs if that's easier to review.

austek:
The automation question behind this PR is now investigated in #7986: bnd's `-jpms-module-info` compared against moditect's `add-module-info` on a 7-module slice of the real build, with the measurements and the remaining open questions linked from there.

Both mechanisms generate descriptors rather than hand-maintaining them, so this PR's approach is superseded either way. Happy to close it.

Two changes in here are worth keeping regardless of which mechanism wins, since they are real defects independent of the choice: the `openshift-model` / `openshift-model-config` split package, and moving `Fabric8ClientInjectionHandler` out of the package it shares with `kube-api-test`. I can open a separate PR for just those if that's useful.

manusa:
For this issue, we're trying to find a more completely automated way to deal with it.
As this spike is showing, the current approach requires boilerplate configuration for each model module which is brittle and hard to maintain.
The idea is to either improve our current maven plugins to add the generated module-info files or find a similar solution.

austek:
Agreed on moditect: one `<jpms.module.name>` per module, 74 at scale, is exactly the brittleness you describe.

The two arms differ sharply there. bnd needs no per-module naming — `io.fabric8.${replace;${project.artifactId};-;.}` in the parent resolves centrally, so the only name properties are the 5 httpclient ones already on main. Measured per-module cost: ~5 `register:` directives for ServiceLoader `provides`. How many `substitute=` options the multi-release-jar deps need is unmeasured.

If improving the plugins is the route, four findings transfer regardless: `999-SNAPSHOT` cannot form a legal automatic module name, so a partial rollout hard-fails jdeps; split packages go first (#8028); `provides` comes from `META-INF/services`; and `module-info` under `META-INF/versions/9/` must be read — bnd doesn't, hence `requires jackson-core`.

This is assigned and milestoned for 8.0.0, so the direction is yours. If a first pass would help I'm glad to take one, equally glad to stay out of the way — just say which.

PR #7920, opened 2026-06-17
fix: make WatchRequestState.startedAtNs volatile (#7919)

### LINKED ISSUE #7919: AbstractWatchManager: WatchRequestState.startedAtNs published via data race (non-volatile write after the volatile store)
Follow-up from #7898 (merged). Low severity, latent thread-safety nit, filing so it isn't lost.

### Hazard

`WatchRequestState.startedAtNs` (added in #7898) is a non-volatile `long`:

```java
// AbstractWatchManager.java:104
long startedAtNs;
```

In `startWatch()` it is written *after* the volatile publication of `latestRequestState`:

```java
// AbstractWatchManager.java:343-344
latestRequestState = new WatchRequestState();       // volatile store: publishes the reference
latestRequestState.startedAtNs = System.nanoTime(); // plain write, happens-after the publish
```

The happens-before edge from the volatile store of `latestRequestState` covers writes that occur *before* it, not a plain field write that occurs *after* it. So a thread that reads the volatile `latestRequestState` and then reads `startedAtNs` is not guaranteed to observe the `System.nanoTime()` value; it may observe the default `0`.

### Where it's read

`watchEnded(...)` runs on the WebSocket receive thread / event loop, a different thread from `startWatch()`:

```java
// AbstractWatchManager.java:496-498
long connectionAgeNs = System.nanoTime() - state.startedAtNs;
if (t == null && resourceVersion.get() != null && !state.messageReceived.get()
    && connectionAgeNs < STALE_RV_CLOSE_THRESHOLD_NS) {
```

If `startedAtNs` is read as a stale `0`, `connectionAgeNs` becomes `System.nanoTime() - 0`, which is not a meaningful connection age (`System.nanoTime()` has an arbitrary, possibly negative origin). Depending on its sign at that moment the guard can either be skipped (the GKE stale-rv workaround silently fails to engage) or fire erroneously (a spurious relist for an informer, or termination for a raw watcher).

### Impact

Low. In practice the receive thread is started from within `start(...)`, which runs after the field is set, so most HttpClient implementations introduce an incidental happens-before and the race isn't observed. It's a latent correctness issue rather than something seen in the wild, but the current ordering relies on that incidental edge instead of guaranteeing visibility.

### Open question / possible directions

Any of these would close the gap, whichever fits the surrounding intent best:

- set `startedAtNs` before the volatile store of `latestRequestState`, so it piggybacks on that publication, or
- initialize it in the `WatchRequestState` constructor (it is conceptually the construction time), or
- make the field `volatile`.

Not blocking; noted as cleanup from the #7898 review.

### Related: stale suppression rationale

Whichever option above is chosen, the `@SuppressWarnings("java:S3077")` rationale comment on `latestRequestState` should be revisited:

```java
// AbstractWatchManager.java:133
@SuppressWarnings("java:S3077") // inner fields (AtomicBoolean, CompletableFuture) are thread-safe
volatile WatchRequestState latestRequestState;
```

The comment justifies the suppression by asserting the inner fields are thread-safe (`AtomicBoolean`, `CompletableFuture`). `startedAtNs` is now a plain `long` among those fields and does not fit that rationale. If the fix makes it `volatile` or sets it before the publish, the comment should be updated to match (or to enumerate the new field), so the rationale stays accurate for a future reader.

manusa:
Follow-up cleanup from #7898 (the #7896 GKE stale-resourceVersion fix).

`WatchRequestState.startedAtNs` was a non-volatile `long` written in `startWatch()` *after* the volatile publication of `latestRequestState`:

    latestRequestState = new WatchRequestState();        // volatile store publishes the reference
    latestRequestState.startedAtNs = System.nanoTime();  // plain write, happens-after the publish

The happens-before edge of the volatile store only covers writes that precede it in program order, so it does not publish the later `startedAtNs` write. The WebSocket receive thread that reads it in `watchEnded()` could therefore observe a stale `0`, making `connectionAgeNs = System.nanoTime() - 0` a meaningless value that could either skip the GKE stale-rv guard or fire it spuriously.

Making the field `volatile` gives the write release semantics and the read acquire semantics, so the receive thread always observes the start time. The `@SuppressWarnings("java:S3077")` rationale on `latestRequestState` is updated to enumerate the new field.

Severity is low/latent: the race is masked by an incidental happens-before from `start()` on most HTTP client implementations and never reproduces on x86, so this is hardening rather than a fix for observed behaviour.

No CHANGELOG entry: `startedAtNs` was introduced in #7898 within the unreleased 7.8-SNAPSHOT cycle, so no released version ever shipped the bug.

Closes #7919

### DIFF
--- a/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
+++ b/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/AbstractWatchManager.java
@@ -101,7 +101,9 @@ public abstract class AbstractWatchManager<T extends HasMetadata> implements Wat
     final CompletableFuture<Void> ended = new CompletableFuture<>();
     final AtomicBoolean started = new AtomicBoolean();
     final AtomicBoolean messageReceived = new AtomicBoolean();
-    long startedAtNs;
+    // volatile: written in startWatch() after the volatile publication of latestRequestState and read
+    // on the WebSocket receive thread in watchEnded(); volatile guarantees that read sees the write.
+    volatile long startedAtNs;
   }
 
   private static final Logger logger = LoggerFactory.getLogger(AbstractWatchManager.class);
@@ -130,7 +132,7 @@ public abstract class AbstractWatchManager<T extends HasMetadata> implements Wat
 
   private final boolean receiveBookmarks;
 
-  @SuppressWarnings("java:S3077") // inner fields (AtomicBoolean, CompletableFuture) are thread-safe
+  @SuppressWarnings("java:S3077") // inner fields (AtomicBoolean, CompletableFuture, volatile long) are thread-safe
   volatile WatchRequestState latestRequestState;
   private final Map<Class<?>, Integer> endErrors = new ConcurrentHashMap<>();
   private AtomicInteger retryAfterSeconds = new AtomicInteger();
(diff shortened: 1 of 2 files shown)


shawkins (approved):
LGTM

ash-thakur-rh (approved):

Issue #7921, opened 2026-06-17
Vert.x event loop blocked when using Fabric8 Kubernetes Client with Netty 4.2.13.Final+


dkbalachandar:
### Description

We are seeing Vert.x `BlockedThreadChecker` warnings when using the Fabric8 Kubernetes Client with the Vert.x HTTP transport after upgrading Netty from **4.2.12.Final** to **4.2.13.Final** (also reproducible with **4.2.15.Final**).

The same application code works correctly with **Netty 4.2.12.Final**. No code changes were made between tests.

The warnings are observed during Kubernetes client connection establishment and appear to originate from the Vert.x/Netty connect path.

Example warning:

```text
Thread[vert.x-eventloop-thread-1] has been blocked for 7496 ms, time limit is 2000 ms
```

The stack trace points to:

```text
io.vertx.core.net.impl.NetClientImpl.connectInternal2
io.netty.bootstrap.Bootstrap.doResolveAndConnect
io.netty.resolver.DefaultAddressResolverGroup.newResolver
```

In another occurrence, the blocked thread was observed during JVM method handle / lambda bytecode generation:

```text
java.lang.invoke.MethodHandleStatics.traceLambdaForm
java.lang.invoke.InvokerBytecodeGenerator.resolveFrom
java.lang.invoke.LambdaForm.compileToBytecode
io.vertx.core.net.impl.NetClientImpl.connectInternal2
```

### Environment

* Fabric8 Kubernetes Client: `7.7.0'
* Vert.x: `4.5.27`
* Java: `21`
* Netty:

  * `4.2.12.Final` → Works
  * `4.2.13.Final` → Reproduces
  * `4.2.15.Final` → Reproduces

All Netty artifacts are aligned to the same version (no mixed Netty versions detected).

### Expected Behavior

No Vert.x blocked thread warnings should occur during Kubernetes client operations.

### Actual Behavior

Vert.x reports blocked event-loop threads during Kubernetes client connection establishment when running with Netty 4.2.13+.

### Additional Information

* Reproducible with the same application code.
* Issue disappears when reverting to Netty 4.2.12.Final.
* Dependency tree confirms all Netty modules are on the same version.
* Looking for guidance on whether this is a known issue/regression and if there are any recommended workarounds besides pinning Netty to 4.2.12.Final or switching HTTP client implementations.

### Fabric8 Kubernetes Client version

7.7.0

### Steps to reproduce

[...]

PR #7923, opened 2026-06-18
fix: move first-connection JDK TLS initialization off the Vert.x event loop

### LINKED ISSUE #7922: httpclient-vertx / httpclient-vertx-5: move first-connection JDK TLS initialization off the Vert.x event loop
### Background

Surfaced while investigating #7921 (with prior context in #6868 and #7907). A user saw `io.vertx.core.impl.BlockedThreadChecker` warnings (`Thread[vert.x-eventloop-thread-1] has been blocked for 7702 ms`) on the first Kubernetes TLS connection. Their specific trigger — forcing Netty 4.2.13+ onto the Vert.x 4.5 transport — is an unsupported dependency convergence and is **not** itself a fabric8 bug. But it exposed a genuine, version-independent hazard: **fabric8 lets the JDK/Netty TLS stack initialize on the Vert.x event-loop thread during the first connection.** On a cold JVM or a constrained node (the reporter is on OKE/k3s; the same reporter's #6868 was a GraalVM native image where this became a hard 10 s connect *timeout*, not just a warning) that one-time work exceeds the 2000 ms `BlockedThreadChecker` limit.

This affects users even on the **supported** Netty 4.1.x pairing — the hazard is version-independent; the Netty override only changed the timing enough to cross the threshold.

### Mechanism (verified against vertx-core 4.5.28 and 5.1.1 sources)

Both HTTP-client modules hand Vert.x a lazily-evaluated SSL context, and Vert.x materializes it **on the event loop** on the first handshake, then caches it:

```
tcp/ChannelProvider.initChannel  (event-loop thread)
  -> SslChannelProvider.createClientSslHandler
    -> SslContextProvider.createClientContext   ->   factory.create()   <- heavy work here
      -> new JdkSslContext -> JdkSslContext.toNegotiator
        -> JdkDefaultApplicationProtocolNegotiator.<clinit> -> ClassLoader.defineClass1   <- the blocked frame
```

- Vert.x 4.5: `SslContextProvider.java:121` is `factory.create()`; `:95` dispatches to `createClientContext` — these match the reporter's stack-trace line numbers exactly.
- Vert.x 5.1: `ClientSslContextProvider.java:82` is the equivalent `factory.create()`.
- In both, `SSLHelper.build()` (which assembles the provider) runs *off* the loop in `executeBlockingInternal`, but `factory.create()` is deferred to connect time **on** the loop. The fix is to make sure the heavy JDK/Netty TLS initialization happens earlier, off the loop.

### Scope

**In scope:** the SSL-context construction leg (the reporter's primary trace, attributable to `VertxHttpClientBuilder.java:137`).

**Out of scope (deliberate decision):** the reporter's other two traces — Netty `DefaultAddressResolverGroup.newResolver` and `LambdaForm.compileToBytecode` on `NetClientImpl.connectInternal2`. These are generic JVM/Netty first-connection warm-up paid by *any* Vert.x app, not fabric8-specific. Eliminating them would require initiating a real network connection during client construction (blocking `build()` on DNS/socket I/O, failing offline, with no well-defined target host), which is an unacceptable side effect for a constructor. We do not address them here.

---

### Fix 1 — `httpclient-vertx` (Vert.x 4.5): build the context once, reuse it

This module already supplies a custom `SslContextFactory`, so we control what `create()` returns. Build the `JdkSslContext` once in `build()` (which runs on the caller's thread, off the event loop) and return that single instance:

```java
// VertxHttpClientBuilder.build(), replacing the lazy lambda currently at lines ~127-148
if (this.sslContext != null) {
  options.setSsl(true);
  // Build the Netty SslContext here, off the Vert.x event loop. Vert.x otherwise calls
  // SslContextFactory.create() lazily on the event loop during the first handshake
  // (SslContextProvider.createClientContext -> factory.create()); returning a pre-built
  // instance keeps that construction -- and the one-time class loading of
  // JdkDefaultApplicationProtocolNegotiator it triggers -- off the loop. See #7921.
  final JdkSslContext jdkSslContext = new JdkSslContext(
      sslContext, true, null, IdentityCipherSuiteFilter.INSTANCE,
      ApplicationProtocolConfig.DISABLED, io.netty.handler.ssl.ClientAuth.NONE,
      protocols, false);
  options.setSslEngineOptions(new JdkSSLEngineOptions() {
    @Override public JdkSSLEngineOptions copy() { return this; }
    @Override public SslContextFactory sslContextFactory() { return () -> jdkSslContext; }
  });
}
```

**Why this is safe / behavior-preserving:** Netty `SslContext` is documented as thread-safe and reusable, and Vert.x already caches one instance per client. Our factory implements only `create()`; Vert.x's `useAlpn/forClient/enabledCipherSuites/applicationProtocols` calls already hit the interface's no-op defaults, so a pre-built singleton is semantically identical (ALPN is disabled, so only one context is ever materialized). One intentional change: a malformed `SSLContext` now fails at `build()` rather than on first connect — same inputs, deterministic either way.

---

### Fix 2 — `httpclient-vertx-5` (Vert.x 5.1): warm the JDK TLS stack off the loop

**Do NOT add a custom `SslContextFactory` here.** #7907 removed it on purpose: Vert.x 5.1's WebSocket client resolves TLS through a per-connection `ClientSSLOptions` that ignores the engine-options factory, so trust must stay in `TrustOptions`/`KeyCertOptions`. Re-adding a factory would re-break WebSocket-over-TLS trust and would only cover the HTTP client anyway.

Instead, force the one-time JDK/Netty TLS initialization off the event loop by building and discarding a JDK client `SslContext` during `build()`. Vert.x still materializes its own context lazily on the loop, but the shared classes/JIT are already warm, so it is cheap:

```java
private static final AtomicBoolean TLS_STACK_WARMED = new AtomicBoolean(false);

/**
 * One-time, off-event-loop warm-up of the JDK TLS stack. Vert.x 5.1 materialises the
 * client SslContext lazily on the event loop on the first handshake
 * (ClientSslContextProvider.createClientContext), which loads
 * JdkDefaultApplicationProtocolNegotiator and compiles SSL-path lambda forms -- enough to
 * trip BlockedThreadChecker on a cold/constrained JVM. Building and discarding a context
 * here (on the caller's thread) loads those JVM-global classes first. Best-effort: must
 * never fail client construction. See #7921. We cannot reuse the instance (unlike the
 * Vert.x 4 module) because #7907 requires trust to flow through TrustOptions/KeyCertOptions,
 * not a custom SslContextFactory.
 */
private void warmTlsStackOffEventLoop() {
  if (this.sslContext == null || !TLS_STACK_WARMED.compareAndSet(false, true)) {
    return;
  }
  try {
    io.netty.handler.ssl.SslContextBuilder.forClient().build();
  } catch (Exception e) {
    logger.debug("TLS stack warm-up skipped: {}", e.getMessage());
  }
}
```

Call it from `build()` right after `applyTlsOptions(...)`. The static guard makes it run once per JVM.

---

### Testing (behavioral, no mocking framework)

Lead with real-TLS behavioral tests against a real self-signed server (the existing pattern in `SslTest` / `Vertx5SslTest`, using `io.vertx.core.net.SelfSignedCertificate`). No Mockito.

**`httpclient-vertx`:**
1. **Behavioral regression (primary):** extend `SslTest` — real self-signed HTTPS server, build the client, perform a real GET, assert it succeeds. Proves pre-building the context preserves end-to-end TLS behavior.
2. **Contract lock-in (mock-free):** this is the one test that actually pins the fix. Reuse the existing real options-capture seam from `VertxHttpClientBuilderTest` (a `VertxHttpClientFactory` subclass whose `additionalConfig(WebClientOptions)` captures the options — a real subclass, not a mock). Configure TLS, build, then assert the pre-build/reuse contract:
   ```java
   JdkSSLEngineOptions engine = (JdkSSLEngineOptions) captured.get().getSslEngineOptions();
   assertThat(engine.sslContextFactory().create())
       .isNotNull()
       .isSameAs(engine.sslContextFactory().create()); // built once, reused; not per-call
   ```
   `@DisplayName("JDK SslContext is built once at client-build time and reused, never constructed per connection")`.

**`httpclient-vertx-5`:**
1. **Behavioral regression (primary):** real TLS HTTP and WebSocket-over-TLS round-trips are already covered by `Vertx5SslTest`, `Vertx5MutualTlsTest`, and the WS-over-TLS tests. They must stay green — that is the guarantee the warm-up neither breaks trust nor throws.
2. **Warm-up safety:** a behavioral test that builds a TLS-configured client twice and exercises a real request, asserting `build()` completes and connects without throwing — confirming the best-effort, once-per-JVM warm-up is harmless and idempotent.

**Real-world payoff (manual, not a CI gate):** reproduce the cold-start under the constrained-container recipe in `CLAUDE.md` (2 cpus, 7 GB, pinned cpuset; optional sibling `stress-ng`) and confirm the `BlockedThreadChecker` warning on the first TLS connect disappears (or drops below threshold) after the change. `BlockedThreadChecker` timing is too flaky to assert in CI.

### ~~CHANGELOG~~

~~User-facing robustness fix → add an entry under both module names referencing the issue, `Fix #7921:`. Note it helps even on the supported Netty 4.1.x pairing, since the hazard is version-independent.~~

### Acceptance criteria

- [ ] vertx-4: SSL context built at `build()`-time and reused; behavioral TLS round-trip test + mock-free same-instance contract test added.
- [ ] vertx-5: off-loop TLS warm-up added without a custom `SslContextFactory`; existing TLS + WS-over-TLS behavioral tests stay green; warm-up safety/idempotency test added.
- [ ] No change to the DNS-resolver / lambda-form connect-path costs (explicitly out of scope).
- ~~CHANGELOG entry referencing #7921.~~

ash-thakur-rh:
## Summary

Moves the first-connection JDK/Netty TLS initialization off the Vert.x event-loop thread in both Vert.x HTTP-client modules. Vert.x materializes the client `SslContext` lazily on the event loop during the first TLS handshake; on a cold or constrained JVM the one-time class loading (`JdkDefaultApplicationProtocolNegotiator`, SSL-path lambda forms) can exceed the 2000 ms `BlockedThreadChecker` threshold.

- **httpclient-vertx (Vert.x 4.5):** the module already supplies a custom `SslContextFactory`, so the `JdkSslContext` is now built once in `build()` (on the caller's thread) and that single instance is returned from the factory, instead of being constructed per `create()` call on the event loop. Netty `SslContext` is thread-safe and reusable, so this is behavior-preserving; the one intentional change is that a malformed `SSLContext` now fails at `build()` rather than on first connect.
- **httpclient-vertx-5 (Vert.x 5.1):** no custom `SslContextFactory` is reintroduced (that was removed in #7907 so trust flows through `TrustOptions`/`KeyCertOptions` for WebSocket-over-TLS). Instead the JDK TLS class-loading path is warmed once per JVM by building and discarding a throwaway `SslContext` during `build()`. Best-effort: it never fails client construction.

## Scope

This covers the SSL-context construction leg surfaced while investigating #7921. The remaining first-connection costs the reporter observed (Netty DNS `DefaultAddressResolverGroup.newResolver` and `LambdaForm.compileToBytecode` on `NetClientImpl.connectInternal2`) are generic JVM/Netty warm-up paid by any Vert.x app, not fabric8-specific, and are out of scope. **#7921 stays open** for the remaining vertx-5 investigation: the reporter's latest traces point at `SslContext.newHandler` / `SslHandler.handlerAdded`, which this warm-up does not cover.

Fixes #7922

## Tests

- vertx-4: existing self-signed TLS round-trip (`SslTest`) stays green; new mock-free contract test asserts the `JdkSslContext` is built once and reused (`sslContextFactory().create()` returns the same instance), which fails if per-connection construction is restored.
- vertx-5: existing `Vertx5SslTest` / `Vertx5MutualTlsTest` / `Vertx5WebSocketTlsTest` stay green (the warm-up neither breaks trust nor throws); new test builds a TLS client twice and runs real HTTPS round-trips to confirm the warm-up is harmless and idempotent.

`BlockedThreadChecker` timing is too flaky to gate in CI; confirming the warning disappears under the constrained-container recipe is a manual step.

### DIFF
--- a/httpclient-vertx-5/src/main/java/io/fabric8/kubernetes/client/vertx5/Vertx5HttpClientBuilder.java
+++ b/httpclient-vertx-5/src/main/java/io/fabric8/kubernetes/client/vertx5/Vertx5HttpClientBuilder.java
@@ -19,6 +19,8 @@ import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.http.HttpClient;
 import io.fabric8.kubernetes.client.http.StandardHttpClientBuilder;
 import io.fabric8.kubernetes.client.http.TlsVersion;
+import io.netty.handler.ssl.SslContextBuilder;
+import io.netty.handler.ssl.SslProvider;
 import io.vertx.core.Vertx;
 import io.vertx.core.VertxOptions;
 import io.vertx.core.file.FileSystemOptions;
@@ -62,6 +64,7 @@ public class Vertx5HttpClientBuilder<F extends HttpClient.Factory>
     extends StandardHttpClientBuilder<Vertx5HttpClient<F>, F, Vertx5HttpClientBuilder<F>> {
 
   private static final Logger logger = LoggerFactory.getLogger(Vertx5HttpClientBuilder.class);
+  private static final AtomicBoolean TLS_STACK_WARMED = new AtomicBoolean(false);
   private static final int MAX_CONNECTIONS = 8192;
 
   // the default for etcd seems to be 3 MB, but we'll default to unlimited, so we have the same behavior across clients
@@ -176,6 +179,7 @@ public class Vertx5HttpClientBuilder<F extends HttpClient.Factory>
 
     if (this.sslContext != null) {
       applyTlsOptions(httpOptions, protocols);
+      warmTlsStackOffEventLoop();
     }
 
     final WebSocketClientOptions wsOptions = createWebSocketClientOptions(protocols);
@@ -258,6 +262,31 @@ public class Vertx5HttpClientBuilder<F extends HttpClient.Factory>
       }
     }
     return vertx;
+  }
+
+  /**
+   * One-time, off-event-loop warm-up of the JDK TLS stack. Vert.x 5.1 materialises the
+   * client SslContext lazily on the event loop on the first handshake
+   * (ClientSslContextProvider.createClientContext), which loads
+   * JdkDefaultApplicationProtocolNegotiator and compiles SSL-path lambda forms — enough to
+   * trip BlockedThreadChecker on a cold/constrained JVM. Building and discarding a context
+   * here (on the caller's thread) loads those JVM-global classes first. Best-effort: must
+   * never fail client construction. See #7921. We cannot reuse the instance (unlike the
+   * Vert.x 4 module) because #7907 requires trust to flow through TrustOptions/KeyCertOptions,
+   * not a custom SslContextFactory. The provider is pinned to {@link SslProvider#JDK} so the
+   * warm-up loads the same classes the client's {@code JdkSSLEngineOptions} path uses, even
+   * when netty-tcnative is on the classpath (which would otherwise make the default provider
+   * OpenSSL and warm the wrong stack).
+   */
+  private static void warmTlsStackOffEventLoop() {
+    if (!TLS_STACK_WARMED.compareAndSet(false, true)) {
+      return;
+    }
+    try {
+      SslContextBuilder.forClient().sslProvider(SslProvider.JDK).build();
+    } catch (Exception e) {
+      logger.debug("TLS stack warm-up skipped: {}", e.getMessage());
+    }
   }
 
   /**
--- a/httpclient-vertx-5/src/test/java/io/fabric8/kubernetes/client/vertx5/Vertx5SslTest.java
+++ b/httpclient-vertx-5/src/test/java/io/fabric8/kubernetes/client/vertx5/Vertx5SslTest.java
@@ -89,6 +89,20 @@ class Vertx5SslTest {
   }
 
   @Test
+  @DisplayName("TLS warm-up is harmless and idempotent: two successive client builds both connect successfully")
+  void tlsWarmUpIsHarmlessAndIdempotent() throws Exception {
+    requestHandler = req -> req.response().end("OK");
+    for (int i = 0; i < 2; i++) {
+      try (HttpClient client = clientFactory.newBuilder().sslContext(null, trustManagers).build()) {
+        HttpRequest request = client.newHttpRequestBuilder().uri("https://localhost:" + port).build();
+        HttpResponse<String> resp = client.sendAsync(request, String.class).get(10, TimeUnit.SECONDS);
+        assertEquals(200, resp.code());
+        assertEquals("OK", resp.bodyString());
+      }
+    }
+  }
+
+  @Test
   @DisplayName("HTTPS rejects an untrusted server certificate instead of falling back to a permissive trust")
   void httpsRejectsUntrustedCertificate() throws Exception {
     requestHandler = req -> req.response().end("OK");
--- a/httpclient-vertx/src/main/java/io/fabric8/kubernetes/client/vertx/VertxHttpClientBuilder.java
+++ b/httpclient-vertx/src/main/java/io/fabric8/kubernetes/client/vertx/VertxHttpClientBuilder.java
@@ -20,6 +20,7 @@ import io.fabric8.kubernetes.client.http.HttpClient;
 import io.fabric8.kubernetes.client.http.StandardHttpClientBuilder;
 import io.fabric8.kubernetes.client.http.TlsVersion;
 import io.netty.handler.ssl.ApplicationProtocolConfig;
+import io.netty.handler.ssl.ClientAuth;
 import io.netty.handler.ssl.IdentityCipherSuiteFilter;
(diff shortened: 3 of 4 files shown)


ash-thakur-rh (commented):
**One optional nit:** `warmTlsStackOffEventLoop()` in `Vertx5HttpClientBuilder` operates only on static state (`TLS_STACK_WARMED`) — could be `private static void` to signal that. Not blocking.

manusa:
I'm going to take this one over so we can merge and then I can complete the investigation on #7921

manusa (approved):
LGTM, thx!

PR #7931, opened 2026-06-22
chore(test): load Mockito as a startup agent to avoid runtime self-attach

manusa:
## Description

Mockito 5's inline mock-maker obtains `Instrumentation` by self-attaching a ByteBuddy agent at runtime. That self-attach (ptrace + an attach socket) is on the [JEP 451](https://openjdk.org/jeps/451) path to being disallowed by the JDK (it already warns on the JDK 21 lane) and is blocked outright by restrictive command sandboxes, so every mocked test failed under such a sandbox with `Could not initialize plugin: ... MockMaker`.

This loads the **ByteBuddy agent** at JVM startup via the surefire `argLine`, so `Instrumentation` is available before any test runs and no self-attach is attempted. `byte-buddy-agent` is declared as a global test dependency (root `<dependencies>`), so its jar is resolved into the local repository for every module and the `-javaagent` can reference it by path — including for modules that do not use Mockito directly.

`byte-buddy-agent` is used rather than `mockito-core` as the agent: a `mockito-core` `-javaagent` puts a second copy of `mockito-core` on the system class loader without `byte-buddy`, which breaks the inline mock-maker (and Spock's Mockito-backed mock maker) with `NoClassDefFoundError: net.bytebuddy.matcher.ElementMatcher`. `byte-buddy-agent` is a self-contained agent jar (just `net.bytebuddy.agent.Installer`), so the mock-maker still loads from the test classpath where `byte-buddy` is present.

To allow a startup agent on the JDK 17/21 lanes, `mockito.version` is unified to 5.23.0: the `java-17` profile's `5.11.0` pin was stale (it was an upgrade back when the default was still on the 4.x line, and the default has since moved past it).

Running the suite under the sandbox also surfaced a pre-existing test-isolation gap: `Config.autoConfigure` reads the standard `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` env vars, so a proxy injected by the environment (a sandbox, or a corporate proxy on a developer box) leaks into `Config` and breaks the proxy-configuration unit tests. The surefire config now clears those vars for test forks.

Together these let `mvn test` / `mvn verify` run cleanly inside the sandbox, so the `.claude/settings.json` carve-out that ran them unsandboxed (the previous approach in this PR) is removed.

Local developer + CI hygiene; no effect on the published artifacts or runtime.


ash-thakur-rh (approved):
LGTM!

Issue #7909, opened 2026-06-12
WebSocket-over-TLS connect hangs to timeout instead of failing fast when the server certificate is untrusted


manusa:
## Summary

When a WebSocket-over-TLS connection is attempted against a server whose certificate is **not** trusted by the configured trust managers, the connect future returned by `WebSocket.Builder#buildAsync` does not complete with the TLS handshake failure. Instead it hangs and the caller only observes a timeout. The underlying `SSLHandshakeException` is never surfaced to the future, so callers get a slow, opaque timeout rather than a fast, descriptive failure.

This is the same propagation gap noted in passing in #7907 ("the handshake failure is not propagated to the connect future, so the client falls through to its own WebSocket timeout instead of failing fast"). #7907 / #7908 fixed the *trust* regression (trusted certs now work over WebSocket); this issue tracks the separate **error-propagation / fail-fast** behavior for *untrusted* certs.

## Observed behavior

Reproduced during the #7908 work with an in-process HTTPS server presenting a self-signed cert, building the client with trust managers that do **not** include it:

- The regular HTTPS request/response path (`sendAsync`) fails promptly with an `ExecutionException` wrapping `SSLHandshakeException` (once retries are disabled).
- The WebSocket path (`buildAsync(...).get(10s)`) does **not** complete within the window: `get` throws `TimeoutException`, i.e. the connect future never received the handshake failure.

Observed on the `httpclient-vertx-5` and `httpclient-jdk` clients. Not yet checked on `httpclient-jetty`, `httpclient-okhttp`, `httpclient-vertx` (4) — see open questions.

## Where

- `httpclient-vertx-5`: `Vertx5HttpClient#buildWebSocketDirect` (`Vertx5HttpClient.java:111-147`) wires an `onFailure` that completes the future exceptionally, but on an untrusted-cert handshake the Vert.x `webSocketClient.connect(...)` future does not appear to fire it promptly.
- The same symptom likely lives behind each client's `buildWebSocketDirect` / `StandardWebSocketBuilder` flow.

## Why it matters

- Slow, misleading diagnostics: a misconfigured CA on a WebSocket-backed feature (`exec`/`attach`/`portForward`/watch) surfaces as a generic timeout instead of a TLS trust error, which is hard to triage.
- It also makes a true fail-fast regression test for the trust path impossible to write at the WebSocket level (the #7908 over-trust guard had to be asserted on the HTTP path for this reason).

## Open questions

[...]

Issue #7915, opened 2026-06-16
Non-transient TLS handshake failures are retried up to the full request-retry backoff limit


manusa:
## Summary

`StandardHttpClient.shouldRetry` treats *any* `IOException` as a transient, retryable failure. Because `javax.net.ssl.SSLException` (and its subtype `SSLHandshakeException`) extends `java.io.IOException`, a deterministic TLS failure (untrusted or mismatched certificate, protocol/cipher mismatch, a proxy `CONNECT` tunnel that never becomes a real TLS endpoint, etc.) is retried up to the full `requestRetryBackoffLimit` before the original error is surfaced. The retries cannot succeed, so they add latency and log noise with no chance of recovery.

This was spun off from the investigation of the flaky test in #7912. The code already carries a TODO acknowledging the gap.

## Where it happens (trace)

`kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/StandardHttpClient.java`:

```java
// shouldRetry(...), throwable branch
} else {
  final Throwable actualCause = unwrapCompletionException(throwable);
  builder.interceptors.forEach((s, interceptor) -> interceptor.afterConnectionFailure(request, actualCause));
  if (actualCause instanceof IOException) {
    // TODO: may not be specific enough - incorrect ssl settings for example will get caught here
    logger.debug(... "should be retried after %d millis because of IOException" ...);
    return retryInterval;
  }
}
```

- The catch-all `instanceof IOException` test is at `StandardHttpClient.java:186`, and the TODO that names this exact gap is at `StandardHttpClient.java:187`.
- The retry loop and backoff live in `AsyncUtils.retryWithExponentialBackoff` and `ExponentialBackoffIntervalCalculator` (`getInterval` = `initialInterval * 2^min(retryIndex, MAX_RETRY_INTERVAL_EXPONENT)`, with `MAX_RETRY_INTERVAL_EXPONENT = 5`).
- Defaults: `Config.DEFAULT_REQUEST_RETRY_BACKOFFLIMIT = 10` and `Config.DEFAULT_REQUEST_RETRY_BACKOFFINTERVAL = 100` (`Config.java:135-136`).

## Concrete impact

[...]

PR #7942, opened 2026-06-25
fix(kubernetes-client-api): fail fast on deterministic TLS trust failures instead of retrying (7933)

### LINKED ISSUE #7933: Fail fast on deterministic TLS trust failures instead of retrying them (~19s drain on WebSocket and HTTP connects)
## Summary

A non-transient TLS **trust** failure (untrusted server certificate, expired cert, hostname mismatch) is retried by the shared HTTP-client retry layer as if it were a transient `IOException`. With the default retry policy this drains the full exponential backoff schedule (~19 s) before the failure ever reaches the caller, instead of failing fast with the real cause.

This is the single root cause behind two existing reports:

- **#7909** — WebSocket-over-TLS connect "hangs to timeout" against an untrusted cert (`exec`/`attach`/`portForward`/watch surface a generic timeout rather than a TLS trust error).
- **#7915** — `StandardHttpClient.shouldRetry` retries on **any** `IOException`, so a deterministic `SSLHandshakeException` is retried up to the backoff limit.

Both are closed in favor of this consolidated issue.

## Root cause

`StandardHttpClient.shouldRetry` (`kubernetes-client-api/.../http/StandardHttpClient.java:186`) retries every `IOException` on the connection-failure path:

```java
} else {
  final Throwable actualCause = unwrapCompletionException(throwable);
  ...
  if (actualCause instanceof IOException) {
    // TODO: may not be specific enough - incorrect ssl settings for example will get caught here   // line 187
    ...
    return retryInterval; // → retried
  }
}
```

`SSLHandshakeException` is an `IOException`, so a deterministic trust failure is retried. With the defaults (`Config.DEFAULT_REQUEST_RETRY_BACKOFFLIMIT = 10`, `DEFAULT_REQUEST_RETRY_BACKOFFINTERVAL = 100`, `MAX_RETRY_INTERVAL_EXPONENT = 5`) the backoff schedule is `100, 200, 400, 800, 1600, 3200, 3200, 3200, 3200, 3200` ms ≈ **19.1 s** across 11 attempts. `AsyncUtils.retryWithExponentialBackoff` (`AsyncUtils.java:72`) only applies a per-attempt `withTimeout` (and only when a request timeout is set); it never bounds the overall retry cycle, so the cycle runs to the limit.

The WebSocket path is the most visible victim because `StandardHttpClient.buildWebSocket` (`StandardHttpClient.java:238`) routes through the same `retryWithExponentialBackoff`/`shouldRetry`, and a connect-time trust failure completes the per-attempt future exceptionally, which `shouldRetry` then retries.

## Evidence

Measured on `main` (`84e5ad9a04`) with a temporary diagnostic added to the shared TCK `AbstractWebSocketTlsTest` (one subclass per client module), connecting to an in-process HTTPS server whose self-signed cert is **not** trusted (`SSLUtils.trustManagers(null, null, false, null, null)`), retries toggled via `RequestConfigBuilder().withRequestRetryBackoffLimit(N)` on `.tag(...)`:

| Client    | retries OFF (library-level)                        | retries = 2 |
|-----------|----------------------------------------------------|-------------|
| jdk       | fail-fast `SSLHandshakeException` (PKIX), ~1157 ms* | ~392 ms     |
| jetty     | fail-fast `SSLHandshakeException` (PKIX), ~659 ms   | ~350 ms     |
| vertx-5   | fail-fast `SSLHandshakeException` (PKIX), ~1465 ms* | ~415 ms     |
| vertx-4   | fail-fast `SSLHandshakeException` (PKIX), ~948 ms   | ~374 ms     |
| okhttp    | fail-fast `SSLHandshakeException` (PKIX), suppressed** | ~353 ms  |

\* First probe per JVM pays the one-time TLS/Vert.x cold-start; warm `retries=2` is faster despite doing two extra retries.
\** okhttp surfaces a primary `ConnectException("Connection refused")` for the IPv6 loopback route (`localhost` → `::1`, a dual-stack test artifact) and stashes the real `SSLHandshakeException("PKIX path building failed …") → ValidatorException → SunCertPathBuilderException` in `getSuppressed()`. On a real cluster endpoint (single reachable address) the `SSLHandshakeException` is primary.

Conclusions:

- **No client hangs at the library level.** All five complete their connect future exceptionally and fast (< 1.5 s) once retries are out of the way. The per-library hang hypothesis from #7909 does not reproduce on current `main` (it predates the #7907/#7908/#7923/#7926 vertx-5 work, and was observed with default retries enabled).
- **The non-propagation is not in the adapters** — every `buildWebSocketDirect` correctly calls `completeExceptionally(t)` (`Vertx5HttpClient.java:137`, `VertxHttpClient.java:96`, `JdkHttpClientImpl.java:323`, `JettyHttpClient.java:159`, `OkHttpWebSocketImpl.java:77`).
- **It is the shared retry layer.** The `retries = 2` column confirms the SSL failure is retried (two extra backoff intervals on top of a warm sub-50 ms handshake). Default policy → ~19 s.

## Scope

- Affects **all five** HTTP clients (jdk, jetty, okhttp, vertx-4, vertx-5), since the cause is in shared `kubernetes-client-api` code.
- Affects **both** the HTTP request path (`sendAsync`/`consumeBytes`, = #7915) and the WebSocket path (`buildWebSocket`, = #7909).
- User-visible impact: a misconfigured cluster CA, expired API-server cert, or hostname mismatch surfaces as a slow (~19 s) generic timeout instead of a prompt, descriptive TLS trust error, on `exec`/`attach`/`portForward`/watch and on ordinary requests.

## Proposed fix

Classify **deterministic TLS-trust failures as terminal (non-retryable)** in the shared `shouldRetry`, so they propagate immediately on both paths.

Add a static predicate in `StandardHttpClient` (or a small `io.fabric8.kubernetes.client.http` helper) and call it from the `actualCause instanceof IOException` branch before returning `retryInterval`; if terminal, `return -1`. Remove the TODO on `StandardHttpClient.java:187`.

Classification — narrow on purpose (per #7915's caution, do **not** make every `SSLException` terminal; resets and `Received fatal alert` can be transient). Treat as terminal only when the cause **and suppressed** tree contains one of these **public** types (no `sun.security.*` dependency — `SunCertPathBuilderException extends java.security.cert.CertPathBuilderException`, so the public supertype catches it):

- `java.security.cert.CertificateException` (untrusted / expired / not-yet-valid)
- `java.security.cert.CertPathValidatorException`
- `java.security.cert.CertPathBuilderException` (PKIX path building failed)
- `javax.net.ssl.SSLPeerUnverifiedException` (hostname / peer mismatch)

The matcher **must walk both `getCause()` and `getSuppressed()`** with a visited-set guard against cyclic causes. A check limited to `instanceof SSLHandshakeException` on the top-level cause would miss okhttp (failure is in `suppressed`) and is too broad (would catch transient resets).

## Acceptance criteria

- Building a client against an untrusted server cert and connecting via **WebSocket** with the **default** retry policy completes the connect future exceptionally in well under the connect timeout (target < 1 s vs. ~19 s today), with the `SSLHandshakeException`/trust cause reachable from the thrown `WebSocketHandshakeException`.
- The same untrusted-cert connection on the **HTTP** path fails promptly with the trust cause under the default retry policy (no ~19 s drain).
- Genuinely transient failures remain retried: a non-trust `SSLException` (e.g. connection reset mid-handshake) and ordinary connect/`IOException` failures still go through the existing backoff.
- Behavior is identical across all five client modules (fix lives in shared code).

## Edge cases

- okhttp dual-stack: deterministic trust failure may appear only in `getSuppressed()` behind a transient primary `ConnectException` (one route down, another cert-untrusted). The tree-walk classifier treats the tree as terminal if a trust signature appears anywhere. Failing fast on the deterministic signal is the safer default for a misconfigured CA; confirm this is acceptable for genuinely mixed trees.
- Cyclic cause chains (self-referential `getCause()`): guard with an identity visited-set.
- Hostname mismatch can surface as `SSLPeerUnverifiedException` or as a `CertificateException` depending on client/JDK; both are covered.
- Expired / not-yet-valid certs (`CertificateExpiredException`/`CertificateNotYetValidException` extend `CertificateException`) are covered and should be terminal.
- Do not regress the existing `429`/`5xx` retry behavior or the connection-failure interceptor callback (`afterConnectionFailure`) — both run before the new check.

## Proposed tests

- TCK `AbstractWebSocketTlsTest`: new negative method — client built with **default** (untrusted) trust managers and **default** retry policy, asserting `buildAsync(...).get(short window)` completes exceptionally with a `WebSocketHandshakeException` whose cause tree contains the trust failure. Pre-fix this drains ~19 s (times out); post-fix < 1 s. Inherited by all five subclasses.
- Shared HTTP TCK: mirror for `sendAsync` — untrusted cert + default retries → prompt `ExecutionException(SSLHandshakeException)`, not a ~19 s retry drain (locks in #7915).
- Unit test for the classifier: direct `SSLHandshakeException(CertPathValidatorException)`; okhttp shape (`ConnectException` with suppressed `SSLHandshakeException`); transient `SSLException("Connection reset")` stays retryable; cyclic-cause guard.

## Out of scope

- A bounded library-independent WebSocket connect/handshake timeout (raised as #7909 open question 3): not needed, since no client actually hangs, and a per-attempt `withTimeout` already exists when a request/connect timeout is set (`WatchConnectionManager`, `PodOperationsImpl`, `PortForwarderWebsocket` all set one). It would also still report a generic timeout rather than the TLS cause.
- The vertx-5 WS-over-TLS trust regression (#7907 / #7908) and first-connection TLS warm-up (#7921 / #7922 / #7926 / #7929) — separate, already addressed; those concern *trusted* connects, not *untrusted* fail-fast.

## Implementation notes

- The fix lives entirely in shared `kubernetes-client-api` (`StandardHttpClient` + classifier); no per-client-module change is needed.
- Regression check after implementing: run the existing per-module TLS suites across all five clients — `*WebSocketTlsTest`, `*SslTest`, `*MutualTlsTest` (jdk, jetty, okhttp, vertx-4, vertx-5) — plus the HTTP-client TCK, since they exercise the shared `shouldRetry`.
- CHANGELOG: add a `Fix #7933` entry under Bug Fixes (reference this consolidated issue, not the closed #7909 / #7915). The IOException-retry behavior is long-standing and shipped, so a user-facing entry is warranted.

Refs #7909, #7915

ash-thakur-rh:
## Summary

- Classify deterministic TLS trust failures (`CertificateException`, `CertPathValidatorException`, `CertPathBuilderException`, `SSLPeerUnverifiedException`) as terminal (non-retryable) in `StandardHttpClient.shouldRetry()`, so they propagate immediately instead of draining the full ~19s exponential backoff schedule
- Add `isTerminalTlsTrustFailure(Throwable)` that walks both `getCause()` and `getSuppressed()` trees with an identity-based cycle guard — handles okhttp's dual-stack shape where the trust failure appears only in suppressed exceptions
- Affects all five HTTP client modules (jdk, jetty, okhttp, vertx-4, vertx-5) on both the HTTP request (`sendAsync`/`consumeBytes`) and WebSocket connect (`buildWebSocket`) paths

Fixes #7933 (consolidates #7909 and #7915)

## Test plan

- [x] 14 new unit tests in `StandardHttpClientTest`: each terminal exception type, suppressed-tree walk, transient SSL still retried, cyclic cause chain, `isTerminalTlsTrustFailure` direct tests, consumeBytes integration
- [x] New WebSocket TCK test in `AbstractWebSocketTlsTest`: untrusted cert fails fast (< 10s, not ~19s) — inherited by all 5 HTTP client subclasses
- [x] New HTTP TCK class `AbstractHttpTlsTest` + 5 subclasses: trusted cert succeeds + untrusted cert fails fast on HTTP path
- [x] Existing SSL/TLS regression suites pass (`Vertx5SslTest`, `Vertx5MutualTlsTest`, `SslTest`)
- [x] `spotless:check` passes across all modified modules


kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractHttpTlsTest.java
```diff
@@ -0,0 +1,89 @@
...
+          .cause()
+          .satisfies(cause -> {
+            assertThat(StandardHttpClient.isTerminalTlsTrustFailure(cause))
+                .as("cause tree should contain a TLS trust failure: %s", cause)
+                .isTrue();
+          });
+      long elapsedMs = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start);
+      assertThat(elapsedMs)
+          .as("should fail fast, not drain the ~19s default backoff schedule")
+          .isLessThan(5_000);
```
manusa: This is the bound I'd ask to put back to 10s. Any value in the 3 to 18s range catches the ~19s drain equally well, and the worry at 5s is the first-HTTPS-handshake cold start in the fork: on the 2-core CI runners under forkCount=1C it's slower than locally, and since jdk/jetty have no rerunFailingTestsCount and these TLS classes are the only HTTPS tests in the fork, JUnit's method ordering can make this timed method the first handshake and eat the full cold start. 10s keeps the margin for free.

kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractWebSocketTlsTest.java
```diff
@@ -84,4 +87,30 @@ public void onMessage(WebSocket webSocket, String text) {
...
+          .cause()
+          .satisfies(cause -> {
+            assertThat(StandardHttpClient.isTerminalTlsTrustFailure(cause))
+                .as("cause tree should contain a TLS trust failure: %s", cause)
+                .isTrue();
+          });
+      long elapsedMs = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start);
+      assertThat(elapsedMs)
+          .as("should fail fast, not drain the ~19s default backoff schedule")
+          .isLessThan(5_000);
```
manusa: Same ask as the HTTP TCK test: please put this back to 10s. The 5s bound costs margin for no extra detection power, and the first handshake cold start in the fork is the flake risk.

kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractHttpTlsTest.java
```diff
@@ -0,0 +1,89 @@
...
+    TrustManager[] untrustedManagers = SSLUtils.trustManagers(null, null, false, null, null);
+    try (HttpClient client = getHttpClientFactory().newBuilder().sslContext(null, untrustedManagers).build()) {
+      long start = System.nanoTime();
+      assertThatThrownBy(() -> client
+          .sendAsync(client.newHttpRequestBuilder().uri(server.url("secure-get")).build(), String.class)
+          .get(10, TimeUnit.SECONDS))
+          .isInstanceOf(ExecutionException.class)
+          .cause()
+          .satisfies(cause -> {
+            assertThat(StandardHttpClient.isTerminalTlsTrustFailure(cause))
```
manusa: Could we assert the real cause here instead of the production predicate? Right now this checks correctness with the same isTerminalTlsTrustFailure that production uses to decide fail-fast, so an over-broad classifier bug would pass prod and test in lockstep. Something like assertThat(cause).isInstanceOfAny(SSLHandshakeException.class, SSLException.class) gives the end-to-end test teeth that don't depend on the classifier. Same applies to the WebSocket test.

manusa (changes requested):
Nice work on this. I read it against #7933 and pulled the branch down to run things, and the core of the change is exactly right: classifying the deterministic trust failures as terminal in the shared shouldRetry is the correct, minimal fix, and doing it with a type-based check that walks both the cause and the suppressed trees is the right call, it stays locale-independent and catches the okhttp dual-stack shape. I confirmed it flows through both the HTTP and WebSocket paths, that the afterConnectionFailure callback and the 429/5xx retry are untouched, and that the new tests genuinely fail without the fix. StandardHttpClientTest passed for me (38/38) and the jdk TCK passed end to end with the untrusted connect failing fast in well under a second.

A couple of things I'd like changed before this goes in.

First, the timing bound. The last commit tightened it from 10s to 5s, and I'd ask to put it back to 10s. The regression you're catching drains about 19s, so any bound in that wide middle range detects it just as well, and 10s gives almost double the headroom. The reason 5s worries me is the cold start of the first HTTPS handshake in a surefire fork: on the 2-core CI runners under forkCount=1C that's noticeably slower than on a dev box, and in the jdk and jetty modules there's no rerunFailingTestsCount and these TLS classes are the only HTTPS tests in the fork, so with JUnit's method ordering the timed untrusted method can end up being the very first handshake and eat the whole cold start. It's a low-probability flake, but the margin was free at 10s and we lose nothing by keeping it.

Second, a smaller one on the untrusted TCK assertions: they check correctness by calling the same isTerminalTlsTrustFailure predicate that production used to fail fast, so an over-broad classifier bug would slip through prod and test together. I'd like the assertion to look at the actual cause instead, something like isInstanceOfAny(SSLHandshakeException.class, SSLException.class), so the end-to-end test has teeth that don't depend on the classifier. The timing check and the unit tests are already solid, this just closes the loop.

Thanks again, this is close.

kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractHttpTlsTest.java
ash-thakur-rh: Done — reverted to 10s in 7a40ddd.

kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractWebSocketTlsTest.java
ash-thakur-rh: Done — reverted to 10s in 7a40ddd, same commit as the HTTP test.

kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/http/AbstractHttpTlsTest.java
ash-thakur-rh: Done — replaced with `isInstanceOf(IOException.class)` plus a test-local `hasSslExceptionInTree` helper that walks the cause/suppressed tree for `SSLException`. The straight `isInstanceOfAny(SSLHandshakeException.class, SSLException.class)` didn't work for okhttp which surfaces `ConnectException` as the primary with `SSLHandshakeException` in suppressed (dual-stack IPv6 loopback). The tree-walk handles all 5 clients and is fully independent of the production classifier. Same commit (7a40ddd).

manusa (approved):
LGTM, thx!

PR #7945, opened 2026-06-26
fix(kubernetes-client): raw API path segments allow request retargeting

GrosQuildu:
## Description

The kubernetes-client DSL builds API URLs by concatenating caller-controlled path segments into request paths without path-segment encoding or traversal validation. When a downstream service accepts a resource name, namespace, API group, metric target, or OpenShift webhook value from an untrusted caller and forwards it to a kubernetes-client instance, the caller can inject `../` segments.

With the OkHttp transport, those dot segments are normalized before the request reaches the server, so an intended namespaced resource request can be retargeted to another Kubernetes API path.

The issue is caused by raw values flowing into [`URLUtils.join`](https://github.com/fabric8io/kubernetes-client/blob/b6c4a60637/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/URLUtils.java#L68) and [`URLUtils.pathJoin`](https://github.com/fabric8io/kubernetes-client/blob/b6c4a60637/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/URLUtils.java#L161-L166), which trim and concatenate path components but do not encode or reject separators. `URLUtils.join` only collapses repeated slashes and query/hash separators; the retargeting demonstrated in the PoC occurs when the OkHttp request builder canonicalizes the resulting URL path.

We verified the same high-level request against the JDK, Jetty, and Vert.x 4 transports, and verified the equivalent low-level request against Vert.x 5. Those transports sent the raw path `/api/v1/namespaces/tenant/pods/../../../nodes/node-1` to a local HTTP server. This makes the concrete PoC OkHttp-specific, while the shared raw path-segment construction remains risky for any deployment path that performs dot-segment normalization downstream.

The vulnerable methods and call sites are:

* `BaseOperation.withName`
* `BaseOperation.inNamespace`
* `OperationSupport.getRootUrlParts`
* `OperationSupport.addNamespacedUrlPathParts`
* `OperationSupport.getResourceUrl`
* `BaseOperation.getCompleteResourceUrl`
* `MetricOperationsImpl.getMetricEndpointUrlWithPlural`
* `BaseClient.getApiGroup`
* `BaseClient.getApiResources`
* `BuildConfigOperationsImpl.withSecret`
* `BuildConfigOperationsImpl.withType`
* `BuildConfigOperationsImpl.trigger`

## Exploit Scenario

An attacker exploits a build or deployment service that uses a privileged kubernetes-client instance to read a caller-selected pod from a caller-selected namespace. The service intends to allow only namespaced pod reads, but it passes the caller's pod name directly to:

```java
client.pods().inNamespace("tenant").withName(<input>);
```

The attacker supplies the name `../../../nodes/node-1`. The client builds a URL under `/api/v1/namespaces/tenant/pods/../../../nodes/node-1`; the OkHttp backend normalizes the dot segments, and the receiving server sees `/api/v1/nodes/node-1`. If the service account has broader permissions than the caller, the attacker can cause the service to access resources outside the intended namespace or resource type.

The following Java PoC uses the public kubernetes-client API and a local HTTP server. It requires a test-scope dependency on `kubernetes-httpclient-okhttp` so the reproducer exercises a supported transport that normalizes dot segments.

```java
package io.fabric8.kubernetes.client.mock;

import com.sun.net.httpserver.HttpServer;
import io.fabric8.kubernetes.client.Config;
import io.fabric8.kubernetes.client.ConfigBuilder;
import io.fabric8.kubernetes.client.KubernetesClientBuilder;
import io.fabric8.kubernetes.client.okhttp.OkHttpClientFactory;
import org.junit.jupiter.api.AfterEach;

import java.io.IOException;
import java.net.InetAddress;
import java.net.InetSocketAddress;
import java.nio.charset.StandardCharsets;
import java.util.concurrent.atomic.AtomicInteger;
import java.util.concurrent.atomic.AtomicReference;

import static org.assertj.core.api.Assertions.assertThat;

class PathSegmentRetargetingOkHttpTest {

  private HttpServer server;

  @AfterEach
  void tearDown() {
    if (server != null) {
      server.stop(0);
    }
  }

  @Test
  void traversalNameRetargetsPodGetToNodePath() throws IOException {
    AtomicInteger requests = new AtomicInteger();
    AtomicReference<String> path = new AtomicReference<>();
    server = HttpServer.create(new InetSocketAddress(InetAddress.getLoopbackAddress(), 0), 0);
    server.createContext("/", exchange -> {
      requests.incrementAndGet();
      path.set(exchange.getRequestURI().getPath());
      byte[] body = "{\"apiVersion\":\"v1\",\"kind\":\"Pod\",\"metadata\":{\"name\":\"node-1\"}}"
          .getBytes(StandardCharsets.UTF_8);
      exchange.getResponseHeaders().add("Content-Type", "application/json");
      exchange.sendResponseHeaders(200, body.length);
      exchange.getResponseBody().write(body);
      exchange.close();
    });
    server.start();

    Config config = new ConfigBuilder()
        .withAutoConfigure(false)
        .withMasterUrl("http://127.0.0.1:" + server.getAddress().getPort())
        .withConnectionTimeout(1000)
        .withRequestTimeout(1000)
        .build();

    try (KubernetesClient client = new KubernetesClientBuilder()
        .withConfig(config)
        .withHttpClientFactory(new OkHttpClientFactory())
        .build()) {
      // THE BUG
      client.pods().inNamespace("tenant").withName("../../../nodes/node-1").get();
    }

    assertThat(requests).hasValue(1);
    assertThat(path).hasValue("/api/v1/nodes/node-1");
    assertThat(path.get()).doesNotContain("../");
  }
}
```

```xml
<dependency>
  <groupId>io.fabric8</groupId>
  <artifactId>kubernetes-httpclient-okhttp</artifactId>
  <scope>test</scope>
</dependency>
```

Temporary test dependency used by the PoC to select the OkHttp client factory.

The PoC was executed with:

```sh
./mvnw -B -ntp -pl kubernetes-tests -am \
  -Dsurefire.failIfNoSpecifiedTests=false \
  -Dtest=PathSegmentRetargetingOkHttpTest test
```

After applying this patch, the same PoC failed before sending the request with `IllegalArgumentException: Path segment must not contain separators or traversal syntax: ../../../nodes/node-1`.

## Threat Model

If the APIs are allowed to accept path traversals, then the threat model should make it clear that the APIs are not safe to use with untrusted inputs.

Otherwise, APIs that take specific path components as inputs should not accept path traversal payloads. The fix ensures that.

## Fix

Encode every typed API path segment before joining it into a URL, and reject segments that contain traversal syntax. The patch adds `URLUtils.encodePathSegment`, uses it for the generic operation builder, metrics paths, discovery paths, and OpenShift webhook paths, and adds focused tests for each affected API surface.

The rejection set uses client-go's disallowed path segment names and contents: `.`, `..`, `/`, and `%`. This helper also rejects empty typed segments. It encodes other URL-reserved characters as path data; for example, `?`, `#`, spaces, and `\` do not become URL delimiters.

Rejecting literal `%` closes encoded and multiply encoded traversal payloads because callers cannot smuggle a later-decoded `%2f`, `%2e`, or `%2e%2e` sequence through a typed path segment.

The [Kubernetes object-name documentation](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/) defines object names as strings that refer to objects in resource URLs, and it lists the object identity components as API group, resource type, namespace, and name. Kubernetes path segment names may not be `.` or `..`, and may not contain `/` or `%`; stricter DNS subdomain and label name forms also exclude `%`.

The [Kubernetes API concepts documentation](https://kubernetes.io/docs/reference/using-api/api-concepts/) places API group, version, resource type, namespace, name, and subresource in the resource URI path.

**Important:** OpenShift [documents BuildConfig webhook URLs](https://docs.redhat.com/en/documentation/openshift_container_platform/4.14/html/builds_using_buildconfig/triggering-builds-build-hooks) as `/apis/build.openshift.io/v1/namespaces/<namespace>/buildconfigs/<name>/webhooks/<secret>/<type>`. OpenShift also [documents generated GitHub webhook template secrets](https://docs.redhat.com/en/documentation/openshift_container_platform/3.9/html/developer_guide/dev-guide-templates) as 40-character alphanumeric values. **We did not identify an official Kubernetes or OpenShift contract that forbids now-rejected literals like `%` as part of the `<secret>`**. Forbidding `%` in secrets is important to prevent bypasses, but may break some users - maybe the secrets should only be encoded.

The patch can be tested with:

```sh
./mvnw -B -ntp -pl kubernetes-client-api,kubernetes-client,kubernetes-tests,openshift-client -am \
  -Dsurefire.failIfNoSpecifiedTests=false \
  -Dtest=URLUtilsTest,OperationSupportTest,BaseClientTest,MetricsTest,BuildConfigOperationsImplTest test
```

---

Paweł Płatek from Trail of Bits in collaboration with OpenAI.


ash-thakur-rh (approved):
Clean security fix with thorough coverage. The validate-reject-encode approach at URL construction boundaries is correct, aligns with Kubernetes naming constraints and client-go's `IsDenylistedPathSegment`, and is applied consistently across all identified attack surfaces. Good test coverage.

**Minor observations (non-blocking):**

1. **`URLUtilsTest.java`**: Consider adding a test case for `encodePathSegment(null)` — the method throws `IllegalArgumentException`, but this behavior is only exercised indirectly via `pathSegmentJoin`'s null guard.

2. **`OperationSupportTest.java`**: Nice coverage of all four segments (name, namespace, plural, subresource). Consider adding a positive-path test asserting that a name with valid-but-unusual characters (e.g., `sha256~abc`, `name:v1.2`) produces the expected encoded URL, to complement the rejection tests.

3. **`BaseClient.java` `encodeApiGroupVersion`**: Splitting on `/` and requiring exactly 2 parts is a good secondary validation on top of `encodePathSegment`. Confirmed safe that bare `"v1"` is short-circuited by the caller on line 304 before reaching this method.

PR #7966, opened 2026-06-26
fix: reject symlinks in ConfigMap file entries

GrosQuildu:
## Description

The `createConfigMapFromDirOrFiles` uses `NOFOLLOW_LINKS` only to decide whether a path is a directory, then reads non-directory entries with `Files.readAllBytes(file)`. A symlink to a regular file is accepted and the target bytes are embedded into the returned ConfigMap.

The kubernetes-client does follow symlinks in general. The explicit use of `NOFOLLOW_LINKS` in this single case raises a question of the expected kubernetes-client behavior in regards to symbolic links.

## Exploit Scenario

A low privileged user creates a symlink `staging/app.properties -> /etc/shadow`.

```java
Path staging = new File("./staging").toPath();
ConfigMap configMap =
        KubernetesResourceUtil.createConfigMapFromDirOrFiles("poc-config", staging);
String leaked = configMap.getData().get("app.properties");
```

The symlink is resolved by kubernetes-client. Unexpected file is leaked.

## Threat Model

Should kubernetes-client not follow symlinks? Or should this behavior be configurable?

Note that the scope of the question is huge:
* `PodUpload.upload` and `PodOperationsImpl.copy` may resolve symlinks on remote pods
* Java generator follows input CRD symlink dirs and writes symlinked .java files with explicit `FOLLOW_LINKS`
* CRD generator output may resolve project-controlled symlinks in CI
* cert/key/truststore, OpenShift fromFile, and many other functionalities

## Fix

If the client is supposed to follow symbolic links, then removing `NOFOLLOW_LINKS` from `createConfigMapFromDirOrFiles` is recommended for consistency, and an entry to threat model is recommended.

If client should not follow symlinks, then harden `createConfigMapFromDirOrFiles` and plan to harden other functionalities around symbolic links handling.

Note that the proposed fix still have race condition (time-of-check-time-of-use) issues.

---

Paweł Płatek from Trail of Bits in collaboration with OpenAI.


GrosQuildu:
Note that the proposed patch may be wrong, and correct idea is to just remove `NOFOLLOW_LINKS`. If so then please just close this PR and followup with a new one.

PR #7955, opened 2026-06-26
fix(java-generator): sanitize CRD schema values

GrosQuildu:
## Description

The CRD Java generator emits multiple schema-controlled values into generated Java source without first encoding them for the Java source context. A malicious CRD file or URL can provide an `integer` enum value that is actually Java syntax or a schema description that uses Java Unicode escapes to break out of a generated Javadoc comment. The generated model then contains attacker-controlled Java code that runs when the downstream application compiles and loads the generated class.

The vulnerable sink is [`JEnum.generateJava`](https://github.com/fabric8io/kubernetes-client/blob/b6c4a6063777e31a2be0587a6ac9de4691636c97/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/JEnum.java#L184-L200):

```java
      String originalName = AbstractJSONSchema2Pojo.escapeQuotes(k);
      Expression valueArgument = new StringLiteralExpr(originalName);
      if (!underlyingType.equals(JAVA_LANG_STRING)) {
        if (underlyingType.equals(JAVA_LANG_LONG) && !originalName.endsWith("L")) {
          valueArgument = new IntegerLiteralExpr(originalName + "L");
        } else {
          valueArgument = new IntegerLiteralExpr(originalName);
        }
      }

      EnumConstantDeclaration decl = new EnumConstantDeclaration();
      decl.addAnnotation(
          new SingleMemberAnnotationExpr(
              new Name("com.fasterxml.jackson.annotation.JsonProperty"),
              new StringLiteralExpr(originalName)));
      decl.setName(constantName);
      decl.addArgument(valueArgument);
```

`IntegerLiteralExpr` is not a numeric parser or encoder. It accepts text that is later printed into the generated Java file. A payload such as `1L); static { ... } //` closes the enum constructor argument and adds a static initializer to the generated enum. The injected code runs when downstream code compiles the generated model and loads the enum class.

The same generated-source boundary is present in schema descriptions. `JObject` renders property descriptions into Javadocs and only replaces literal `*/`. Java compilers process Unicode escapes before comment parsing, so a description containing `\u002a/ static { ... } /\u002a` becomes `*/ static { ... } /*` during compilation and turns comment text into a class-body static initializer.

[...]

PR #7976, opened 2026-06-29
fix(java-generator): detect code injection in generated source via structural validation

andreaTP:
Supersedes and closes #7955.

Detects code injection in the generated Java sources via structural validation: every generated class is re-parsed (with Java Unicode escape preprocessing enabled so it matches what `javac` sees) and a structural mismatch aborts generation before any source is written. Property descriptions additionally neutralize Unicode escape introducers.

Happy to hear feedback!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### 7.9-SNAPSHOT
 
 #### Bugs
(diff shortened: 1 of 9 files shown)


manusa (approved):
LGTM, thx!

PR #7978, opened 2026-06-30
fix(java-generator): escape schema-controlled values at generated-source sinks

manusa:
Follow-up to #7976, hardening the java-generator code-injection fix.

#7976 added a structural re-validation that detects injection in the generated sources. This fixes the root cause: every schema-controlled value is now emitted as a fully escaped Java string literal at the sink, so a malicious value can never break out of its literal in the first place.

Sinks now escaped via `StringEscapeUtils.escapeJava` (the escaper already used for `@Pattern`, `@JsonPropertyDescription` and defaults):
- `JCRObject`: `@Group`, `@Version`, `@Singular`, `@Plural`
- `JEnum`: the string enum value and its `@JsonProperty`
- `JObject`: the property-name `@JsonProperty` and the `@JsonPropertyOrder` field list

The structural validation from #7976 stays as defense in depth. The `escapeQuotes` helper, which only escaped quotes and never backslashes (the gap behind the Unicode-escape bypass), is removed now that it has no callers.

This also fixes a latent false positive: a benign value containing a backslash (such as a Windows registry path) previously produced an invalid Java escape that aborted generation, and now compiles.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ### 7.9-SNAPSHOT
 
 #### Bugs
-* Fix #7955: (java-generator) Malicious CRD schema values can no longer inject executable code into the generated Java sources. Each generated class is re-parsed and structurally validated before it is written (with Java Unicode escape preprocessing enabled to match `javac`), so any enum value, name or description that alters the generated structure aborts generation; property descriptions additionally neutralize Unicode escape introducers
+* Fix #7955: (java-generator) Malicious CRD schema values can no longer inject executable code into the generated Java sources. Schema-controlled values (enum values, CRD group/version/names, property names, descriptions and defaults) are emitted as fully escaped Java string literals, so a value carrying a Unicode-escaped quote cannot break out of its literal once `javac` decodes it. As a defense in depth, each generated class is also re-parsed and structurally validated before it is written (with Java Unicode escape preprocessing enabled to match `javac`), aborting generation on any residual structural mismatch
 
 #### Improvements
 
--- a/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
+++ b/java-generator/core/src/main/java/io/fabric8/java/generator/nodes/AbstractJSONSchema2Pojo.java
@@ -147,10 +147,6 @@ public abstract class AbstractJSONSchema2Pojo {
     return sanitized;
   }
 
-  public static String escapeQuotes(String str) {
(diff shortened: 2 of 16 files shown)


andreaTP:
I do not oppose, but I think those checks are redundant.
LLM proposed the changes to me too, I believe they are not useful.

Even when we materialize, in memory, an "hacked" source code file, there is no chance it would cause any damage/subtract any information (is still just source code).

> a Windows registry path

This seems to be fixing something, at the same time I don't know how much it will be used in practice.

manusa:
Fair, #7976 already blocks the real attack since the bad CRD never gets written, so this is mostly belt and suspenders.

Couple of small things though. On "just source code": downstream projects compile what we generate and load the classes in their build, so an injected `static {}` block does run on their side. We reproduced it in the initial review, the initializer fired and wrote a marker file on class load. #7976 covers it by rejecting now, but it's not inert.

The part that actually changes behavior is a false positive: a legit value with a backslash like `HKEY\Software` comes out as `"HKEY\Software"`, fails to re-parse, and the whole CRD gets rejected as "possible code injection". Escaping fixes that.

And honestly nothing breaks here (CI's green, the only golden diff is the `@Version` comma spacing), it just makes the generator harder to break later. We've been bitten by the lenient route before, so I'd rather close it at the sink and keep your check as the backstop. Your call though.

andreaTP:
> downstream projects compile what we generate

correct, only after the source file is being written, which is not the case.

> honestly nothing breaks here
> Escaping fixes that.

True, I'm afraid of introducing accidental breakage in user code, any time we introduce an additional escaping we risk to roll a(possibly) major change for the users.

Again, not really fussy, just sharing how I see this :shrug: if you prefer to have this additional input escaping go for it.

manusa:
The other problem that this tackles is that `LLM proposed the changes to me too` means that the random security engineer who runs a check on the repo will also get this and will land a false positive and trigger a whole lot of paperwork and processes.

I prefer the risk to rollback this change (now that's isolated) if there are reported breakages due to the additional escaping than dealing with security issues that involve a lot of paperwork and are expedited (if you're not already suffering from this, I'm pretty sure you'll know what I mean soon enough 😅 😝)

BTW, I definitely need to sync with you and the state of everything endive 🥬 ;)

andreaTP:
> the random security engineer who runs a check on the repo will also get this

I buy this as a sadly compelling argument :cry: 

> BTW, I definitely need to sync with you and the state of everything endive 🥬 ;)

You know where to find me! Happy to catch up!

PR #7784, opened 2026-05-13
fix(model-generator): emit KubernetesResource service file directly from plugin

manusa:
## Summary

Replaces the Sundrio Velocity `TemplateTransformationProcessor` — a `javac` annotation processor that produced the `META-INF/services/io.fabric8.kubernetes.api.model.KubernetesResource` file consumed at runtime by `KubernetesDeserializer.registerClassesFromClassLoaders()` — with direct file emission from `openapi-model-generator-maven-plugin` at `generate-sources` time.

## Why

Under parallel Maven builds (`-T 1C`, enabled in CI by #7675 / #7768 / #7769) multiple threads share the in-process compiler JVM and race on the classloader resource read inside Velocity's `RuntimeInstance.setDefaultProperties()`, producing `IOException: Stream closed` failures across model modules (#7774). Forking the compiler (`<fork>true</fork>`) removed the race locally but inflated the model-generator build wall-clock by ~2.55× (7:13 → 18:27 on a 16-core box), regressing most of what #7675 delivered.

## How

- `openapi-model-generator-maven-plugin` (`ModelGenerator`) gathers FQNs of every generated class with `apiVersion != null`, unions them with a new `additionalKubernetesResources` setting, and writes the sorted UTF-8 service file to `${project.basedir}/src/generated/resources/META-INF/services/io.fabric8.kubernetes.api.model.KubernetesResource`. The new file is committed alongside `src/generated/java`.
- A new `add-generated-resources` execution on `build-helper-maven-plugin` (inherited by all model modules, including extensions) registers `src/generated/resources` as a build resource so the file is packaged into the jar.
- The `@TemplateTransformations` block is removed from `model_class_annotations.mustache` and the corresponding imports from `ModelGenerator`. All ~2442 generated classes lose those 5 lines.
- The only hand-written `KubernetesResource` contributor (`KubernetesList`) is registered via the new `<additionalKubernetesResources>` plugin setting in `kubernetes-model-core/pom.xml`; its own `@TemplateTransformations` annotation is stripped.
- `io.sundr:sundr-codegen-velocity-nodeps` and `io.sundr:transform-annotations` are removed from the root `dependencyManagement`, the `kubernetes-model-generator` parent, and 9 leaf module poms (`kubernetes-client-api`, `kubernetes-client`, `openshift-client-api`, `openshift-client`, plus 5 extension model/client poms).
- `manifest.vm` is deleted.
[...]

PR #7979, opened 2026-06-30
feat(model): commit sundrio Builder/Fluent to VCS, skip APT in default build (#5622)

### LINKED ISSUE #5622: Push Sundrio generated code to the version control system
## Description

In order to improve developer performance and better understand the changes introduced in the generated code by changes in the Sundrio code-generator, we need to store the generated code in Git.

This is similar to what we're already doing with the `src/generated/java` we have from jsonschema processing.

It might require executing sundrio from somewhere else than a the annotation processing environment.

ash-thakur-rh:
## Summary

Commits sundrio-generated `*Builder.java` and `*Fluent.java` files to `src/generated/java` and disables sundrio APT during normal builds. Builder/Fluent classes compile as plain source, eliminating the APT cost that dominates model module build time.

Fixes #5622

## Approach

**Default build** (no `-Pgenerate`):
- Compiler runs with Lombok only (no Sundrio APT)
- Committed Builder/Fluent in `src/generated/java` compile as plain source
- Build time drops from ~44s to ~5-15s per model module

**Generate profile** (`-Pgenerate`, via `make generate-model`):
- Sundrio APT runs in a separate `generate-sundrio` compiler execution with `proc:only` — generates files but does not compile them
- Antrun copies only the module-specific `*Builder.java` and `*Fluent.java` back to `src/generated/java`
- `proc:only` prevents sundrio from generating stripped-down copies of dependency Fluent classes (e.g., `PodTemplateSpecFluent` from `kubernetes-model-core`) that would shadow the correct versions from dependency jars
- Default-compile (Lombok only) handles actual compilation using the committed/copied Builder/Fluent files

**Modules updated**: `kubernetes-model-core`, `kubernetes-model-batch`

## Key design decisions

1. **`proc:only` for sundrio execution**: Sundrio regenerates Fluent classes for all referenced types (not just the current module's). These stripped-down copies lack methods like `withNewSpec()` and shadow correct versions from dependency jars. `proc:only` generates files without compiling them, so only the module-specific files are kept (via antrun filter) and the rest are discarded.

2. **`process-sources` phase for generate+copy**: The openapi-model-generator wipes `src/generated/java` during `generate-sources` to regenerate model classes. The sundrio execution and antrun copy run in `process-sources` (after model regeneration, before `default-compile`) so Builder/Fluent files exist when compilation starts.

3. **Antrun copy with batch filter**: Only `**/batch/**/*Builder.java` and `**/batch/**/*Fluent.java` (or module-appropriate patterns) are copied — non-module Fluent classes generated by sundrio are discarded.

4. **`.gitattributes`**: `**/src/generated/** linguist-generated=true` collapses generated diffs in PR review.

## Prior work

- #7784 removed sundrio Velocity / `@TemplateTransformation` from model generation, so the only sundrio APT output is now `*Builder.java` / `*Fluent.java`
- The existing `src/generated/java` source root and `make generate-model` flow are reused with no Makefile changes

## Type of change

## Checklist


ash-thakur-rh:
## Why antrun copy instead of generating directly to `src/generated/java`

We initially tried pointing `<generatedSourcesDirectory>` directly at `src/generated/java` to avoid the copy step. This doesn't work for two reasons:

### 1. The compiler wipes `generatedSourcesDirectory` before APT runs

`maven-compiler-plugin` cleans the `generatedSourcesDirectory` before starting annotation processing to ensure no stale generated files remain. Since `src/generated/java` also contains the model classes (e.g., `Job.java`, `CronJob.java`) that Sundrio needs as input, wiping the directory destroys the very source files that APT needs to process. No input → no output.

### 2. Javac's Filer API cannot overwrite existing files

Even if the directory weren't wiped, javac's annotation processing uses the `Filer.createSourceFile()` API which **throws `FilerException` when a file already exists** at the target path. Since the committed Builder/Fluent files already live in `src/generated/java`, Sundrio cannot write new versions there.

### The solution

Generate to the default `target/generated-sources/annotations` (a clean, empty directory after `mvn clean`), then use antrun to copy only the module-specific Builder/Fluent files back to `src/generated/java`. This avoids both the wipe problem and the Filer overwrite limitation, while filtering out stripped-down copies of dependency Fluent classes that Sundrio also generates (e.g., a `PodTemplateSpecFluent` missing `withNewSpec()` that would shadow the correct version from `kubernetes-model-core`).

ash-thakur-rh:
closing in favor of https://github.com/fabric8io/kubernetes-client/pull/7980

PR #7980, opened 2026-06-30
feat(kubernetes-model-batch): commit sundrio Builders to VCS, skip APT in default build (#5622)

manusa:
## Summary

Reference proof-of-concept for #5622, scoped to a single module (`kubernetes-model-batch`): snapshot sundrio's annotation-processor output (`*Builder.java` / `*Fluent.java`) into committed source and skip sundrio APT in the default build, so the Builders/Fluents compile as plain javac source. This removes the per-build APT round-trip that dominates model-module build time.

This is an independent take from #7979 and deliberately avoids the antrun copy-with-filter that PR uses. The key finding: with the reactor building dependencies first, sundrio's APT already emits **only the module's own Builders** (no shallow "shadow" copies of referenced types like `PodTemplateSpecFluent`), so no filtering and no copy step are needed.

## Mechanism

A dedicated, committed source root `src/generated-builders/java` holds the Builders/Fluents, kept separate from the POJOs in `src/generated/java`:

- **Default build** (no `-Pgenerate`): `maven-compiler-plugin` names only the Lombok processors, so sundrio APT does not run and the committed Builders compile as plain source. Lombok still runs, keeping `equals`/`hashCode`/`toString`.
- **`-Pgenerate`** (via `make generate-model`): the compiler's `generatedSourcesDirectory` points at `src/generated-builders/java`, and it names the Lombok + sundrio processors so they load from the **compile classpath** — sundrio then resolves referenced Builders from dependency jars rather than emitting shallow shadow copies.

`build-helper` adds `src/generated-builders/java` as a source root. Supporting config (license excludes, `sonar.exclusions`, `javadoc.sourcepath`, `.gitattributes`, and the `generate-model.yml` drift guard) is extended to the new directory.

## Why a separate directory (vs. writing into `src/generated/java`)

`maven-compiler-plugin` deletes its `generatedSourcesDirectory` before annotation processing. Pointing it at `src/generated/java` deletes the POJOs that are sundrio's input. A separate directory makes that wipe land on exactly the Builders — which is what you want on a regen — while the POJOs are untouched. The same wipe makes regeneration idempotent and removes orphaned Builders when a POJO is deleted (verified with a removal test).

## Evidence (local, `kubernetes-model-batch`)

- `-Pgenerate` regenerates 44 Builders/Fluents, **byte-identical** to the previous APT output, **zero** shadow files.
- Default build runs **zero** sundrio generation, produces a jar with a **byte-identical class list**, and the module test passes.
- `license:check` and `spotless:check` pass.

## Corrections vs. the issue's original plan

- The issue's `<proc>none</proc>` would disable Lombok too, silently dropping `equals`/`hashCode`/`toString`. The default build runs Lombok-only instead.
- No copy/antrun and no Makefile stale-file cleanup are needed: the compiler's `generatedSourcesDirectory` wipe plus the openapi generator's own `deleteDirectory` of `src/generated/java` handle staleness.

## Scope and follow-ups (tracked on #5622)

This is a single-module POC. Lifting the wiring into the model parent pom for all model modules (atomically with committing their Builders), the modules with hand-written `@Buildable` sources, a parent-level `generate` profile, and the option to make the default-build processor selection less brittle (`annotationProcessorPaths` = lombok only) are follow-ups described on the issue.

## Commits

1. **config** — build wiring + license/sonar/javadoc/`.gitattributes`/CI-guard.
2. **generated** — the 44 committed Builder/Fluent files.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,7 @@
 # eol style
 *.sh    text eol=lf
 *.bat   text eol=crlf
+
+# Generated source trees (openapi POJOs + sundrio Builders/Fluents): collapse diffs in PR review
+**/src/generated/java/** linguist-generated=true
+**/src/generated-builders/java/** linguist-generated=true
--- a/.github/workflows/generate-model.yml
+++ b/.github/workflows/generate-model.yml
@@ -59,7 +59,7 @@ jobs:
         run: |
           git add .
           IGNORED_PATTERNS="validation-schema.json"
-          INCLUDE_PATTERNS="kube-schema.json\|src/generated/java\|src/generated/resources\|openapi/schemas/"
+          INCLUDE_PATTERNS="kube-schema.json\|src/generated/java\|src/generated-builders/java\|src/generated/resources\|openapi/schemas/"
           MODIFIED_FILES_COUNT=$(git status | grep -v "$IGNORED_PATTERNS" | grep "$INCLUDE_PATTERNS" | wc -l)
           if [ $MODIFIED_FILES_COUNT -eq "0" ]; then
             echo "No Schema files modified. Build Success.";
--- a/kubernetes-model-generator/kubernetes-model-batch/pom.xml
+++ b/kubernetes-model-generator/kubernetes-model-batch/pom.xml
@@ -43,6 +43,50 @@
     </dependency>
   </dependencies>
 
+  <build>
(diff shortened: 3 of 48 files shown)


ash-thakur-rh (approved):
LGTM!

PR #7985, opened 2026-07-02
feat(kubernetes-model-core): commit sundrio Builders to VCS, skip APT in default build

manusa:
Second slice of #5622: converts `kubernetes-model-core` to commit sundrio's
generated Builders/Fluents and skip sundrio annotation processing (APT) in the
default build, following the mechanism merged for `kubernetes-model-batch` in #7980.

## What changes
- **`src/generated-builders/java`** (new committed source root): 590 files
  (295 `*Builder.java` + 295 `*Fluent.java`), covering the 289 generated POJOs
  and the 6 hand-written `@Buildable` classes (`AnyType`, `Quantity`,
  `IntOrString`, `GenericKubernetesResource`, `KubernetesList`,
  `runtime.RawExtension`).
- **`pom.xml`**: the default build runs Lombok only (sundrio APT off) so the
  committed Builders compile as plain source; the `generate` profile redirects
  the compiler's `generatedSourcesDirectory` to `src/generated-builders/java`
  and names the Lombok + sundrio processors (classpath-loaded) for
  regeneration. `build-helper` adds the new source root.

No root-level changes are needed: the license/sonar/javadoc excludes,
`.gitattributes`, and the `generate-model` drift guard are already glob-based
(`**/src/generated-builders/java/**`) from #7980.

## Verification
- `-Pgenerate` regenerates the Builders **byte-identical** to the previous APT
  output (`diff -r` exit 0); no shadow files, no POJO drift.
- The default build runs **zero** sundrio APT; the produced jar's class list is
  **byte-identical** to before.
- `spotless:check` + `license:check` pass; `mvn clean test` -> 292 tests pass.
- Orphan removal verified: a stale Builder in the committed dir is wiped by the
  compiler's pre-APT clean on regeneration.

## Performance
Measured locally (`mvn -o clean compile`, Java 17 dev host, offline).

**Inner loop / CI** (default build — sundrio APT eliminated), median of 3:
~29.5s -> ~19.9s (**~33% faster**).

**Outer loop / `make generate-model`** (`-Pgenerate`), 12 iterations each (1 warm-up discarded):

| Config | median | trimmed mean* | min | max | stdev |
|---|---:|---:|---:|---:|---:|
| before (main) | 33.01s | 33.13s | 32.16 | 34.55 | 0.83 |
| after (this PR) | 33.16s | 33.31s | 31.99 | 41.27 | 2.38 |

*trimmed = drop min+max. Median delta **+0.15s (+0.5%)** — within noise; committing the
Builders does not increase regeneration cost. (The larger "after" stdev is a single
outlier, not a systematic shift.)

The two commits split the generated snapshot from the build-config switch so
each can be reviewed separately.

Relates to #5622


manusa:
Hi @ash-thakur-rh, I think this is a solid second step towards #5622 please review to see everything is OK.

ash-thakur-rh (approved):
LGTM!

PR #8007, opened 2026-07-21
refactor(kubernetes-model-generator): drop committed-Builders gate

manusa:
Completes #5622's follow-up #1. With `kubernetes-model-common` converted in #8006, all 49
model modules under `kubernetes-model-generator` have their sundrio Builders committed, so
the transitional per-module gate has nothing left to gate.

- the `committed-generated-builders` profile and its `<file><exists>` activation are gone;
  its `<build>` (build-helper `add-source` + the lombok-only `annotationProcessorPaths`)
  moves into the parent's unconditional `<build>`
- `sundrio.generatedBuildersDirectory` now points at `${basedir}/src/generated-builders/java`
  directly instead of being overridden per module
- the `generate` profile is unchanged, and its `combine.self="override"` no longer depends on
  profile declaration order, since it now overrides the base `<build>` rather than a sibling
  profile. The ordering warning that guarded that is removed.
- `Makefile`: the 10 `extensions/*/model` modules are discovered instead of enumerated

Net -12 lines.

### Scope, which is wider than "model modules"

`extensions/pom.xml` parents off this pom, so the unconditional config also reaches
`openapi/maven-plugin`, `openapi/validator` and the `extensions/*/client`, `*/tests`,
`*/examples` modules - all except `open-virtual-network`'s `client` and `tests`, which parent
off the root pom directly and are untouched.

None of those modules declares `@Buildable`, so there is no behavioural change today. The
consequence to know about is forward-looking: adding `@Buildable` to one of them will no
longer produce a Builder. I verified this rather than assuming it - a probe `@Buildable` class
in `extensions/volcano/client` generates a Builder/Fluent before this change and none after.
It is not a silent trap in practice, because the Builder's absence is a compile error the
moment anything references it; the pom documents it and points at the fix (put the type in a
`*/model` module, or give that module its own empty `<annotationProcessorPaths>`).

### Verification

Whole-reactor before/after, same base commit:

- all **194 jars**: class CRC-32 lists **identical** (99,212 entries); all **194 OSGi
  manifests** identical
- full build with tests: **3313 tests, 0 failures, 0 errors**. `kube-api-test` fails
  identically on an unmodified tree (it starts a real kube-apiserver) and inherits from the
  root pom, not this parent
- the only sundrio APT left anywhere is in `kubernetes-client-api` and `openshift-client-api`,
  which are root-pom descendants and out of scope here
- `make openapi-generate-java-classes`: 3x BUILD SUCCESS with **zero drift** across every
  `src/generated/java` and `src/generated-builders/java`
- the Makefile's discovered list is byte-identical to the enumerated one it replaces, and the
  empty-result guard was tested by running the discovery where nothing matches
- full jar entry lists (resources included, not just classes) compared for both `openapi/*`
  modules, since `maven-core` puts a second annotation processor (sisu) on that classpath:
  identical, so nothing was silently dropped
- `license:check` and `spotless:check` pass

### Known cosmetic side effect

Under `-Pgenerate`, modules with nothing to generate get an empty `src/generated-builders/java`,
because maven-compiler-plugin creates its `generatedSourcesDirectory`. It is untracked and git
does not record empty directories. Gating it away would reintroduce the per-module activation
this PR removes, so it is documented in the pom instead.

### Follow-up not included here

`make generate-javadoc-links` has no CI guard, and a stale `doc/javadoc-links/*/element-list`
fails silently (offline `<offlineLink>` package lists, `javadoc.doclint=none`, and `make javadoc`
never fails on an unresolvable link). It needs a full build to check without false positives, so
it does not fit `generate-model.yml`. Tracked as a follow-up on #5622 rather than as its own issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,19 +43,15 @@ openapi-generate-java-classes:
 	# Test dependency needed for model de/serialization validation
 	mvn $(MAVEN_ARGS) clean install -pl . -pl zjsonpatch
 	cd kubernetes-model-generator && mvn $(MAVEN_ARGS) -Pgenerate clean install
-	cd extensions && mvn $(MAVEN_ARGS) \
-		-pl . \
-		-pl certmanager/model \
-		-pl chaosmesh/model \
-		-pl istio/model \
-		-pl knative/model \
-		-pl open-cluster-management/model \
-		-pl open-virtual-network/model \
-		-pl tekton/model \
-		-pl verticalpodautoscaler/model \
-		-pl volcano/model \
-		-pl volumesnapshot/model \
-		-Pgenerate clean install
+	# Every extensions/**/model module, discovered rather than enumerated: this list also decides what
+	# the generate-model.yml drift check can see, so a hand-written one fails open — an extension model
+	# module missing from it is never regenerated and its committed Builders go stale silently.
+	# Discovery has to be guarded or it fails open the same way: an empty result would collapse to
+	# `-pl .`, which builds only the aggregator and reports success having regenerated nothing.
+	cd extensions && \
+		MODELS="$$(find . -path '*/target' -prune -o -path '*/model/pom.xml' -print | sed 's|/pom.xml$$||;s|^\./||' | sort | tr '\n' ',' | sed 's/,$$//')" && \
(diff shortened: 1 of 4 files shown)


ash-thakur-rh (approved):
LGTM!

PR #8014, opened 2026-07-24
fix(kube-api-test): fix env var parsing and add startupTimeout to @EnableKubeAPIServer

### LINKED ISSUE #7807: kube-api-test sample tests hit 60s startupTimeout under -T 1C contention on shared CI runners
Investigation issue tracking the broader pattern of kube-api-test sample tests hitting `startupTimeout` under CI contention. The mTLS `/readyz` fix in #7802 addresses one specific cause (the unauthenticated 403 window during RBAC bootstrap) but does NOT address apiserver process startup being slow to bind under heavy reactor parallelism.

## Evidence

During the 13 PR #7802 reruns, `JUnitExtensionKubeConfigUpdateTest` / `KubernetesMutationHookHandlingTest` and other kube-api-test sample tests hit the 60 s `startupTimeout` once (rerun #6, Java 11 Linux, job [76331448501](https://github.com/fabric8io/kubernetes-client/actions/runs/25955268530/job/76331448501)):

```
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 60.22 s <<< FAILURE! -- in io.fabric8.kubeapitest.sample.KubernetesMutationHookHandlingTest
io.fabric8.kubeapitest.KubeAPITestException: Kube API Server did not start properly
	at io.fabric8.kubeapitest.process.ProcessReadinessChecker.pollWithTimeout(ProcessReadinessChecker.java:140)
	at io.fabric8.kubeapitest.process.ProcessReadinessChecker.waitUntilReady(ProcessReadinessChecker.java:128)
	at io.fabric8.kubeapitest.process.KubeAPIServerProcess.waitUntilReady(KubeAPIServerProcess.java:109)
	at io.fabric8.kubeapitest.KubeAPIServer.start(KubeAPIServer.java:61)
```

Exact `Time elapsed: 60.22 s` — the timeout fired at the boundary. The probe is mTLS now, so RBAC 403 is not the cause; the apiserver process actually wasn't ready in 60 s.

A second instance surfaced in **#7834** (`KubernetesValidationHookHandlingTest`, `Time elapsed: 60.17 s`) — same boundary-crossing shape, plus a separate (since-fixed) `cleanEtcdData()` re-entry bug that was masking the real failure in `afterAll`.

## Likely cause

`-T 1C` reactor parallelism on GH Actions' 2-vCPU runners with `forkCount=1C` puts the runner under heavy compute pressure. Spawning `kube-apiserver` + `etcd` + going through the full post-start hook cascade can exceed 60 s under that pressure. Local instrumented startup on idle takes ~2.6 s for `/readyz` to hit 200, so the slowdown factor under CI contention is 20×+.

## What's been done

**#7836** raises the default `startupTimeout` 60 000 → 120 000 ms (env var `KUBE_API_TEST_STARTUP_TIMEOUT` still overrides) and makes `KubeAPIServer.stop()` idempotent so the cleanup-race noise from #7834 can no longer mask the real failure in `afterAll`. The bump halves the CI boundary-crossing risk; it does not reduce the underlying ~20× contention factor.

**#7856** makes `ProcessReadinessChecker.pollWithTimeout` observe an `abortCheck` supplier so the startup wait aborts promptly when `kube-apiserver` or `etcd` exits unexpectedly during startup, instead of stalling the full `startupTimeout` (now 120 s). `onExit` handlers also log the child's exit code at ERROR so the failure is triageable from surefire output.

## Alternatives weighed and not adopted in #7836

- **Reduce parallelism for the kube-api-test module** (`forkCount=1` in surefire) — doesn't help against the dominant sibling-Maven-module pressure from the `-T 1C` reactor, slows the module for limited gain, and leaves the timeout boundary risk unchanged.
- **Serialize kube-api-test sample tests via `@ResourceLock`** — only serializes within one JVM/fork; doesn't help across surefire forks or sibling modules; same boundary risk.

## Open questions

- Empirical: do post-#7836 CI runs still flake at the 120 s boundary? If they do, the underlying contention warrants attention beyond another timeout bump.

## Out of scope

The mTLS `/readyz` fix in #7802 (which addresses the orthogonal RBAC-403 sub-case) is independent and stays.

ash-thakur-rh:
## Summary

- **Fix ClassCastException in env var parsing**: `KubeAPIServerConfigBuilder.finalConfigValue()` used `Class.cast()` to convert env var strings to `Integer`/`Boolean`, which always throws `ClassCastException` since `cast()` only does reference-type widening, not parsing. Replaced with a single generic `finalConfigValue` that takes a `Function<String, T>` parser — callers pass `Integer::valueOf`, `parseStrictBoolean`, or `Function.identity()`.
- **Strict Boolean parsing**: Non-canonical values like `"yes"`, `"1"`, `"on"` are rejected with `KubeAPITestException` (consistent with Integer error path). Only `"true"` / `"false"` (case-insensitive) are accepted.
- **Add `startupTimeout` to `@EnableKubeAPIServer`**: The `withStartupTimeout()` builder method existed but was unreachable from the annotation. Added `startupTimeout` attribute (default `-1` = not set) and wired it through `KubeAPIServerExtension.annotationToConfig()`, so users can write `@EnableKubeAPIServer(startupTimeout = 180000)`.
- **Uniform validation**: `startupTimeout` must be positive — validated in `build()` so all three entry points (annotation, builder API, env var) are protected.
- **Injectable env reader for testability**: Package-private `withEnvReader(UnaryOperator<String>)` lets tests substitute `System::getenv`, following the same pattern as `Config.getHomeDir` in this project.

Fixes #7807

## Test plan

- [x] `parseStrictBoolean` parses `"true"`/`"false"` (case-insensitive), rejects non-canonical values and empty string
- [x] Env var `KUBE_API_TEST_STARTUP_TIMEOUT` correctly parsed as Integer through `build()`
- [x] Env var `KUBE_API_TEST_OFFLINE_MODE` correctly parsed as Boolean through `build()`
- [x] Explicit builder value takes precedence over env var
- [x] Invalid integer env var throws `KubeAPITestException` with env var name and value
- [x] Non-canonical boolean env var throws `KubeAPITestException` with env var name and value
- [x] Non-positive `startupTimeout` rejected from builder API
- [x] `@EnableKubeAPIServer(startupTimeout = 180000)` produces config with `startupTimeout == 180000`
- [x] `@EnableKubeAPIServer` (no startupTimeout) falls through to builder default (120000)
- [x] `@EnableKubeAPIServer(startupTimeout = 0)` rejected with `KubeAPITestException`
- [x] All tests isolated from host environment via `withEnvReader`
- [x] 28 unit tests pass, formatting and license headers clean

### DIFF
--- a/junit/kube-api-test/core/src/main/java/io/fabric8/kubeapitest/KubeAPIServerConfigBuilder.java
+++ b/junit/kube-api-test/core/src/main/java/io/fabric8/kubeapitest/KubeAPIServerConfigBuilder.java
@@ -19,6 +19,7 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Function;
+import java.util.function.UnaryOperator;
 
 public final class KubeAPIServerConfigBuilder {
 
@@ -33,6 +34,7 @@ public final class KubeAPIServerConfigBuilder {
 
   private final List<String> apiServerFlags = new ArrayList<>(0);
   private boolean updateKubeConfig = false;
+  private UnaryOperator<String> envReader = System::getenv;
 
   private String apiTestDir;
   private String apiServerVersion;
@@ -88,7 +90,7 @@ public final class KubeAPIServerConfigBuilder {
     if (currentValue != null) {
       return currentValue;
     }
-    String envValue = System.getenv(envVariable);
+    String envValue = envReader.apply(envVariable);
     if (envValue == null) {
       return defaultValue;
     }
@@ -135,6 +137,11 @@ public final class KubeAPIServerConfigBuilder {
 
   public KubeAPIServerConfigBuilder withStartupTimeout(Integer startupTimeout) {
     this.startupTimeout = startupTimeout;
+    return this;
+  }
+
+  KubeAPIServerConfigBuilder withEnvReader(UnaryOperator<String> envReader) {
+    this.envReader = envReader;
     return this;
   }
 
--- a/junit/kube-api-test/core/src/main/java/io/fabric8/kubeapitest/KubeAPIServerConfigBuilder.java
+++ b/junit/kube-api-test/core/src/main/java/io/fabric8/kubeapitest/KubeAPIServerConfigBuilder.java
@@ -68,6 +68,7 @@ public final class KubeAPIServerConfigBuilder {
     this.apiServerVersion = finalConfigValue(this.apiServerVersion, KUBE_API_TEST_API_SERVER_VERSION, null);
     this.waitForEtcdHealthCheckOnStartup = finalConfigValue(this.waitForEtcdHealthCheckOnStartup,
         KUBE_API_TEST_WAIT_FOR_ETCD_HEALTH_CHECK, false);
+    // 120s: idle startup ~2.6s scales ~20× under -T 1C CI contention (#7807)
     this.startupTimeout = finalConfigValue(this.startupTimeout, KUBE_API_TEST_STARTUP_TIMEOUT, 120_000);
 
     return new KubeAPIServerConfig(apiTestDir, apiServerVersion, offlineMode, apiServerFlags,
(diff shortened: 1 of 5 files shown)


csviri (approved):
LGTM

manusa (approved):
LGTM, thx!

PR #8015, opened 2026-07-24
chore(ci): add Java 25 LTS to CI build matrix

### LINKED ISSUE #7427: Add Java 25 LTS to CI build matrix
## Description

The current CI build pipelines only test against Java LTS versions 11, 17, and 21. With Java 25 now released as the latest LTS version, we should add it to the build matrix to ensure compatibility for users adopting this new LTS.

## Background

Currently, the following workflows test with specific Java versions:

| Workflow | Java Version(s) |
|----------|-----------------|
| `build.yml` | 11, 17, 21 |
| `windows-build.yml` | 17 |
| `e2e-tests.yml` | 11 |
| `e2e-httpclient-tests.yml` | 17 |
| `sonar.yml` | 17 |

The project compiles with `maven.compiler.release=11`, meaning it should remain compatible with Java 11+ at runtime. Adding Java 25 to the matrix ensures the library works correctly for users on the latest LTS.

## Scope

Add Java 25 LTS to the CI build matrix:

1. **Primary CI workflow (`build.yml`)**: Add Java 25 to the matrix alongside existing LTS versions (11, 17, 21)
2. **Windows build (`windows-build.yml`)**: Consider adding Java 25 testing
3. **Evaluate and fix any compatibility issues** that may arise

## Potential Compatibility Concerns

- **Sundrio annotation processor** (v0.230.1): Builder generation may need verification with Java 25
- **Jackson** (v2.20.0): Should support Java 25
- **HTTP clients**: Vertx, Jetty, OkHttp compatibility
- **Deprecated API removals**: Any APIs removed in Java 25

## Files to Update

- `.github/workflows/build.yml` - Add Java 25 to the matrix
- `.github/workflows/windows-build.yml` - Add Java 25 testing (optional)

## Acceptance Criteria

- [ ] Add Java 25 to the build matrix in `.github/workflows/build.yml`
- [ ] Verify all tests pass with Java 25
- [ ] Fix any compatibility issues discovered
- [ ] Update `.github/workflows/windows-build.yml` to include Java 25

ash-thakur-rh:
## Summary

Build and test on Java 25 LTS.

- CI: add Java 25 to `build.yml` (Linux) and `windows-build.yml`
- JDK 23+ no longer discovers annotation processors on the classpath by default (JDK-8321314), which silently turned off Lombok and Sundrio. The root pom now sets `maven.compiler.proc=full` and forwards it to the invoker ITs, so building needs JDK 17.0.11+
- Karaf ITs: turn off Karaf's local console, which calls `Subject.getSubject` and throws on JDK 23+ (apache/karaf#2214). The tests don't use it
- java-generator Gradle plugin ITs are disabled on Java 22+, where Gradle 8.5 can't run (#8104)

Fixes #7427

### DIFF
--- a/crd-generator/maven-plugin/src/it/test-dependencies/pom.xml
+++ b/crd-generator/maven-plugin/src/it/test-dependencies/pom.xml
@@ -39,6 +39,7 @@
         <configuration>
           <source>1.8</source>
           <target>1.8</target>
+          <proc>full</proc>
         </configuration>
       </plugin>
     </plugins>
--- a/java-generator/it/src/it/cert-manager/pom.xml
+++ b/java-generator/it/src/it/cert-manager/pom.xml
@@ -64,6 +64,14 @@
   <build>
     <plugins>
       <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>@maven.compiler.plugin.version@</version>
+        <configuration>
+          <proc>full</proc>
+        </configuration>
+      </plugin>
+      <plugin>
         <groupId>io.fabric8</groupId>
         <artifactId>java-generator-maven-plugin</artifactId>
         <version>@project.version@</version>
--- a/java-generator/it/src/test/java/io/fabric8/java/generator/gradle/plugin/BuildWithGradlePluginIT.java
+++ b/java-generator/it/src/test/java/io/fabric8/java/generator/gradle/plugin/BuildWithGradlePluginIT.java
@@ -16,6 +16,8 @@
 package io.fabric8.java.generator.gradle.plugin;
 
 import org.gradle.testkit.runner.BuildResult;
+import org.junit.jupiter.api.condition.DisabledForJreRange;
+import org.junit.jupiter.api.condition.JRE;
 import org.junit.jupiter.api.extension.RegisterExtension;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
@@ -25,6 +27,7 @@ import java.util.stream.Stream;
 
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+@DisabledForJreRange(min = JRE.JAVA_25, disabledReason = "Gradle 8.x does not support JDK 25+")
 public class BuildWithGradlePluginIT {
   @RegisterExtension
   final ITGradleRunnerExtension gradleRunner = new ITGradleRunnerExtension();
--- a/openshift-client-api/pom.xml
+++ b/openshift-client-api/pom.xml
@@ -29,6 +29,7 @@
   <name>Fabric8 :: Openshift :: Java Client API</name>
 
   <properties>
+    <sundrio.generatedBuildersDirectory>${basedir}/src/generated-builders/java</sundrio.generatedBuildersDirectory>
     <osgi.require-capability>
       osgi.extender;
       filter:="(osgi.extender=osgi.serviceloader.registrar)"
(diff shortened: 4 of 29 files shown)


crd-generator/maven-plugin/src/it/test-dependencies/pom.xml
```diff
@@ -39,6 +39,7 @@
         <configuration>
           <source>1.8</source>
           <target>1.8</target>
+          <proc>full</proc>
```
manusa, line 43: I think you've mentioned (internally) that this is required by Lombok and JDK23+. Can we add a comment on top of this so we are aware why this was added in case we eventually remove Lombok?

openshift-client-api/pom.xml
```diff
@@ -168,6 +169,37 @@
...
+        <configuration>
+          <annotationProcessorPaths>
+            <path>
+              <groupId>org.projectlombok</groupId>
+              <artifactId>lombok</artifactId>
+              <version>${lombok.version}</version>
+            </path>
+          </annotationProcessorPaths>
+        </configuration>
+      </plugin>
```
manusa, line 203: We might not actually need these builders.
Let's make sure that the builders are public API and people downstream are using them.
AFAIR these are just internal wirings to make ConfigBuilder work.

crd-generator/maven-plugin/src/it/test-dependencies/pom.xml
ash-thakur-rh: done!

openshift-client-api/pom.xml
ash-thakur-rh: These builders are not public API, no downstream consumer imports or references the classes directly. However, all of them are required for compilation:

  - SundrioConfigFluent / SundrioOpenShiftConfigFluent: superclasses of ConfigFluent and OpenShiftConfigFluent, providing the fluent setter methods that the public builders
  inherit.
  - SundrioConfigBuilder: used internally by `ConfigBuilder.toSundrioConfig()` and the Config constructor.
  - SundrioOpenShiftConfigBuilder: extends SundrioOpenShiftConfigFluent (Sundrio always emits both as a unit for @Buildable).
  - RequestConfigFluent, LeaderElectionConfigFluent: superclasses that RequestConfigBuilder and LeaderElectionConfigBuilder extend.
  - EditableRunConfig: returned by `RunConfigBuilder.build()`.
  - CustomResourceFluent: extended by Tekton's generated CustomResourceBuilder and ResourcesFluent.

In short: they are internal wiring, but none can be removed, they are compile-time dependencies of either public API builders or downstream extension modules.

java-generator/it/src/it/cert-manager/pom.xml
```diff
@@ -63,6 +63,15 @@
 
   <build>
     <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>@maven.compiler.plugin.version@</version>
+        <configuration>
+          <!-- Required by Lombok on JDK 23+: annotation processing is disabled by default (JDK-8321314) -->
```
andreaTP, line 71: We are not using lombok in the `java-generator` afaik

java-generator/it/src/test/java/io/fabric8/java/generator/gradle/plugin/BuildWithGradlePluginIT.java
```diff
@@ -25,6 +27,7 @@
 
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+@DisabledForJreRange(min = JRE.JAVA_25, disabledReason = "Gradle 8.x does not support JDK 25+")
```
andreaTP, line 30: Uh, this is surprising!

java-generator/it/src/it/cert-manager/pom.xml
ash-thakur-rh: this is being used in integration test only.

java-generator/it/src/test/java/io/fabric8/java/generator/gradle/plugin/BuildWithGradlePluginIT.java
ash-thakur-rh: yes, gradle 9 supports jdk 25. We will make the change of gradle 9 update here as part of other issue.

java-generator/it/src/it/cert-manager/pom.xml
andreaTP: checked, you are right, tnx for getting back

java-generator/it/src/test/java/io/fabric8/java/generator/gradle/plugin/BuildWithGradlePluginIT.java
andreaTP: :+1:

andreaTP (approved):

manusa (approved):
LGTM, thx!

manusa:
I rebased onto main and pushed a follow-up commit for the review:
- `maven.compiler.proc=full` is now set once in the root pom and forwarded to the invoker ITs, so the committed Sundrio builders and the per-pom `<proc>` overrides aren't needed (Sundrio's output is byte-identical to the committed files).
- The Karaf ITs run on Java 25 again: only Karaf's local console hits `Subject.getSubject` (fix pending in apache/karaf#2214), and our tests don't need it.
- Gradle ITs are disabled from Java 22, where Gradle 8.5 can't run. Dropped the CHANGELOG entry.

Issue #7986, opened 2026-07-03
Make all packages compatible with Jigsaw


thoroughsoft:
### Describe the bug

Jigsaw (modules) is an integrated part of Java since version 9 (released 2017).
Most fabric8 components are still not "compatible" with this specification.
That is the reason this ticket is filed as a bug, not a feature request.

Building a module based app using fabric8 forces you to use automatic naming using the file name.
This generates the following warning when compiling using maven:
[WARNING] * Required filename-based automodules detected: [list of all jar files]. Please don't publish this project to a public artifact repository! *

As a minimum, Automatic-Module-Name should be set in the jar manifest.

### Fabric8 Kubernetes Client version

7.8.0

### Steps to reproduce

Compile anything that uses a fabric8 jar as a dependency in a module based project.

### Expected behavior

No module dependencies using file names.

### Runtime

Kubernetes (vanilla)

### Kubernetes API Server version

1.36

### Environment

Windows

### Tasks

- [ ] Investigate possible ways to automate the generation of the module metadata so that this hasn't have to be maintained by hand.
   Proposals such as https://github.com/fabric8io/kubernetes-client/pull/7911 should be discarded since they are not automation-driven
   If there's not an existent Maven Plugin or well-maintained popular public (FOSS) tooling available, we might want to implement our own plugin such as the one we use to generate Java Models, etc.
- [ ] Draft follow-up tasks within the same issue #7986 so that it can be implemented (make sure to include acceptance criteria, tests, identify edge and corner cases, CI integration, etc.)

austek:
Adopt bnd if converting the 9 already-bnd-wired `packaging=jar` modules to `packaging=bundle` — dropping the published `-bundle` classifier karaf's `feature.xml` references — is acceptable; otherwise use moditect for those 9. Everything else follows from that answer.

bnd needs one instruction on the existing `maven-bundle-plugin` config plus a central name macro: 59 of 74 published modules unchanged, 68 of 74 after the conversion (6 have no bnd wiring). Gated on pinning bndlib >= 6.4.0, untested: at 6.3.1 a `serviceLoaderUses` NPE leaves `kubernetes-model-core` unbuildable and `uses` unreachable.

moditect gets `requires` names right, including multi-release jars where bnd emits `requires jackson-core`, but costs a name property per module and a split-package exclusion list; last release 2025-07-20, checked 2026-08-09.

[...]

PR #8028, opened 2026-08-09
fix: eliminate split packages in openshift-model and kube-api-test (7986)

austek:
## Description

Two packages are currently split across published artifacts. Split packages are legal on the classpath and tolerated by OSGi, but the JPMS module system rejects them, so the affected artifacts cannot be placed on the module path together. This surfaced while investigating #7986, but both are defects in their own right and independent of whichever descriptor-generation mechanism that issue settles on — hence a standalone PR.

**`openshift-model` / `openshift-model-config`** — the OSGi `Export-Package` wildcard `io.fabric8.openshift.api.model**` swept the `io.fabric8.openshift.api.model.config.*` classes into `openshift-model` from the shared generated source tree, so both bundles shipped copies. Now excluded. `openshift-model`'s own bytecode does not reference those packages (after the change its manifest names them in neither `Export-Package` nor `Import-Package`), and `openshift-model` already declares a dependency on `openshift-model-config`, so consumers continue to resolve them transitively from the module that owns them.

**`kube-api-test` / `kube-api-test-client-inject`** — `Fabric8ClientInjectionHandler` sat in `io.fabric8.kubeapitest.junit`, the same package core populates. Moved to `io.fabric8.kubeapitest.junit.inject`, with the `ServiceLoader` registration updated. `ClientInjectionHandler` and `KubeConfigInjectionHandler` were already public, so the move needed only two added imports.

Verified before and after with a scan comparing package sets across each jar pair:

```
before:  SPLIT io.fabric8.openshift.api.model.config.v1        openshift-model  openshift-model-config
         SPLIT io.fabric8.openshift.api.model.config.v1alpha2  openshift-model  openshift-model-config
         SPLIT io.fabric8.kubeapitest.junit                    kube-api-test    kube-api-test-client-inject
after:   OK no split packages   (both pairs)
```

`kube-api-test` core 47 tests and `kube-api-test-client-inject` 6 tests pass, including the `ServiceLoader`-driven injection tests that would fail if the move or the service registration were wrong.

## Type of change

The class move is source-breaking for anyone naming `io.fabric8.kubeapitest.junit.Fabric8ClientInjectionHandler` directly. In normal use it is discovered through `ServiceLoader` and never referenced by name. Called out under breaking changes in the CHANGELOG. Happy to split this PR if you would rather take the `openshift-model` fix on its own.

## Checklist

On the unticked boxes: no new unit tests, because neither change adds behaviour — the existing `kube-api-test` suites already cover the moved class through `ServiceLoader`, and the packaging property is verified by jar content rather than by a test. Sonar and the Kubernetes/OpenShift runs are left for CI. If you would like the split-package check turned into a permanent test rather than a one-off verification, that is task 5 of #7986 and I am happy to take it there.


manusa (approved):
LGTM, thx!

PR #8020, opened 2026-07-30
feat: move baseline to java 17

ash-thakur-rh:
## Description

Move the project's Java baseline from 11 to 17. The root `pom.xml` compiler settings (`maven.compiler.source/target/release`) were already set to 17, but CI workflows, release pipelines, integration tests, documentation, and dependabot config still referenced Java 11 (or even 1.8 for CRD generator ITs).

This PR aligns everything with the Java 17 baseline:

- **CI workflows** (`build.yml`, `release.yaml`, `release-snapshots.yaml`, `e2e-tests.yml`, `e2e-crc-okd-tests.yml`, `e2e-chaos-tests.yml`, `javadocs.yml`): removed Java 11 from matrices, updated all setup steps to Java 17
- **dependabot.yml**: removed JUnit 6 major-version ignore (Java 17 satisfies its requirement), updated comments from "Java 11" to "Java 17"
- **Root pom.xml**: set `karaf.itest.skip` default to `true` and removed the now-redundant `java-17` profile (was only activating on JDK 17+ to skip karaf itests)
- **CRD generator IT poms** (11 files): updated `<source>1.8</source>/<target>1.8</target>` to `17`
- **Gradle integration tests** (3 `build.gradle` files): `JavaLanguageVersion.of(11)` → `of(17)`
- **Documentation** (`AGENTS.md`/`CLAUDE.md`, CRD generator READMEs, skill file): updated all Java 11 references to Java 17

Fixes #8009

## Type of change
 - [x] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [x] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [ ] I Added [CHANGELOG](https://github.com/fabric8io/kubernetes-client/blob/main/CHANGELOG.md) entry regarding this change
 - [ ] I have implemented unit tests to cover my changes
 - [x] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/main/doc/CHEATSHEET.md) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [ ] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

Issue #8009, opened 2026-07-22
Upgrade Java Baseline to Java 17


manusa:
Update the kubernetes client to have baseline as Java 17

Issue #6081, opened 2024-06-26
Set Java baseline to 11


manusa:
## Description

Part of #5778 

Starting from version 7.0.0, the Fabric8 Kubernetes Client will require at minimum Java 11 to work.

## Tasks
- [x] Bump Maven Compiler source and target versions
  https://github.com/fabric8io/kubernetes-client/blob/50497aac5be2b3550a9cca10e5ace5eaccdb59f4/pom.xml#L174-L175
- [x] Remove Maven compiler settings overrides
  https://github.com/fabric8io/kubernetes-client/blob/8622505401460c0189059e4b7d8824db87d7837c/junit/kube-api-test/pom.xml#L37-L41
  https://github.com/fabric8io/kubernetes-client/blob/8622505401460c0189059e4b7d8824db87d7837c/httpclient-jdk/pom.xml#L32
  https://github.com/fabric8io/kubernetes-client/blob/8622505401460c0189059e4b7d8824db87d7837c/httpclient-jetty/pom.xml#L32
  https://github.com/fabric8io/kubernetes-client/blob/8622505401460c0189059e4b7d8824db87d7837c/httpclient-tests/pom.xml#L32
- [x] Remove Java-11 specific profile in Maven pom.xml files and merge with defaults
  https://github.com/fabric8io/kubernetes-client/blob/50497aac5be2b3550a9cca10e5ace5eaccdb59f4/pom.xml#L1509
  Pay attention to what was done in #5190
- [x] Remove 8 from build pipeline
  https://github.com/fabric8io/kubernetes-client/blob/3c889e7752fa2b841c8764a62c1754df0bdf963f/.github/workflows/build.yml#L47
- [x] Change to 11 in e2e-tests pipeline
  https://github.com/fabric8io/kubernetes-client/blob/e3fa9c82a2c5a5ebdbca41334a2181a5b4df1253/.github/workflows/e2e-tests.yml#L55-L58
  https://github.com/fabric8io/kubernetes-client/blob/e3fa9c82a2c5a5ebdbca41334a2181a5b4df1253/.github/workflows/e2e-tests.yml#L90
  https://github.com/fabric8io/kubernetes-client/blob/e3fa9c82a2c5a5ebdbca41334a2181a5b4df1253/.github/workflows/e2e-tests.yml#L121
[...]

PR #5874, opened 2024-04-09
Run CI with Java 17 and 21

andreaTP:
## Description

Supersedes closes #3612

## Type of change

 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [ ] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [x] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [x] Code contributed by me aligns with current project license: [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0)
 - [ ] I Added [CHANGELOG](https://github.com/fabric8io/kubernetes-client/blob/main/CHANGELOG.md) entry regarding this change
 - [ ] I have implemented unit tests to cover my changes
 - [ ] I have added/updated the [javadocs](https://www.javadoc.io/doc/io.fabric8/kubernetes-client/latest/index.html) and other [documentation](https://github.com/fabric8io/kubernetes-client/blob/main/doc/CHEATSHEET.md) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=fabric8io_kubernetes-client) report
 - [ ] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

PR #8051, opened 2026-09-10
fix(karaf): restore OSGi service activation under Java 17, and follow-ups from the baseline move

### LINKED ISSUE #8050: Karaf feature is broken under Java 17: ASM 8 can't weave Java 17 bytecode
Since the Java 17 baseline (#8009), our bundles are compiled to class file v61. The Karaf feature pins Aries SPI-Fly 1.3.0 and ASM 8.0.1, and ASM 8 can't read v61 — SPI-Fly's weaving hook throws `Unsupported class file major version 61`, so `ServiceLoader` call sites are never woven, `HttpClientUtils.getHttpClientFactory` finds no `HttpClient.Factory`, and `ManagedKubernetesClient` fails to activate. `KubernetesClient` is unusable under OSGi.

This went unnoticed because the `java-17` profile has skipped the Karaf ITs on JDK 17+ since #5874, and #8020 removed the last Java 11 CI job that ran them.

Separately, the `scr` feature we ship pins Felix SCR 2.0.6, which provides `osgi.component` 1.3 while our bundle requires `>=1.5` — it can't activate the client either. It only works today because Karaf independently installs 2.2.18.

manusa:
## Description

Fixes #8050

Follow-up to #8020. The Karaf feature has been broken since the Java 17 baseline landed, and the review of that PR turned up a set of loose ends around the baseline move. One commit per concern.

### `fix(karaf)`: restore OSGi service activation under Java 17

Our bundles now compile to class file v61, but the feature pinned Aries SPI-Fly 1.3.0 and ASM 8.0.1 — ASM 8 can't read v61, so SPI-Fly's weaving hook fails, `ServiceLoader` call sites are never woven, `HttpClientUtils` finds no `HttpClient.Factory`, and `ManagedKubernetesClient` never activates.

- **SPI-Fly 1.3.0 → 1.3.7, ASM 8.0.1 → 9.10.1.** These have to move together because SPI-Fly imports ASM over OSGi (it does not embed it): the imported range moved from `[8.0,9)` in 1.3.0 to `[9.6,10)` in 1.3.7. Both properties are only read by the feature descriptor, so they now sit in `platforms/karaf/pom.xml` next to `karaf.version`.
- **Removed the `scr` feature this repository defined.** It pinned Felix SCR 2.0.6, which *exported* the Declarative Services API itself. 2.2.18 and later export none of it and import it mandatorily, so a single-bundle `scr` feature is no longer self-contained — Karaf's own ships `org.osgi.service.component/1.5.1`, `org.osgi.util.function/1.2.0` and `org.osgi.util.promise/1.3.0` alongside it. Ours was also versioned `${project.version}`, which always outranks Karaf's `4.4.11`, so an unversioned `scr` request selected the incomplete copy and suppressed Karaf's conditionals (`scr:list`, `scr:info`, the MBean). `kubernetes-client` now depends on the distribution's `scr` feature, and the itests resolve Karaf's standard repository so the offline closure contains those bundles.
- **Removed the `java-17` profile**, which set `karaf.itest.skip=true` on every JDK 17+ build and is what hid all of the above. Added in #5874 when the JDK 11 job still ran these tests; #8020 removed that job, so they ran nowhere.

### `chore(karaf)`: tidy up the integration test harness

Now that the ITs actually execute, several dormant problems are live on every build.

- The `lib/endorsed` patch-module options derived their path from a `karaf.version` system property surefire never set, so they defaulted to `4.4.1` while the container ships `4.4.11`. The JVM ignores patch-module paths that don't exist, so they had never done anything — and the Karaf specs locator was correspondingly unavailable, logging **47** `NoClassDefFoundError: org/apache/karaf/specs/locator/OsgiLocator` framework errors. Passing the property through drops that to **0**.
- `org.ops4j.pax.url.mvn.repositories` is now replaced rather than appended to. The shipped list still points at the decommissioned `oss.sonatype.org` OSSRH snapshot host, and appending re-added Central a second time.
- The unpacked container is kept only under `-Dkaraf.itest.keepRuntimeFolder=true`. It was ~420MB per run that an incremental build never reclaimed; CI runners are ephemeral.
- Karaf ITs are skipped on the Windows build (~3x slower there, plus a spurious `Terminal has been closed` error annotation on a green job). The Linux matrix covers 17 and 21.
- `make sonar` passes `-Psonar` to its `install` half — that profile is what sets `karaf.itest.skip` for the Sonar job, but it was only applied to `sonar:sonar`, which already skips tests.

### `docs(java17)`: v8 migration guide and baseline enforcement

- Adds `doc/MIGRATION-v8.md`, covering the two consequences that aren't obvious from "the baseline moved": the Maven/Gradle plugins and the annotation processor now need a Java 17+ JVM to run the *build*, independently of what the consuming project targets; and the OSGi bundles declare `osgi.ee=JavaSE 17`, which is a *resolution* requirement.
- Collapses the duplicated #8009 CHANGELOG entry to one line under breaking changes (matching how #6081 recorded the previous baseline move) and adds the migration-guide pointer.
- `maven-enforcer-plugin` had its only rule commented out; a `requireJavaVersion [17,)` rule now reports the requirement instead of failing with `invalid target release: 17`. `maven-plugin-plugin` records the same in the generated descriptor, so downstream users on an old JVM get a readable Maven error rather than `UnsupportedClassVersionError`.

### `docs`: cleanups

- The CRD generator READMEs' cross-JDK sections described a JDK 17 Maven / JDK 21 project while the remedies below still handed the reader a JDK 17 toolchain, which cannot load the JDK 21 classes described. Examples moved to 21.
- The model-update skill told the reader to switch JDKs between steps and then named 17 for both.
- The CRD generator IT poms interpolate `@maven.compiler.source@` / `@maven.compiler.target@` instead of hardcoding `17`, as the java-generator and openapi IT poms already do.

### Verification

7/7 Karaf ITs pass on **JDK 17** and **JDK 21**, with no skip override needed.

- `target/features-repo/org/` previously held only `apache ow2 snakeyaml yaml`; it now also contains `osgi/` with the three DS API bundles.
- Container feature state previously recorded both `scr/999.0.0.SNAPSHOT` and `scr/4.4.11`; now only `scr/4.4.11`.
- Container `ERROR` lines across a full run: **95 → 5** (the remainder is the jline shutdown message).
- `target/exam` after a run: **419MB → 0**.
- Enforcer verified to reject JDK 11 with the intended message and pass on 17; `crd-generator-maven-plugin`'s descriptor carries `<requiredJavaVersion>17</requiredJavaVersion>`; the filtered IT poms resolve to `<source>17</source>` and `project-scan` passes.

One caveat on the above: the pax-exam container resolves bundles from the local repository, so a local run validates whatever `999-SNAPSHOT` is installed rather than the reactor output. CI builds the reactor first, so its result is the authoritative one.

<details>
<summary>Considered and dropped: <code>karaf-maven-plugin:verify</code></summary>

Binding the `verify` goal in the features module looked like the systemic fix for a descriptor that is never validated. Measured against the two bugs here, it catches neither: an ASM pinned too low to read our bytecode resolves fine and only fails later in SPI-Fly's weaving hook, and an incomplete `scr` feature still resolves because Karaf's repository must be on the descriptor list for `scr` to be found at all — and that repository also supplies the DS bundles the incomplete feature omitted. It only catches a bundle GAV that does not exist.

It also breaks the build: the features module declares no Maven dependency on the ~100 bundles its descriptor references, so under `-T 1C` and a clean local repository `verify` runs before those artifacts are installed. Making it correct means declaring every bundle as a dependency, which is a design change rather than a drive-by. Dropped.
</details>

## Type of change

## Checklist

### DIFF
--- a/.claude/skills/k8s-model-update/SKILL.md
+++ b/.claude/skills/k8s-model-update/SKILL.md
@@ -29,11 +29,9 @@ The process has four user-confirmation checkpoints so nothing ships without revi
 
 ## Prerequisites
 
-- **Minimum Java 17** — required to build and run the project.
-- **Java 17 is the preferred JDK for model generation** (`make generate-model`). JDK 25 is NOT supported.
-- **`make format` requires minimum Java 17** — the Spotless formatter and license-header tooling need JDK 17+.
+- **Java 17** — required to build the project, to generate the model (`make generate-model`) and to run `make format`. JDK 25 is NOT supported.
 
-Before starting, verify the active JDK version (`java -version`). Switch JDK versions between steps as needed (e.g., JDK 17 for generation, JDK 17+ for formatting).
+Before starting, verify the active JDK version (`java -version`).
 
 ---
 
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -58,7 +58,10 @@ jobs:
           cache: 'maven'
       - name: Build Project
         shell: bash
-        run: ./mvnw ${MAVEN_ARGS} install
+        # The Karaf integration tests are skipped here: they take ~3x longer on Windows than on
+        # Linux and the Karaf shell emits a spurious "Terminal has been closed" error annotation
+        # on shutdown. The Linux build matrix covers them on both Java 17 and 21.
+        run: ./mvnw ${MAVEN_ARGS} install -Dkaraf.itest.skip=true
       - name: Drop in-repo artifacts before cache save
         # Don't let io.fabric8:* jars installed by this run leak into the cache that future
         # runs restore. Only third-party dependencies should be cached.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,14 +6,14 @@
 * Fix #8050: (karaf) `KubernetesClient` can be activated as an OSGi service again under the Java 17 baseline. Aries SPI-Fly (1.3.0) and ASM (8.0.1) predate Java 17 bytecode, so SPI-Fly's weaving hook failed with `Unsupported class file major version 61`, `ServiceLoader` call sites were never woven, and no `HttpClient.Factory` could be found. The feature now ships Aries SPI-Fly 1.3.7 and ASM 9.10.1, which moves SPI-Fly's ASM import range from `[8.0,9)` to `[9.6,10)`. The `scr` feature the Karaf feature repository used to define (pinning Felix SCR 2.0.6, which provides `osgi.extender=osgi.component` 1.3 while the client requires 1.5+, and which shadowed Karaf's own `scr` feature) has been removed: `kubernetes-client` now depends on the `scr` feature provided by Karaf, so the Declarative Services API bundles and the `scr:*` shell commands come from the distribution. The Karaf integration tests no longer skip on JDK 17+, so this is covered by CI again
 
 #### Improvements
-* Fix #8009: Java baseline upgraded to Java 17, enabling access to modern language features, better performance, and long-term support
 
 #### Dependency Upgrade
 
 #### New Features
 
 #### _**Note**_: Breaking changes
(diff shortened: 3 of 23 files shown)


ash-thakur-rh (approved):
LGTM!

PR #8106, opened 2026-09-15
fix(gradle): Update gradle API version and enable gradle IT tests

### LINKED ISSUE #8104: Run the java-generator Gradle plugin ITs on Java 22+
`SimpleIT` and `BuildWithGradlePluginIT` run Gradle `gradle-api.version` (8.5), which can't run on Java 22+, so they're disabled there since #8015. Moving to Gradle 9.1+ (first with Java 25 support) would bring them back on the whole CI matrix.

ash-thakur-rh:
Update gradle API version and enable gradle IT tests


ash-thakur-rh:
note: check if current plugin version runs with gradle 9 or not?

PR #8108, opened 2026-09-15
fix: prevent requests hanging when retry processing throws

### LINKED ISSUE #8105: Exceptions in async retry processing can leave requests hanging
### Describe the bug

[`AsyncUtils.retryWithExponentialBackoff`](https://github.com/fabric8io/kubernetes-client/blob/f8cc68fa9446ecb31e228ce4c1a255f36d83bc79/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/AsyncUtils.java#L64-L94) can leave its returned future pending indefinitely when retry processing throws an exception. Synchronous client operations waiting on that future can hang even when a request timeout is configured.

There are three affected paths:

- **The action throws during a scheduled retry**, before returning a future. `Utils.schedule` captures the exception in its own future, which the retry helper ignores.
- **The retry decision (`shouldRetry`) throws.** The exception is captured in the future returned by `whenComplete`, which is also ignored.
- **Response cleanup (`onCancel`) throws before a retry.** This has the same discarded-`whenComplete` problem: neither the retry nor completion of the returned result occurs.

For example:

```text
Attempt 1 returns a future that fails with a retryable error
Attempt 2 throws before returning a future
The exception is captured by the scheduler, but the caller's result never completes
```

A token provider throwing while preparing the second request can trigger this sequence. A custom `afterConnectionFailure` interceptor throwing can trigger the retry-decision case.

The timeout applies to each attempt's future, not the overall result. It cannot help when the action throws before that future exists, or when a retry callback throws after the attempt has already completed.

### Expected behavior and underlying fix

Once the helper has returned a result future, an exception that aborts retry processing must complete that result exceptionally instead of leaving it pending.

The fix needs to propagate failures from both scheduled retry execution and the completion callback that decides, cleans up, and schedules retries. Observing only the scheduler future fixes the action case but misses exceptions captured inside `whenComplete`.

Regression coverage should include a throwing retry action, a throwing retry decision, and throwing cleanup before retry. The retry-decision coverage should include an already-failed attempt: `whenComplete` preserves the original failure if its callback also throws, so exception propagation must account for that behavior.

7.9.0. The same retry code is also present on `main`.

### Relevant code

- [Retry orchestration in AsyncUtils](https://github.com/fabric8io/kubernetes-client/blob/f8cc68fa9446ecb31e228ce4c1a255f36d83bc79/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/AsyncUtils.java#L64-L94)
- [Exception capture in Utils.schedule](https://github.com/fabric8io/kubernetes-client/blob/f8cc68fa9446ecb31e228ce4c1a255f36d83bc79/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/Utils.java#L472-L486)
- [Unbounded wait in OperationSupport](https://github.com/fabric8io/kubernetes-client/blob/f8cc68fa9446ecb31e228ce4c1a255f36d83bc79/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/OperationSupport.java#L489-L496)

arnav-ag:
## Description

Fixes #8105.

Exceptions from a scheduled retry action, retry decision, or response cleanup can leave the request future pending indefinitely. Propagate these failures to the result future, including callback failures after an already-failed attempt, so callers receive an exception instead of hanging.

Added regressions for each failure path. All 12 `AsyncUtilsTest` cases pass using JUnit Platform with freshly compiled affected sources and existing build dependencies. The normal Maven test command is blocked by missing local `999-SNAPSHOT` dependencies.

The license check passed. Java formatting stalled while provisioning formatter dependencies and was stopped; CI still needs to verify formatting.

## Type of change

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Chore (non-breaking change which doesn't affect codebase; test, version modification, documentation, etc.)

## Checklist


manusa (approved):
LGTM, thx!

